Issue #1, opened 2016-07-25
Initial app setup 


angaither:
- [x] create a laravel app
 - [x] and mock the phoenix tables
- [x] `reportback`
 - [x] `reportback_items`
 - [x] `reportback_log`
 - [x] `kudos`

I think the only initial thing we want to add to the table, is the `northstar_id`

sbsmith86:
Laravel app created! 85999585f898c19230e94b194dde9d912172084c

sbsmith86:
@angaither just to make sure we need to add `northstar_id` to the `reportback` table, right?

sbsmith86:
@angaither actually i think `northstar_id` should just be `user_id` and then `id` in the user table would be the Northstar Id, similar to how we did it in gladiator.

sbsmith86:
Removed the steps to create the POST endpoint and the step to POST back to phoenix. I broke those steps out into their own issues #10, #11 , #12

sbsmith86:
Gonna close this. See the comment above.

PR #4, opened 2016-07-27
Add migrations

sbsmith86:
#### What's this PR do?

This PR adds migrations to the Rogue app to start to flush out the db structure. The goal is to have the db work just like it currently does now. 

Tables added:
* `reportback`
* `reportback_log`
* `reportback_item` formally known as `dosomething_reportback_file`
* `kudos`

Set up the app and run the migrations. The migrations should run smoothly. Compare the tables created to their corresponding tables in phoenix and the column settings should match up.

Wasn't sure about how to handle northstar ids. I decided that, like in gladiator, since the user table will be storing northstar ids, we should just call columns that store northstar ids `user_id` 

To keep things consistent, I put `user_id` wherever `uid` is stored so we can have references to the northstar user on those tables, open to thoughts about that.

#### Relevant tickets
Addresses #1 

/cc @weerd @DFurnes @aaronschachter

### DIFF
--- a/database/migrations/2016_07_26_175053_create_reportback_table.php
+++ b/database/migrations/2016_07_26_175053_create_reportback_table.php
@@ -17,7 +17,7 @@ class CreateReportbackTable extends Migration
             $table->string('northstar_id')->index()->comment('Users northstar id');
             $table->integer('drupal_id')->index()->comment('The phoenix users.uid that reported back.');
             $table->integer('campaign_id')->index()->comment('The campaign node id that the user has reported back for.');
-            $table->integer('campaign_run_nid')->index()->nullable()->comment('The campaign run id that the user reported back for.');
+            $table->integer('campaign_run_id')->index()->nullable()->comment('The campaign run id that the user reported back for.');
             $table->integer('quantity')->comment('The quantity of reportback_nouns reportback_verbed.');
             $table->text('why_participated')->nullable()->comment('Why the user participated.');
             $table->integer('num_participants')->nullable()->comment('The number of participants, if applicable.');
--- a/database/migrations/2016_07_26_175053_create_reportback_table.php
+++ b/database/migrations/2016_07_26_175053_create_reportback_table.php
@@ -14,10 +14,10 @@ class CreateReportbackTable extends Migration
     {
         Schema::create('reportback', function (Blueprint $table) {
             $table->increments('id')->comment('Primary key. Maps to rbid');
-            $table->string('northstar_id')->comment('Users northstar id');
-            $table->integer('drupal_id')->comment('The phoenix users.uid that reported back.');
-            $table->integer('campaign_id')->comment('The campaign node id that the user has reported back for.');
-            $table->integer('campaign_run_nid')->nullable()->comment('The campaign run id that the user reported back for.');
+            $table->string('northstar_id')->index()->comment('Users northstar id');
+            $table->integer('drupal_id')->index()->comment('The phoenix users.uid that reported back.');
+            $table->integer('campaign_id')->index()->comment('The campaign node id that the user has reported back for.');
+            $table->integer('campaign_run_nid')->index()->nullable()->comment('The campaign run id that the user reported back for.');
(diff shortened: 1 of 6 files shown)


database/migrations/2016_07_26_175053_create_reportback_table.php
```diff
@@ -0,0 +1,41 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('reportback', function (Blueprint $table) {
+            $table->increments('id')->comment('Primary key. Maps to rbid');
+            $table->string('user_id')->comment('The rogue users.id that correlates to their Northstar id');
+            $table->integer('uid')->comment('The phoenix users.uid that reported back.');
```
angaither: let's call this `drupal_id` to keep things [consistent with northstar](https://github.com/DoSomething/northstar/search?utf8=%E2%9C%93&q=drupal_id)

database/migrations/2016_07_26_175053_create_reportback_table.php
```diff
@@ -0,0 +1,41 @@
...
+            $table->integer('nid')->comment('The node.nid that the user has reported back for.');
```
angaither: I'm also wondering about these names, and maybe the crew will have a lot of feelings and can debate this later, but I'd rather go with `node_id` or something. hmmm

database/migrations/2016_07_26_175053_create_reportback_table.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('reportback', function (Blueprint $table) {
+            $table->increments('id')->comment('Primary key. Maps to rbid');
+            $table->string('user_id')->comment('The rogue users.id that correlates to their Northstar id');
```
DFurnes: We've used `northstar_id` for this elsewhere so it might be nice to do that for consistency.

database/migrations/2016_07_26_175053_create_reportback_table.php
sbsmith86: @angaither ah i think that makes sense, since we are out of the context of drupal we should be more specific

DFurnes:
Looks nice to me. 😄 We'll probably want to add database indexes too.

sbsmith86:
@DFurnes Not sure what you mean about adding database indexes

sbsmith86:
@angaither updated based on your comments!

DFurnes:
@sbsmith86 For columns that we're frequently querying by, [adding database indexes](https://laravel.com/docs/5.2/migrations#creating-indexes) will speed up lookups once we have a lot of records in here (for example, so MySQL can find all reportbacks for a particular `northstar_id` without having to check each row's value in that column). You can probably see which ones we have indexes on in Dr00ps and mirror those choices.

sbsmith86:
@DFurnes oh yes, sure sure. will do.

sbsmith86:
@angaither @DFurnes Update column names and added indexes. Can you take another look?

database/migrations/2016_07_26_175053_create_reportback_table.php
```diff
@@ -0,0 +1,41 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('reportback', function (Blueprint $table) {
+            $table->increments('id')->comment('Primary key. Maps to rbid');
+            $table->string('northstar_id')->index()->comment('Users northstar id');
+            $table->integer('drupal_id')->index()->comment('The phoenix users.uid that reported back.');
+            $table->integer('campaign_id')->index()->comment('The campaign node id that the user has reported back for.');
+            $table->integer('campaign_run_nid')->index()->nullable()->comment('The campaign run id that the user reported back for.');
```
angaither: `id` not `nid`

sbsmith86: @angaither ack updated!

angaither:
:shipit: 🚀

PR #6, opened 2016-07-28
Models

sbsmith86:
#### What's this PR do?
* Add `.editorconfig` file.
* Moves models into `Rogue/Models` directory/namespace.
* Adds models for each table.

#### Relevant tickets
Addresses #1


DFurnes:
No diff since it was only renamed, but the User model class is still in the root `\Rogue` namespace (rather than `\Rogue\Models`).

composer.json
```diff
@@ -17,7 +17,8 @@
     },
     "autoload": {
         "classmap": [
-            "database"
+            "database",
+            "app/Models"
```
DFurnes: This shouldn't be necessary since the models are all loaded in the PSR-4 `Rogue\` namespace defined below.

sbsmith86:
@DFurnes ahhh yep. good catch! Wasn't ready to really deal with users yet so I didn't even look at this file haha.

DFurnes:
👍 ⛵

PR #7, opened 2016-07-29
Seeders

sbsmith86:
#### What's this PR do?

* Update table names to be plural to follow laravel conventions...also makes more sense. 
* Added a new `id` field to the `reportback_items` table 
* Made all the seeders so we have something to work with. NOTE: all fake data in there, I think when we start connecting to other APIs we can update to see with real RBs.

Set up the app, run `php artisan migrate && php artisan db:seed` and all should go well. 

#### Relevant tickets
Addresses #1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,10 +6,10 @@
     "type": "project",
     "require": {
         "php": ">=5.5.9",
-        "laravel/framework": "5.2.*"
+        "laravel/framework": "5.2.*",
+        "fzaninotto/faker": "^1.6"
     },
     "require-dev": {
-        "fzaninotto/faker": "~1.4",
         "mockery/mockery": "0.9.*",
         "phpunit/phpunit": "~4.0",
         "symfony/css-selector": "2.8.*|3.0.*",
--- a/composer.json
+++ b/composer.json
@@ -7,9 +7,9 @@
     "require": {
         "php": ">=5.5.9",
         "laravel/framework": "5.2.*",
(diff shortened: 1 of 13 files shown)


composer.json
```diff
@@ -6,10 +6,10 @@
     "type": "project",
     "require": {
         "php": ">=5.5.9",
-        "laravel/framework": "5.2.*"
+        "laravel/framework": "5.2.*",
+        "fzaninotto/faker": "^1.6"
```
DFurnes: Faker should be a dev dependency.

database/seeds/ReportbackItemSeeder.php
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Run the database seeds.
+     *
+     * @return void
+     */
+    public function run()
+    {
+        $faker = Faker\Factory::create();
+
+        $reportbacks = Reportback::all();
```
DFurnes: Not necessary for merge, but it could make sense to do this with model factories in the `ReportbackSeeder` seeder. Check out the "Adding Relations To Models" section in the [Model Factories](https://laravel.com/docs/5.2/testing#model-factories) docs.

DFurnes:
Looks good! One tiny comment, otherwise 🚢!

database/seeds/ReportbackItemSeeder.php
sbsmith86: @DFurnes oh shit that is cool

PR #17, opened 2016-08-04
Laravel backups

blisteringherb:
#### What's this PR do?
I've added the [Laravel Backups package](https://docs.spatie.be/laravel-backup/v3/introduction) and its dependencies (AWS, S3, Slack notifications) so that we can use `php artisan` commands to more easily backup the database and files from the Rogue app. This is a proof of concept, and I'd like to eventually put these packages into each Laravel app that we currently have and any future ones.  It's currently set up to backup both locally and to send the backups to s3.  

At the moment, the settings will put a notification in the Slack #devops room when the backup has failed and when an unhealthy backup is found. We can add additional notifications for positive outcomes in the future if desired. By adding `slack` in the notifications events, you can test out the Slack notification for successful backups.

You can either run `php artisan backup:run`as the root user from the command line on the production node, or view the output from the (future) Jenkins job to make sure the backup command is working. You can also look at the `rogue.dosomething.org` folder in the ds-rogue-prod S3 bucket in the AWS console to verify that the backups are going to the right place.

At the moment, the `php artisan backup:list` command isn't working with s3, and I've filed a ticket with the package maintainer to figure out the issue. https://github.com/spatie/laravel-backup/issues/183 The inability to list the backups isn't a blocker at the moment for getting this into production so that we can push the backups to s3 ASAP.

Having these drush-like commands in Laravel projects is an important part of being able to manage the db without having to use `mysqldump` or `mysql` with all of the parameters to connect to the db. Adding the packages to S3 is also an important part of eventually taking all of the storage off of the node and putting into more resilient and redundant storage in S3.

#### Relevant tickets
Fixes #

#### Checklist
- [ x ] Tested on staging.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,8 @@
         "spatie/laravel-backup": "^3.8",
         "maknz/slack": "^1.7",
         "maknz/slack-laravel": "^1.0",
-        "league/flysystem-aws-s3-v3": "~1.0"
+        "league/flysystem-aws-s3-v3": "~1.0",
+        "aws/aws-sdk-php-laravel": "^3.1"
     },
     "require-dev": {
         "fzaninotto/faker": "^1.6",
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
         "predis/predis": "~1.0",
         "spatie/laravel-backup": "^3.8",
         "maknz/slack": "^1.7",
-        "maknz/slack-laravel": "^1.0"
+        "maknz/slack-laravel": "^1.0",
+        "league/flysystem-aws-s3-v3": "~1.0"
     },
     "require-dev": {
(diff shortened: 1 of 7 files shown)


angaither:
@blisteringherb can you update the readme with the new command and fix the style errors?

otherwise looks 👌!

mshmsh5000:
I agree—style fixes and then 🍕 

Just to ask the dumbest question I can: [This package](https://docs.spatie.be/laravel-backup/v3/introduction) doesn't provide recovery steps/artisan recovery commands, right? The backup process is fully automated, but the restoration process would be manual?

blisteringherb:
That's correct, the restore process would be manual. @mshmsh5000

mshmsh5000:
Ok thanks.

PR #18, opened 2016-08-04
WORK IN PROGRESS: Start api

sbsmith86:
#### What's this PR do?

I am starting work on the API for Rogue. The first step is to create a POST endpoint `/reportbacks` that can be hit to store reportbacks. The first iteration of this app will be to have this thing function exactly how drupal is functioning now. We have copied the database schema from drupal and functionally are thinking about  it the same way. However, we will be looking to make updates and that could mean updates to the database schema which is why I felt the repository stuff made sense here. 

Here is my main question:

When a reportback is posted we have to do three things - Create a record in the `reportback` table, create a record in the `reportback_items` table, create a record in the `reportback_log` table. Would it make sense to have just a service class that does all that logic? Or should all that logic live in a Repository (as i have it now)?


app/Http/Middleware/VerifyCsrfToken.php
```diff
@@ -12,6 +12,6 @@ class VerifyCsrfToken extends BaseVerifier
      * @var array
      */
     protected $except = [
-        //
+        'api/*',
```
DFurnes: You can exclude this middleware by putting the API endpoints in the [`api` middleware group](https://github.com/DoSomething/rogue/blob/db087e225a3dbddf2e66f0bf318d1d079229fcbd/app/Http/Kernel.php#L34-L36), rather than [`web`](https://github.com/DoSomething/rogue/blob/db087e225a3dbddf2e66f0bf318d1d079229fcbd/app/Http/Kernel.php#L26-L32).

app/Http/Transformers/ReportbackTransformer.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Transform resource data.
+     *
+     * @param User $user
+     * @return array
+     */
+    public function transform(Reportback $reportback)
+    {
+        return [
+            'id' => (string) $reportback->id,
```
DFurnes: I think these would make sense to handle with [attribute casting](https://laravel.com/docs/5.2/eloquent-mutators#attribute-casting) on the model itself, so we don't have to remember to cast them everywhere we use in the codebase.

app/Repositories/ReportbackRepository.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Create a new reportback
+     *
+     * @param  array $data
+     * @return \Gladiator\Models\Message
+     */
+    public function create($data)
+    {
+        // Store reportback
+        $reportback = new Reportback;
```
DFurnes: Totally not necessary, but you could combine these three lines into `$reportback = Reportback::create($data)`.

DFurnes:
Looks nice to me! 👍 I'm still not sure that the repository abstraction will end up being necessary, but it doesn't hurt to keep it around in case.

app/Http/Transformers/ReportbackTransformer.php
weerd: @sbsmith86 is the `id` not coming in from the database as a string?

DFurnes:
ALTHOUGH! I was just looking through this again with @chloealee to see how she'd integrate the Phoenix API endpoints, and it does kinda make sense to have the `ReportbackRepository` be responsible for all the local database stuff, and then let the `ReportbackService` handle sending off requests to the repository & Phoenix API (like all the higher level orchestration).

app/Http/Transformers/ReportbackTransformer.php
sbsmith86: it comes back as a string. I was just trying to be as explicit as possible in the transformer, but doesn't seem necessary

weerd: Yeah, I'd say you only really need to cast if it's coming back as one thing and you need to set it as something else, but if it's the same type, maybe don't worry about it? :shrug:

sbsmith86:
@DFurnes ok great! thanks for your 👀 and comments. I think this is all making sense to me now.

app/Repositories/ReportbackRepository.php
weerd: @DFurnes I thought you hate facades? `(new Reportback)->create($data)` alternatively, haha.

app/Services/ReportbackService.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Rogue\Services;
```
weerd: this sounds like a cool company name `Rogue Services`, haha

sbsmith86: Sounds like an evil corporation! BWAHAHA

app/Repositories/ReportbackRepository.php
DFurnes: Not a facade! 😉 Reportback is a real class with a static method.

The reason that facades bother me is the class (`\Auth` for example) doesn't _really_ exist, and Laravel then maps the "missing" class to a real one (`\Illuminate\Auth\AuthManager` or `\Illuminate\Contracts\Auth\Guard`) at runtime via PHP's Reflection API and the service container. So it can make it confusing to figure out what's happening behind the scenes.

weerd: ah! gotcha. 🎯

angaither: @_@

app/Services/ReportbackService.php
angaither: @deadlybutter

weerd: haha! 🙀

deadlybutter: giphy

Issue #12, opened 2016-07-29
POST reportbacks back to Phoenix when they get stored


sbsmith86:
To get this thing up an running, when a reportback gets stored in Rogue, we will simply POST it back to phoenix so all the functionality in phoenix will remain the same while we make updates in this app. 

### Questions
 - what is the current status on being able to POST reportbacks back to phoenix?

sbsmith86:
I don't think we are doing this anymore unless we get the file storage thing handled. I will close for now and reopen when we see where we are at next sprint.

PR #22, opened 2016-08-11
Phoenix service

### LINKED ISSUE #21: Queue failed POST requests to Rogue
If a POST request fails to the app, we should queue those requests so we can run them again.

chloealee:
#### What's this PR do?
Sets up basic functionality to allow connection from rogue to phoenix, using [northstar-php's](https://github.com/dosomething/northstar-php) Rest Api Client.

Sets up basic functionality in to user Rest Api Client to make calls to phoenix. 

Added a new function `raw` in `Phoenix.php` that is needed if we POST back to phoenix and need to add Cookie and X-CSRF-Token to header. This might not be needed right now since it's not basic setup - let me know if I should remove! 

@DFurnes @sbsmith86 is there a way to test this as is? Or should I test with the `POST https://www.dosomething.org/api/v1/auth/login` endpoint perhaps to make sure it's working? 

#### Relevant tickets
Sets up service to connect to phoenix to allow for https://github.com/DoSomething/rogue/issues/12

#### Checklist
- [ ] Tested on staging.


.env.example
```diff
@@ -1,27 +0,0 @@
-APP_ENV=local
```
DFurnes: This file shouldn't have been deleted from Git. It's a template so anyone can set up their own `.env` when they first set up their local app.

Vagrantfile
```diff
@@ -0,0 +1,71 @@
+# -*- mode: ruby -*-
```
DFurnes: I'm not sure how this got into the repo, but it shouldn't be necessary since we use Homestead for this project.

app/Providers/PhoenixServiceProvider.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Rogue\Providers;
+
+use Rogue\Services\Phoenix\Phoenix;
+use Illuminate\Support\ServiceProvider;
+
+class PhoenixServiceProvider extends ServiceProvider
```
DFurnes: Since you're not injecting any custom dependencies or config values, this shouldn't be necessary. Laravel can resolve the `\Rogue\Services\Phoenix` without it being explicitly registered in a service provider.

app/Services/Phoenix/Phoenix.php
```diff
@@ -0,0 +1,103 @@
...
+	public function __construct()
+	{
+        $this->base_uri = config('services.phoenix.uri') . '/api/' . config('services.phoenix.version') . '/';
+
+        $headers = [
+           'Content-type' => 'application/json',
+           'Accept' => 'application/json',
+        ];
+
+        parent::__construct($this->base_uri, $headers);
```
DFurnes: I don't think that `RestApiClient` takes [more than one parameter](https://github.com/DoSomething/northstar-php/blob/821b6ebe02939b241c150ca672e306a37c54f3ed/src/Common/RestApiClient.php#L41) so I believe you can remove this second argument & the `$headers` array above.

DFurnes: It also doesn't look like `$this->base_uri` is a field on the base class, so you can make that into a normal variable `$url = config(...` before passing it to the parent constructor. It'll be stored on the RestApiClient's Guzzle instance.

database/migrations/2016_06_29_205018_add_oauth_fields_to_users.php
```diff
@@ -0,0 +1,45 @@
+<?php
```
sbsmith86: I'm not sure what this is/why we need it? I was thinking we would do Auth stuff separately since we haven't dealt with users yet. Is it related to the phoenix stuff?

DFurnes:
The `composer.phar` file can be removed from the repo as well. Composer is installed as a global dependency on Homestead and our build servers.

app/Providers/PhoenixServiceProvider.php
chloealee: ah makes sense, thanks @DFurnes ! So just delete this whole file?

DFurnes: Yep, and remove it from the providers array in `config/app.php`.

chloealee: Cool, thanks!

chloealee:
Cool - all set and deleted @DFurnes !

database/migrations/2016_06_29_205018_add_oauth_fields_to_users.php
chloealee: I don't think that auth for phoenix is available yet but this was added when integrating northstar-php and can be used later to authorize against northstar - is that correct @DFurnes ?

.env.example
```diff
@@ -24,4 +24,4 @@ MAIL_HOST=mailtrap.io
 MAIL_PORT=2525
 MAIL_USERNAME=null
 MAIL_PASSWORD=null
-MAIL_ENCRYPTION=null
+MAIL_ENCRYPTION=null
```
DFurnes: Not a big deal, but there should be a newline at the end of the file or Git gets angryyy.

app/Services/Phoenix/Phoenix.php
```diff
@@ -0,0 +1,98 @@
...
+use DoSomething\Northstar\Resources\NorthstarUserCollection;
+use DoSomething\Northstar\Resources\NorthstarClientCollection;
+use GuzzleHttp\Cookie\CookieJar;
+use Cache;
+
+class Phoenix extends RestApiClient
+{
+	/**
+	 * Create a new Phoenix API client.
+	 * @param array $config
```
DFurnes: This param doesn't exist!

database/migrations/2016_06_29_205018_add_oauth_fields_to_users.php
DFurnes: Yep, these are the two bundled migrations with Northstar PHP, for storing user access/refresh tokens and client access tokens respectively. They're not used unless authenticating via Northstar, so safe to remove for now.

DFurnes:
Looks like StyleCI is upset about indentation (check that you have [EditorConfig](http://editorconfig.org) installed in your Sublime Text!) and a few unused imports at the top of the Phoenix API class that can be removed.

database/migrations/2016_06_29_205018_add_oauth_fields_to_users.php
sbsmith86: @DFurnes ahhh i see. hmm i guess we can leave them there then! Just didn;t want to get too ahead of ourselves

.env.example
chloealee: ah yes! I thought I had the package that it did that automatically but looks like I'll have to re-install that - thank you!

chloealee:
that is so cool that is detects unused imports at the top of the file! I didn't realize that but totally makes sense what it meant now - thank you!!

chloealee:
@DFurnes is it possible there is something wrong with my Style Ci? It doesn't seem to change the spacing when I change it and seems a little different from my actual code.

DFurnes:
It looks like you might be set to 3-space indents. I can come take a look IRL!

chloealee:
@DFurnes hmm in sublime my indents are 4-spaces!

chloealee:
@DFurnes thank you so much for the help! All commits squashed now :)

app/Services/Phoenix/Phoenix.php
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * Returns a token for making authenticated requests to the Drupal API.
+     *
+     * @return array - Cookie & token for authenticated requests
+     */
+    private function authenticate()
+    {
+        $authentication = Cache::remember('drupal.authentication', 30, function () {
+            $payload = [
+                'username' => env('PHOENIX_USERNAME'),
```
DFurnes: Oh sorry, just noticed this: these should be added to the `config/services.php` file like the URL setting, and then referenced here using the `config()` helper method. This allows Laravel to cache those values between requests.

DFurnes:
Noticed one last tiny thing, but otherwise looks perfect! 🎯

app/Services/Phoenix/Phoenix.php
chloealee: ah I see that makes sense! Thanks for the explanation for why to use this instead of just the value in the .env file! Just changed and will merge :)

chloealee:
More issues with Style Ci:
screen shot 2016-08-15 at 11 07 26 am

According to [this post](https://github.com/dingo/api/pull/1057#issuecomment-226555890), it says it's a bug with Github. Going to try to close and re-open the commit per their advice and just want to document here!

chloealee:
Hmm @DFurnes the above trick didn't seem to work - have you seen this before?

DFurnes:
@chloealee Hmm, you could try amending your last commit and then force-pushing, but I think you're probably safe to just merge if you want.

chloealee:
ok thank you! Tried the amend and forcing the push but still the same error :-/ Going to just merge.

PR #24, opened 2016-08-16
Rogue authentication via Northstar

### LINKED ISSUE #9: Authentication via Northstar
Allow admins to authenticate via Northstar. 

[See example](https://github.com/DoSomething/gladiator/pull/12) of how we did this in gladiator 

Or is there a better way of doing this now, with the `Northstar-php` package? @DFurnes @weerd

chloealee:
#### What's this PR do?
- Adds [Northstar-PHP](https://github.com/DoSomething/northstar-php)
- Uses [`northstar` authentication driver](https://github.com/DoSomething/northstar-php#laravel-authentication)
- Adds role middleware (`CheckRole.php`) to only allow admin users to use Rogue.

@DFurnes is there a way to test this yet even though other parts are not fully built out? 

#### Notes
`resources/views/auth/unauthorized.blade.php` is just a placeholder for now! 

#### Relevant tickets
Fixes #9 

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/Auth/AuthController.php
```diff
@@ -31,13 +31,20 @@ class AuthController extends Controller
...
+    /**
+     * Create a new authentication controller instance.
+     * @param Guard $guard
      */
-    public function __construct()
+    public function __construct(Guard $guard)
     {
-        $this->middleware($this->guestMiddleware(), ['except' => 'logout']);
+        $this->guard = $guard;
+        $this->middleware('role:admin', ['except' => 'logout']);
```
DFurnes: This probably doesn't work, right? Since it'd be gating the `login` action to only allow logged-in admin users.

app/Http/Middleware/CheckRole.php
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Handle an incoming request.
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \Closure $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next)
+    {
+        $roles = array_slice(func_get_args(), 2);
```
DFurnes: Here's the snippet for getting this with a spread parameter for the function:

```php
public function handle($request, Closure $next, ...$roles)
{
// ... now $roles is the same thing gotten by array_slice(fund_get_args(), 2);
```

config/auth.php
```diff
@@ -66,8 +66,8 @@
 
     'providers' => [
         'users' => [
-            'driver' => 'eloquent',
-            'model' => Rogue\Models\User::class,
+            'driver' => 'northstar',
+            'model' => App\User::class,
```
DFurnes: I believe this should still be `Rogue\Models\User::class`. `App\User::class` is just provided as a placeholder in the readme for whatever the given application's user model is named.

config/services.php
```diff
@@ -23,6 +23,14 @@
         'secret' => env('MANDRILL_SECRET'),
     ],
 
+    'northstar' => [
+        'grant' => 'password',
+        'url' => env('NORTHSTAR_URL'),
+        'client_id' => env('NORTHSTAR_CLIENT_ID'),
+        'client_secret' => ent('NORTHSTAR_CLIENT_SECRET'),
+        'scope' => ['user', 'admin'],
```
DFurnes: For now, let's request `['user', 'role:staff', 'role:admin']` as default scopes for Rogue.

app/Http/Controllers/Auth/AuthController.php
chloealee: Ah right! So would I have to create a `UsersController.php` that would look something similar to [this](https://github.com/DoSomething/aurora/blob/master/app/Http/Controllers/UsersController.php#L1-L27)?

config/auth.php
chloealee: ahh duh!

config/services.php
chloealee: Ah got it! And is this because it is assigning roles to the user instead of having a user and an admin?

DFurnes: Yep, so these determine what kind of abilities the access token that Northstar grants will have. The `user`, `role:staff`, and `role:admin` will allow the token to act with as much authority as the user has on their account (so like `role:admin` allows you to act as an admin, but only if your Northstar user has the admin role).

The `admin` scope grants privileges and is more intended for "machine" clients that wouldn't have a role attached to them (yet! it's probably something we should change).

app/Http/Controllers/Auth/AuthController.php
DFurnes: Right! So you could use `role:admin,staff` to limit access to another controller to only admin or staff user accounts (but anyone should be allowed to log in).

chloealee: ahh makes total sense!

DFurnes:
This looks good! 🔏 

> @DFurnes is there a way to test this yet even though other parts are not fully built out?

Yes! Try adding an `index` route to the `UsersController` that you made in this PR (it can just return `"Yay you're allowed in!"` as a string or something). You should then be able to hook this up to Northstar QA & verify that logins and protected routes all work as expected. :)

resources/views/auth/anauthorized.blade.php
```diff
@@ -0,0 +1,2 @@
+<h1>You Don't have the proper priviledges to do this!</h1>
```
DFurnes: OH! One little thing. This file has a typo in the name: `anauthorized.blade.php`. 👾

chloealee: oh shoot - thanks for catching that!

chloealee:
Cool, thanks @DFurnes ! When you say, hook up to Northstar QA, do you mean change the `NORTHSTAR_URL` in my `.env` to `http://northstar-qa.dosomething.org` and update the NORTHSTAR_CLIENT_ID and NORTHSTAR_CLIENT_SECRET?

DFurnes:
Yes!

chloealee:
@DFurnes I _think_ this is working now? When I try to visit `http://rogue.dev:8000/api/v1/users` it then redirects me to `http://rogue.dev:8000/api/v1/users` - is this Northstar-php magic prompting the user to sign in first to make sure they have the correct credentials to view this info?

DFurnes:
Yep! I'm pretty sure [this](https://github.com/DoSomething/northstar-php/blob/821b6ebe02939b241c150ca672e306a37c54f3ed/src/Laravel/LaravelOAuthRepository.php#L79-L93) is what you're seeing.

chloealee:
Great, thanks @DFurnes ! Is it ok to merge?

DFurnes:
Yep, looks good to me! 👍

PR #25, opened 2016-08-18
Hooks up S3 to save reportback item photo

### LINKED ISSUE #26: Save reportback item file in S3
When Phoenix sends a reportback to Rogue, grab the file from `$data` and store in S3.

chloealee:
#### What's this PR do?
- Adds `file_url` column to the `reportback_items` table to store the URL where the reportback item image is located on S3. 
- Adds functionality to save reportback item photo in S3 when reportback is received from Phoenix.
- Adds s3 testing to `testCreatingNewReportback` test.

- Run `phpunit` and all tests should still pass. 
- Manually, post to `http://rogue.dev:8000/api/v1/reportbacks` in Postman with below as reference:
screen shot 2016-08-25 at 3 12 10 pm
- Check S3 and you should see images uploaded to correct bucket. Works with both file uploads and base64 strings. 
- Database should also reflect the url of where these images live. 

Follows logic used to [upload user profile photos to S3 in Northstar](https://github.com/DoSomething/northstar/blob/dev/app/Services/AWS.php).

#### Relevant tickets
Fixes #26 

#### Checklist
- [ ] Tested on staging.


app/Services/AWS.php
```diff
@@ -0,0 +1,93 @@
...
+            $data = $this->base64StringToDataString($file);
+        } else {
+            $data = file_get_contents($file->getPathname());
+        }
+
+        $extension = pathinfo($file, PATHINFO_EXTENSION);
+        if (empty($extension)) {
+            $extension = $this->guessExtension($data);
+        }
+
```
chloealee: @DFurnes I know this isn't ideal but this is the only way that I could get both the test to pass and it to work when I tested manually. 

When using `pathinfo`, this works correctly in the test but when testing manually, this returns an empty string. When using `guessExtension`, this works correctly when testing manually but made the tests fail because it read the extension of `jpegs` as `bin` and the extension of `png` as `txt`. Let me know if you think there is a better way of doing this!

chloealee:
ohhh @DFurnes now I see that there is a way to mock an uploaded file - I'll work on doing this so we don't have code specific to tests and real situations!

chloealee:
@DFurnes this is ready now for a final look!

app/Services/AWS.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Store a reportback item (image) in S3.
+     *
+     * @param \Symfony\Component\HttpFoundation\File\UploadedFile|string $file
+     *  File object, or a base-64 encoded data URI
+     * @param string $filename - Filename to write image to
+     *
+     * @return string - URL of stored image
+     */
+    public function storeReportbackItem($file, $filename)
```
DFurnes: It might make sense to just call this method `storeImage` since it's not really specific to reportback items.

app/Services/AWS.php
```diff
@@ -0,0 +1,90 @@
...
+        $extension = $this->guessExtension($data);
+
+        // Make sure we're only uploading valid image types
+        if (! in_array($extension, ['jpeg', 'png'])) {
+            throw new UnprocessableEntityHttpException('Invalid file type. Upload a JPEG or PNG.');
+        }
+
+        // Add a unique timestamp (e.g. uploads/folder/filename-1456498664.jpeg) to
+        // uploads to prevent AWS cache giving the user an old upload.
+        $path = '/uploads/' . env('S3_BUCKET') . '/' . $filename . '-' . time() . '.' . $extension;
```
DFurnes: I think the folder name should be passed in as a parameter (or hard-coded as `reportback-items` if not) here. It doesn't seem super helpful to put the images within a folder named for the bucket they're already being stored in.

tests/ReportbackApiTest.php
```diff
@@ -22,6 +23,18 @@ class ReportbackApiTest extends TestCase
...
+        $this->fileSystem->shouldReceive('put')->andReturn(true);
+
+        // Create an uploaded file.
+        $path = storage_path('images/huskycorgi.jpeg');
+        $original_name = 'huskycorgi.jpeg';
+        $mime_type = 'image/jpeg';
+        $size = 2476;
+        $error = null;
+        $test = true;
+        $file = new \Illuminate\Http\UploadedFile($path, $original_name, $mime_type, $size, $error, $test);
```
DFurnes: 👍! It could be nice to make a `$this->mockFile` method on the TestCase so this can be re-used elsewhere.

tests/ReportbackApiTest.php
```diff
@@ -22,6 +23,18 @@ class ReportbackApiTest extends TestCase
...
     public function testCreatingNewReportback()
     {
+        // Mock sending image to AWS.
+        $this->fileSystem->shouldReceive('put')->andReturn(true);
+
+        // Create an uploaded file.
+        $path = storage_path('images/huskycorgi.jpeg');
+        $original_name = 'huskycorgi.jpeg';
+        $mime_type = 'image/jpeg';
+        $size = 2476;
```
DFurnes: Is this the size of the `huskycorgi.jpeg` image? Not sure if it'd make sense to infer from the path if so. Definitely not a blocker if tests are passing, but perhaps a good future improvement!

chloealee: you're right - doesn't need it! I will delete now.

chloealee:
@DFurnes ready for a final final look now :)

DFurnes:
🚢 ➡️

app/Http/Transformers/ReportbackItemTransformer.php
```diff
@@ -19,6 +19,7 @@ public function transform(ReportbackItem $reportbackItem)
             'id' => (string) $reportbackItem->id,
             'reportback_id' => $reportbackItem->reportback_id,
             'file_id' => $reportbackItem->file_id,
+            'file_url' => $reportbackItem->file_url,
```
weerd: Thoughts on maybe changing this to be a media item and inside it it contains the id, url, etc... kinda like it does now in the RB items resource endpoint? 
image

Having objects that contain further properties helps avoid having to have a long list of properties with special prefixes `file_something`... 

We could keep it named file if that ends up being preferred, but something more like:
```
'media' => [
  'id' => $reportbackItem->file_id,
  'url' => $reportbackItem->file_url,
]
```

thoughts?

chloealee: I think this is a great idea - will update now!

Issue #11, opened 2016-07-29
Create a PUT endpoint that updates reportbacks


sbsmith86:
Create an endpoint in Rogue that an app can POST to to store reportbacks.  
- [ ] Use the endpoint at `/reportback`
- [ ] Update reportbacks that we already have in the `reportbacks` table when the `quantitity` or `why_participated` values changed and create a log recore
- [ ] Create log record when the reportback gets updated in `reportback_logs`

ngjo:
due date: aug 30

sbsmith86:
Closing with #30

PR #30, opened 2016-08-31
Update reportbacks

### LINKED ISSUE #20: /reportbacks POST endpoint API validation
Validate POST requests coming into the `/reportbacks` POST endpoint. 

Rules:
  - **northstar_id** or **drupal_id**: (string|int).
    One of these should be required. 
  - **campaign_id**: (int) required.
  - **campaign_run_id**: (int) required.
  - **quantity**: (int) required.
  - **why_participated**: (string) required
  - **file_id**: (int) required.
  - **caption**: (string).
  - **source**: (string).

sbsmith86:
#### What's this PR do?
If a reportback already exists, this PR updates the reportback with the data that is passed in the requests. 

Right now "Already exists" means there is a record in the `reportback` table with a unique combination of `drupal_id` or `northstar_id`, `campaign_id`, and `campaign_run_id`. Eventually we could use `id` to determine if a reportback exists. But since we are maintaining two different systems, a reportback might have different ids in Phoenix and Rogue and the systems are unaware of the ids set in the other system. I would be up to discussing if we need to like store `rbid` in rogue or `rogue_reportback_id` in drupal in the meantime. I think it might be useful especially when working with reportback items.

I put some comments on some other things I updated and why. I also left out some stuff that I also think we need to think through like updating reviewed status per item when it is reviewed (how do we identify the corresponding items in phoenix and rogue?) and or an API endpoints that just deal with items. 

run `phpunit` make sure tests pass. Also make a requests to update a value of a reportback that already exists. 

#### Relevant tickets
Fixes #20 #11

### DIFF
--- a/app/Http/Controllers/Api/ReportbackController.php
+++ b/app/Http/Controllers/Api/ReportbackController.php
@@ -3,7 +3,7 @@
 namespace Rogue\Http\Controllers\Api;
 
 use Rogue\Models\Reportback;
-use Illuminate\Http\Request;
+use Rogue\Http\Requests\ReportbackRequest;
 use Rogue\Services\ReportbackService;
 use Rogue\Http\Transformers\ReportbackTransformer;
 
@@ -30,8 +30,9 @@ class ReportbackController extends ApiController
      * @param  \Illuminate\Http\Request  $request
      * @return \Illuminate\Http\Response
      */
-    public function store(Request $request)
+    public function store(ReportbackRequest $request)
     {
+        // @TODO - instead should probably just have a method that gets northstar_id by default from a drupal_id if that is the only thing provided and then use that to find the reportback.
         $userId = $request['northstar_id'] ? $request['northstar_id'] : $request['drupal_id'];
         $type = $request['northstar_id'] ? 'northstar_id' : 'drupal_id';
 
--- a/app/Http/Controllers/Api/ReportbackController.php
+++ b/app/Http/Controllers/Api/ReportbackController.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Http\Controllers\Api;
 
+use Rogue\Models\Reportback;
 use Illuminate\Http\Request;
 use Rogue\Services\ReportbackService;
 use Rogue\Http\Transformers\ReportbackTransformer;
@@ -23,15 +24,27 @@ class ReportbackController extends ApiController
     }
 
     /**
-     * Store a newly created resource in storage.
-     * @todo upsert reportbacks by default.
+     * Store a newly created resource in storage or
+     * update a reportback if it already exists.
+     *
      * @param  \Illuminate\Http\Request  $request
      * @return \Illuminate\Http\Response
      */
-    public function create(Request $request)
+    public function store(Request $request)
     {
-        $reportback = $this->reportbackService->create($request->all());
+        $userId = $request['northstar_id'] ? $request['northstar_id'] : $request['drupal_id'];
+        $type = $request['northstar_id'] ? 'northstar_id' : 'drupal_id';
 
-        return $this->item($reportback);
+        $reportback = $this->reportbackService->exists($request['campaign_id'], $request['campaign_run_id'], $userId, $type);
+
+        if (! $reportback) {
+            $reportback = $this->reportbackService->create($request->all());
+
+            return $this->item($reportback);
+        } else {
+            $updatedReportback = $this->reportbackService->update($reportback, $request->all());
+
+            return $this->item($updatedReportback);
+        }
     }
 }
--- a/app/Http/Controllers/Api/ReportbackController.php
+++ b/app/Http/Controllers/Api/ReportbackController.php
@@ -36,7 +36,7 @@ class ReportbackController extends ApiController
         $userId = $request['northstar_id'] ? $request['northstar_id'] : $request['drupal_id'];
         $type = $request['northstar_id'] ? 'northstar_id' : 'drupal_id';
 
-        $reportback = $this->reportbackService->exists($request['campaign_id'], $request['campaign_run_id'], $userId, $type);
+        $reportback = $this->reportbackService->getReportback($request['campaign_id'], $request['campaign_run_id'], $userId, $type);
 
         if (! $reportback) {
             $reportback = $this->reportbackService->create($request->all());
--- a/app/Http/Controllers/Api/ReportbackController.php
+++ b/app/Http/Controllers/Api/ReportbackController.php
@@ -38,14 +38,18 @@ class ReportbackController extends ApiController
 
         $reportback = $this->reportbackService->getReportback($request['campaign_id'], $request['campaign_run_id'], $userId, $type);
 
-        if (! $reportback) {
+        $updating = ! is_null($reportback);
+
+        if (! $updating) {
             $reportback = $this->reportbackService->create($request->all());
 
-            return $this->item($reportback);
+            $code = 200;
         } else {
-            $updatedReportback = $this->reportbackService->update($reportback, $request->all());
+            $reportback = $this->reportbackService->update($reportback, $request->all());
 
-            return $this->item($updatedReportback);
+            $code = 201;
         }
+
+        return $this->item($reportback, $code);
     }
 }
--- a/app/Repositories/ReportbackRepository.php
+++ b/app/Repositories/ReportbackRepository.php
@@ -120,7 +120,7 @@ class ReportbackRepository
      *
      * @return \Rogue\Models\Reportback|null
      */
-    public function getUserReportback($campaignId, $campaignRunId, $userId, $type = 'northstar_id')
+    public function getReportbackByUser($campaignId, $campaignRunId, $userId, $type = 'northstar_id')
     {
         try {
             if ($type === 'drupal_id') {
--- a/app/Repositories/ReportbackRepository.php
+++ b/app/Repositories/ReportbackRepository.php
@@ -20,7 +20,7 @@ class ReportbackRepository
      *
      * @todo Handle errors better during creation.
      * @param  array $data
-     * @return \Rogue\Models\Reportback
+     * @return \Rogue\Models\Reportback|null
      */
     public function create($data)
     {
--- a/app/Repositories/ReportbackRepository.php
+++ b/app/Repositories/ReportbackRepository.php
@@ -20,24 +20,63 @@ class ReportbackRepository
      *
      * @todo Handle errors better during creation.
      * @param  array $data
-     * @return \Gladiator\Models\Message
+     * @return \Rogue\Models\Reportback
      */
     public function create($data)
     {
-        // Store reportback
         $reportback = Reportback::create($data);
 
-        // Store reportback item photo in S3.
-        $data['file_url'] = $this->aws->storeImage($data['file'], $data['file_id']);
+        if ($reportback) {
+            $reportback = $this->addItem($reportback, $data);
 
-        // Store reportback item.
-        $reportback->items()->create(array_only($data, ['file_id', 'file_url', 'caption', 'status', 'reviewed', 'reviewer', 'review_source', 'source', 'remote_addr']));
+            // Record transaction in log table.
+            $this->log($reportback, $data, 'insert');
 
+            return $reportback;
+        }
+
+        return null;
+    }
+
+    /**
+     * Update an existing reportback.
+     *
+     * @param  \Rogue\Models\Reportback $reportback
+     * @param  array $data
+     *
+     * @return \Rogue\Models\Reportback
+     */
+    public function update($reportback, $data)
+    {
+        if (isset($data['file'])) {
+            $reportback = $this->addItem($reportback, $data);
+        }
+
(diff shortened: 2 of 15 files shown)


app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -23,15 +24,28 @@ public function __construct(ReportbackService $reportbackService)
...
+     * update a reportback if it already exists.
+     *
      * @param  \Illuminate\Http\Request  $request
      * @return \Illuminate\Http\Response
      */
-    public function create(Request $request)
+    public function store(ReportbackRequest $request)
     {
-        $reportback = $this->reportbackService->create($request->all());
+        // @TODO - instead should probably just have a method that gets northstar_id by default from a drupal_id if that is the only thing provided and then use that to find the reportback.
```
weerd: I like the sound of that cause then you would never need the `type` to know the difference. It's always ends up as a `northstar_id`. But cool for a later thing :) Good to note it!

app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -23,15 +24,28 @@ public function __construct(ReportbackService $reportbackService)
...
+        $userId = $request['northstar_id'] ? $request['northstar_id'] : $request['drupal_id'];
+        $type = $request['northstar_id'] ? 'northstar_id' : 'drupal_id';
 
-        return $this->item($reportback);
+        $reportback = $this->reportbackService->exists($request['campaign_id'], $request['campaign_run_id'], $userId, $type);
```
weerd: Totally a nitpick, but I read `$this->reportbackService->exists()` as, "determine if the Reportback Service exists and is up and running". Maybe eventually rename this method to `find()`? Also feel free to ignore me 😬

app/Repositories/ReportbackRepository.php
```diff
@@ -20,24 +20,63 @@ public function __construct(AWS $aws)
      *
      * @todo Handle errors better during creation.
      * @param  array $data
-     * @return \Gladiator\Models\Message
+     * @return \Rogue\Models\Reportback
```
weerd: hehe `Gladiator` 💂🏼

app/Http/Controllers/Api/ReportbackController.php
DFurnes: Yeah agreed! We have `normalize` and `resolve` methods on the Registrar in Northstar that handle this kind of "how do I find this thing from a bunch types of identifiers" problem and it seems to work pretty elegantly.

app/Http/Controllers/Api/ReportbackController.php
DFurnes: Yeah, or `getReportback` perhaps? To me, `exists` kinda implies a boolean response.

app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -23,15 +24,28 @@ public function __construct(ReportbackService $reportbackService)
...
+
+        if (! $reportback) {
+            $reportback = $this->reportbackService->create($request->all());
+
+            return $this->item($reportback);
+        } else {
+            $updatedReportback = $this->reportbackService->update($reportback, $request->all());
+
+            return $this->item($updatedReportback);
```
DFurnes: Another nitpicky suggestion, but these two `return`s could be pulled out from the conditional so there's always a "happy path" to the return at the end (and then it's just returning the same `$reportback` either way).

app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -23,15 +24,28 @@ public function __construct(ReportbackService $reportbackService)
...
-        $reportback = $this->reportbackService->create($request->all());
+        // @TODO - instead should probably just have a method that gets northstar_id by default from a drupal_id if that is the only thing provided and then use that to find the reportback.
+        $userId = $request['northstar_id'] ? $request['northstar_id'] : $request['drupal_id'];
+        $type = $request['northstar_id'] ? 'northstar_id' : 'drupal_id';
 
-        return $this->item($reportback);
+        $reportback = $this->reportbackService->exists($request['campaign_id'], $request['campaign_run_id'], $userId, $type);
+
+        if (! $reportback) {
+            $reportback = $this->reportbackService->create($request->all());
```
weerd: I might be confusing myself, but what's the difference here between what the Repository would do, versus the Service? I would have assumed this would be the Respository, but I could be jumbling it up in my head...

app/Http/Controllers/Api/ReportbackController.php
DFurnes: Another nice thing would be to keep track of which branch of the conditional is run (say, with an `$updating` boolean) that can then be passed along as either a 200 or 201 status code to the response method. Here's an [example](https://github.com/DoSomething/northstar/blob/e6255f2fbc5cd162587867adb2457c07bfd81b51/app/Http/Controllers/UserController.php#L123-L125) of a similar kind of thing on Northstar's `POST users/` endpoint.

app/Repositories/ReportbackRepository.php
DFurnes: Since this method can return null, we should specify that in the `@return` docblock:

```php
/**
 * ...
 * @return \Rogue\Models\Reportback|null
 */
```

app/Repositories/ReportbackRepository.php
```diff
@@ -50,4 +89,57 @@ public function create($data)
...
+    {
+        try {
+            if ($type === 'drupal_id') {
+                $reportback = Reportback::where([
+                    ['drupal_id', '=', $userId],
+                    ['campaign_id', '=', $campaignId],
+                    ['campaign_run_id', '=', $campaignRunId],
+                ])->firstOrFail();
+            } else {
+                $reportback = Reportback::where([
```
weerd: Kinda duplicating these here. The `drupal_id` vs `northstar_id` is the only diff between them. Maybe just define the array separate and use conditional for the id. Something like this:

```php
$parameters = [];
// do stuff to fill out the $parameters array along w/ conditional for ID.

$reportback = Reportback::where($parameters)->firstOrFail();
``

File: app/Repositories/ReportbackRepository.php [hunk 7]
```diff
@@ -50,4 +89,57 @@ public function create($data)
...
+     * @param string|int $userId
+     * @param string $type
+     *
+     * @return \Rogue\Models\Reportback|null
+     */
+    public function getReportbackByUser($campaignId, $campaignRunId, $userId, $type = 'northstar_id')
+    {
+        try {
+            if ($type === 'drupal_id') {
+                $reportback = Reportback::where([
```
DFurnes: Not sure if it's more or less readable, but these two cases could be combined & enforced with a guard:

```php
if (! in_array($type, ['northstar_id', 'drupal_id']) {
    throw new \Exception('Invalid user ID type provided.');
}

return Reportback::where([
    $type => $userId,
    'campaign_id' => $campaignId,
    'campaign_run_id' => $campaignRunId,
])->firstOrFail();
```

app/Services/ReportbackService.php
```diff
@@ -28,4 +36,37 @@ public function create($data)
...
+     * @param array $data
+     *
+     * @return \Rogue\Models\Reportback $reportback.
+     */
+    public function update($reportback, $data)
+    {
+        $reportback = $this->reportbackRepository->update($reportback, $data);
+
+        return $reportback;
+    }
```
weerd: yeah, the Service here isn't really doing much to help out the Repository... just acting as a middleman. You can probs inject the Repository in the `ReportbackController` along w/ the Service for where that's needed instead only passing the Service and needing duplicate methods like this one to reach the Repository... 

I might be missing something though. Thoughts?

app/Repositories/ReportbackRepository.php
DFurnes: Oh rekt, haha. Disregard my comment below. 😇

database/migrations/2016_06_29_205018_create_oauth_client_table.php
```diff
@@ -3,7 +3,7 @@
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Database\Migrations\Migration;
 
-class CreateOAuthClientTable extends Migration
+class CreateOauthClientTable extends Migration
```
weerd: It took me too long to realize what the change here was... haha ;)

DFurnes:
Left some arguably nitpicky suggestions, but looking good to me! I definitely agree that having a service + repository may be overkill, but doesn't really hurt to keep around for a bit in case. ⛵

weerd:
@DFurnes I like having both, and makes sense to do so I think, but not necessarily have repeated methods where one just passes off responsibility to the other without actually making any changes. If that's the case, just also `use` the repository as well in the controller ;) Hopefully I explained that correctly before 😬

app/Http/Controllers/Api/ReportbackController.php
sbsmith86: makes sense!

app/Http/Controllers/Api/ReportbackController.php
sbsmith86: The idea is that the service would handle all business logic around creating a reportback and the repository handles all database level transactions. 

So for example in this iteration of the project, when a reportback gets created we send it back to phoenix. The service would handle connecting to the phoenix service and doing that request but won't have to worry about all the logic around adding a reportback to the db (which involves updating a few tables). 

Or if you wanted to like, send an event out that a reportback was created, you would do it in the service class.

app/Http/Controllers/Api/ReportbackController.php
sbsmith86: oh that sounds cool.

app/Http/Controllers/Api/ReportbackController.php
weerd: Gotcha! Sounds good. Thanks for clarifying 💃

app/Repositories/ReportbackRepository.php
sbsmith86: yea I can definitely clean this up. But the idea is to eventually resolve everything to use `northstar_id` so i don't think we will even need the distinction in this function. But doesn't hurt to make it better now.

app/Services/ReportbackService.php
sbsmith86: yea right now this is just a middleman, but see  my comment above, we will need this service to handle all the business logic around creating a reportback (i.e, sending it back to phoenix which @chloealee is working on now, or sending events out to the system that something happened, or emails when stuff gets updated, i dunno) Maybe it is over kill at the moment, but I liked the pattern and felt like it allowed me to wrap my head around the separation of concerns.

database/migrations/2016_06_29_205018_create_oauth_client_table.php
sbsmith86: oh yea @DFurnes this was a change i had to make in these migrations that northstar-php made. When I was trying to reset my db I was getting errors that the `CreateOAuthClientTable` didn't exist, even though it did. It was because the file is defined as `create_oauth_client_table` and the casing was off on `oauth` being considered one word. Not sure if this is something you have run into before.

DFurnes: Oh whaaaaa 😮 Yeah, looks like I did that for the one migration and not the other.

sbsmith86:
@DFurnes @weerd yea i totally agree, it is overkill at the moment (and after that talk about code smells maybe this is that one where you develop too much for future stuff) but I really do believe that eventually there will be more business logic around creating and updating reportbacks that we might want a service class to handle.

DFurnes:
Yeah, I think there's a good argument for the separation. Emphasis on "may be" in my comment! 😇

sbsmith86:
@DFurnes @weerd Made some updates, can you take another look?

app/Repositories/ReportbackRepository.php
```diff
@@ -50,4 +89,53 @@ public function create($data)
...
+        }
+
+        $parameters = [
+            $type => $userId,
+            'campaign_id' => $campaignId,
+            'campaign_run_id' => $campaignRunId,
+        ];
+
+        try {
+            return Reportback::where($parameters)->firstOrFail();
```
DFurnes: You could avoid this try/catch by using the `->first()` Eloquent method instead, since that will just return `null` when there isn't a matching record for the query.

tests/ReportbackApiTest.php
```diff
@@ -83,6 +84,10 @@ public function testErrorOnDuplicateReportback()
             'remote_addr' => '207.110.19.130',
         ]);
 
-        $this->assertEquals(500, $response->status());
+        $this->assertEquals(200, $response->status());
```
DFurnes: This is totally subjective, but I kinda like using Laravel's test case methods for things like this:

```php
$this->call('POST', $this->reportbackApiUrl, [ ... ]);

$this->seeJsonSubset([
    'data' => [
        'quantity' => 2000,
    ],
]);
```

DFurnes:
Two little suggestions, but otherwise 👍 from me!

weerd:
¡OLÉ!

PR #31, opened 2016-09-06
Post reportback to phoenix

### LINKED ISSUE #12: POST reportbacks back to Phoenix when they get stored

chloealee:
#### What's this PR do?
- After a reportback is created in Rogue, we immediately `POST` the reportback back to Phoenix. 
- Creates a `failed_logs` table that records all reportbacks that fail to post back to Phoenix. 
- Creates a test to make sure data is getting sent in correct format to Phoenix. 
- Overrides some Northstar-php to get token and cookie to be able to post to Phoenix. 

- All tests should pass when running `phpunit`.
- Running test in Postman like below screen shot should still return reportback:
screen shot 2016-08-25 at 3 12 10 pm

Only thing that I didn't test was to make sure that the reportback is actually in the Phoenix database (for some reason, I can't connect to the database in SequelPro because of SSL connection fail!)

#### Relevant tickets
Fixes #12 

#### Checklist
- [ ] Tested on staging.


app/Models/FailedLog.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Rogue\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
+class FailedLog extends Model
```
DFurnes: Not necessary for this pull request, but I wonder if we should be handling this with [Laravel queues](https://laravel.com/docs/5.3/queues) since they have good support for [handling re-trying failed jobs](https://laravel.com/docs/5.3/queues#dealing-with-failed-jobs) built in.

angaither: not that's ♨️ ness

app/Providers/PhoenixServiceProvider.php
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    /**
+     * Register the application services.
+     *
+     * @return void
+     */
+    public function register()
+    {
+        $this->app->singleton('phoenix', function ($app) {
```
DFurnes: It shouldn't be necessary to make this class into a singleton (since it doesn't have to maintain any state between calls made during a single request), so you can probably remove this service provider & let the service container handle finding and instantiating the class.

app/Services/Phoenix/Phoenix.php
```diff
@@ -13,9 +13,9 @@ class Phoenix extends RestApiClient
      */
     public function __construct()
     {
-        $url = config('services.phoenix.uri') . '/api/' . config('services.phoenix.version') . '/';
+        $this->url = config('services.phoenix.uri') . '/api/' . config('services.phoenix.version') . '/';
```
DFurnes: You don't have to store this on a property, since it gets set as the default base URI for the client. Setting it as a local variable `$url` and passing it along to the RestApiClient's constructor should be enough!

app/Services/Phoenix/Phoenix.php
```diff
@@ -31,15 +31,14 @@ private function authenticate()
                 'password' => config('services.phoenix.password'),
             ];
 
-            $response = $this->post($url . 'auth/login', $payload, false);
-
-            $body = json_decode($response->getBody()->getContents(), true);
-            $session_name = $body['session_name'];
-            $session_value = $body['sessid'];
+            $response = $this->post($this->url . 'auth/login', $payload, false);
```
DFurnes: So each of these calls can be simplified: `$response = $this->post('auth/login', $payload, false);` 😄

app/Services/Phoenix/Phoenix.php
```diff
@@ -31,15 +31,14 @@ private function authenticate()
...
+            // $body = json_decode($response->getBody()->getContents(), true);
```
DFurnes: 👾

app/Services/Phoenix/Phoenix.php
```diff
@@ -69,6 +68,21 @@ private function getAuthenticationCookie()
...
+     * Send a POST request to save a copy of the reportback in Phoenix.
+     *
+     * @param string $nid
+     * @param array $body
+     * @param bool $withAuthorization - Should this request be authorized?
+     * @return object|false
+     */
+    public function postReportback($nid, $body = [])
+    {
+        $response = $this->post($this->url . 'campaigns/' . $nid . '/reportback', $body, $withAuthorization = true);
```
DFurnes: The `$withAuthorization` parameter will default to true, so that can be omitted here.

app/Services/ReportbackService.php
```diff
@@ -34,6 +35,38 @@ public function create($data)
     {
         $reportback = $this->reportbackRepository->create($data);
 
+        // POST reportback back to phoenix.
+        $body = [
+            'uid' => $reportback->drupal_id,
+            'nid' => $reportback->campaign_id,
+            'quantity' => $reportback->quantity,
+            'why_participated' => $reportback->why_participated,
+            'file_url' => $reportback['items'][0]['file_url'],
```
DFurnes: Hmm, since this is an Eloquent model, can this be re-written `$reportback->items()->first()->file_url`? Or maybe these should be computed properties on the Reportback model (that gets each field from the first item)? It'd be nice to avoid working on things as plain old arrays whenever possible, and it feels like it could be a little dangerous just "reaching in" from here.

app/Services/ReportbackService.php
```diff
@@ -34,6 +35,38 @@ public function create($data)
...
+            'caption' => $reportback['items'][0]['caption'],
+            'source' => $reportback['items'][0]['source'],
+        ];
+
+        $phoenixResponse = $this->phoenix->postReportback($reportback->campaign_id, $body);
+
+        // If POST to Phoenix fails, record in failed_logs table.
```
DFurnes: Yeah, let's definitely think about using queued jobs for this. Seems like a perfect fit for what we're doing. 🎫

DFurnes: That'll also make it easy to make these requests asynchronous in the future if we have load issues!

tests/TestCase.php
```diff
@@ -36,6 +36,7 @@ public function setUp()
 
         // Setup mocks
         $this->fileSystem = $this->mock('Illuminate\Filesystem\Filesystem');
+        $this->phoenix = $this->mock('Rogue\Services\Phoenix\Phoenix');
```
DFurnes: Two very nit-picky suggestions:

1. Let's use PHP's `::class` here so it's easy to see whether these resolve or not, like this `$this->mock(\Rogue\Services\Phoenix\Phoenix::class);`
2. The Phoenix class looks like it's "extra" namespaced! Let's remove the [extraneous `\Phoenix` namespace](https://github.com/DoSomething/rogue/blob/0f92aed400f28c9d79a394e1289bbb00604467ad/app/Services/Phoenix/Phoenix.php#L3) from the class so we don't need to type that twice.

app/Services/ReportbackService.php
chloealee: awesome, very cool, @DFurnes ! @angaither should I do this for this PR or in interest of getting this to keep moving, make an issue for this to iterate on this next?

tests/TestCase.php
chloealee: ah yes I don't know why I did that!

chloealee:
@DFurnes ready for another look! I will make a separate issue to update and include Laravel queues!

tests/TestCase.php
```diff
@@ -36,7 +36,7 @@ public function setUp()
 
         // Setup mocks
         $this->fileSystem = $this->mock('Illuminate\Filesystem\Filesystem');
```
DFurnes: Ack sorry one more – can you update this one to use `::class` as well? 😁

chloealee: @DFurnes ah yes, wasn't sure about that - changing now!

DFurnes:
🚢

PR #35, opened 2016-09-12
Queue

### LINKED ISSUE #34: Replace failed_logs table with Laravel queues and use handling for retrying failed jobs
Instead of using our custom `failed_logs` table, replace this with Laravel queues and utilize their handling for retrying failed jobs. 

Rogue will be responsible for knowing about failed requests to Phoenix and then also retrying failed jobs.

chloealee:
#### What's this PR do?
- Refactors to use Laravel queues and error handling.
- Updates test to mock the job that sends reporback from Rogue->Phoenix instead of the Phoenix class.
- Drops the `failed_logs` table.
- Adds the `failed_jobs` table. 

- Make sure the tests are passing by running `phpunit` in the vagrant box.
- To manually test:
  - Log into your local. 
  - Sign up for a campaign. 
  - Go to your "My Account" page. 
  - See that you are signed up for that campaign.
  - Post reportback via postman for that campaign and campaign run, with your `drupal_id`. 
  - Refresh your "My Account" page and make sure the campaign has been moved from "YOU'RE DOING" to "YOU DID"
  Example Postman request:
screen shot 2016-09-12 at 3 06 15 pm

Question -  @sbsmith86 when we are updating the reportback, do we want to dispatch the job to send this to Phoenix as well?

Note from Shae for future planning: 
Shae Smith [11:27 AM]  
I don’t know if this is needed at the moment, but we could also use the event system in laravel to send stuff to queues since we might want a way of notifiying a bunch of systems that something happened in rogue. but I think this approach is fine for now.
https://dosomething.slack.com/archives/rogue/p1473348457000138

#### Relevant tickets
Fixes #34 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Jobs/SendReportbackToPhoenix.php
+++ b/app/Jobs/SendReportbackToPhoenix.php
@@ -7,24 +7,22 @@ use Illuminate\Queue\SerializesModels;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Rogue\Services\Phoenix;
+use Rogue\Models\Reportback;
 
 class SendReportbackToPhoenix extends Job implements ShouldQueue
 {
     use InteractsWithQueue, SerializesModels;
 
-    protected $nid;
-    protected $body;
-
+    protected $reportback;
     /**
      * Create a new job instance.
      *
      * @return void
      */
-    public function __construct($nid, $body)
+    public function __construct(Reportback $reportback)
     {
         $this->phoenix = new Phoenix;
-        $this->nid = $nid;
-        $this->body = $body;
+        $this->reportback = $reportback;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


sbsmith86:
@chloealee it depends on how #6972 is implemented. If you save the update to drupal first and then send it to rogue then we don't need to send it back. And, I sorta feel like that flow is ok for now, since we don't want to add more work to this project.

chloealee:
@sbsmith86 cool, makes sense!

sbsmith86:
The code looks fine to me, also.

chloealee:
Great! Can I 🚢 ?

sbsmith86:
sure!

app/Services/ReportbackService.php
```diff
@@ -14,14 +14,14 @@ class ReportbackService
...
      */
-    protected $phoenix;
+    protected $sendReportbackToPhoenix;
 
-    public function __construct(ReportbackRepository $reportbackRepository, Phoenix $phoenix)
+    public function __construct(ReportbackRepository $reportbackRepository, SendReportbackToPhoenix $sendReportbackToPhoenix)
     {
         $this->reportbackRepository = $reportbackRepository;
-        $this->phoenix = $phoenix;
+        $this->sendReportbackToPhoenix = $sendReportbackToPhoenix;
```
DFurnes: Just got around to looking at this now, so sorry to comment after merge. 😁 It looks like this injected class and property aren't used anywhere? Dispatching by creating a new instance inline (like you did below) is pretty standard convention so this should be safe to remove.

app/Jobs/SendReportbackToPhoenix.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $phoenix = new Phoenix;
```
DFurnes: Definitely not necessary, but it may be cleaner to use the `$this->phoenix` that you created above rather than instantiating a second client as a local variable (or just remove the property above!).

app/Services/ReportbackService.php
chloealee: ah shoot yes, I forgot to delete that, thanks @DFurnes ! I'll make the changes and submit a new PR!

PR #40, opened 2016-09-20
Update rb items endpoint

### LINKED ISSUE #39: Create /items endpoint to update reportbackitems
In order for Rogue to update reportback items' `status` when they are reviewed in Phoenix, we need to have an endpoint to update a collection of reportback items. 

Refs: https://github.com/DoSomething/phoenix/issues/6972

chloealee:
#### What's this PR do?
- Adds `PUT /items` endpoint to update the status (and more in the future) of reportback items.
- Adds a test to make sure this endpoint is working correctly and updating reportback item(s) status correctly. 

- Run `phpunit` and all tests should pass. 
- To test manually, pick a reportback item or multiple reportback items in the database you want to update and send a request similar to the following:
screen shot 2016-09-20 at 2 49 16 pm
 Then check the database to make sure the status has changed to 'approved'

- When testing manually, if sending more than one reportback item to update, request data must be entered in x-ww-form-urlencoded or raw section. 
- **NOTE** this update assumes that the reportback item is already in Rogue. 

#### Relevant tickets
Fixes #39 

Refs https://github.com/DoSomething/phoenix/issues/6972 and is the Rogue side of the work to close this issue. 

#### Checklist
- [ ] Tested on staging.


app/Http/Requests/ReportbackItemRequest.php
```diff
@@ -0,0 +1,34 @@
...
+        // return [
+        //     'rogue_reportback_item_id' => 'required|int',
+        //     'quantity' => 'int',
+        //     'why_participated' => 'string',
+        //     'caption' => 'string',
+        //     'source' => 'string',
+        //     'status' => 'string',
+        // ];
+    }
+}
```
chloealee: @sbsmith86 do we need to write validation rules for a collection?

chloealee:
Another thing I just thought of! Do we already have a gate so only admins can send requests to Rogue's endpoints?

app/Http/Requests/ReportbackItemRequest.php
sbsmith86: I'm not sure how it works with collections, but if you don't validate the request here, you can/should do checks that the things you are expecting to exists in the request do, before using them.

app/Repositories/ReportbackRepository.php
```diff
@@ -134,4 +135,20 @@ public function getReportbackByUser($campaignId, $campaignRunId, $userId, $type
...
+     * Updates an individual reportbackitem or many reportback items.
+     *
+     * @param array $data
+     *
+     * @return
+     */
+    public function updateReportbackItems($data)
+    {
+        foreach ($data as $reportbackItem) {
+            $rbItem = ReportbackItem::where(['id' => $reportbackItem['rogue_reportback_item_id']])->first();
```
sbsmith86: For example (from my comment above), YOu should check that `$reportbackItem['rogue_reportback_item_id']` exists and is not empty here as a safety measure.

sbsmith86: And below you should check for `$reportbackItem['status']`

app/Repositories/ReportbackRepository.php
```diff
@@ -134,4 +135,20 @@ public function getReportbackByUser($campaignId, $campaignRunId, $userId, $type
...
+            $rbItem->status = $reportbackItem['status'];
+            $rbItem->save();
+        }
+    }
```
sbsmith86: This function should return something, either the updated item or null if something failed (i.e those properties above don't exist in the request)

app/Services/ReportbackService.php
```diff
@@ -67,4 +67,16 @@ public function getReportback($campaignId, $campaignRunId, $userId, $type)
...
+    /*
+     * Handles all business logic around update a reportbackitem(s).
+     *
+     * @param array $data
+     *
+     * @return
+     */
+    public function updateReportbackItems($data)
+    {
+        $this->reportbackRepository->updateReportbackItems($data);
```
sbsmith86: This would then receive the return value (which you will set from my comment above) and set that to a variable that you pass to the controller like

```
$item = $this->reportbackRepository->updateReportbackItems($data);
return $item
```

tests/ReportbackApiTest.php
```diff
@@ -99,4 +101,38 @@ public function testUpdatingReportback()
...
+    {
+        // Create a reportback item and save to the reportback_items table.
+        $reportback = factory(Reportback::class)->create();
+        $reportbackItem = new ReportbackItem;
+        $reportbackItem->reportback_id = $reportback->id;
+        $reportbackItem->id = $this->faker->randomNumber(4);
+        $reportbackItem->status = 'pending';
+        $reportbackItem->save();
+
+        // dd($reportbackItem->id);
```
sbsmith86: can remove this comment ;)

app/Repositories/ReportbackRepository.php
chloealee: oh right, thank you!

app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -52,4 +54,15 @@ public function store(ReportbackRequest $request)
...
+
+    /**
+     * Update reportbackitem(s) that already exists.
+     *
+     * @param \Illuminate\Http\Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function updateReportbackItems(Request $request)
+    {
+        $this->reportbackService->updateReportbackItems($request->all());
```
sbsmith86: (sorry for the out of order comments) This should receive a return value from `updateReportbackItems()` and then it should use a Transformer to format the response to the client. 

A Transformer is a class that allows you to format the response that an API returns to a client using the API in a way that makes it more usable. For example, when returning a date from the database, you might want that date to show in the API response formatted in a readable way. I am sorta confused about how this endpoint is retuning any formatted response but it seems like it is? The reason why I suggest using a Transformer is because it is best practice and allows use to follow the separation of concerns pattern.

I already created a `ReportbackItemTransformer` [class](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/ReportbackItemTransformer.php) that I think you can use here. This class basically just defines what the API response should look like from the rogue api when returning `ReportbackItem` models.

I think the steps for doing this should be pretty straight forward: 

In the `ReportbackController` class you pretty much follow the same thing that is being done for reportbacks except: 

1) Create a new `protected` variable like `$itemTransformer` 
2) Set that in the constructor
3) add a `return` statement in this function that does something like `return $this->item($item, $code, $itemTransformer);`
4) when you are returning a collection of items you will probably have to do something like `return $this->collection($items, $code, $itemTransformer);

I hope that makes sense! I can swing by if you have any questions.

chloealee:
@sbsmith86 ah okay good to know! I'll flag to get on Jen's radar then!

app/Http/Controllers/Api/ReportbackController.php
chloealee: Thank you so much for providing such detailed instructions and also providing context/explanation behind this! Totally makes sense to do this - I wasn't exactly sure what to return - if it should be a message "Updated" or the item and was going to come back to it but totally forgot to! You're right to return the object so there is proof that the reportback item has been updated.

I followed your instructions and for step three added: 
```
        $item = $this->reportbackService->updateReportbackItems($request->all());

        return $this->item($item, 201, $this->itemTransformer);
```

One error that I'm getting right now: 
```
ErrorException in ResourceAbstract.php line 145:
Argument 1 passed to League\Fractal\Resource\ResourceAbstract::setMeta() must be of the type array, object given, called in /home/vagrant/DoSomething.org/rogue/app/Http/Controllers/Api/ApiController.php on line 65 and defined
```

sbsmith86: @chloealee sorry! My instructions above are just for the what is done in your method in the Controller. The Service and Repository should just return the instance of the item or items  that was updated and maybe `null` if something went wrong.

chloealee: @sbsmith86 ah yes sorry - I changed everything in the Service and Repository as well to return the `$item` or `null` if there was an error! Just added this to the code in the Controller but still getting the same response as above? 
```
    public function updateReportbackItems(Request $request)
    {
        $item = $this->reportbackService->updateReportbackItems($request->all());

        if ($item === null ) {
            $code = 404;
        } else {
            $code = 201;
        }

        return $this->item($item, $code, $this->itemTransformer);
    }
```

sbsmith86: @chloealee in this instance is `$item` a collection? (in which case you need to do something like in step 4) You would probably have to do a check like::

```
Pseduo code:

if $item is an instance of Item {
 return $this->item($item, $code, $this->itemTransformer);
} else {
return $this->collection($item, $code, $this->itemTransformer);
}
```

chloealee: Thanks, @sbsmith86! Added that logic in but still getting the same error message. It's also strange because when I do `dd(gettype($items)` it says it is an array (which is what it is looking for)?

chloealee: oh wait, I was looking at the wrong thing! I feel like the error message is saying something that I don't want to mess with (`League\Fractal\Resource\ResourceAbstract::setMeta()`)?

app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -52,4 +56,31 @@ public function store(ReportbackRequest $request)
...
+            $code = 404;
+        } else {
+            $code = 201;
+        }
+
+        $meta = [];
+
+        if (count($items) === 1) {
+            $item = $items[0];
+
```
chloealee: @sbsmith86 is this ok to do? Looks a little strange but wasn't sure how to do a better way! Otherwise if I did a check in `ReportbackRepository.php` in the `updateReportbackItems` function, it'd be a lot of repeat code :-/

chloealee:
@sbsmith86 another note - this code assumes that the reportback item is already in Rogue (since we'll be migrating all existing reportbacks/items from Phoenix to Rogue when we go live?). Just want to double check to make sure this is okay as is too!

sbsmith86:
@chloealee ok yea, I mean by the time you are in the reviewing stage that item should definitely be created. 

looks good to me 👍  great work!

sbsmith86:
@chloealee oh wait there was one other comment I made about the `item` vs `collection code.

chloealee:
thanks @sbsmith86 ! Do you mean the comment about `item` vs `collection` to the controller [here](https://github.com/DoSomething/rogue/pull/40/files#diff-8176f4ab3f170558ede2abda4c6bbf7eR81)?

sbsmith86:
@chloealee yep

chloealee:
@sbsmith86 I think I made that change already? Unless I'm missing another comment?

sbsmith86:
@chloealee you did [this](https://github.com/DoSomething/rogue/pull/40/files#r79716606) already? the code looks the same.

chloealee:
@sbsmith86 I don't think the link directed to a comment! Can you re-link?

sbsmith86:
@chloealee https://github.com/DoSomething/rogue/pull/40/files#r79716606

chloealee:
@sbsmith86 sorry for all the back and forth - I'm just really confused! All I'm seeing is this:
screen shot 2016-09-21 at 3 17 16 pm


Not sure what you want me to change?

app/Http/Controllers/Api/ReportbackController.php
sbsmith86: @chloealee I don't know what is going on, what I wrote here disappeared.

I was just saying that we can simplify this by always returning it as a collection cause it looks like the repository always returns an array of reportback items even if it is only one. We can opt to change this later.

chloealee:
@sbsmith86 updated!

sbsmith86:
Great 👍 💯

PR #41, opened 2016-09-28
tests for submitting reportbacks with emojis

katiecrane:
#### What's this PR do?
Added 4 new tests!
- `testPostingReportbackWithNormalOldEmoji`
- `testPostingReportbackWithNormalNewEmoji`
- `testPostingReportbackWithEncodedOldEmoji`
- `testPostingReportbackWithEncodedNewEmoji`

Also added some helper methods in `TestCase.php` in order to condense the tests that exist and make writing future ones easier. These were incorporated into the tests that already existed as well.

Make sure I did the encoded emoji correctly please! I followed the same style to test submitting a reportback as happens in `testCreatingNewReportback` which I think should be what we want here.
Also make sure when I condensed the existing tests I didn't inadvertently change what they are testing.

#### Relevant tickets
no issue created as of now

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/tests/ReportbackApiTest.php
+++ b/tests/ReportbackApiTest.php
@@ -43,25 +43,16 @@ class ReportbackApiTest extends TestCase
      */
     public function testUpdatingReportback()
     {
+        // Create a reportback so that one exists
         $reportback = factory(Reportback::class)->create();
 
-        $response = $this->call('POST', $this->reportbackApiUrl, [
-            'northstar_id' => $reportback->northstar_id,
-            'drupal_id' => $reportback->drupal_id,
-            'campaign_id' => $reportback->campaign_id,
-            'campaign_run_id' => $reportback->campaign_run_id,
-            // Change quanitity.
-            'quantity' => 2000,
-            'why_participated' => $this->faker->paragraph(3),
-            'num_participated' => null,
-            'file_id' => $this->faker->randomNumber(4),
-            'caption' => $this->faker->sentence(),
-            'source' => 'runscope',
-            'remote_addr' => '207.110.19.130',
-        ]);
+        // Change the quantity
+        $reportback->quantity = 2000;
 
-        $this->assertResponseStatus(201);
+        // Post the update
+        $this->updateReportback($reportback->toArray());
 
+        // Make sure we see the update in the response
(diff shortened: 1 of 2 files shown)


tests/ReportbackApiTest.php
```diff
@@ -134,4 +134,204 @@ public function testUpdatingReportbackItem()
...
+
+    /**
+     * Test posting a reportback item containing an unencoded old emoji
+     *
+     * @return void
+     */
+    public function testPostingReportbackWithNormalOldEmoji()
+    {
+        // Mock sending image to AWS.
+        $this->fileSystem->shouldReceive('put')->andReturn(true);
```
katiecrane: The lines like this will probably have to be changed once #42 is approved. I figure it makes sense to get approval on that first so I only have to change these ones once.

sbsmith86:
I am smelling a code smell around these reportback tests. Seems like a lot of repeated code, with the only things changing being what we are sending in the content of the reportback. I wonder if we could extract some of the repeated bits into a method on `TestCast.php` and just pass in an object with the content of the reportback in all of the tests. I feel like I can see instances in the future of use wanting to test different content in a reportback and having to repeat this a lot. Thoughts?

katiecrane:
@sbsmith86 Yeah I completely agree with that. I can try to pull out as much as possible into helper functions, I agree that this will/should come up a lot in the future!

katiecrane:
@sbsmith86 I pulled out some functionality into `TestCase.php` and started by condensing one test, `testPostingReportbackWithEncodedNewEmoji` just to avoid making the same fixes 4 times. 

I broke things into 3 functions in `TestCase.php` for flexibility. I thought about combining the last 2 fucntions, but the second one is specific to posting a NEW reportback, so I thought it would be better to keep it separate.

sbsmith86:
Looks good. I think we will want those other tests to be updated with the new functions you created, though right? If it's on the radar to fix, then merge away! Otherwise, I would just take the time to do it now.

katiecrane:
@sbsmith86 yup! just wanted to do it in one place first for review so I can do the rest all in one go 😄  update to come!

sbsmith86:
Nice! 👍

PR #42, opened 2016-09-28
use storage facade to mock sending files to aws

katiecrane:
#### What's this PR do?
We now use the `Storage` facade to interact with AWS so we can use more laravel 🔮  and mock the call to put the uploaded files on AWS. Doing this also allowed us to delete some things.

I think the best way to test is to first try to post a reportback and make sure your file still goes to AWS. Then, remove all the aws/s3 credentials from your `.env` and make sure the tests pass. This insures that the mock is working as expected. Don't forget to add your credentials back in.

#### Relevant tickets
no tickets created

#### Checklist
- [ ] Tested on staging.

cc: @chloealee and shoutout to @DFurnes for his Otwellian Magic

### DIFF
--- a/tests/ReportbackApiTest.php
+++ b/tests/ReportbackApiTest.php
@@ -27,7 +27,8 @@ class ReportbackApiTest extends TestCase
     public function testCreatingNewReportback()
     {
         // Mock sending image to AWS.
-        $this->fileSystem->shouldReceive('put')->andReturn(true);
+        Storage::shouldReceive('put')
+                    ->andReturn(true);
 
         // Mock job that sends reportback back to Phoenix.
         $this->expectsJobs(Rogue\Jobs\SendReportbackToPhoenix::class);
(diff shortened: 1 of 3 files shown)


tests/ReportbackApiTest.php
```diff
@@ -43,11 +44,11 @@ public function testCreatingNewReportback()
             'quantity'         => $this->faker->numberBetween(10, 1000),
             'why_participated' => $this->faker->paragraph(3),
             'num_participants' => null,
-            'file_id' => $this->faker->randomNumber(4),
-            'caption' => $this->faker->sentence(),
-            'source' => 'runscope',
-            'remote_addr' => '207.110.19.130',
-            'file' => $file,
+            'file_id'          => $this->faker->randomNumber(4),
```
sbsmith86: Is StyleCI ok with this spacing? I think I get errors when I try to align arrow notation.

sbsmith86:
👍

tests/ReportbackApiTest.php
katiecrane: The checks are passing so I guess it's okay? I historically have had lots of issues getting StyleCI to be happy with me

PR #43, opened 2016-09-29
Sends updated reportback and new reportback items to phoenix

### LINKED ISSUE #44: Send new and updated RB items from Rogue to Phoenix
We want new and updated RBs to be sent from Rogue to Phoenix so we're accounting for all RBs!

chloealee:
#### What's this PR do?
- Sends updated reportback and new reportback items to Phoenix 
- Fixes bug that takes information from first reportback item instead of reportback item that was most recently created
- Updates comment and test function name to add more specificity. 

- In Postman, create a new reportback. Make sure this is in both Rogue and Phoenix. 
- Create a new reportback item. Make sure this is also in both Rogue and Phoenix. 
- Update the `why_participated` field and make sure this is reflected in Rogue and Phoenix. 

#### Relevant tickets
Fixes #44

With help from @katiecrane ! 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Jobs/SendReportbackToPhoenix.php
+++ b/app/Jobs/SendReportbackToPhoenix.php
@@ -32,6 +32,7 @@ class SendReportbackToPhoenix extends Job implements ShouldQueue
     public function handle()
     {
         $phoenix = new Phoenix;
+
         $reportbackItem = $this->reportback->items()->orderBy('created_at', 'desc')->first();
 
         $body = [
--- a/app/Jobs/SendReportbackToPhoenix.php
+++ b/app/Jobs/SendReportbackToPhoenix.php
@@ -32,15 +32,16 @@ class SendReportbackToPhoenix extends Job implements ShouldQueue
     public function handle()
     {
         $phoenix = new Phoenix;
+        $reportbackItem = $this->reportback->items()->orderBy('created_at', 'desc')->first();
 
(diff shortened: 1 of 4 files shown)


app/Jobs/SendReportbackToPhoenix.php
```diff
@@ -33,14 +33,16 @@ public function handle()
     {
         $phoenix = new Phoenix;
 
+        $reportbackItem = $this->reportback->items()->orderBy('created_at', 'desc')->first();
```
sbsmith86: Why do we only want to send one reportback item back to phoenix?

chloealee: This is for sending a new reportback item back to Phoenix. Since you can only create on reportback at a time, every time a new reportback item is created, it will send back to Phoenix from Rogue. This code makes sure to take the most recently created (before it was taking the first reportback item created for that reportback and sending that information instead of the one that was just made).

sbsmith86: I guess I'm wondering if there would ever be an instance when we would want all items returned so phoenix could do whatever it wants with all the items after something has been updated. But that is not a blocker for merge, we could always figure that out if it comes up.

sbsmith86:
👍

PR #48, opened 2016-10-11
Remove file_id column

### LINKED ISSUE #33: Remove file_id from reportback_items table
We do not need to store this drupal `fid` anymore because rogue does it's own file storage and phoenix will be able to send requests to rogue that target items based on their `id` in the `reportback_items` table.

sbsmith86:
#### What's this PR do?

* Removes references to `file_id` in rogue which was referencing the `fid` as it was stored in drupal. We do not need to store this anymore since rogue does it's own file storage. 
* Updates the previously defined `Reaction` to `Reportback` many-to-many relationship to be a `Reaction` to `ReportbackItem` relationship since reactions actually live on items.
* Added a step to the `ReportbackItem` seeder that will give the item a reaction so we have some data to play with.

pull down this branch, reset, re-migrate and re-seed your database.

`php artisan migrate:refresh --seed` 

all should run well, and you should not see any `file_id` columns in your db.

You can also try out the relationship by grabbing a Reaction from the database and running something like: 

`dd($reaction->reportackItems);` 

You should see all the items that have that reaction.

#### Relevant tickets
Fixes #33

### DIFF
--- a/app/Http/Transformers/ReportbackItemTransformer.php
+++ b/app/Http/Transformers/ReportbackItemTransformer.php
@@ -19,7 +19,6 @@ class ReportbackItemTransformer extends TransformerAbstract
             'id' => (string) $reportbackItem->id,
             'reportback_id' => $reportbackItem->reportback_id,
             'media' => [
-                'id' => $reportbackItem->file_id,
                 'url' => $reportbackItem->file_url,
             ],
             'caption' => $reportbackItem->caption,
--- a/app/Models/Reaction.php
+++ b/app/Models/Reaction.php
@@ -11,13 +11,13 @@ class Reaction extends Model
      *
      * @var array
      */
-    protected $fillable = ['id', 'northstar_id', 'drupal_id', 'file_id', 'taxonomy_id'];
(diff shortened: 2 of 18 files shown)


chloealee:
@sbsmith86 I think this line will break now that we don't have `file_id` https://github.com/DoSomething/rogue/blob/master/app/Repositories/ReportbackRepository.php#L106

sbsmith86:
@chloealee ya good catch!

weerd:
Apart from Chloe's comment that needs to be updated, looks good! 👍 

Maybe do a quick full repo search for `file_id` to make sure there aren't any stragglers!

sbsmith86:
@chloealee ya i had missed a few things! Thanks for point it out.

* I updated aws to create filenames that follow the pattern `{campaign_id}-129389123.jpg`
* Removed references to `file_id` in the tests (I also had to do some clean up there, we were not mocking the job to send to phoenix so it was actually trying to hit phoenix)
* Removed references to `file_id` in the documentation.

chloealee:
this is awesome @sbsmith86 !

PR #49, opened 2016-10-14
issue_template.md

sbsmith86:
#### What's this PR do?
Adds the issue template from phoenix so we can write better issues. We can always update this, but I find it helpful.


katiecrane:
looks familiar! 👍 🎉

PR #53, opened 2016-10-18
User authentication

### LINKED ISSUE #50: User authentication
_FEATURE OVERVIEW_
-------------------
#### User Story

As an admin, I want to be be able to log into rogue so that I can review reportbacks. 

We need to create a log in interface that admins can use to log in to rogue via gateway. It sounds like this is now done by sending the user to northstar where they log in and then get redirected back to rogue as an authenticated user. 

We need to figure out a way to only allow admins to be authenticated on rogue, not just any user.

sbsmith86:
#### What's this PR do?

Adds login functionality to Rogue through Gateway which authenticates users. Users that login are stored in the `users` table with a role. I set up so middleware that looks for `admin` and `staff` roles on routes.

* Update the user table to remove unused columns since we are authenticating in Northstar.
* Note: I left the `id` field in the user table because Gateway was having issues creating a user without an `id` column on the `users` table. This seems like a Laravel requirement even though all we care about is `northstar_id`
* Implements Gateway authentication in Rogue
* Defines middleware

if you pull down rogue and either have it point to northstar-qa you should be able to go through the login flow. 

1) I added some tests but I am having issues with how to test this stuff. Trying to return and instance of the PSR-7 Redirect Response gives me this error when trying to run `$this->visit()`. Happy to discuss further if anyone has any other ideas.
```
Error: Call to undefined method Symfony\Component\HttpFoundation\Response::getTargetUrl()
```

2) I set this up testing my dev with northstar-qa. I'm not sure if that matters or not, but if using a dev instance of northstar, I am guessing you will have to set up your dev as client for your local northstar.

#### Relevant tickets
Fixes #50 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -31,8 +31,7 @@ class AuthController extends Controller
      *
      * @var string
      */
-    protected $redirectTo = '/';
-
+    protected $redirectTo = '/reportbacks';
 
     /**
      * Where to redirect users after logout.
@@ -40,17 +39,6 @@ class AuthController extends Controller
      * @var string
      */
     protected $redirectAfterLogout = '/';
-
-
-    /**
-     * Create a new authentication controller instance.
-     *
-     * @return void
-     */
-    public function __construct()
-    {
-        // $this->middleware($this->guestMiddleware(), ['except' => 'logout']);
-    }
 
     /**
      * Handle a login request to the application.
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -3,8 +3,6 @@
 namespace Rogue\Http\Controllers\Auth;
 
 use Rogue\Http\Controllers\Controller;
-use Illuminate\Foundation\Auth\ThrottlesLogins;
-use Illuminate\Foundation\Auth\AuthenticatesAndRegistersUsers;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
@@ -20,8 +18,6 @@ class AuthController extends Controller
     | a simple trait to add these behaviors. Why don't you explore it?
     |
     */
-
-    use AuthenticatesAndRegistersUsers, ThrottlesLogins;
 
     /**
      * Where to redirect users after login / registration.
--- a/app/Http/Controllers/Auth/AuthController.php
+++ b/app/Http/Controllers/Auth/AuthController.php
@@ -2,14 +2,11 @@
 
 namespace Rogue\Http\Controllers\Auth;
 
-use Rogue\User;
-use Validator;
 use Rogue\Http\Controllers\Controller;
 use Illuminate\Foundation\Auth\ThrottlesLogins;
 use Illuminate\Foundation\Auth\AuthenticatesAndRegistersUsers;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
-
 
 class AuthController extends Controller
 {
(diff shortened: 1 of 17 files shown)


app/Http/Controllers/Auth/AuthController.php
```diff
@@ -2,11 +2,11 @@
 
 namespace Rogue\Http\Controllers\Auth;
 
-use Rogue\User;
-use Validator;
 use Rogue\Http\Controllers\Controller;
 use Illuminate\Foundation\Auth\ThrottlesLogins;
 use Illuminate\Foundation\Auth\AuthenticatesAndRegistersUsers;
```
DFurnes: You can remove these two traits (`ThrottlesLogins`, `AuthenticatesAndRegistersUsers`, and the corresponding `use` statements below in the class). They don't do anything with the Northstar login methods.

app/Http/Kernel.php
```diff
@@ -23,14 +28,6 @@ class Kernel extends HttpKernel
      * @var array
      */
     protected $middlewareGroups = [
-        'web' => [
```
DFurnes: What's the reasoning for moving these into the global middleware stack? I think it makes sense to only apply these on web routes (since API routes don't need sessions, cookies, or CSRF tokens).

app/Models/User.php
```diff
@@ -2,36 +2,27 @@
...
 
-class User extends Authenticatable
+class User extends Model implements AuthenticatableContract, NorthstarUserContract
 {
-    /**
-     * The attributes that are mass assignable.
-     *
-     * @var array
-     */
-    protected $fillable = [
```
DFurnes: I think it makes sense to keep this property defined as an empty array to protect against mass-assignment attacks if we ever exposed this model to form input somehow.

resources/views/layouts/nav.blade.php
```diff
@@ -0,0 +1,29 @@
+<nav class="navigation -white -floating">
+    <a class="navigation__logo" href="/"><span>DoSomething.org</span></a>
+    <div class="navigation__menu">
+        @if (Auth::user())
+            <ul class="navigation__primary">
+                <li>
+                    <a href="#">
+                        <strong class="navigation__title">One Fish</strong>
```
DFurnes: 🐟

tests/UserTest.php
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Test that northstar authorization is called when hiting the /login route
+     *
+     * @return void
+     */
+    public function testLogin()
+    {
+        $mock = $this->mock(Northstar::class)
+            ->shouldReceive('authorize')
+            ->andReturn(302);
```
DFurnes: I think it's safe to remove the `->andReturn(302);` and just not mock a return value.

DFurnes:
> 1) I added some tests but I am having issues with how to test this stuff. Trying to return and instance of the PSR-7 Redirect Response gives me this error when trying to run $this->visit(). Happy to discuss further if anyone has any other ideas.

Ah yep, I've run into this before. Laravel's testing utilities expect all controller responses to be Laravel response objects, so it's trying to call a Laravel method even though it's a PSR-7 response. I think it's safe to just not do any inspection of the responses here.

> 2) I set this up testing my dev with northstar-qa. I'm not sure if that matters or not, but if using a dev instance of northstar, I am guessing you will have to set up your dev as client for your local northstar.

We have a seeded `trusted-test-client` client in Northstar that can be used for local dev so you don't have to manually create a client for every local app (see [Aurora's `.env`](https://github.com/DoSomething/aurora/blob/ce6c9fb1f528a16dc7d569cae0c8ed0e5154a9ec/.env.example#L15-L16), for example).

app/Http/Kernel.php
sbsmith86: That makes sense!

sbsmith86:
@DFurnes Thanks for your comments! Made those updates. Also thanks for letting me know about the `trusted-test-client` setting. Very cool.

PR #54, opened 2016-10-21
Reactions endpoint

### LINKED ISSUE #52: Reactions endpoint
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want a way to like/unlike reportback items in Rogue via an endpoint that records this information. 

Steps: 
1. Remove `taxonomy_id` and `drupal_id` from `reactions` table. 
2. Add `reportback_item_id` to `reactions` table.
3. Create an endpoint that mimics current reactions endpoint to record `northstar_id` and `reportback_item_id` of which reportback items have been 'liked.' 

#### Why This Matters
Required in order to support reactions in Rogue.

chloealee:
#### What's this PR do?
- Deletes `reaction_reportback_item` pivot table.
- Updates relation between `reportback_item` and `reaction`.
- Deletes `taxonomy_id` and `drupal_id` from `reactions` table.
- Adds `reportback_item_id` to `reactions` table.
- Creates `/reactions` endpoint that uses Laravel's `SoftDelete` function. This endpoint works as follows:
  - When the endpoint is hit, if the `reportback_item` does not have a `reaction` from the particular user with the `northstar_id`, it creates a new record of the reaction in the `reactions` table, saving the user's `northstar_id` and the `reportback_item_id`. 
  - If there is a record of the reaction for this `reportback_item` from this user and the `deleted_at` column is `null`, the reaction gets soft deleted and a timestamp gets saved to the `deleted_at` column and `updated_at` column.
  - If there is a record of the reaction from this user and the `deleted_at` column has a timestamp, the reaction gets restored - the timestamp from the `deleted_at` column gets removed and `null` is saved to that value. A new timestamp gets added to the `updated_at` column. 
  - This means that we won't have a full history of how many times a user liked/unliked a reportback item but we can see if a user likes a reportback item and then unlikes it. 
- Creates `ReactionTransformer.php` that returns the `Reaction` object and an `action` status in metadata if the `reportback_item` was `liked` or `unliked`.

- In Postman, hit the `/reactions` endpoint with a `northstar_id` and `reportback_item_id`. 
- You should see a new record created in the `reactions` table. 
- Hit the `/reactions` endpoint again and you should see the timestamp in the `updated_at` and `deleted_at` columns. 
- Hit the endpoint once more and you should see an updated timestamp in the `updated_at` column and `null` in the `deleted_at` column. 
- To test the validation rules, send request without `northstar_id` and/or `reportback_item_id`. Also make sure you have the header `Accept` => `application/json`. You should get an error like this: `{"northstar_id":["The northstar id field is required."]}`

#### Relevant tickets
Fixes #52 

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/Api/ReactionController.php
```diff
@@ -0,0 +1,43 @@
...
+        } else {
+            // If the reportback_item is "liked" by this user, soft delete the reaction. Otherwise, restore the reaction.
+            if (is_null($existingReaction->deleted_at)) {
+                $existingReaction->delete();
+            } else {
+                $existingReaction->restore();
+            }
+        }
+
+        // TODO: When we build out Phoenix and decide on what we want in the response, we will need to add a Reaction Transformer.
```
sbsmith86: The reason I ask that is that we could just send back the entire reaction model to whatever app is requesting it with some metadata about the action that was performed (deleted, created, etc). That way the client can decide what it wants to do with that information on it's own. 

Is there a specific concern about what will be needed on the phoenix end that needs to be talked through?

chloealee: @sbsmith86 yes we can totally do that now! Just wasn't sure what info. we'd want to return in the response. Working on that now :)

chloealee: @sbsmith86 would the meta data look something like the below?
```
{
    "metadata" :{
        "action": "unliked",
    },
    "data" :{
        "reaction_id": 1,
        "northstar_id": 1001,
        "reportback_item_id": 356,
        ...
    }
}
```

sbsmith86: Ya. I think Fractal just uses the key `meta` but that is the basic structure

chloealee: cool, thanks @sbsmith86 ! Ready for another 👀 !

app/Http/Controllers/Api/ReactionController.php
```diff
@@ -0,0 +1,66 @@
...
+            $code = 200;
+            $action = 'liked';
+        } else {
+            // If the reportback_item was previously "liked" by this user, soft delete the reaction. Otherwise, restore the reaction.
+            if (is_null($reaction->deleted_at)) {
+                $code = 201;
+                $action = 'unliked';
+                $reaction->delete();
+            } else {
+                $code = 201;
```
chloealee: ah great point - just changed!

app/Http/Transformers/ReactionTransformer.php
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Transform the resource data.
+     *
+     * @param \Rogue\Models\Reaction $reaction
+     * @return array
+     */
+    public function transform(Reaction $reaction)
+    {
+        return [
+            'reaction_id' => $reaction->id,
```
sbsmith86: I think we want to cast `reaction_id` and `reportback_item_id` as `strings` in the transformer. Since the id represents a record not something you would do math on, so it isn't really an `int`. Same thing we did in the transformer [here](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/ReportbackTransformer.php#L28)

app/Http/routes.php
```diff
@@ -23,5 +23,6 @@
     Route::get('reportbacks', 'Api\ReportbackController@index');
     Route::post('reportbacks', 'Api\ReportbackController@store');
     Route::put('items', 'Api\ReportbackController@updateReportbackItems');
+    Route::post('reactions', 'Api\ReactionController@store');
```
sbsmith86: Not necessary for merge but we should organize these endpoints a little better  in this file. Something to keep in mind as we build out the API

app/Http/Transformers/ReactionTransformer.php
chloealee: Got it, makes sense! Updated :)

app/Http/routes.php
chloealee: Yes, agreed!

sbsmith86:
Looks good! Great work!

chloealee:
Thanks, Shae!!

PR #62, opened 2016-11-01
Image crop & rotate

### LINKED ISSUE #61: Crop and Rotate images
_FEATURE OVERVIEW_
-------------------
#### User Story

When a reportback is sent to rogue we should store the original image file and also crop, rotate and store the new file separately. We the need to send that edited image back to phoenix so that nothing breaks on the phoenix side which is expecting cropped images.

#### Why This Matters

The gallery layout is currently based on square images and we don't want to break the gallery on the user facing side.

sbsmith86:
#### What's this PR do?

* Stores the original image in S3 and also crops and rotates it and stores that new image in a new `edited_file_url` column on the item so we can have both. 

* Exposes both file urls in the response like:

image

* Sends the edited file back to phoenix.

I tried to make the commits clear, so you should be able to review commit by commit. 

Run the unit tests!

Post a reportback (or new item) to rogue with cropping properites (`crop_x`, `crop_y`, `crop_height`, `crop_width`, `crop_rotate`). The orginal file should get stored in `file_url` and the cropped file should be stored in `edited_file_url`

Started a helper file to start extracting some helper functions. I put the image cropping stuff in there cause it didn't feel like anything should really own that. But open to other thoughts on that. 

Also when you pull down this branch you might need to run `composer dump-autoload` to make sure that helper class is pulled in. 

#### Relevant tickets
Fixes #61

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/ReportbackRepository.php
+++ b/app/Repositories/ReportbackRepository.php
@@ -6,11 +6,21 @@ use Rogue\Models\Reportback;
 use Rogue\Models\ReportbackLog;
 use Rogue\Models\ReportbackItem;
 use Rogue\Services\AWS;
+use Intervention\Image\Facades\Image;
+use finfo;
+use Symfony\Component\HttpFoundation\File\MimeType\ExtensionGuesser;
 
 class ReportbackRepository
 {
+    /**
+     * Instance of \Rogue\Services\AWS;
+     *
+     */
     protected $AWS;
 
+    /**
+     * Constructor
+     */
     public function __construct(AWS $aws)
     {
         $this->aws = $aws;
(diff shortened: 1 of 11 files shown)


app/Repositories/ReportbackRepository.php
```diff
@@ -9,8 +9,19 @@
 
 class ReportbackRepository
 {
+    /**
+     * Instance of \Rogue\Services\AWS;
```
DFurnes: Nitpicky, but we've used PHPDoc `@var` annotations on fields elsewhere:

```php
/**
 * AWS service class.
 *
 * @var \Rogue\Services\AWS
 */
```

app/Repositories/ReportbackRepository.php
```diff
@@ -9,8 +9,19 @@
...
+     */
     protected $AWS;
 
+    /**
+     * Array of properties needed for cropping and rotating.
```
DFurnes: Likewise, adding a `@var array` below the description here.

app/Services/ReportbackService.php
```diff
@@ -13,6 +13,11 @@ class ReportbackService
      */
     protected $reportbackRepository;
 
+    /**
+     * Contructor
```
DFurnes: `Contructor` → `Constructor` 🤓 👾

app/helpers.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Given an array of key/value pairs, create a new array with only the
+ * keys given.
+ *
+ * @param  associative array  $data
+ * @param  array $desiredKeys
+ *
+ * @return array
+ */
+function extract_values_by_keys($array, $desiredKeys)
```
DFurnes: I think you might be able to use Laravel's [`array_only`](https://laravel.com/docs/5.3/helpers#method-array-only) for this.

sbsmith86: welp. yes. that seems perfect! thanks

sbsmith86:
@DFurnes Thanks for your 👀 ! Made those updates. Is this good to go?

PR #64, opened 2016-11-01
pull s3 bucket from config

### LINKED ISSUE #63: don't hardcode AWS bucket into url
_BUG_
-------------------
#### Current Behavior
Currently part of the AWS url that points to the bucket is [hardcoded](https://github.com/DoSomething/rogue/blob/a76b8edb1ad9e3858efd0e088da8d581345463e2/app/Services/AWS.php#L48). 

#### Desired Behavior
We have this in the `.env` as `S3_BUCKET`, so let's pull it from there instead.

#### Why This Matters
We want to be able to use different buckets in different environments.

katiecrane:
#### What's this PR do?
Pull the S3 bucket from the config with `config('filesystems.disks.s3.bucket')` the same way we get the public url.

Just make sure it ends up building the correct url.

#### Relevant tickets
Fixes #63

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Services/AWS.php
+++ b/app/Services/AWS.php
@@ -45,7 +45,7 @@ class AWS
             throw new HttpException(500, 'Unable to save image to S3.');
         }
 
-        return config('filesystems.disks.s3.public_url') . '/ds-rogue-prod' . $path;
+        return config('filesystems.disks.s3.public_url') . '/' . config('filesystems.disks.s3.bucket') . $path;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


sbsmith86:
YASSS 🎉

angaither:
@katiecrane just make sure this is set in the default version so we don't get faillllz locally

katiecrane:
@angaither you mean set a default in `env.example` or add it like this:  `'bucket' => env('S3_BUCKET', 'ds-rogue-qa')`?

sbsmith86:
👍

PR #77, opened 2016-11-29
only send newly added files to Phoenix

### LINKED ISSUE #76: only send an image with a reportback if a new image was submitted
_BUG_
-------------------
#### Current Behavior
Rogue always sends the most recently added file on a reportback over to Phoenix, even if the most recent update was just to the quantity.

#### Desired Behavior
Only send a file to Phoenix if one was just added.

#### Steps to Replicate
Turn Rogue on and submit a reportback via Phoenix web. Then, update just the quantity. You will see the image you submitted upon creation will appear twice now.

#### Why This Matters
We don't want to tell Phoenix that a new image was submitted if it wasn't.

katiecrane:
#### What's this PR do?
Reportback images no longer duplicate with every single update to a reportback! 

However, there is a case where the image is still duplicated which stems from how Phoenix checks to see if a new image was added. As far as I can tell, that case is just when you:
1. create a reportback or submit an update with a new image
2. hit the permalink page
3. hit back to return to the campaign page, do not refresh and do not wait to long!
4. change the something in the form but do not add a new image
5. submit again
6. return to the campaign page to see the duplicated image
I couldn't find any differences in the `$form_state` or anything in Phoenix that would hint at it being a new image rather than a duplicated one. 

Submit a new reportback, and make sure you can make updates to the quantity and the why without the image doubling.

A potential problem with this is that once https://github.com/DoSomething/phoenix/pull/7198 is merged, all calls to the Phoenix endpoint will end up in Rogue. The Phoenix endpoint does not require a `caption` to be provided in any cases, and here we assume that one is given if there is a file, which is what the web form requires.

#### Relevant tickets
Fixes #76

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Jobs/SendReportbackToPhoenix.php
+++ b/app/Jobs/SendReportbackToPhoenix.php
@@ -20,7 +20,7 @@ class SendReportbackToPhoenix extends Job implements ShouldQueue
      *
      * @return void
      */
-    public function __construct(Reportback $reportback, $hasFile = TRUE)
+    public function __construct(Reportback $reportback, $hasFile = true)
     {
         $this->reportback = $reportback;
         $this->hasFile = $hasFile;
--- a/app/Jobs/SendReportbackToPhoenix.php
+++ b/app/Jobs/SendReportbackToPhoenix.php
@@ -47,7 +47,7 @@ class SendReportbackToPhoenix extends Job implements ShouldQueue
         if ($this->hasFile) {
             $reportbackItem = $this->reportback->items()->orderBy('created_at', 'desc')->first();
             $body['file_url'] = is_null($reportbackItem->edited_file_url) ? $reportbackItem->file_url : $reportbackItem->edited_file_url;
-            $body['caption'] = $reportbackItem->caption;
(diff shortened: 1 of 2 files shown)


app/Jobs/SendReportbackToPhoenix.php
```diff
@@ -33,18 +35,22 @@ public function handle()
...
-            'file_url' => is_null($reportbackItem->edited_file_url) ? $reportbackItem->file_url : $reportbackItem->edited_file_url,
-            'caption' => $reportbackItem->caption,
-            'source' => $reportbackItem->source,
         ];
 
+        // Data that everything except an update without a file will have
+        if ($this->hasFile) {
+            $reportbackItem = $this->reportback->items()->orderBy('created_at', 'desc')->first();
+            $body['file_url'] = is_null($reportbackItem->edited_file_url) ? $reportbackItem->file_url : $reportbackItem->edited_file_url;
+            $body['caption'] = $reportbackItem->caption;
```
angaither: yup

katiecrane: word, on it

chloealee:
👍

PR #78, opened 2016-11-30
Get user info

### LINKED ISSUE #71: Pull in user information into rogue using Northstar or Cached data
_FEATURE OVERVIEW_
-------------------
#### User Story

Eventually we will want to be able to expose important user information in rogue. To do this we should implement a system in rogue that when wanting to display user information will first look in Cache and grab the user info from there. If it isn't found there, make the call to Northstar to grab the user info and also store that information in cache so that the next time we need that users information it will be there (or not be there depending on when we expire the cache.)

#### Why This Matters

When we start to build out the admin interface for Rogue we will want to be able to display things like user name, email, address, etc.

sbsmith86:
#### What's this PR do?

Creates the system of grabbing user information in Rogue. 

First we try to grab user information that we stored in cache, if they aren't found there we grab them directly from Northstar. 

This work is based largely on @weerd 's work in Gladiator. The Gladiator system is a little more complicated since we actually store users in Gladiator, but in Rogue we just store references to northstar ids in the reportbacks table. 

So what this sets up is a `Registrar` service that handles the logic of resolving users in northstar vs cache. We also have a `CacheRepository` class that handles the storing/grabbing cache logic [(based on this)](https://github.com/DoSomething/gladiator/blob/master/app/Repositories/CacheStorage.php) 

This also sets up a view at `/users` to just list the users in the users table, which will really only be admins but helps to provide an interface for this work. 

I started writing tests for this, but decided I needed to get this merged in more. I will flag that work for the future. 

#### Relevant tickets
Fixes #71 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -2,7 +2,8 @@
 
 namespace Rogue\Http\Controllers;
 
-use DoSomething\Gateway\Northstar;
+// use DoSomething\Gateway\Northstar;
+use Rogue\Services\Registrar as Registrar;
 
 class UsersController extends Controller
 {
@@ -12,8 +13,19 @@ class UsersController extends Controller
      */
     protected $northstar;
 
-    public function __construct(Northstar $northstar)
+    public function __construct(Registrar $registrar)
     {
-        $this->northstar = $northstar;
+        // $this->northstar = $northstar;
+        $this->registrar = $registrar;
+    }
+
+    /**
+     * For now, just a test to have route to send authenticated/logged out users to.
+     */
+    public function index()
+    {
+
+        dd($this->registrar->find('559442cca59dbfc9578b4bf4'));
+        // dd($this->registrar);
     }
 }
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -3,6 +3,7 @@
 namespace Rogue\Http\Controllers;
 
 use Rogue\Services\Registrar as Registrar;
+use Rogue\Models\User;
 
 class UsersController extends Controller
 {
@@ -18,10 +19,16 @@ class UsersController extends Controller
     }
 
     /**
-     * List of Rogue users.
+     * Display a listing of the resource.
+     *
+     * @return \Illuminate\Http\Response
      */
     public function index()
     {
+        $ids = User::all()->pluck('northstar_id')->all();
 
+        $users = $this->registrar->findAll($ids);
+
+        return view('users.index', compact('users'));
     }
 }
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -24,10 +24,9 @@ class UsersController extends Controller
      */
     public function index()
     {
-
+        // $this->registrar->find('559442cca59dbfc9578b4bf4');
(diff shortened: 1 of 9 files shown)


app/Services/Registrar.php
```diff
@@ -0,0 +1,115 @@
...
+     */
+    public function find($id)
+    {
+        $user = $this->cache->retrieve($id);
+
+        if (! $user) {
+            $user = $this->northstar->getUser('id', $id);
+
+            // @TODO - How long should we store users in Cache?
+            $this->cache->store($user->id, $user);
```
angaither: can we push updates here so it's only updated when it needs to be?

app/helpers.php
```diff
@@ -28,9 +28,14 @@ function edit_image($image, $coords)
...
-    return $newTransactionId;
+    if ($transactionId) {
+        $transactionIdParts = explode('-', $transactionId);
+        $incrementedStep = $transactionIdParts[1] + 1;
+        $newTransactionId = $transactionIdParts[0] . '-' . $incrementedStep;
+
+        return $newTransactionId;
+    }
+
+    return null;
```
angaither: should this return null if there isn't a transaction id? or should it create one?

app/Http/Controllers/UsersController.php
```diff
@@ -12,8 +13,22 @@ class UsersController extends Controller
...
+    }
+
+    /**
+     * Display a listing of the resource.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function index()
+    {
+        $ids = User::all()->pluck('northstar_id')->all();
```
angaither: this probably needs a pager, rather than displaying all users

sbsmith86: I could see a need for that in the future. But this is only a page that displays users in the `Users` table which is really only admins, which is a small set of people right now. I mainly put this here to just have a visual display in the app of getting user information.

app/Services/Registrar.php
sbsmith86: I'm not sure what you mean? Like only store the user if they have updated information?

app/Http/Controllers/UsersController.php
angaither: oh my bad, I thought this was like gladiator, where we had to kill the whole user page

app/helpers.php
sbsmith86: hmm ya i guess we could create one instead. The reason I did this was mainly for testing. When trying to test the API, the tests were failing cause we don't send over a transaction ID with the fake requests. But I could see an instance where requests are coming in without the header for reals. But in those cases, would we want rogue to be responsible for creating the header here? Or would we want to not accept the request without that header?

DFurnes:
Looks good to me! 👍 This makes me want to bundle some caching logic into Gateway in the future as well. Would be nice to have all of that standardized between apps.

app/Services/Registrar.php
sbsmith86: oooo hmm yea. I like that idea! I'm not clear on how we can tell in Gateway that a user has updated information.

app/helpers.php
angaither: Yeah, I'm wondering about this for the migration script as well...

PR #83, opened 2016-12-05
check if isset so it can still forward

angaither:
#### What's this PR do?
update my previous PR that didn't check if things were set

👍 

I'm a dummy!

#### Relevant tickets
fixes #phoenix deploys (hopefully)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Services/ReportbackService.php
+++ b/app/Services/ReportbackService.php
@@ -41,7 +41,7 @@ class ReportbackService
 
         // POST reportback back to Phoenix, unless told not to.
         // If request fails, record in failed_jobs table.
-        if (! $data['do_not_forward']) {
+        if (! isset($data['do_not_forward'])) {
             dispatch(new SendReportbackToPhoenix($reportback));
         }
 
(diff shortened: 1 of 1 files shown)


chloealee:
👍

app/Services/ReportbackService.php
```diff
@@ -41,7 +41,7 @@ public function create($data, $transactionId)
 
         // POST reportback back to Phoenix, unless told not to.
         // If request fails, record in failed_jobs table.
-        if (! $data['do_not_forward']) {
+        if (! isset($data['do_not_forward'])) {
```
katiecrane: won't `! isset($data['do_not_forward'])` always evaluate to `false` if `$data['do_not_forward']` is set as `true` or `false`?

angaither: yeah, but I think that is ok? because the only time it is being set is in the [migration script](https://github.com/DoSomething/phoenix/pull/7226/files#diff-9d15d4c1e9153ce45a524cc0e8900b4bR51) 
the rest of the values don't have that included...

katiecrane: > if that value is ever set, don’t send it back to phoenix

^this

PR #85, opened 2016-12-08
send remote address to Phoenix

### LINKED ISSUE #84: send the remote address back to Phoenix
_BUG_
-------------------
#### Current Behavior
Rogue does not send the remote address that it gets from Phoenix back to Phoenix.

#### Desired Behavior
Send it back so it doesn't get overwritten!

#### Steps to Replicate
Do anything that sends a reportback to Rogue. Check the request you sent back to Phoenix. What?! No `remote_addr`?!?

#### Why This Matters
Data.

katiecrane:
#### What's this PR do?
Rogue was already [setting the `remote_addr`](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/ReportbackItemTransformer.php#L31) it was getting from Phoenix. Now, we also send this value back to Phoenix.

Try hardcoding this address as something and make sure it makes it to Phoenix.

This works in conjunction with https://github.com/DoSomething/phoenix/pull/7243, but I don't think the order of merging/deploying matters to not break anything. Both must be merged for it to actually work as expected though.

#### Relevant tickets
Fixes #84

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Jobs/SendReportbackToPhoenix.php
+++ b/app/Jobs/SendReportbackToPhoenix.php
@@ -49,6 +49,7 @@ class SendReportbackToPhoenix extends Job implements ShouldQueue
             $body['file_url'] = is_null($reportbackItem->edited_file_url) ? $reportbackItem->file_url : $reportbackItem->edited_file_url;
             $body['caption'] = isset($reportbackItem->caption) ? $reportbackItem->caption : null;
             $body['source'] = $reportbackItem->source;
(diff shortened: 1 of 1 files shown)


katiecrane:
is it cool with everyone if I do a styleCI PR after this gets merged? it has a lot of complaints outside of the file edited in this PR

chloealee:
awesome! 👍

chloealee:
also re you comment above - that is fine with me!

PR #87, opened 2016-12-14
Rogue api to new tables

### LINKED ISSUE #86: Send reportbacks posted via API to new tables
_FEATURE OVERVIEW_
-------------------
#### User Story

When we send reportbacks to rogue we want them to be stored in the new tables ( #80 ) we created to start storing reportbacks in this new system.

Reportbacks get stored in rogue the same way the get stored in phoenix.

sbsmith86:
#### What's this PR do?

Sorry for the rather large PR. 

In a nutshell, this sets up an endpoint at `/posts` that stores a "photo post" in the new rogue db schema. 

When the `/posts` endpoint is hit we use call the `store()` method on the `PostsController` 

The `PostsController` sends the request to the `PostService`. Right now we only have one post "type" so we use a `PhotoRepository` to store the photo. There could be a world, in the near future, where we have different types of posts and my thought was we would use the `PostService` to resolve which repo to use to store the post. 

I am interested in thinking more about how to structure the Post logic. If we have different types of posts but they all will generally implement the same functionality like create, delete, read.

I need to write tests for this.

I need to work on how updating posts work.  

#### Relevant tickets
Fixes #86 

#### Checklist
- [ ] Tested on staging.


app/Jobs/SendPostToPhoenix.php
```diff
@@ -0,0 +1,65 @@
...
+     * @var bool
+     */
+    protected $hasFile;
+
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct(Photo $photo, $hasFile = true)
```
weerd: Do we always know it's a `$photo`? Shouldn't this just be `$post` since it _could_ be a variety post types? Or is this because it's a specific class for Drupal Phoenix?

weerd: Maybe just rename to `SendPhotoPostToPhoenix` class?

app/Models/Photo.php
```diff
@@ -11,7 +11,21 @@ class Photo extends Model
      *
      * @var array
      */
-    protected $fillable = ['event_id', 'signup_id', 'northstar_id', 'file_url', 'edited_file_url', 'caption', 'status'];
+    protected $fillable = ['event_id', 'signup_id', 'northstar_id', 'file_url', 'edited_file_url', 'caption', 'status', 'source', 'remote_addr'];
```
weerd: This is random, but one thing I've started to try in my Laravel projects is when the `$fillable` array seems like it's gonna have a lot of stuff, I switch to use `$guarded` to just specify the items _not_ mass-assignable, which is usually a much smaller set of items. Feel free to ignore; just a suggestion cause it kind of became a habit to always use `$fillable` just cause that how we do it elsewhere, but doesn't necessarily have to be the default 😉

app/Repositories/PhotoRepository.php
```diff
@@ -0,0 +1,79 @@
...
+     * Constructor
+     */
+    public function __construct(AWS $aws, Registrar $registrar)
+    {
+        $this->aws = $aws;
+        $this->registrar = $registrar;
+    }
+
+    /**
+     * Create an event.
```
weerd: Creating an `Event` or a `Photo`?

weerd:
@sbsmith86 what's the point of `Event` vs `Signup` again? I'm a little confused on that part I think... Overall I think it's looking pretty cool though!

app/Jobs/SendPostToPhoenix.php
sbsmith86: Oh yea we could rename this job. It's really just a temporary thing, we normally won't be sending stuff back to phoenix, just for now until rogue powers the gallery and admin stuff.

weerd: Gotcha, ok!

sbsmith86:
@weerd So we have and `events` table which stores all atomic events within the system (something gets reviewed, posted, updated, tagged, etc) 

the `signups` table stores that someone signed up for a particular campaign/campaign run, and all posts live under a signup. So you can be a user who signs up for campaign X and submits multiple posts for that campaign. When you signup, there is a signup event created. When you post, there is a post event created.

weerd:
Cool cool! That makes sense :)

app/Repositories/PhotoRepository.php
```diff
@@ -0,0 +1,79 @@
...
+use Rogue\Services\Registrar;
+
+class PhotoRepository
+{
+    /**
+     * AWS service class instance.
+     *
+     * @var \Rogue\Services\AWS
+     */
+    protected $AWS;
```
DFurnes: This is lowercase `$this->aws` elsewhere!

app/Http/routes.php
```diff
@@ -27,19 +27,24 @@
...
+    // reactions
     Route::post('reactions', 'Api\ReactionController@store');
 
-    // /reportbacks
+    // reportbacks
     Route::get('reportbacks', 'Api\ReportbackController@index');
     Route::post('reportbacks', 'Api\ReportbackController@store');
 });
+
+// v2 routes
```
DFurnes: Exciting!

app/Http/Controllers/Api/PostsController.php
```diff
@@ -0,0 +1,96 @@
...
+
+        // @TODO - should we eventually throw an error if a signup doesn't exist before a post is created? I create one here because we haven't implemented sending signups to rogue yet, so it will have to create a signup record for all posts.
+        if (is_null($signup)) {
+            $signup = $this->signups->create($request->all());
+        }
+
+        $post = $this->posts->create($request->all(), $signup->id, $transactionId);
+
+        if ($post) {
+            $code = 200;
```
DFurnes: Does this do anything? I might be missing something but it looks like the `PostService` should always be returning something truthy here, so this conditional would always run. Also isn't the default value for the `$this->item` method's code parameter already 200?

app/Http/Controllers/Api/PostsController.php
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        return $this->resolvePostTransformer($post, $code);
```
DFurnes: This makes me think that these should be separate endpoints, or at least a nested resource within a standard `PostTransformer` used for the shared fields that all types of posts have. A huge benefit of having a transformer class is a consistent response format, which might not be true with more than one transformer for a route.

app/Http/Controllers/Api/PostsController.php
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request)
+    {
+        $transactionId = incrementTransactionId($request);
```
DFurnes: This feels like something that'd make sense as a (global?) middleware that reads the `X-Transaction-Id` header on any incoming request and if found it increments it by one, rather than needing to remember to add these lines to each relevant controller and service class.

app/Http/Transformers/PhotoTransformer.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public function transform(Photo $photo)
+    {
+        return [
+            'event_id' => $photo->event_id,
+            'signup_id' => $photo->signup_id,
+            'northstar_id' => $photo->northstar_id,
+            'campaign_id' => $photo->signup->campaign_id,
+            'campaign_run_id' => $photo->signup->campaign_run_id,
+            'post' => [
```
DFurnes: Following up on my above comment about the transformer – I feel like this `post` object would make a lot more sense as the "top level" item, and then a nested `photo` (etc.) for any fields specific to a particular type of thing.

DFurnes:
I don't have any specific suggestion to improve this, but I had some trouble jumping around the code between the controller/service/repository/model & figuring out how all the code fit together. I'm sure it'll all help manage complexity in the future, but as a "first time reader" it was a bit overwhelming. 😬

app/Http/Controllers/Api/PostsController.php
sbsmith86: Ah i love that idea. I might hold off on that as a separate issue since it would be a little out of the scope of this work. But i will document that.

app/Http/Controllers/Api/PostsController.php
sbsmith86: lol yea.. i don't remember what the was. Probably from some earlier version.

app/Http/Transformers/PhotoTransformer.php
sbsmith86: @DFurnes I definitely like the idea of having a standard `PostTransformer` . I am going back and forth on each post type having it's own endpoint. It would certainly make things a little more straightforward because we would never have to do the work of figuring out what we trying to store/grab/return. But then at the same time I feel like all of these things are really the same, it's just they store different stuff, so I like just having one endpoint to deal with them. 

It's also weird because right now we only have photos so I might have shot myself in the foot with too much future thinking.

sbsmith86:
@DFurnes yea, so I definitely might have over complicated things. I was trying to follow some sort of pattern that would allow me to wrap my head around where things go. But maybe it is worth it to just keep it simple until it gets more complicated. I don't want anyone to be that confused jumping into the code.

app/Http/Transformers/PhotoTransformer.php
DFurnes: I definitely like the direction things are going with the "post" generalization! I think if you were able to even just have a `photo => [ ... ], video => null` style "nested" items for the more specific reportback types it'd feel pretty tidy.

DFurnes:
@sbsmith86 Yeah, I think it could also just be a problem with trying to read it all as a diff rather than in an actual code editor too, so I was scrolling around a ton more than I would otherwise.

weerd:
@sbsmith86 I do like a lot of the ideas you've thought of so far for this and there's defs ways to keep it simple, but versatile to allow for the future stuff(z)! Looking good! 👍

angaither:
@sbsmith86 when you come back, can you add these new endpoints to the documentation folder

PR #94, opened 2017-01-09
Polymorphic post

sbsmith86:
#### What's this PR do?

This is a proof of concept PR that implements a [Polymorphic](https://laravel.com/docs/5.3/eloquent-relationships#polymorphic-relations) Relationship between "Posts" and a post sub-type like a `photo`. 

Before a "Post" wasn't a real thing which means we didn't have a unifying way to deal with different types of posts. While they are different content, they serve the similar purpose and we didn't have a great way of tying them together. With this work, we can do things like have a `PostTransformer` for all posts so they have a standard response regardless of type. Or, we can find all posts for a signup across all types. 

#### In a nutshell
- we have a `Post` table that holds the thing that in common with all posts: `event_id`, `signup_id` 
- There is a polymorphic relationship defined on the `Post` and `Photo` model.

_To create a post you first create the photo, or the subtype item and then just do:_

```php
        $post = new Post([
            'event_id' => $postEvent->id,
            'signup_id' => $signupId,
        ]);
        $post->content()->associate($photo);
        $post->save();
```
_which creates a record in the Posts table:_

image

_To use the relationship:_ 

```php
$post = Post::where([
'event_id' => 123
]);

$type = $post->post_type;
$file = $post->content->file_url;
$caption = $post->content->caption;
```

_And here is an example of what the API response would look like after posting to the API to create a post:_ 

image

Based a lot of this work on [this blog post. ](https://webscope.co.nz/polymorphic-relations-with-eloquent-and-laravel-5)

### DIFF
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -2,11 +2,11 @@
 
 namespace Rogue\Http\Controllers\Api;
 
-use Rogue\Services\PostService;
-use Rogue\Http\Transformers\PostTransformer;
-use Illuminate\Http\Request;
 use Rogue\Models\Signup;
+use Illuminate\Http\Request;
+use Rogue\Services\PostService;
 use Rogue\Repositories\SignupRepository;
+use Rogue\Http\Transformers\PostTransformer;
 
 class PostsController extends ApiController
 {
@@ -43,7 +43,6 @@ class PostsController extends ApiController
         // Now we have one PostTransformer to handle returning a Post to the API request.
         $this->transformer = new PostTransformer;
     }
-
 
     /**
      * Store a newly created resource in storage.
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -3,8 +3,8 @@
 namespace Rogue\Http\Controllers\Api;
 
 use Rogue\Models\Signup;
-use Illuminate\Http\Request;
 use Rogue\Services\PostService;
+use Rogue\Http\Requests\PostRequest;
 use Rogue\Repositories\SignupRepository;
 use Rogue\Http\Transformers\PostTransformer;
 
@@ -51,7 +51,7 @@ class PostsController extends ApiController
      * @return \Illuminate\Http\Response
      */
     // @TODO - Validate post request.
-    public function store(Request $request)
+    public function store(PostRequest $request)
     {
         $transactionId = incrementTransactionId($request);
 
@@ -66,7 +66,7 @@ class PostsController extends ApiController
             $signup = $this->signups->create($request->all());
         }
 
-        // Send the data to the PostService class which will handle delgating
+        // Send the data to the PostService class which will handle determining
         // which type of post we are dealing with and which repostitory to use to // actually create the post.
         $post = $this->posts->create($request->all(), $signup->id, $transactionId);
 
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -2,11 +2,11 @@
 
 namespace Rogue\Http\Controllers\Api;
 
-use Rogue\Models\Signup;
-use Illuminate\Http\Request;
 use Rogue\Services\PostService;
-use Rogue\Repositories\SignupRepository;
 use Rogue\Http\Transformers\PostTransformer;
+use Illuminate\Http\Request;
+use Rogue\Models\Signup;
+use Rogue\Repositories\SignupRepository;
 
 class PostsController extends ApiController
 {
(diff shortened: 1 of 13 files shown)


app/Http/Controllers/Api/PostsController.php
```diff
@@ -62,35 +66,10 @@ public function store(Request $request)
...
-            case 'Rogue\Models\Photo':
-                $this->transformer = new PhotoTransformer;
-
-                return $this->item($model, $code);
-
-                break;
-            default:
-                break;
-        }
+        return $this->item($post);
```
DFurnes: 🎉

app/Http/Transformers/PostTransformer.php
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Transform resource data.
+     *
+     * @param \Rogue\Models\Photo $photo
+     * @return array
+     */
+    public function transform(Post $post)
+    {
+        return [
+            'event_id' => $post->event_id,
```
DFurnes: Is this field needed now that there's the embedded `event` resource?

app/Services/PostService.php
```diff
@@ -48,4 +39,17 @@ public function create($data, $signupId, $transactionId)
...
+    protected function resolvePostRepository($type)
+    {
+        switch ($type) {
+            case 'post_photo':
+                $this->repository = app('Rogue\Repositories\PhotoRepository');
+
+                break;
+
+            default:
+                break;
```
DFurnes: This should probably throw an exception.

sbsmith86:
@DFurnes What I liked about this is that we use `signup_id` and `event_id` as sorta the common things amongst all posts. That way we really can just let each sub-type define the content it holds itself without assuming anything about what the content will look like. It feels very flexible this way.

app/Http/Transformers/PostTransformer.php
sbsmith86: ah good point.

app/Services/PostService.php
sbsmith86: another good point

weerd:
https://cloud.githubusercontent.com/assets/1700409/21785043/fa426478-d68b-11e6-841a-9d2ff83f43cd.png

@sbsmith86 just curious, if this is a record on the post table, why are you prefixing `post_` to the `id` and the `type`?

sbsmith86:
@weerd The `posts` table holds a reference to a post subtype. So that record there says there was a post submitted. `post_type` tells us it is a `Photo` post and `post_id` tell us which record in the `photos` table it correlates to. This table and those columns are required to make polymorphic relationships work. In the [documentation](https://laravel.com/docs/5.3/eloquent-relationships#polymorphic-relations), they outline a scenario where you allow comments on different type of content and you uses a comments table to store the relationship between a comment and which type of content it is on.

DFurnes:
> What I liked about this is that we use signup_id and event_id as sorta the common things amongst all posts. That way we really can just let each sub-type define the content it holds itself without assuming anything about what the content will look like. It feels very flexible this way.

That seems like a good argument to me. 👍

> @weerd The posts table holds a reference to a post subtype. So that record there says there was a post submitted. post_type tells us it is a Photo post and post_id tell us which record in the photos table it correlates to.

It looks like a common convention is to make the polymorphic relationship a "-able" word, like `commentable` for things that can accept comments or `taggable` for things that can be tagged. Maybe `postable_type` and `postable_id` would make it more immediately clear that its a relationship?

sbsmith86:
I wasn't a huge fan of the "-able" suffix and it seemed a little silly lol and after some reading and understanding these relationships a little more, the documentation really outlines how to define a "has-a" relationship but we want to define something more like an "is-a" relationship. LIke this post "is-a" photo...it which case looking for `post_type` seemed to make more sense to mean.

weerd:
I was definitely confused about it sans the suffix. Regardless of what it's called, something to distinguish the post record `id` from the `post_id` as those being two separate things. From someone just jumping in to see these references it was confusing 😐

sbsmith86:
@weerd i'm not sure we can get rid of any prefix. They are required to define the relationship. We can make it `item_id` and `item_type` maybe? and then you would call `$post->item_type` instead

weerd:
Oh @sbsmith86, I didn't mean to get rid of it. Just either use the established `-able` standard like @DFurnes mentioned or something else to help distinguish it. `item` could also work... :man_shrugging:

sbsmith86:
@DFurnes do you think you can give this another look for approval? I made the updates you suggested, added some lightweight validation which can been flushed out more later ( i don't want this PR to get too big, and I update the polymorphic column to use the `-able` suffix for clarity.

app/Http/Requests/PostRequest.php
```diff
@@ -0,0 +1,73 @@
...
+     * Determines the validation ruleset to use for a particular type of post.
+     *
+     * @param  object $request
+     * @return array|null
+     */
+    protected function getRules($request)
+    {
+        switch ($request->event_type) {
+            case 'post_photo':
+                return array_merge($this->rules, $this->setPhotoRules());
```
DFurnes: I'm curious why `$this->rules` is a property but `$this->setPhotoRules()` is a function. If you'd like to keep it as a function, would it make sense to rename it to `$this->getPhotoRules()` since it doesn't mutate anything?

sbsmith86: yep makes sense

PR #97, opened 2017-01-13
signups api endpoint

katiecrane:
#### What's this PR do?
Creates the signups `POST` endpoint in Rogue oooOOOOoooOOOOoooo :crystal_ball:

Request data format example:
```
"northstar_id" => "5639066ba59dbfe6598b4567"
"campaign_id" => "6"
"campaign_run_id" => "1740"
"quantity" => "200"
"source" => "the-fox-den"
"do_not_forward" => "TRUE"
"why_participated" => "bcuz I luv endpointz"
"created_at" => "2016-12-23 16:50:36"
"updated_at" => "2017-01-16 16:50:36"
"photo" => [
    0 => [
      "source" => "photo-source"
      "remote_addr" => "testin yo"
      "caption" => "LOOK AT THAT CAT"
      "file" => UploadedFile {#164
        -originalName: "tongue-cat.jpg"
        ...}
      "event_type" => "post_photo"
      "northstar_id" => "5639066ba59dbfe6598b4567"
    ]
    1 => [
      "source" => "second-source"
      "remote_addr" => "second home yknow?"
      "caption" => "lil chickens"
      "file" => UploadedFile {#173
        -originalName: "chickens.png"
        ...}
      "event_type" => "post_photo"
      "northstar_id" => "5639066ba59dbfe6598b4567"
    ]
]
```

Example response:
```
{
  "data": {
    "id": 160,
    "northstar_id": "5639066ba59dbfe6598b4567",
    "campaign_id": "6",
    "campaign_run_id": "1740",
    "quantity": null,
    "quantity_pending": "200",
    "why_participated": "bcuz I luv endpointz",
    "created_at": "2017-01-19T19:11:14+0000",
    "updated_at": "2017-01-19T19:11:14+0000",
    "post": {
      "data": [
        {
          "signup_id": 160,
          "northstar_id": "5639066ba59dbfe6598b4567",
          "campaign_id": 6,
          "campaign_run_id": 1740,
          "content": {
            "type": "post_photo",
            "media": {
              "url": "https://s3.amazonaws.com/ds-rogue-qa/uploads/reportback-items/160-1484853074.jpeg",
              "edited_url": null
            },
            "caption": "LOOK AT THAT CAT",
            "status": "pending",
            "created_at": "2017-01-19T19:11:15+0000",
            "updated_at": "2017-01-19T19:11:15+0000"
          },
          "event": {
            "data": {
              "event_id": "290",
              "event_type": "post_photo",
              "submission_type": "",
              "created_at": "2017-01-19T19:11:14+0000",
              "updated_at": "2017-01-19T19:11:14+0000"
            }
          }
        },
        {
          "signup_id": 160,
          "northstar_id": "5639066ba59dbfe6598b4567",
          "campaign_id": 6,
          "campaign_run_id": 1740,
          "content": {
            "type": "post_photo",
            "media": {
              "url": "https://s3.amazonaws.com/ds-rogue-qa/uploads/reportback-items/160-1484853075.png",
              "edited_url": null
            },
            "caption": "lil chickens",
            "status": "pending",
            "created_at": "2017-01-19T19:11:16+0000",
            "updated_at": "2017-01-19T19:11:16+0000"
          },
          "event": {
            "data": {
              "event_id": "291",
              "event_type": "post_photo",
              "submission_type": "",
              "created_at": "2017-01-19T19:11:15+0000",
              "updated_at": "2017-01-19T19:11:15+0000"
            }
          }
        }
      ]
    },
    "event": {
      "data": {
        "event_id": "289",
        "event_type": "signup",
        "submission_type": "user",
        "created_at": "2017-01-19T19:11:14+0000",
        "updated_at": "2017-01-19T19:11:14+0000"
      }
    }
  }
}
```

Make sure the format of the request and response make sense. Did I make any silly mistakes?

#### Relevant tickets
Fixes #

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/Api/SignupsController.php
```diff
@@ -0,0 +1,75 @@
+<?php
+
+namespace Rogue\Http\Controllers\Api;
+
+use Illuminate\Http\Request;
+use Rogue\Services\SignupService;
+use Rogue\Repositories\PhotoRepository;
+use Rogue\Http\Transformers\SignupTransformer;
+
+// use Rogue\Http\Requests;
```
DFurnes: 👾

app/Http/Controllers/Api/SignupsController.php
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Create a controller instance.
+     *
+     * @param  PostContract  $posts
+     * @return void
+     */
+    public function __construct(SignupService $signups, PhotoRepository $photo)
+    {
+        $this->signups = $signups;
+        $this->photo = $photo;
```
DFurnes: Should `$this->photo` be changed to `$this->photos` so it's plural like `$this->signups`?

app/Http/Controllers/Api/SignupsController.php
```diff
@@ -0,0 +1,75 @@
...
+
+        if ($signup) {
+            $code = '200';
+        }
+
+        // Transform the data to return it
+        $this->transformer = new SignupTransformer;
+
+        // check to see if there is a reportback too aka we are migratin'
+        if (array_key_exists('photo', $request->all())) {
```
DFurnes: This could also be written as `$request->has('photo')` which might read nicer.

app/Http/Controllers/Api/SignupsController.php
```diff
@@ -0,0 +1,75 @@
...
+            // create the photo and tie it to this signup
+            foreach ($request->all()['photo'] as $photos) {
```
DFurnes: Same here with `$request->input('photo')` instead of `$request->all()['photo']`.

app/Http/Controllers/Api/SignupsController.php
```diff
@@ -0,0 +1,75 @@
...
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request)
+    {
+        $transactionId = incrementTransactionId($request);
+
+        $signup = $this->signups->create($request->all(), $transactionId);
+
+        if ($signup) {
+            $code = '200';
```
DFurnes: It looks like the `item` method on the base API controller [expects the status code to be an integer](https://github.com/DoSomething/rogue/blob/b7e816b7c8e1afdad8335d4754b24a2f3251c4d7/app/Http/Controllers/Api/ApiController.php#L43). I'm also curious what scenarios `$signup` won't be defined in – should there be a different code if so?

DFurnes: It looks like the SignupService's `create` method always creates a new signup in the database. If so, would it make sense to return a `201` status code here to indicate that?

app/Jobs/SendSignupToPhoenix.php
```diff
@@ -0,0 +1,49 @@
...
+use Rogue\Services\Registrar;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+class SendSignupToPhoenix extends Job implements ShouldQueue
+{
+    use InteractsWithQueue, SerializesModels;
+
+    protected $signup;
```
DFurnes: Docblock! 📝

app/Jobs/SendSignupToPhoenix.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $phoenix = new Phoenix;
```
DFurnes: These could both be injected from the service container using type-hinting on the `handle` method:

```php
/**
 * Execute the job.
 *
 * @return void
 */
public function handle(Phoenix $phoenix, Registrar $registrar)
{
    $drupal_id = $registrar->find($this->signup->northstar_id)->drupal_id;
    // ...
```

It doesn't make a difference right now since neither of those classes have their own injected dependencies, but it'll makes things easier later on down the road if you ever do. 😄

app/Services/Phoenix.php
```diff
@@ -39,4 +39,17 @@ public function postReportback($body = [])
...
+     * Send a POST request to save a copy of the signup in Phoenix.
+     *
+     * @param array $body
+     * @return object|false
+     */
+    public function postSignup($body = [], $campaign_id)
+    {
+        $response = $this->post('campaigns/' . $campaign_id . '/signup', $body);
+
+        return is_null($response) ? null : $response;
```
DFurnes: Is this ternary necessary? It seems like it'll return `null` only if `$response` is already null.

app/Services/SignupService.php
```diff
@@ -0,0 +1,51 @@
...
+     * @return Illuminate\Database\Eloquent\Model $model
+     */
+    public function create($data, $transactionId)
+    {
+        $signup = $this->signup->create($data);
+
+        // Add new transaction id to header.
+        request()->headers->set('X-Request-ID', $transactionId);
+
+       // @TODO: call class to send back to Phoenix here once it exists
```
DFurnes: Has this `TODO` been done?

app/Http/Controllers/Api/SignupsController.php
katiecrane: @DFurnes [elsewhere](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Api/ReportbackController.php#L49-L57) it looks like we are returning `200` when an item is created and `201` when it is updated, that's what I went off of for consistency. Am I looking at this wrong? Also good catch on removing the quotes!!

app/Http/Controllers/Api/SignupsController.php
katiecrane: :bowing_woman:

app/Services/Phoenix.php
katiecrane: 🙈

app/Http/Controllers/Api/SignupsController.php
DFurnes: That's backwards! We should be using [`201 Created`](https://httpstatuses.com/201) when a new resource is created, and [`200 OK`](https://httpstatuses.com/200) for general "success" responses.

app/Http/Controllers/Api/SignupsController.php
```diff
@@ -0,0 +1,72 @@
...
+     */
+    protected $photos;
+
+    /**
+     * Create a controller instance.
+     *
+     * @param  PostContract  $posts
+     * @return void
+     */
+    public function __construct(SignupService $signups, PhotoRepository $photos)
```
sbsmith86: We should be using the new `PostService` instead of the `PhotoRepostitory` now. The `PostService` would resolve which repository to use to create the resource based on the `event_type` passed in the request.

app/Http/Transformers/SignupTransformer.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    /**
+     * Transform resource data.
+     *
+     * @param \Rogue\Models\Signup $signup
+     * @return array
+     */
+    public function transform(Signup $signup)
+    {
+        return [
```
sbsmith86: I feel like we should embed the `posts` that were created in this response.

app/Jobs/SendSignupToPhoenix.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->signup = $signup;
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle(Phoenix $phoenix, Registrar $registrar)
```
sbsmith86: Could you type hint `$phoenix` and `Registrar` in the constructor so they are instantiated when the class gets built? Similar to what we do in `SendReportbackToPhoenix`

app/Repositories/PhotoRepository.php
```diff
@@ -39,18 +40,22 @@ public function __construct(AWS $aws, Registrar $registrar)
      * @param  int $signupId
      * @return \Rogue\Models\Photo|null
      */
-    public function create(array $data, $signupId)
+    public function create(array $data, Signup $signup)
```
sbsmith86: Why do we need the whole `$signup` here? I believe that `northstar_id` should already be in the data array and all you really need is the `signupId` to associate with the item

app/Jobs/SendSignupToPhoenix.php
DFurnes: Dependency injection should happen [in the `handle` method](https://laravel.com/docs/5.3/queues#class-structure) for jobs. The constructor is for parameters that get serialized into the job (the specific signup that's being processed and is passed in when you dispatch `new SendSignupToPhoenix($signup)`)

sbsmith86: That should be updated in the other job also then.

katiecrane: @sbsmith86 is it cool if I handle updates not used by the signup endpoint in separate PRs? This is pretty big and I don't want it to get too scattered focus wise or be too much to test all at once. If that's cool I'll make issues for the other changes so they don't get forgotten!

sbsmith86: fine with me

app/Repositories/PhotoRepository.php
katiecrane: @sbsmith86 I'm not sure the best way to do this part. I structured the request (up in the write up) so that photos are in their own array with all the data that lives on a `photo` and not a `signup`. Since `signup` needs `northstar_id` I kept that outside and it seemed weird to me to have it in each `photo` array as well. Although now I'm thinking it would be easier to just have `northstar_id` both outside for the signup and then with the data for each `photo` as well so the code doesn't need to be altered since this is just for the purposes of the migration anyway. Does that sound reasonable?

sbsmith86: I think your original thought sounds reasonable, actually, but i believe this will break the `/posts` endpoint that is still sending `$signupId`

katiecrane: I think that maybe adding the `northstar_id` into the `photo` array in the `SignupsController`  before creating the post might be the nicest way to do it. That way the migration requests can still be a little nicer but the `/posts` endpoint is still happy

sbsmith86: sounds good!

app/Http/Controllers/Api/SignupsController.php
katiecrane: got this working! so cool!!

app/Http/Transformers/SignupTransformer.php
katiecrane: @sbsmith86 this is updated now!

sbsmith86: looks good!

app/Repositories/PhotoRepository.php
```diff
@@ -44,7 +44,11 @@ public function create(array $data, $signupId)
     {
         $postEvent = Event::create($data);
 
-        $fileUrl = $this->aws->storeImage($data['file'], $data['campaign_id']);
+        if (array_key_exists('campaign_id', $data)) {
+            $fileUrl = $this->aws->storeImage($data['file'], $data['campaign_id']);
+        } else {
+            $fileUrl = $this->aws->storeImage($data['file'], $signupId);
+        }
```
katiecrane: @sbsmith86 I wanted to make sure that this part looks okay, I'm not sure if it matters what the file name is? Since that's all the `campaign_id` is used for it made sense to me to not go out of our way to pass the `campaign_id` when we wouldn't otherwise

sbsmith86: hmm thanks for pointing this out! I think we should somehow standardize file names (That way it is easier to search if we ever need to) I don't really care what the standard is, but it should be something we can easily identify, maybe we just use `signupId` as the standard instead of `campaign_id`?

katiecrane: @sbsmith86 yeah, I think `signupId` makes sense in the new world, but I think we need the if/else until Phoenix is sending to the signups endpoint to not break how this works currently

sbsmith86: i don't think so just pass the `signupId` into the `crop` function or set it as a property of the class and then you can use it however you want.

katiecrane: @sbsmith86 oooooo right I got confused and forgot that currently we use `ReportbackRepository.php`, thank you!

katiecrane: @sbsmith86 updated!

PR #107, opened 2017-01-20
changes quantity to quantity_pending and removes hard coded uid

chloealee:
#### What's this PR do?
- Adds the `Registrar` in `SendPostToPhoenix.php` in order to send `drupal_id` (right now it is hard coded). 
- Changes `quantity` to send `quantity_pending` instead since `quantity` will be `NULL` and we want to send user's reported quantity to Phoenix. 

Hit `/posts` endpoint and all should work correctly.

### DIFF
--- a/app/Jobs/SendPostToPhoenix.php
+++ b/app/Jobs/SendPostToPhoenix.php
@@ -6,6 +6,7 @@ use Rogue\Models\Post;
 use Rogue\Services\Phoenix;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Contracts\Queue\ShouldQueue;
+use Rogue\Services\Registrar;
 
 class SendPostToPhoenix extends Job implements ShouldQueue
 {
@@ -40,15 +41,16 @@ class SendPostToPhoenix extends Job implements ShouldQueue
      *
      * @return void
      */
-    public function handle()
+    public function handle(Registrar $registrar)
     {
         $phoenix = new Phoenix;
+        $drupal_id = $registrar->find($this->post->signup->northstar_id)->drupal_id;
 
         // Data that every post will have
         $body = [
-            'uid' => 12345, // Grab drupal_id from northstar object?
+            'uid' => $drupal_id,
             'nid' => $this->post->signup->campaign_id,
-            'quantity' => $this->post->signup->quantity,
+            'quantity' => $this->post->signup->quantity_pending,
             'why_participated' => $this->post->signup->why_participated,
         ];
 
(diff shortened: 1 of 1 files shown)


app/Jobs/SendPostToPhoenix.php
```diff
@@ -40,15 +41,16 @@ public function __construct(Post $post, $hasFile = true)
...
         $phoenix = new Phoenix;
+        $drupal_id = $registrar->find($this->post->signup->northstar_id)->drupal_id;
 
         // Data that every post will have
         $body = [
-            'uid' => 12345, // Grab drupal_id from northstar object?
+            'uid' => $drupal_id,
             'nid' => $this->post->signup->campaign_id,
-            'quantity' => $this->post->signup->quantity,
+            'quantity' => $this->post->signup->quantity_pending,
```
katiecrane: So for the time being, `quantity_pending` is the same as Phoenix's `quantity`? If so I think this all looks good!

chloealee: Yes as I understand it! Otherwise, it sends `NULL` to Phoenix to save in the database. `quantity_pending` is the quantity that the user has submitted.

app/Jobs/SendPostToPhoenix.php
```diff
@@ -40,21 +41,22 @@ public function __construct(Post $post, $hasFile = true)
...
             'why_participated' => $this->post->signup->why_participated,
         ];
 
         // Data that everything except an update without a file will have
         if ($this->hasFile) {
-            $body['file_url'] = is_null($this->post->content->edited_file_url) ? $this->photo->file_url : $this->post->content->edited_file_url;
+            $body['file_url'] = is_null($this->post->content->edited_file_url) ? $this->post->file_url : $this->post->content->edited_file_url;
```
chloealee: @sbsmith86 @katiecrane found another small typo and figured I'd just put it in here instead of making a separate PR for it!

chloealee: oh wait - I don't think the above makes sense either - working on this now!

app/Jobs/SendPostToPhoenix.php
```diff
@@ -40,21 +41,22 @@ public function __construct(Post $post, $hasFile = true)
      *
      * @return void
      */
-    public function handle()
+    public function handle(Registrar $registrar)
     {
         $phoenix = new Phoenix;
+        $drupal_id = $registrar->find($this->post->signup->northstar_id)->drupal_id;
```
sbsmith86: 🙏

PR #108, opened 2017-01-23
Allow signup updates

katiecrane:
#### What's this PR do?
Added the ability for `quantity` and `why_participated` on a signup to be updated. This allows for Posts to be sent from Phoenix. I think we want to be careful about only using this functionality when there was actually a change in either of those values. If we think that should happen inside the `SignupRepository` `update` function then I am glad to add it there!

Also, regarding `events`, I think the Posts/Events relationship needs to be updated to mimic the Signup/Events relationship here. The `signup_id` column for post events will still have the `signup_id` because `posts` all have that. I'm happy to do that in the future but I thought it was out of scope for this update.

Make sure that when you add a new post that you are able to update the `quantity` and `why_participated` on the signup. Call something like this in `PostsController`: 
```
$this->signups->update($signup, $request->all());
```
Is this how it should be structured?

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/SignupRepository.php
+++ b/app/Repositories/SignupRepository.php
@@ -15,13 +15,15 @@ class SignupRepository
      */
     public function create($data)
     {
+        // Create the post_signup event
         $event = Event::create([
             'northstar_id' => $data['northstar_id'],
-            'event_type' => 'signup',
+            'event_type' => 'post_signup',
             'submission_type' => 'user',
         ]);
 
-        $signup = $event->signup()->create([
+        // Create the signup
+        $signup = Signup::create([
             'event_id' => $event->id,
             'northstar_id' => $data['northstar_id'],
             'campaign_id' => $data['campaign_id'],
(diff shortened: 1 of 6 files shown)


chloealee:
this looks great, @katiecrane ! The only thing is I think we'll want a new `PUT` endpoint for this to access the update functionality?

app/Repositories/SignupRepository.php
```diff
@@ -15,13 +15,15 @@ class SignupRepository
      */
     public function create($data)
     {
+        // Create the post_signup event
         $event = Event::create([
             'northstar_id' => $data['northstar_id'],
-            'event_type' => 'signup',
+            'event_type' => 'post_signup',
```
sbsmith86: What is `post_signup` supposed to signify? If this is just for creating a signup I think we should not use the "post" terminology cause it seems confusing. A post is different than a signup

sbsmith86: Also when we create this signup, I think we should also store the other elements like quantity and why participated in the event record, right?

katiecrane:
Going to use this for updates via the `/signups` endpoint, so closing this

PR #112, opened 2017-01-30
Adds /reviews endpoint to update photo(s)'s status when it gets reviewed and updates signups/photos 

### LINKED ISSUE #100: Update posts using /posts endpoint
_FEATURE OVERVIEW_
-------------------
#### User Story

The `/posts` endpoint in rogue doesn't update posts, it only creates them. The posts endpoint should both update the post content (like the caption) as well as update stuff like `why_participated` and `quantity` on the signup record. Additionally we should log changes in the `events` table, similar to how we were using the `reportback_logs` table.

chloealee:
#### What's this PR do?
- Adds update functionality to update a signup (`why_participated` and `quantity`)
- Adds update functionality to add a new photo post to a signup 
- Adds `/reviews` endpoint to update a photo(s)'s status when reviewed by an admin

This should be tested with [corresponding Phoenix work](https://github.com/DoSomething/phoenix/pull/7295).

1.Change Rogue API version number field to v2 in the Phoenix Rogue API settings.
2.  Run `drush ubdb` in drupal to reflect new database changes.

Make sure all information matches up between Phoenix and Rogue databases for below scenarios:
- Create a new signup (RB) and photo (reportback item) from Phoenix web.
- Add another photo (rb item).
- Add another photo (rb item) and update the signup (reportback's quantity and why_participated).
- Only update the signup (reportback's quantity and why_participated).
- Review a photo (reportback item's status as an admin).
- Cron job for reportbacks and reportback items.

A few notes: 
1. If we're adding a new photo AND changing the signup's `quantity` and/or `why_participated`, this is all one event because we don't have a good way differentiate  if it is only adding a new photo or adding new photo and changing `quantity` and/or `why_participated` at the moment e.g. is it sufficient to use character count to see if `why_participated` has changed? This will be for decisions to make in the future cc @ngjo @jessleenyc 
2. In this PR, the signup's `event_id` that gets returned is the original `event_id` when the signup was created, not the `event_id` of when the signup was updated. A new event is still created when a signup update occurs. However, the `event_type` is a `post_photo` since we mentioned not wanting to complicate this now. We don't have a great way of tying this signup update back to the signup. @sbsmith86 @katiecrane Do you think this is ok for now until we make more decisions? 
3. We don't have a `reviewer` column in the `photos` table like we do in the `reportback_items` table and don't have information on who reviewed it. However, when a photo is reviewed, I've changed the `submission_type` of the event to be `admin review`. I can add: 
  - add more info here instead of the hard coded `admin review` (admin's name/email/northstar id/etc) in this column 
  - add `reviewer` to the `photo` table (mimicking the `reportback_items` table)?
  - change `event_type` to `admin review` (wasn't sure about this since we talked about keeping `event_type`s as only `post_photo` for now and make decisions about this later cc @sbsmith86 @katiecrane 
4. When an admin reviews a photo, right now we have no idea which photo it is. It only creates a new event and the only info we have is that it is a `submission_type` `admin_review`. Do we want to add something to the `events` table to refer to which post is it? Then it is a little tricky though because we'd have to add `photo_id` or `video_id` etc? cc @sbsmith86 @katiecrane 

#### Relevant tickets
Fixes #100 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -56,20 +56,25 @@ class PostsController extends ApiController
         $transactionId = incrementTransactionId($request);
 
         // @TODO - Remove. This is temporary. Just hardcoding some params in the request that the client would normally pass. But we assume everything is a photo post from a user at the moment.
-        $request['event_type'] = 'post_photo';
+        // $request['event_type'] = 'post_photo';
         $request['submission_type'] = 'user';
 
         $signup = $this->signups->get($request['northstar_id'], $request['campaign_id'], $request['campaign_run_id']);
 
+        $updating = ! is_null($signup);
+
         // @TODO - should we eventually throw an error if a signup doesn't exist before a post is created? I create one here because we haven't implemented sending signups to rogue yet, so it will have to create a signup record for all posts.
-        if (is_null($signup)) {
+        if (! $updating) {
             $signup = $this->signups->create($request->all());
+
+            // Send the data to the PostService class which will handle determining
+            // which type of post we are dealing with and which repostitory to use to actually create the post.
+            $post = $this->posts->create($request->all(), $signup->id, $transactionId);
+        } else {
+            $post = $this->posts->update($signup, $request->all(), $transactionId);
         }
 
-        // Send the data to the PostService class which will handle determining
-        // which type of post we are dealing with and which repostitory to use to actually create the post.
-        $post = $this->posts->create($request->all(), $signup->id, $transactionId);
-
+        // QUESTION: what do we want to return here per scenario?
         return $this->item($post);
     }
 }
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -8,6 +8,7 @@ use Rogue\Services\PostService;
 use Rogue\Http\Requests\PostRequest;
 use Rogue\Repositories\SignupRepository;
 use Rogue\Http\Transformers\PostTransformer;
+use Rogue\Http\Transformers\PhotoTransformer;
 
 class PostsController extends ApiController
 {
@@ -31,6 +32,11 @@ class PostsController extends ApiController
     protected $transformer;
 
     /**
+     * @var \Rogue\Http\Transformers\PhotoTransformer
+     */
+    protected $photoTransformer;
+
+    /**
      * Create a controller instance.
      *
      * @param  PostContract  $posts
@@ -43,6 +49,7 @@ class PostsController extends ApiController
 
         // Now we have one PostTransformer to handle returning a Post to the API request.
         $this->transformer = new PostTransformer;
+        $this->photoTransformer = new PhotoTransformer;
     }
 
     /**
@@ -111,8 +118,9 @@ class PostsController extends ApiController
             $code = 201;
         }
 
-        dd($reviewedPosts);
+        $meta = [];
 
-        // TODO: how do we want to return this data?
+        // @TODO: we'll need to change the transformer here depending on what type of post.
+        return $this->collection($reviewedPosts, $code, $meta, $this->photoTransformer);
     }
 }
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -2,8 +2,8 @@
 
 namespace Rogue\Http\Controllers\Api;
 
-use Illuminate\Http\Request;
 use Rogue\Models\Signup;
+use Illuminate\Http\Request;
 use Rogue\Services\PostService;
 use Rogue\Http\Requests\PostRequest;
 use Rogue\Repositories\SignupRepository;
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Http\Controllers\Api;
 
+use Illuminate\Http\Request;
 use Rogue\Models\Signup;
 use Rogue\Services\PostService;
 use Rogue\Http\Requests\PostRequest;
@@ -71,9 +72,13 @@ class PostsController extends ApiController
             // which type of post we are dealing with and which repostitory to use to actually create the post.
             $post = $this->posts->create($request->all(), $signup->id, $transactionId);
 
+            $code = 200;
+
             return $this->item($post);
         } else {
             $post = $this->posts->update($signup, $request->all(), $transactionId);
+
+            $code = 201;
 
             if (isset($request['file'])) {
                 return $this->item($post);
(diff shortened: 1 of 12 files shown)


app/Repositories/PhotoRepository.php
```diff
@@ -76,6 +76,71 @@ public function create(array $data, $signupId)
...
+    {
+        // Update the signup's quantity and why_participated data.
+        // We will always update these since we can't tell if this has been changed in a good way yet.
+        $data['quantity_pending'] = $data['quantity'];
+        $signup->fill(array_only($data, ['quantity_pending', 'why_participated']));
+        $signup->save();
+
+        // If there is a file, create a new photo post.
+        if (isset($data['file'])) {
+            return $this->create($data, $signup->id);
```
chloealee: @katiecrane Just wanted to flag this for you in the code! We chatted about this briefly last week re: whether or not we need the `array_key_exists` statement similar to [this line](https://github.com/DoSomething/rogue/blob/master/app/Repositories/ReportbackRepository.php#L134) with a broken photo - let me know if we do!

chloealee:
@sbsmith86 @katiecrane just fixed the style ci fixes - ready for another look when you have a chance!

app/Http/Controllers/Api/PostsController.php
```diff
@@ -56,20 +64,61 @@ public function store(PostRequest $request)
...
+    }
+
+    /**
+     * Update a post(s)'s status when reviewed.
+     *
+     * @param \Illuminate\Http\Request $request
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function reviews(Request $request)
```
sbsmith86: I sorta feel like reviews should eventually have their own controller, repo, form request class, and stuff since we are considering them like a new api resource. I don't want use to conflate the review endpoint logic with the post logic. Fine to do that in a separate Issue/PR though.

app/Http/Requests/PostRequest.php
```diff
@@ -46,7 +46,7 @@ protected function getRules($request)
     {
         switch ($request->event_type) {
             case 'post_photo':
-                return array_merge($this->rules, $this->setPhotoRules());
+                return array_merge($this->rules, $this->getPhotoRules());
```
sbsmith86: 👏  Thanks!

app/Http/Controllers/Api/PostsController.php
chloealee: ah good point @sbsmith86 ! Maybe I can tackle this when I work on #117

app/Repositories/PhotoRepository.php
```diff
@@ -76,6 +76,72 @@ public function create(array $data, $signupId)
...
+     *
+     * @param \Rogue\Models\Photo $signup
+     * @param array $data
+     *
+     * @return \Rogue\Models\Photo
+     */
+    public function update($signup, $data)
+    {
+        // Update the signup's quantity and why_participated data.
+        // We will always update these since we can't tell if this has been changed in a good way yet.
```
sbsmith86: Can you add a `@TODO` here to remove this logic when we are no longer supporting the phoenix-ashes campaign template

app/Repositories/PhotoRepository.php
```diff
@@ -76,6 +76,72 @@ public function create(array $data, $signupId)
...
+        $data['quantity_pending'] = $data['quantity'];
+        $signup->fill(array_only($data, ['quantity_pending', 'why_participated']));
+        $signup->save();
+
+        // If there is a file, create a new photo post.
+        if (isset($data['file'])) {
+            return $this->create($data, $signup->id);
+        } else {
+            // If it doesn't add a new photo, a new event won't be created. Create the event here for an updated signup.
+            // @TODO: right now, an updated signup's event_type is recorded as post_photo. Depending on future decisions, we might want to update this.
```
sbsmith86: Is this TODO still true re: `post_photo` being used as the event type for updating signups?

app/Repositories/PhotoRepository.php
```diff
@@ -76,6 +76,72 @@ public function create(array $data, $signupId)
...
+     * @param array $data
+     *
+     * @return
+     */
+    public function reviews($data)
+    {
+        $reviewedPhotos = [];
+
+        foreach ($data as $review) {
+            if ($review['rogue_event_id'] && ! empty($review['rogue_event_id'])) {
```
sbsmith86: should this be an `isset()` check? Otherwise it looks like both of these statements are checking if it is false. so like `if (isset($review['rogue_event_id']) && ! empty($review['rogue_event_id']))`

sbsmith86: Actually looks like you can just use `empty()` http://stackoverflow.com/questions/4559925/why-check-both-isset-and-empty

app/Repositories/PhotoRepository.php
chloealee: @sbsmith86 yes it is! But I'll change this in #115 - thank you for the reminder!

app/Repositories/PhotoRepository.php
```diff
@@ -76,6 +76,72 @@ public function create(array $data, $signupId)
...
+                $post = Post::where(['event_id' => $review['rogue_event_id']])->first();
+                $photo = Photo::where(['id' => $post->postable_id])->first();
```
sbsmith86: I _think_ you can just use the relationship after you grab the `$post` so something like `$post->content()` should give you the photo record.

app/Services/PostService.php
```diff
@@ -39,7 +40,59 @@ public function create($data, $signupId, $transactionId)
...
+    /**
+     * Handles all business logic around updating the posts(s)'s status after being reviewed.
+     *
+     * @param array $data
+     *
+     * @return
+     */
+    public function reviews($data)
+    {
+        $this->resolvePostRepository($data[0]['event_type']);
```
sbsmith86: is this always where event type will be in the array?

app/Http/Controllers/Api/PostsController.php
sbsmith86: @chloealee sounds good!

app/Services/PostService.php
chloealee: @sbsmith86 since this is a nested array and admins have the ability to review multiple posts at once, I am just checking what type it is for the first post in the array. Right now, they will always be photos since that's the only functionality we have. 

Short answer: Yes. Even if there is only one photo being reviewed, it is still in a nested array. 

Something to think about in the future: will admins be able to review photos and videos (or any other type) all at once? If so, this logic will need to change. 

Should I add a @TODO note here regarding this?

sbsmith86: @chloealee yea that would be a good reminder. I could imagine that in the future.

chloealee: updated!

documentation/README.md
```diff
@@ -2,6 +2,8 @@
 This is the Rogue API, it is used to capture activity from members.
 All `POST` and `PUT` endpoints require an api key (`X-DS-Rogue-API-Key`) in the header to be submitted with the request. 
 
+Refer to this [document](https://github.com/DoSomething/rogue/wiki/API) for the most updated documentation.
```
sbsmith86: ah so I might have been confusing. I think the full documentation, with the params and the expected response should be in here and then link to this documentation from the wiki

chloealee: ah gotcha @sbsmith86 ! Just added more robust documentation and linked the documentation from the wiki.

PR #119, opened 2017-02-02
Test for creating a photo post

sbsmith86:
This does a simple test that we can create a photo when hitting the `/posts` endpoint

In working on this test I found some issues in the code that had to be fixed.

1. There were tests failing on the user functionality trying to log a user in. The middleware was failing when mocking the Northstar class and trying to call `asUser` on it to select which grant type to use. Furnes suggested just updating the gateway package that had work done to not require that call anymore

2. Validation error handling wasn't working correctly in the custom exception hander I created

3. We couldn't rollback migrations after data was in the database because foreign key constraints were failing. 

This PR fixes those 3 things as well as adds the test. 

#### TODO

There are definitely more tests we can write out around, validation, updating, reviewing. But I was thinking this could be just a first step.


chloealee:
this looks great, @sbsmith86 ! I think there are just some StyleCI fixes :)

PR #120, opened 2017-02-03
Reviewing reportback items bug

chloealee:
#### What's this PR do?
- If a post cannot be found by `event_id`, then it is in the old data schema. Send this to the `updateReportbackItems` function and deal with following old logic (when we sent reviewed reportback items to `/items`). 
- Before returning to the controller, remove any Reportback Items objects from `$reviewedPosts` array and put in it's own array, `$reportbackItems`.
- If `$reviewedPosts` is empty, this means all the reviewed items are Reportback Item objects.  Return only if all reviewed items are reportback item objects. Then we will send all to the Reportback Items Controller. 
- If `$reviewedPosts` is not empty, this means there are at least one reviewed items that is a Post object. The `$reviewedPost` array omits reportback item objects so the Post transformer doesn't break and Phoenix puts in `dosomething_rogue_failed_task_log`.  

- Find a `rogue_event_id` in Pheonix's `dosomething_rogue_reportbacks` that doesn't exist in Rogue's `events` table but _does_ exist in the `reportback_items` table. 
- Find the corresponding `fid` and set to `pending`.
- Go to the campaign and review the fid. 
- Check to make sure the corresponding reportback item's status in Rogue is updated.

### DIFF
--- a/app/Repositories/PhotoRepository.php
+++ b/app/Repositories/PhotoRepository.php
@@ -145,8 +145,7 @@ class PhotoRepository
                         return null;
                     }
                 } else {
-                    $formatReview = [$review];
-                    $reportbackItem = $this->reportbackRepository->updateReportbackItems($formatReview);
+                    $reportbackItem = $this->reportbackRepository->updateReportbackItems([$review]);
                     array_push($reviewedPhotos, $reportbackItem);
                 }
             } else {
--- a/app/Repositories/PhotoRepository.php
+++ b/app/Repositories/PhotoRepository.php
@@ -146,9 +146,13 @@ class PhotoRepository
                         return null;
                     }
                 } else {
-                    $formatReview = [$review];
-                    $reportbackItem = $this->reportbackRepository->updateReportbackItems($formatReview);
-                    array_push($reviewedPhotos, $reportbackItem);
+                    if ($review['status'] && ! empty($review['status'])) {
+                        $formatReview = [$review];
+                        $reportbackItem = $this->reportbackRepository->updateReportbackItems($formatReview);
+                        array_push($reviewedPhotos, $reportbackItem);
+                    } else {
+                        return null;
+                    }
                 }
             } else {
                 return null;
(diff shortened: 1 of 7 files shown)


app/Repositories/PhotoRepository.php
```diff
@@ -122,19 +130,24 @@ public function reviews($data)
...
 
-                    array_push($reviewedPhotos, $post->content);
+                        array_push($reviewedPhotos, $post->content);
+                    } else {
+                        return null;
+                    }
                 } else {
-                    return null;
+                    $formatReview = [$review];
+                    $reportbackItem = $this->reportbackRepository->updateReportbackItems($formatReview);
```
sbsmith86: You can do this in one line `$this->reportbackRepository->updateReportbackItems([$review]);`

sbsmith86:
@chloealee Thanks for your work on this! I am a little concerned we are now having Reportback Item classes and functionality mixed in with the Post classes and functionality. I don't want to hold up us getting this bug fixed so that reviewing can resume (and I might not be grasping the complexity of this situation) 

Would it be possible to do something like create a `ReviewController` that gets an incoming request, check the `event_id`, see if it finds it as a post or reportback item, and depending on which one it is, send it to the correct service that kicks off the review flow (including returning the response through the correct transformer) for that thing?

If that is possible, I would love for us to go down something like that route so that we don't have so much to clean up in the future when this won't be a problem.

chloealee:
@sbsmith86 ah that is a great idea and totally agree that it's a little iffy to have Reportback Item classes mixed in with Post classes! I think that's a great solution to create a `ReviewController` and will work on that now / let you know if I run into anything tricky!

sbsmith86:
@chloealee awesome! let me know if I can help in anyway

chloealee:
@sbsmith86 this is updated now with a Reviews Controller! Two notes: 

- This will be updated when I tackle using `reviews` (to add a `resolvePostRepository` function etc.).
- Right now, if first checks if there are reportback item objects and if so, _only_ returns the transformed reportback item objects. If there are reportback item objects _and_ photo objects being reviewed, it will update both, however, will only return the data for the updated reportback item objects.  
- Otherwise, it'll return the updated photo objects.

chloealee:
@sbsmith86 yes! Thank you for this great suggestion!

PR #124, opened 2017-02-09
handle missing images, add signup_id to events table, populate event columns

### LINKED ISSUE #122: Broken images are not handled
_BUG_
-------------------
#### Current Behavior
Broken images are not handled and causes errors.

#### Desired Behavior
Set `file_url` to `default`.

#### Steps to Replicate
Try to send a file with no data.

#### Why This Matters
For the migration 🎉

### LINKED ISSUE #123: Add `signup_id` to `events` table
_FEATURE OVERVIEW_
-------------------
#### User Story
We want to be able to easily get all the activity under a particular signup.

#### Why This Matters
^ and Data!

I think you would have to check every table that has a `signup_id` column and then mush everything together and then you have a ["One could say that...but should one?"](https://www.youtube.com/watch?v=V1U8TXcJwqA) situation.

katiecrane:
#### What's this PR do?
This is several small fixes in one! These are things that were either blocking migration work, or that needed to be fixed to make sure the migration is working correctly.

1. Handles receiving broken images
2. Adds `signup_id` to `events` table and populates this field
3. Populates the `quantity`, `why_participated`, etc. fields in the `events` table for `signup` events

👀 
1. Broken images shouldn't break anything and should get `file_url` set as `default`
2. When a `signup` or `post_photo` event is created, the `signup_id` should appear in the `events` table.
3. When a `signup` event occurs, the submitted data should appear in the `events` table.

#### Relevant tickets
Fixes #122
Fixes #123

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/PhotoRepository.php
+++ b/app/Repositories/PhotoRepository.php
@@ -50,11 +50,11 @@ class PhotoRepository
 
         if (isset($data['file'])) {
             $fileUrl = $this->aws->storeImage($data['file'], $signupId);
+
+            $editedImage = $this->crop($data, $signupId);
         } else {
             $fileUrl = 'default';
         }
-
-        $editedImage = $this->crop($data, $signupId);
 
         $photo = Photo::create([
             'file_url' => $fileUrl,
--- a/app/Repositories/PhotoRepository.php
+++ b/app/Repositories/PhotoRepository.php
@@ -42,6 +42,9 @@ class PhotoRepository
      */
     public function create(array $data, $signupId)
     {
+        // Include signup_id when we create the Event
+        $data['signup_id'] = $signupId;
+
         // Don't send quantity and why_participated - we don't want this to live on the post_photo event.
         $postEvent = Event::create(array_except($data, ['quantity', 'why_participated']));
 
--- a/app/Repositories/PhotoRepository.php
+++ b/app/Repositories/PhotoRepository.php
@@ -45,7 +45,11 @@ class PhotoRepository
         // Don't send quantity and why_participated - we don't want this to live on the post_photo event.
         $postEvent = Event::create(array_except($data, ['quantity', 'why_participated']));
 
-        $fileUrl = $this->aws->storeImage($data['file'], $signupId);
+        if (isset($data['file'])) {
+            $fileUrl = $this->aws->storeImage($data['file'], $signupId);
+        } else {
+            $fileUrl = 'default';
+        }
 
         $editedImage = $this->crop($data, $signupId);
 
--- a/app/Repositories/SignupRepository.php
+++ b/app/Repositories/SignupRepository.php
@@ -45,6 +45,10 @@ class SignupRepository
             $signup->save(['timestamps' => false]);
         }
 
+        // Now that the signup exists, set the signup_id on the signup event
+        $event->signup_id = $signup->id;
+        $event->save();
+
         return $signup;
     }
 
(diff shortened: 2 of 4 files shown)


app/Repositories/PhotoRepository.php
```diff
@@ -42,10 +42,17 @@ public function __construct(AWS $aws, Registrar $registrar)
...
         $postEvent = Event::create(array_except($data, ['quantity', 'why_participated']));
 
-        $fileUrl = $this->aws->storeImage($data['file'], $signupId);
+        if (isset($data['file'])) {
+            $fileUrl = $this->aws->storeImage($data['file'], $signupId);
+        } else {
+            $fileUrl = 'default';
+        }
 
         $editedImage = $this->crop($data, $signupId);
```
chloealee: @katiecrane this is a dumb question but just want to double check! If `$data['file']` is not set, is it possible it'll have crop values? Just want to make sure for when it goes to the [`crop`](edit_image) function that this doesn't break either.

app/Repositories/PhotoRepository.php
```diff
@@ -42,10 +42,17 @@ public function __construct(AWS $aws, Registrar $registrar)
      */
     public function create(array $data, $signupId)
     {
+        // Include signup_id when we create the Event
+        $data['signup_id'] = $signupId;
```
sbsmith86: what is this used for?

app/Repositories/SignupRepository.php
```diff
@@ -43,6 +45,10 @@ public function create($data)
             $signup->save(['timestamps' => false]);
```
sbsmith86: Sorry i know this isn't a part of you PR, but is this saving the signup without timestamps? Why do we want to do this?

katiecrane: @sbsmith86 no worries! this saves the signup with the timestamps that were passed with the rest of the data, if any were. so for the migration we are able to keep the original timestamps.

katiecrane: if no timestamp is passed it saves like normal!

app/Repositories/PhotoRepository.php
katiecrane: @chloealee great question! It hasn't caused any crashes for me, but we definitely don't need to try to set `$editedImage` for broken ones, so I'm pulling it up into the `if`. Good catch!

app/Repositories/PhotoRepository.php
katiecrane: @sbsmith86 It gets passed to `Event::create` with all the other data, since we now have a `signup_id` column on `events`

app/Repositories/SignupRepository.php
```diff
@@ -43,6 +45,10 @@ public function create($data)
...
         }
 
+        // Now that the signup exists, set the signup_id on the signup event
```
sbsmith86: Ah this is getting complicated now. We first create the event, then the signup, the add the signup id to the event.  I wonder if this is something we can refactor later. Just noting it.

sbsmith86:
looks good to me! I will let @chloealee give the final approval once her comments are addressed

app/Repositories/SignupRepository.php
katiecrane: @sbsmith86 agreed, I was trying to figure out a better way to do it too and was blanking, but I haven't slept on it yet

PR #130, opened 2017-02-13
/activity endpoint

### LINKED ISSUE #129: Create /activity endpoint
_FEATURE OVERVIEW_
-------------------
#### User Story
- As a developer, I want to expose the `/activity` endpoint. 
- For now, we want:
   - Pagination
   - Filters:
     - `campaign_id`
     - `campaign_run_id`
     - `count`
     - `page` 
- The return response should look like the sample response in the [Rogue Wiki](https://github.com/DoSomething/rogue/wiki/AP). 

#### Why This Matters
This is important so Quasar can have access to all activity in Rogue and easily get data needed.

chloealee:
#### What's this PR do?
- Creates the `/activity` endpoint that returns information following [this structure](https://github.com/DoSomething/rogue/wiki/API).
  - Two minor changes to follow the same naming as the table column names under `posts`: 
    - Instead of `post_id`, the endpoint returns `postable_id`.
    - Instead of `type`, the endpoing returns `postable_type`.
- Allows unauthenticated users to hit this endpoint. 
- Allows filtering by: 
  - `campaign_id`
  - `campaign_run_id`
  - `limit` (count)
  - `page`
  - Note: to filter by `campaign_id` and `campaign_run_id`, must have `filter` first. `limit` and `page` do not need `filter`.
    - e.g. `/api/v2/activity?filter[campaign_id]=47,49&filter[campaign_run_id]=1771&limit=2`
- Pagination returns 20 records per page. 

Test out `/api/v2/activity` with different filters. Pagination will be at the bottom of the response. Make sure response looks like the sample response in the [Wiki Page](https://github.com/DoSomething/rogue/wiki/API). 

You might need to fake some data (in the `events` table, add manually add a `signup_id`) since this is a new migration. 

#### Relevant tickets
Fixes #129

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/Traits/FiltersRequests.php
```diff
@@ -0,0 +1,55 @@
...
+        $filters = array_intersect_key($filters, array_flip($indexes));
+
+        // You can filter by multiple values, e.g. `filter[source]=agg,cgg`
+        // to get records that have a source value of either `agg` or `cgg`.
+        foreach ($filters as $filter => $values) {
+            $values = explode(',', $values);
+            if (count($values) > 1) {
+                // For the first `where` query, we want to limit results... from then on,
+                // we want to append (e.g. `SELECT * (WHERE _ OR WHERE _ OR WHERE _)` and (WHERE _ OR WHERE _))
+                    $query->where(function ($query) use ($values, $filter) {
```
DFurnes: Ahhh, smart! 👍

DFurnes: Indentation does look a little off on this line, though! 👾

app/Http/Transformers/ActivityTransformer.php
```diff
@@ -0,0 +1,104 @@
...
+     * Include the post.
+     *
+     * @param \Rogue\Models\Signup $signup
+     * @return \League\Fractal\Resource\Collection
+     */
+    public function includePosts(Signup $signup)
+    {
+        $post = $signup->posts;
+
+        return new \League\Fractal\Resource\Collection($post, function ($post) {
```
DFurnes: You should be able to use the existing `PostTransformer` and `EventTransformer` for these!

```php
    /**
     * Include the post.
     *
     * @return League\Fractal\ItemResource
     */
    public function includePosts(Signup $signup)
    {
        $post = $signup->posts;

        return $this->item($post, new PostTransformer);
    }
```

DFurnes:
Definitely not necessary for merge, but it'd also be lovely to see some functional tests that demonstrate this behavior… both for easier reviewing & ensuring we don't get regressions in the future!

app/Http/Transformers/ActivityTransformer.php
chloealee: @DFurnes yes! The reason I didn't use the `PostTransformer` and `EventTransformer` here is because as far as I understand, we might be changing the responses for those endpoints and might even just return status codes instead of full responses. cc @sbsmith86 @katiecrane let me know if I misunderstood though and should change this!

chloealee:
Ah yes, thanks for the reminder, @DFurnes ! I'll talk to @ngjo about making a card for tests!

app/Http/Transformers/ActivityTransformer.php
sbsmith86: @chloealee as far as I know we aren't changing the responses for those (but I guess we could if we find we don't really need to return so much), but also it shouldn't matter since we definitely want to include the whole post and event in this response. So I think using the existing transformers should be fine.

app/Http/routes.php
```diff
@@ -56,3 +56,5 @@
     // signups
     Route::post('signups', 'Api\SignupsController@store');
 });
+
+Route::get('api/v2/activity', 'Api\ActivityController@index');
```
sbsmith86: Is this outside of the `v2` group because of the middleware?

app/Http/Transformers/ActivityTransformer.php
chloealee: @sbsmith86 ok! I'll just need to change the existing transformers around a little to fit the format we want.

app/Http/routes.php
chloealee: @sbsmith86 I put it outside so a user doesn't have to be authorized to hit this endpoint

sbsmith86: @chloealee ah ya. I wonder if we should specify that a route needs the `api` middleware in the controller instead of defining it on the group so that all these can be grouped together.

chloealee: @sbsmith86 great call - changed!

chloealee:
@sbsmith86 @DFurnes @katiecrane also just updated documentation!

app/Http/Controllers/Api/ActivityController.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Rogue\Http\Controllers\Api;
+
+use Rogue\Models\Signup;
+use Illuminate\Http\Request;
+use App\Http\Controllers\Controller;
+// use Rogue\Http\Transformers\ActivityTransformer;
```
katiecrane: Should this get removed?

app/Http/Transformers/EventTransformer.php
```diff
@@ -16,9 +16,17 @@ class EventTransformer extends TransformerAbstract
     public function transform(Event $event)
     {
         return [
-            'event_id' => (string) $event->id,
+            'event_id' => $event->id,
```
sbsmith86: why don't we want to return this as a string anymore?

app/Http/Controllers/Api/ActivityController.php
chloealee: ah yes, thanks for catching that!

app/Http/Transformers/EventTransformer.php
chloealee: ah I actually didn't notice that it was a string there at first! But it's an `INT` in the database, is it better to reflect what it is in the database?

documentation/endpoints/signups.md
```diff
@@ -119,85 +119,64 @@ Example request body:
...
-            }
+            "remote_addr": null,
+            "post_source": null,
+            "created_at": "2017-01-18T20:33:41+0000",
+            "updated_at": "2017-01-18T20:33:41+0000"
           }
         },
+      ],
+    },
+    "events": {
```
katiecrane: So if a Signup has `update_signup` events the will be included here as well?

chloealee: Any event that is associated with that signup id will be in the events section @katiecrane

documentation/endpoints/posts.md
```diff
@@ -40,29 +40,21 @@ Example Response:
 ```
 {
   "data": {
-    "signup_id": 31,
```
sbsmith86: So, this might not be needed for merge since we don't really know what people need, but I feel like we still want to return the signup, northstar id, campaign id, etc that the post is related to.

documentation/endpoints/signups.md
```diff
@@ -119,85 +119,64 @@ Example request body:
...
           "content": {
-            "type": "post_photo",
             "media": {
-              "url": "https://s3.amazonaws.com/ds-rogue-qa/uploads/reportback-items/160-1484853074.jpeg",
-              "edited_url": null
+              "url": "https://s3.amazonaws.com/ds-rogue-prod/uploads/reportback-items/1334-1484771619.jpeg",
+              "edited_url": "https://s3.amazonaws.com/ds-rogue-prod/uploads/reportback-items/edited_1334-1484771621.jpeg"
             },
-            "caption": "LOOK AT THAT CAT",
+            "caption": "caption",
```
sbsmith86: lol

documentation/endpoints/posts.md
chloealee: Totally! I just removed that since we're using all these Transformers now to populate the `/activity` endpoint instead of a separate `ActivityTransformer` and followed the structure that we wanted to be returned according to the wiki :)

documentation/endpoints/signups.md
chloealee: hahah sorry @katiecrane I didn't realize I did that! Was just copying and pasting :)

katiecrane: hahahaha
icon-tongue-cat

DFurnes: 🐱 👀

app/Http/Transformers/EventTransformer.php
sbsmith86: Actually, I think laravel returns it as a string anyway. We might have just been making it explicit. Disregard. 

I think in general things that are like identifiers should be strings since they don't behave like numbers that you would do math on or anything.

documentation/endpoints/signups.md
katiecrane: cool! I was trying to figure out if any of this will change anything for the migration, but we just need to grab `signup_id` and then the event id for the posts, and I thiiiink this actually makes it easier to do so!

sbsmith86:
Looks good to me!

documentation/endpoints/signups.md
chloealee: ok great!!

app/Http/Transformers/EventTransformer.php
chloealee: ah that makes sense, thanks for the explanation of that!

app/Http/Controllers/Api/ActivityController.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Rogue\Http\Controllers\Api;
+
+use Rogue\Models\Signup;
+use Illuminate\Http\Request;
+use App\Http\Controllers\Controller;
```
DFurnes: Is this importing a real class?

app/Http/Controllers/Api/ReviewsController.php
```diff
@@ -43,6 +43,8 @@ class ReviewsController extends ApiController
      */
     public function __construct(PostService $posts, ReportbackService $reportbacks)
     {
+        $this->middleware('api');
```
DFurnes: Sorry if I missed this elsewhere, but what's the rationale for moving the `api` middleware group onto the individual controllers? It seems like it makes it a bit easier to forget.

app/Http/Transformers/SignupTransformer.php
```diff
@@ -26,41 +27,44 @@ class SignupTransformer extends TransformerAbstract
...
      *
      * @param \Rogue\Models\Signup $signup
-     * @return \League\Fractal\Resource\Collection
+     * @return \League\Fractal\Resource\Item
      */
-    public function includePost(Signup $signup)
+    public function includeEvents(Signup $signup)
     {
-        $post = $signup->posts;
+        $event = Event::where('signup_id', $signup->id)->get();
```
DFurnes: Can we use the [`$signup->event` relationship](https://github.com/DoSomething/rogue/blob/6ddc46438d152597553ce418daa1e0997fd4cb4e/app/Models/Signup.php#L16-L22) here? Or if it's a one-to-many, should the model be updated?

app/Http/Controllers/Api/ActivityController.php
chloealee: ah great call! I'll delete :)

app/Http/Controllers/Api/ReviewsController.php
chloealee: @DFurnes that was suggested here by @sbsmith86 so the `Routes` file would be a bit cleaner.

app/Http/Transformers/SignupTransformer.php
chloealee: I think a single `signup` belongs to an `event` but is associated with many `events` through the `post` relationship. Each `post` belongs to an `event` so in turn the `signup` doesn't necessarily `haveMany` `events`, but has many `posts` that have `events`. I'm not sure if I'm thinking about this correctly though? cc @sbsmith86 @katiecrane

app/Http/Controllers/Api/ReviewsController.php
sbsmith86: @DFurnes i thought it was weird that we had to leave routes out of the `v2` group because we didn't want the api middleware to apply.

app/Http/Transformers/SignupTransformer.php
sbsmith86: A signup will have many events now (which is why we put signup id on the events table), post events, update events, review events, etc. So ya we might need to update the relationship on the model to has many. I think we want to be in a world where we can grab `signup->events` and see all the things that happened under a signup.

chloealee: ah I see! Updated both the Signup model and the `SignupTransformer` to reflect this :) @DFurnes @sbsmith86

chloealee:
@DFurnes @sbsmith86 @katiecrane just a note - I think this breaks the `PostAPITest` but I will be fixing this in the PR for the `ActivityTest`s that I write as soon as this is merged!

app/Models/Signup.php
```diff
@@ -26,11 +26,11 @@ class Signup extends Model
     ];
 
     /**
-     * Each signup belongs to an event.
+     * Get the events associated with this signup.
      */
     public function event()
```
DFurnes: I'm sorry this is kinda nitpicky, but this should be `events()` now that we're returning a collection. 😬

app/Http/Transformers/SignupTransformer.php
```diff
@@ -26,10 +26,12 @@ class SignupTransformer extends TransformerAbstract
      */
     public function transform(Signup $signup)
     {
+        $signup_event = Event::where('id', $signup->event_id)->get();
+
         return [
             'signup_id' => $signup->id,
             'signup_event_id' => $signup->event_id,
-            'submission_type' => $signup->event->submission_type,
+            'submission_type' => $signup_event[0]->submission_type,
```
DFurnes: You should be able to just grab the first item from the relationship, rather than re-querying. 

```php
$signup->events->first()->submission_type
```

PR #139, opened 2017-02-21
Activity tests

### LINKED ISSUE #136: /activity unit tests
write unit tests for the new `/activity` endpoint

chloealee:
#### What's this PR do?
- Fixes `PostApiTest` to pass according to what is newly returned (changes to Transformer).
- Uses `DatabaseMigrations` in `TestCase.php`
- Adds unit tests for `/activity` endpoint to test:
  - `limit` query param
  - `page` query param
  - `campaign_id` query param
  - `campaign_run_id` query param
  - combination of query params with expected response to not return anything

- Run `vendor/bin/phpunit` to make sure tests are passing

#### Relevant tickets
Fixes #136 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -26,7 +26,6 @@ class SignupTransformer extends TransformerAbstract
      */
     public function transform(Signup $signup)
     {
-        dd($signup->events);
         return [
             'signup_id' => $signup->id,
             'signup_event_id' => $signup->event_id,
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -26,6 +26,7 @@ class SignupTransformer extends TransformerAbstract
      */
     public function transform(Signup $signup)
     {
+        dd($signup->events);
         return [
             'signup_id' => $signup->id,
             'signup_event_id' => $signup->event_id,
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -3,9 +3,9 @@
 use Faker\Generator;
 use Rogue\Models\User;
 use Rogue\Models\Event;
+use Rogue\Models\Signup;
 use Rogue\Models\Reaction;
 use Rogue\Models\Reportback;
-use Rogue\Models\Signup;
 
 /*
 |--------------------------------------------------------------------------
@@ -18,7 +18,7 @@ use Rogue\Models\Signup;
 |
 */
 
-// Event - Signup Factory
+// Event Factory
 $factory->define(Event::class, function (Generator $faker) {
     return [
         'northstar_id' => str_random(24),
@@ -31,7 +31,7 @@ $factory->define(Signup::class, function (Generator $faker) {
     return [
         'campaign_id' => $this->faker->randomNumber(4),
         'campaign_run_id' => $this->faker->randomNumber(4),
-        'event_id' => function() {
+        'event_id' => function () {
             return factory(Event::class)->create()->id;
         },
     ];
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -5,6 +5,7 @@ use Rogue\Models\User;
 use Rogue\Models\Event;
 use Rogue\Models\Reaction;
 use Rogue\Models\Reportback;
+use Rogue\Models\Signup;
 
 /*
 |--------------------------------------------------------------------------
@@ -21,6 +22,18 @@ use Rogue\Models\Reportback;
 $factory->define(Event::class, function (Generator $faker) {
     return [
         'northstar_id' => str_random(24),
+        'submission_type' => 'signup',
+    ];
+});
+
+// Signup Factory
+$factory->define(Signup::class, function (Generator $faker) {
+    return [
+        'campaign_id' => $this->faker->randomNumber(4),
+        'campaign_run_id' => $this->faker->randomNumber(4),
+        'event_id' => function() {
+            return factory(Event::class)->create()->id;
+        },
     ];
 });
 
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -2,9 +2,9 @@
 
 use Faker\Generator;
 use Rogue\Models\User;
+use Rogue\Models\Event;
 use Rogue\Models\Reaction;
 use Rogue\Models\Reportback;
-use Rogue\Models\Event;
 
 /*
 |--------------------------------------------------------------------------
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -18,7 +18,7 @@ use Rogue\Models\Signup;
 |
 */
 
-// Event Factory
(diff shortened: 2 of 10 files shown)


tests/ActivityApiTest.php
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Test for retrieving a user's activity with limit query param.
+     *
+     * GET /activity?limit=8
+     * @return void
+     */
+    public function testActivityIndexWithLimitQuery()
+    {
+        $response = $this->json('GET', $this->activityApiUrl . '?limit=8');
```
DFurnes: The `$this->json` method returns the `TestCase` as a "self" reference, so it shouldn't be saved as `$response`.

tests/ActivityApiTest.php
```diff
@@ -0,0 +1,127 @@
...
+
+        $this->assertResponseStatus(200);
+
+        $response = $this->decodeResponseJson();
+
+        $this->assertEquals(8, $response['meta']['pagination']['per_page']);
```
DFurnes: This is totally subjective, but I really like the `seeJsonSubset` test method for stuff like this. It feels like it makes it a bit easier to see exactly what you're expecting the response to look like:

```php
$this->json('GET', $this->activityApiUrl . '?limit=8');

$this->seeJsonSubset([
  'meta' => [
    'pagination' => [
      'per_page' => 8
    ]
  ]
]);  
```

tests/ActivityApiTest.php
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Test for retrieving a user's activity with campaign_id query param.
+     *
+     * GET /activity?filter[campaign_id]=47
+     * @return void
+     */
+    public function testActivityIndexWithCampaignIdQuery()
+    {
+        $event = factory(Event::class)->create();
```
DFurnes: You should create the corresponding signup as part of the model factory, so you don't have to do it by hand in each test method! Check out the "Relations & Attribute Closures" section in the [Using Factories](https://laravel.com/docs/5.4/database-testing#using-factories) documentation.

tests/TestCase.php
```diff
@@ -1,9 +1,14 @@
 <?php
 
 use Rogue\Jobs\SendReportbackToPhoenix;
+use Illuminate\Foundation\Testing\DatabaseMigrations;
 
 abstract class TestCase extends Illuminate\Foundation\Testing\TestCase
 {
+    use DatabaseMigrations;
+
+    static protected $isInitialTest = true;
```
DFurnes: You can remove this for now! We added it when testing out that database transactions stuff, but it's not doing anything anymore.

tests/ActivityApiTest.php
chloealee: oohh yes I love this!

database/factories/ModelFactory.php
```diff
@@ -16,6 +18,25 @@
 |
 */
 
+// Event Factory
+$factory->define(Event::class, function (Generator $faker) {
+    return [
+        'northstar_id' => str_random(24),
+        'submission_type' => 'signup',
```
chloealee: @DFurnes hardcoding this for now since we need it in the SignupTransformer

database/factories/ModelFactory.php
```diff
@@ -16,6 +18,25 @@
...
+    return [
+        'campaign_id' => $this->faker->randomNumber(4),
+        'campaign_run_id' => $this->faker->randomNumber(4),
+        'event_id' => function() {
+            return factory(Event::class)->create()->id;
+        },
+    ];
+});
+
 // Reportback Factory
```
chloealee: @DFurnes this doesn't seem to work quite right at associating `$signup->events`

app/Http/Transformers/SignupTransformer.php
```diff
@@ -26,6 +26,7 @@ class SignupTransformer extends TransformerAbstract
      */
     public function transform(Signup $signup)
     {
+        dd($signup->events);
```
chloealee: @DFurnes It looks like the Relations and Attribute Closure I made in the ModelFactory isn't quite right. It gets mad and returns this as an empty collection and breaks with `'submission_type' => $signup->events->first()->submission_type,` below.

database/factories/ModelFactory.php
DFurnes: Makes sense! In the future it could make sense to have different [factory states](https://laravel.com/docs/5.4/database-testing#writing-factories) for each "type" of event. For example, we do this in Northstar so we can easily make User models w/ different roles.

database/factories/ModelFactory.php
DFurnes: Ahhh, yeah! 💥

I think the problem is that there's no `signup_id` set on the embedded event, so when the "has many" relationship tries to look for events with that column, it can't find any.

app/Http/Transformers/SignupTransformer.php
DFurnes: Yeah, I'm a little confused by the relationship between events and signups. The way we're trying to set a single `event_id` column in the Signup factory makes sense for a "belongs to" relationship (where each signup belongs to a single event).

For a "has many" relationship, I think you'd want to do the chained `->each()` calls on the returned factory to make the events like you suggested on Slack.

DFurnes: That being said I'm a little confused by how all these relationships are set up. It seems like a kinda tangled web. 🕸 It might be nice to chat it out IRL at some point.

chloealee: ok thanks, Dave!! I will give that a shot and yes would love to chat more IRL about the relationships!

tests/ActivityApiTest.php
```diff
@@ -0,0 +1,133 @@
...
+     * GET /activity?filter[campaign_id]=47
+     * @return void
+     */
+    public function testActivityIndexWithCampaignIdQuery()
+    {
+        $signup = factory(Signup::class)->create();
+        $event = Event::all()->first();
+        $event->signup_id = $signup->id;
+        $event->save();
+
```
chloealee: @DFurnes updated and got it to work this way. I couldn't use something like this to work:

```
$users = factory(App\User::class, 3)
           ->create()
           ->each(function ($u) {
                $u->posts()->save(factory(App\Post::class)->make());
            });
```

because in the Signup Factory, if I don't have 
```        
        'event_id' => function() {
            return factory(Event::class)->create()->id;
        },
```
it gets mad but then I think we are creating two events which is incorrect too? I'm not sure if the original way reads cleaner and is safer to create a signup instead of `Event::all()first->();` line?

DFurnes: Yeah, `Event::all()->first()` definitely seems a little dangerous. I'm still a little confused how the relationship between these two records works... it seems weird that both tables need a `RELATED_id` column to mark the relation.

Let's go with what you have here for now, but make a ticket to revisit later?

chloealee: Yeah it is a little strange - I'll make a ticket now. Thanks, Dave!

PR #147, opened 2017-02-28
Reviews table

### LINKED ISSUE #113: /reviews unit tests
write unit tests for the `/reviews` endpoint

### LINKED ISSUE #117: Send reviews information to the reviews table
_BUG_
-------------------
#### Current Behavior
- Right now, we are not utilizing the reviews table when reviews are being made. Instead, we are just creating an `review_photo` event type and adding it to the events table. 

#### Desired Behavior
- We want to track these in the reviews table.

chloealee:
#### What's this PR do?
- When a photo post is reviewed, a review is added to the reviews table. 
- Updates `Reviews` table to include `postable_id` and `postable_type`.
- Updates relations between reviews and posts.
- Updates documentation (now this endpoint requires a `reviewer` and `event_type`).

- Post a review to the `/reviews` endpoint. 
  - Make sure an event with the  correct information is created.
  - Make sure a review with the correct information is created.

#### Relevant tickets
Fixes #117 
Fixes #113 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -105,9 +105,9 @@ class PostsController extends ApiController
      */
     public function reviews(Request $request)
     {
-        $reviewedPosts = $this->posts->reviews($request->all());
+        $reviewedPost = $this->posts->reviews($request->all());
 
-        if (empty($reviewedPosts)) {
+        if (empty($reviewedPost)) {
             $code = 404;
         } else {
             $code = 201;
@@ -116,6 +116,6 @@ class PostsController extends ApiController
         $meta = [];
 
         // @TODO: we'll need to change the transformer here depending on what type of post.
-        return $this->collection($reviewedPosts, $code, $meta, $this->photoTransformer);
+        return $this->item($reviewedPost, $code, $meta, $this->photoTransformer);
     }
 }
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -8,7 +8,6 @@ use Rogue\Services\PostService;
 use Rogue\Http\Requests\PostRequest;
 use Rogue\Repositories\SignupRepository;
 use Rogue\Http\Transformers\PostTransformer;
-use Rogue\Http\Transformers\PhotoTransformer;
 
 class PostsController extends ApiController
 {
@@ -32,11 +31,6 @@ class PostsController extends ApiController
     protected $transformer;
 
     /**
-     * @var \Rogue\Http\Transformers\PhotoTransformer
-     */
-    protected $photoTransformer;
-
-    /**
      * Create a controller instance.
      *
      * @param  PostContract  $posts
@@ -51,7 +45,6 @@ class PostsController extends ApiController
 
         // Now we have one PostTransformer to handle returning a Post to the API request.
         $this->transformer = new PostTransformer;
-        $this->photoTransformer = new PhotoTransformer;
     }
 
     /**
@@ -115,7 +108,6 @@ class PostsController extends ApiController
 
         $meta = [];
 
-        // @TODO: we'll need to change the transformer here depending on what type of post.
-        return $this->item($reviewedPost, $code, $meta, $this->photoTransformer);
+        return $this->item($reviewedPost, $code, $meta, $this->transformer);
     }
 }
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -88,26 +88,4 @@ class PostsController extends ApiController
             }
         }
     }
-
-    /**
-     * Update a post(s)'s status when reviewed.
-     *
-     * @param \Illuminate\Http\Request $request
-     *
-     * @return \Illuminate\Http\Response
-     */
-    public function reviews(Request $request)
-    {
-        $reviewedPost = $this->posts->reviews($request->all());
-
-        if (empty($reviewedPost)) {
-            $code = 404;
-        } else {
-            $code = 201;
-        }
-
-        $meta = [];
-
-        return $this->item($reviewedPost, $code, $meta, $this->transformer);
-    }
 }
--- a/app/Http/Controllers/Api/PostsController.php
+++ b/app/Http/Controllers/Api/PostsController.php
@@ -105,11 +105,6 @@ class PostsController extends ApiController
      */
     public function reviews(Request $request)
     {
-        $this->validate($request, [
-            '*.rogue_event_id' => 'required',
-            '*.status' => 'required',
-        ]);
-
         $reviewedPosts = $this->posts->reviews($request->all());
 
         if (empty($reviewedPosts)) {
--- a/app/Http/Controllers/Api/ReviewsController.php
+++ b/app/Http/Controllers/Api/ReviewsController.php
@@ -74,10 +74,9 @@ class ReviewsController extends ApiController
             $post = Post::where(['event_id' => $review['rogue_event_id']])->first();
             array_push($photos, $review);
         }
-        if ($photos) {
-            $reviewedPhotos = $this->posts->reviews($photos);
-            $reviewedPhotosCode = $this->code($reviewedPhotos);
-        }
+
+        $reviewedPhotos = $this->posts->reviews($photos);
+        $reviewedPhotosCode = $this->code($reviewedPhotos);
 
         $meta = [];
 
--- a/app/Http/Controllers/Api/ReviewsController.php
+++ b/app/Http/Controllers/Api/ReviewsController.php
@@ -77,18 +77,12 @@ class ReviewsController extends ApiController
                 array_push($photos, $review);
             } else {
                 $review['rogue_reportback_item_id'] = $review['rogue_event_id'];
-                array_push($reportbackItems, array_only($review, ['rogue_reportback_item_id', 'status']));
+                array_push($reportbackItems, array_only($review, ['rogue_reportback_item_id', 'status', 'reviewer']));
             }
         }
 
         if ($reportbackItems) {
-            //Create a new review for each of the reviewed items.
-            foreach ($reportbackItems as $reportbackItem) {
-                $review = Review::create()
-            }
-
             $reviewedReportbackItems = $this->reportbacks->updateReportbackItems($reportbackItems);
-
             $reviewedReportbackItemsCode = $this->code($reviewedReportbackItems);
         }
 
--- a/app/Http/Controllers/Api/ReviewsController.php
+++ b/app/Http/Controllers/Api/ReviewsController.php
@@ -64,7 +64,7 @@ class ReviewsController extends ApiController
         $this->validate($request, [
             '*.rogue_event_id' => 'required',
             '*.status' => 'required',
-            '*.admin_northstar_id' => 'required',
+            '*.reviewer' => 'required',
         ]);
 
         $reportbackItems = [];
@@ -84,7 +84,7 @@ class ReviewsController extends ApiController
         if ($reportbackItems) {
             //Create a new review for each of the reviewed items.
             foreach ($reportbackItems as $reportbackItem) {
-                $event = Review::create()
+                $review = Review::create()
             }
 
             $reviewedReportbackItems = $this->reportbacks->updateReportbackItems($reportbackItems);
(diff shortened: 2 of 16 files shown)


app/Repositories/PhotoRepository.php
```diff
@@ -154,12 +155,44 @@ public function reviews($data)
...
                 if ($review['status'] && ! empty($review['status'])) {
                     // @TODO: update to add more details in the event e.g. admin who reviewed, admin's northstar id, etc.
                     $review['submission_type'] = 'admin';
 
-                    Event::create($review);
+                    // Create the Event.
+                    $event = Event::create([
+                        'signup_id' => $post->signup_id,
+                        // Do we want the user's northstar id or the admin's?
+                        'northstar_id' => $post->northstar_id,
```
chloealee: @sbsmith86 @katiecrane @DFurnes do we want the user's northstar id or the admin's? Right now, this is the user's but I'm thinking maybe it makes more sense to have the admin's?

app/Repositories/ReportbackRepository.php
```diff
@@ -179,11 +181,48 @@ public function updateReportbackItems($data)
...
 
                 if ($reportbackItem['status'] && ! empty($reportbackItem['status'])) {
+                    // @TODO: update to add more details in the event e.g. admin who reviewed, admin's northstar id, etc.
+                    $reportbackItem['submission_type'] = 'admin';
+
+                    // Create the Event.
+                    $event = Event::create([
+                        'signup_id' => 0,
+                        // Do we want the user's northstar id or the admin's?
+                        'northstar_id' => $rb->northstar_id,
```
chloealee: @sbsmith86 @katiecrane @DFurnes same with this! Should we change it to the admin's northstar id?

app/Repositories/ReportbackRepository.php
```diff
@@ -179,11 +181,47 @@ public function updateReportbackItems($data)
         foreach ($data as $reportbackItem) {
```
DFurnes: Do we want to allow mass-reviewing reportbacks? It doesn't feel super RESTful (which isn't the end of the world), and it also seems like we might be working ourselves into the same corner we were in in Phoenix (where multiple reviewers couldn't safely be working on the same campaign without conflicts).

app/Models/ReportbackItem.php
```diff
@@ -28,4 +28,12 @@ public function reactions()
     {
         return $this->hasMany('\Rogue\Models\Reaction');
     }
+
+    /**
+     * Each reportback item has one review.
+     */
+    public function review()
```
sbsmith86: @DFurnes yea they are. ReportbackItem was from the old DB structure and `Post` is the new. At one point we had reviews coming in from both so we had to continue to support both structures which is one of the reasons we turned off sending to rogue so that we could just get all of the reportbacks into rogue as a `Post`. So I don't think this PR needs to support `ReportbackItem`s anymore @chloealee

app/Repositories/ReportbackRepository.php
sbsmith86: @DFurnes in the new interface, you would just review one at a time, but in Phoenix you review a bunch and then submit. This is one of the reasons why I brought up what I did in Slack. I would like to not have to support the Phoenix interface at all in Rogue.

app/Models/ReportbackItem.php
chloealee: @sbsmith86 ah totally right I wasn't thinking - will update now!

app/Repositories/ReportbackRepository.php
ngjo: yes to what dave said! in rogue we'd want multiple reviewers to be able to safely work on the same campaign w/o conflicts

chloealee:
@DFurnes @sbsmith86 @katiecrane finally finished with the unit tests. This is ready for another 👀 !

chloealee:
Right now, Phoenix no longer sends over the `reviewer` to this endpoint but this PR now requires it. We don't have to update Phoenix since it'll never use this endpoint anymore, right? cc @DFurnes @sbsmith86 @katiecrane

app/Http/Controllers/Api/ReviewsController.php
```diff
@@ -76,13 +77,12 @@ public function reviews(Request $request)
                 array_push($photos, $review);
             } else {
                 $review['rogue_reportback_item_id'] = $review['rogue_event_id'];
-                array_push($reportbackItems, array_only($review, ['rogue_reportback_item_id', 'status']));
+                array_push($reportbackItems, array_except($review, ['rogue_event_id']));
```
katiecrane: @chloealee is this still used ever?

chloealee: @katiecrane after our conversation, I guess it won't be since we won't ever be reviewing in Phoenix and sending to Rogue! Should I delete all the reportback item logic?

katiecrane: @chloealee my understanding is that we can get rid of all of that

app/Repositories/PhotoRepository.php
```diff
@@ -154,17 +155,48 @@ public function reviews($data)
...
                     $review['submission_type'] = 'admin';
 
-                    Event::create($review);
-
-                    $post->content->status = $review['status'];
-                    $post->content->save();
-
-                    array_push($reviewedPhotos, $post->content);
+                    // Create the Event.
+                    $event = Event::create([
```
sbsmith86: should we hold off on this until we do the work of moving the even creation logic into a model event? Or we can move forward and fix later.

app/Repositories/ReportbackRepository.php
sbsmith86: @chloealee yea i think we shouldn't allow this.

app/Repositories/ReportbackRepository.php
```diff
@@ -179,11 +179,14 @@ public function updateReportbackItems($data)
...
             if (isset($reportbackItem['rogue_reportback_item_id']) && ! empty($reportbackItem['rogue_reportback_item_id'])) {
```
sbsmith86: And we don't need to support reportback item ids anymore.

app/Repositories/PhotoRepository.php
chloealee: @sbsmith86 I don't feel strongly either way! Maybe keep for now since there is a test to make sure an event is created and we'll want to keep this test even when we switch to creation logic into a model event?

app/Repositories/ReportbackRepository.php
chloealee: @sbsmith86 @katiecrane removed all the added reportback item logic!

app/Http/Controllers/Api/ReviewsController.php
```diff
@@ -67,35 +67,21 @@ public function reviews(Request $request)
...
-            }
+            array_push($photos, $review);
         }
-
-        if ($reportbackItems) {
-            $reviewedReportbackItems = $this->reportbacks->updateReportbackItems($reportbackItems);
-            $reviewedReportbackItemsCode = $this->code($reviewedReportbackItems);
-        }
-
         if ($photos) {
```
katiecrane: @chloealee is this always true?

tests/ReviewsApiTest.php
```diff
@@ -0,0 +1,142 @@
...
+        ]);
+    }
+
+    /** Test that a PUT request to /reviews works for multiple reviews.
+     *
+     *
+     * PUT /reviews
+     * @return void
+     */
+    public function testUpdatingMultipleReviews()
```
sbsmith86: So I'm not sure this is functionality we really want. In the new interface, an admin would hit approve/reject one at a time instead of how it works now where you change a bunch of statuses and then submit them. We storta talked about it in a comment chain before. I don't know if it is definitely need to merge this PR but we should only expect to be reviewing posts one at a time.

app/Repositories/ReportbackRepository.php
```diff
@@ -179,14 +179,9 @@ public function updateReportbackItems($data)
...
                 $rbItem = ReportbackItem::where(['id' => $reportbackItem['rogue_reportback_item_id']])->first();
```
katiecrane: @chloealee does this part get used?

chloealee: @katiecrane this is old code!

tests/ReviewsApiTest.php
chloealee: ah that's right - sorry! Should I go ahead and update that here @sbsmith86 ?

app/Repositories/ReportbackRepository.php
katiecrane: @chloealee I don't know how I did this so sorry

tests/ReviewsApiTest.php
chloealee: I may as well - I'll do that now!

app/Http/Controllers/Api/ReviewsController.php
```diff
@@ -55,48 +55,19 @@ public function __construct(PostService $posts, ReportbackService $reportbacks)
...
      *
      * @return \Illuminate\Http\Response
      */
-    public function reviews(Request $request)
+    public function reviews(ReviewsRequest $request)
     {
-        $this->validate($request, [
-            '*.rogue_event_id' => 'required',
-            '*.status' => 'required',
-        ]);
```
chloealee: @sbsmith86 @katiecrane moved this out to it's own `ReviewsRequest.php` since we aren't dealing with multiple reportback items anymore. 

@sbsmith86 I got it to work (you were right, my headers were wrong in Postman!) to return a `500` response code. However, the message is not filled in automatically like it is when I don't follow validation rules in PostRequest. I also didn't follow validation rules in the ReactionRequest and no message is returned there either, just a `500` code. Is this ok to just return a `500` or should it be also returning the message?

app/Http/Requests/ReviewsRequest.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
chloealee: @sbsmith86 @katiecrane here is the new validation (instead of in the Reivews Controller!)

app/Http/Controllers/Api/PostsController.php
```diff
@@ -121,6 +116,6 @@ public function reviews(Request $request)
         $meta = [];
 
         // @TODO: we'll need to change the transformer here depending on what type of post.
-        return $this->collection($reviewedPosts, $code, $meta, $this->photoTransformer);
+        return $this->item($reviewedPost, $code, $meta, $this->photoTransformer);
```
sbsmith86: We have the `PostTransformer` that should be used across all post types.

tests/ReviewsApiTest.php
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * PUT /reviews
+     * @return void
+     */
+    public function testUpdatingASingleReview()
+    {
+        // Create a photo.
+        $photo = factory(Photo::class)->create();
+
+        // Create a post and associate the photo with it.
```
sbsmith86: I wonder if we should move this work into a model event also. So when a type of thing is created we also automatically create the post for it. Not needed for merge, just a thought.

sbsmith86:
Looks good! I just had one comment about using the `PostTransformer`instead of the `PhotoTransformer`

chloealee:
Awesome, thanks @sbsmith86 ! Just took out the photo transformer and replaced with the post transformer (thanks for that!) and updated documentation.

app/Http/Controllers/Api/ReviewsController.php
```diff
@@ -3,9 +3,9 @@
 namespace Rogue\Http\Controllers\Api;
 
 use Rogue\Models\Post;
-use Illuminate\Http\Request;
 use Rogue\Services\PostService;
 use Rogue\Services\ReportbackService;
+use Rogue\Http\Requests\ReviewsRequest;
 use Rogue\Http\Transformers\PhotoTransformer;
```
sbsmith86: Sorrrrry 😬  I think you can use the `PostTransformer` here to?

app/Http/Controllers/Api/PostsController.php
chloealee: @sbsmith86 I actually don't think we even need this here? I don't think it's being used!

chloealee: wait sorry - I am totally confusing myself. It should go through to the Post Controller but it looks like it skips and goes right to the Post Service. I will fix this!

chloealee: @sbsmith86 or is it ok for it to go straight to the Post Service?

chloealee:
thanks for being so patient @sbsmith86 ! Just updated so it returns a post and uses the post transformer and also updated the test and documentation to reflect this!

PR #150, opened 2017-03-02
Exposes total reaction count in GET /activity endpoint

### LINKED ISSUE #149: Return total reaction count for posts in GET /activity response
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want to expose the number of reactions for a post in the `GET /activity` endpoint response.

chloealee:
#### What's this PR do?
- Updates relation of reaction to be with photo instead of post. 
- Updates `POST /reactions` endpoint to accept `reactionable_id` and `reactionable_type` instead of `postable_id` and `postable_type`. 
- Re-names `postable_id` and `postable_type` columns in the reactions table to be `reactionable_id` and `reactionable_type`.
- Exposes total reaction count in the `GET /activity` endpoint response. 
- Updates documenation.

- Hit `GET /activity` endpoint and the response should look like this: 
```
{
  "data": [
    {
      "signup_id": 66,
      "signup_event_id": 814,
      "submission_type": "user",
      "northstar_id": "5589c991a59dbfa93d8b45ae",
      "campaign_id": 47,
      "campaign_run_id": 1771,
      "quantity": null,
      "quantity_pending": null,
      "why_participated": null,
      "signup_source": "chloe-testing",
      "created_at": "2017-02-16T17:42:41+0000",
      "updated_at": "2017-02-16T17:42:41+0000",
      "posts": {
        "data": [
          {
            "postable_id": 352,
            "post_event_id": 816,
            "submission_type": "user",
            "postable_type": "Rogue\\Models\\Photo",
            "content": {
              "media": {
                "url": "https://s3.amazonaws.com/ds-rogue-prod/uploads/reportback-items/18-1487193185.jpeg",
                "edited_url": "https://s3.amazonaws.com/ds-rogue-prod/uploads/reportback-items/edited_18-1487193185.jpeg"
              },
              "caption": "new1",
              "status": "still working?",
              "total_reactions": 3,
              "remote_addr": "207.110.19.130",
              "post_source": "runscope",
              "created_at": "2017-02-15T21:13:05+0000",
              "updated_at": "2017-03-01T20:14:45+0000"
            }
          }
        ]
      },
      "events": {
        "data": [
          {
            "event_id": 814,
            "event_type": "signup",
            "submission_type": "user",
            "quantity": null,
            "quantity_pending": null,
            "why_participated": null,
            "caption": null,
            "status": null,
            "source": "chloe-testing",
            "remote_addr": null,
            "reason": null,
            "created_at": "2017-02-16T17:42:41+0000",
            "updated_at": "2017-02-16T17:42:41+0000"
          }
        ]
      }
    }
  ],
  "meta": {
    "pagination": {
      "total": 1,
      "count": 1,
      "per_page": 20,
      "current_page": 1,
      "total_pages": 1,
      "links": []
    }
  }
}
```
- Hit the `POST /reactions` endpoint with `reactionable_id` and `reactionable_type` instead of `postable_id` and `postable_type` and you should receive a response that looks like: 
```
{
  "data": {
    "reaction_id": "29",
    "northstar_id": "12345",
    "reactionable_id": "88",
    "reactionable_type": "photo",
    "created_at": {
      "date": "2017-03-14 20:58:07.000000",
      "timezone_type": 3,
      "timezone": "UTC"
    },
    "updated_at": {
      "date": "2017-03-15 15:23:43.000000",
      "timezone_type": 3,
      "timezone": "UTC"
    },
    "deleted_at": {
      "date": "2017-03-15 15:26:57.000000",
      "timezone_type": 3,
      "timezone": "UTC"
    }
  },
  "meta": {
    "action": "unliked",
    "total_reactions": 2
  }
}
```
#### Relevant tickets
Fixes #149 

#### Checklist
- [ ] Tested on staging.


app/helpers.php
```diff
@@ -33,6 +34,18 @@ function edit_image($image, $coords)
...
 /**
+ * Query for total reactions for a post.
+ *
+ * @param int $postableId
+ * @param int $postableType
+ * @return int total count
+ */
+function getTotalPostReactions($postableId, $postableType)
+{
+    return Reaction::where(['postable_id' => $postableId, 'postable_type' => $postableType])->count();
```
DFurnes: This will make a separate "count" query for each model in the results (so if there are 5 posts returned from a request, it'll make 5 separate SQL queries to get each of their reactions). This is pretty rough from a performance standpoint (imagine if there were 50 posts being requested!)

Try looking into querying the posts [with the count of reactions](https://laravel.com/docs/5.4/eloquent-relationships#counting-related-models).

chloealee: @DFurnes ah great point! I'm not sure if this works with the polymorphic relation though? When I do 
```
    return Post::withCount(['reactions' => function ($query) {
        $query->where(['postable_id' => $postableId, 'postable_type' => $postableType]);
    }])->get();
```
I get the error
```
{
  "error": {
    "code": 500,
    "message": "Array to string conversion"
  }
}
```
It gets mad in the HasOneOrMany.php file, with `$this->localKey` (line 412) being 
```
array:1 [
  0 => "event_id"
]
``` 
instead of an expected string. I think it's because of the polymorphic relationship after reading more about `$this->localKey` [here](http://stackoverflow.com/questions/26803411/what-is-local-key-in-laravel-eloquent)?

chloealee: Actually, I'm thinking now that there's an issue with the relations between the Reaction and the Post?

app/Models/Post.php
```diff
@@ -53,7 +53,22 @@ public function signup()
      */
     public function reactions()
     {
-        return $this->hasMany(Reaction::class);
+        // return $this->morphToMany(Reaction::class, 'postable');
+        return $this->morphMany(Reaction::class, 'postable');
+    }
```
chloealee: @DFurnes we update the model relation last week but after looking at the docs and the examples, I think that we actually don't have a many to many polymorphic relation and just a regular polymorphic relation between post and reactions. Looking at the [polymorphic example](https://laravel.com/docs/5.2/eloquent-relationships#polymorphic-relations), it seems like our reactions are more like the example's "likes" rather than the "tags" in the [many to many polymorphic example](https://laravel.com/docs/5.2/eloquent-relationships#many-to-many-polymorphic-relations) since we don't have different reaction types like a tag will have different tags. 

I wanted to check with you to see if this seems reasonable to you before going down a rabbit hole? The only thing is when I die dump `$post->reactions` in the Post Transformer, I'm still getting this error 
```
local.ERROR: exception 'ErrorException' with message 'array_key_exists(): The first argument should be either a string or an integer' in /home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:2633
``` 
so maybe I'm still a little off with the relations?

app/Models/Reaction.php
```diff
@@ -26,8 +26,9 @@ class Reaction extends Model
     /**
      * Each reaction belongs to a post.
      */
-    public function post()
+    public function postable()
     {
-        return $this->belongsTo(Post::class);
+        // return $this->morphedByMany(Post::class, 'postable');
+        return $this->morphTo();
```
chloealee: @DFurnes also updated the relationship here!

app/Models/Post.php
DFurnes: > I wanted to check with you to see if this seems reasonable to you before going down a rabbit hole?

Yes! I think you're 100% on the right track. Are you getting the `array_key_exists` error if you switch to plain one-to-one "morph"? It looks like that's happening in Eloquent's `getAttribute` method, so it'd make sense if it was just because it was looking for a nonexistent column.

DFurnes: Do the arguments for the failing `getAttribute` method call appear in the stack trace for that exception? If so, that might also give us some good ideas.

chloealee: @DFurnes ah progress! When I change to just `return $this->morphTo(Reaction::class, 'postable');` it returns `null` when I die dump `$post->reactions` which makes sense however I added a reaction for a post and it still says `null`. 

With `return $this->morphMany(Reaction::class, 'postable');` and the `array_key_exists` error, the `$key` is `id`, which is a string so doesn't quite add up to the error message?

chloealee: ah that would make sense too like you were saying that it's trying to get `id` of a Post but a Post doesn't have an `id`

chloealee: hm - right now we have `postable_id` and `postable_type` in the Reactions table - should this be changed to `reactionable_id` and `reactionable_type` and then in the model relations have `reactionable` instead of `postable`?

DFurnes: Hmm, good call. I think that makes sense… 🤔

chloealee: ok I will update and play around with that to see if it works better!

chloealee:
@DFurnes this is ready for a look! I don't think we need to do the model function anymore (`::withCount`) with the new relation between a reaction and a photo instead of a reaction and a post. Since we're able to grab the reaction count by `count($photo->reactions)` in the Reaction Controller and `count($post->content->reactions)` in the Post Transformer, it should eliminate the slow querying.

app/Http/Controllers/Api/ReactionController.php
```diff
@@ -61,7 +63,15 @@ public function store(ReactionRequest $request)
             }
         }
 
-        $totalReactions = $this->getTotalPostReactions($postableId, $postableType);
+        $reactionableType = explode('\\', $reactionableType)[2];
```
DFurnes: Is this to grab the non-namespaced class name? I'm a little worried what happens if a well-formatted string isn't passed here... e.g. if an application accidentally passed `reactionable_type => 'photo'`.

Other applications shouldn't have to know about the internal Rogue class names, so it might make sense to have a mapping of `'photo' => \App\Models\Photo::class` here to allow the API to just use always "short" names & transform them to the appropriate namespaced class in the controller or service class.

app/Http/Controllers/Api/ReactionController.php
```diff
@@ -61,7 +63,15 @@ public function store(ReactionRequest $request)
...
+
+        // @TODO: as we add more post types, we should break the below into a helper function and add different cases.
+        if ($reactionableType === 'Photo') {
+            $photo = Photo::where('id', $reactionableId)->first();
+            $totalReactions = count($photo->reactions);
+        } else {
+            throw new HttpException(405, 'Not a valid post type');
```
DFurnes: It'd make sense to validate this as part of the rules in your `ReactionRequest` form request class:

```diff
+            'reactionable_id' => 'required|int',
-            'reactionable_type' => 'required',
+            'reactionable_type' => 'required|in:photo',
```

app/Http/Transformers/PostTransformer.php
```diff
@@ -27,6 +27,7 @@ public function transform(Post $post)
                 ],
                 'caption' => $post->content->caption,
                 'status' => $post->content->status,
+                'total_reactions' => count($post->content->reactions),
```
DFurnes: This may secretly still be making an extra SQL query – if a relationship isn't [eager loaded](https://laravel.com/docs/5.4/eloquent-relationships#eager-loading) it'll only make the call for the "related" model once it's requested, which would still mean we might only load reactions once we render the `PostTransformer`.

It could be helpful to add some logging to the [database query event](https://laravel.com/docs/5.4/database#listening-for-query-events) so we can see exactly what database queries are made on a typical request.

app/Models/Photo.php
```diff
@@ -22,4 +22,12 @@ public function post()
     {
         return $this->morphOne('Rogue\Models\Post', 'postable');
     }
+
+    /**
+     * Get all of the photo's reactions.
+     */
+    public function reactions()
+    {
+        return $this->morphMany('Rogue\Models\Reaction', 'reactionable');
```
DFurnes: Total nitpick, but let's use `::class` notation when referencing classes. That way we can catch any silent errors that might otherwise occur from typos. It also saves you a few keystrokes! 😄

```diff
-        return $this->morphMany('Rogue\Models\Reaction', 'reactionable');
+        return $this->morphMany(Reaction::class, 'reactionable');
```

app/Models/Post.php
```diff
@@ -49,11 +49,17 @@ public function signup()
...
+     *
+     * @param int $postableId
+     * @param int $postableType
+     * @return int total count
      */
-    public function reactions()
+    public static function withReactionCount($postableId, $postableType)
     {
-        return $this->hasMany(Reaction::class);
+        return self::withCount(['reactions' => function ($query) use ($postableId, $postableType) {
```
DFurnes: It's not necessary for this pull request, but let's still see if we can make this work!

Feel free to comment it out in the meantime and add a `@TODO` 😄

chloealee: @DFurnes ok! I think this would be in the Photo Model though now? Is that right?

DFurnes: Yes, I think you're right!

chloealee: Ok cool - I will make a note there now, thank you!

chloealee: also for the future, here is some code I was fidgeting with:
```
    /**
     * Query for total reactions for a photo.
     *
     * @param int $reactionableId
     * @return int total count
     */
    public static function withReactionCount($reactionableId)
    {

        // $posts = Post::withCount(['votes', 'comments' => function ($query) {
        //     $query->where('content', 'like', 'foo%');
        // }])->get();

        // return self::withCount('reactions')->get();

        // this returns all Photos in an array that we can iterate through and call `->reaction_count` on each
        return self::withCount(['reactions', 'id' => function ($query) use ($reactionableId) {
            $query->where('id', $reactionableId);
        }])->get();

        // this gets us the correct photo but we can't call ->reactions_count on it
        // return self::where('id', $reactionableId)->withCount('reactions')->get();

    }
```
The problem that I was running into was returning the specific photo queried by the `reactionableId` - I could only get this to return all Photos and then call `->reaction_count` on one photo at a time (e.g. `$photos[0]->reaction_count`) so we'd still have to query again by `reactionableId`.

DFurnes: Yeah, it's tricky with all the relationships. 😕 We'll see!

app/Http/Transformers/PostTransformer.php
chloealee: @DFurnes ah I think you're right! Matt helped me turn on a log in my terminal and it looks like it is still making the extra SQL query:
screen shot 2017-03-15 at 2 15 02 pm
In this case, I don't think eager loading will help since we're doing this in the Post Transformer with only 1 post at a time? 

I'm trying to think about how we can use the `::withCount` query in this case - would we want to go back to [this function](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/SignupTransformer.php#L61-L66) and load the `$posts::withCount` here? But then we'd have to have another query here as well?

DFurnes: Oh, nice! 🏆 And correct, you'll need to either do the `withCount` or any eager-loads when making the initial Eloquent query. Let me know if you want to pair on that!

chloealee: @DFurnes got it! I'd love to pair if you have time! I'm free so let me know when works best for you :)

chloealee:
@DFurnes okay - all tests are passing and ready for a final look before merge! 

Just noting here, Dave and I chatted IRL about the `withCount` or any eager-loads when making the initial Eloquent query we discussed in [this comment](https://github.com/DoSomething/rogue/pull/150#discussion_r106249559). Dave suggested to update the seeder to make 500k + records and hitting the `GET /activity` endpoint to see how long it takes and then tackle the `withCount`/eager load issue in a separate PR.

config/app.php
```diff
@@ -215,7 +215,7 @@
         'Validator' => Illuminate\Support\Facades\Validator::class,
         'View' => Illuminate\Support\Facades\View::class,
         'Image' => Intervention\Image\Facades\Image::class,
-
+        'photo' => \Rogue\Models\Photo::class,
```
DFurnes: Hmm, this feels awkward to me but let's merge and think about it later!

chloealee: @DFurnes ok I will merge! I wasn't sure how to do the mapping you mentioned [here](https://github.com/DoSomething/rogue/pull/150#discussion_r106202388) and thought this might be the way to do it? Should I make an issue for it?

Issue #151, opened 2017-03-06
Create Campaigns Overview Page


sbsmith86:
_FEATURE OVERVIEW_
-------------------
#### User Story

Create a "Campaigns Overview" page at `/campaigns` that will show all campaigns we have posts for grouped by cause space. Each campaign should show a count of approved, pending, rejected, and deleted posts. 

#### Additional Information (optional)

Should look like this: 
![image](https://cloud.githubusercontent.com/assets/1700409/23633800/28321de4-0295-11e7-86a7-d8b7cfd70a16.png)

ngjo:
![](https://github.trello.services/images/mini-trello-icon.png) [(5) ROGUE HOME As a reviewer, I want to be able to see campaigns sorted by staff picks and then cause space -- "Campaigns List"](https://trello.com/c/R8mIjSni/163-5-rogue-home-as-a-reviewer-i-want-to-be-able-to-see-campaigns-sorted-by-staff-picks-and-then-cause-space-campaigns-list)

PR #152, opened 2017-03-06
Campaign Overview Page Components

sbsmith86:
#### What's this PR do?

* Updates the auth flow to send authenticated users to `/campaigns` so they get taken directly to the overview page.
* Creates a `campaign_overview.blade.php` template to hold the overview page.
* Creates a `CampaignsController` to pass the overview data to the view
* Adds the overview component to `app.js` and adds it to the page when the `overviewContainer` is on the page
  - **questions** Is this the best way to do this? this will start a pattern of needing to look for a element on a page to know which page we are on and which components to use on it. 
* Creates a `CampaignTable` component that is basically this: 
image

* Uses hardcoded data past to the `window.STATE` object in the controller and then iterates over that data to show different tables. 

Pull down, build the frontend, log in, you should get to `/campaigns` it should look like this: 

image

Full disclosure: I'm not sure we need to use react on this page. This will be a relatively static page that will be mostly used to navigate into the app. But it was a good exercise for me in using react. 

#### Relevant tickets
Addresses #151

#### Checklist
- [ ] Tested on staging.


resources/assets/app.js
```diff
@@ -1,17 +1,16 @@
...
-  let message = `Hey ${props.message}`;
-  let subtitle = `${props.subtitle}`;
-
-  return <div className="container__block -narrow"><p>{ message }</p><h2>{subtitle}</h2></div>;
-};
+import CampaignOverview from './components/CampaignOverview'
 
 ready(() => {
-  ReactDom.render(<Greeter {...window.STATE} />, document.getElementById('app'));
+  const overviewContainer = document.getElementById('overviewContainer');
```
katiecrane: Ooooo I forget if I saw this in any of the tutorials I watched but this makes it so much easier to read when you call `render` later. I'm a big fan

resources/assets/app.js
```diff
@@ -1,17 +1,16 @@
...
+
+  if(overviewContainer)
+  {
```
DFurnes: This is the most nitpicky of nitpicks, but our [style guide](https://github.com/airbnb/javascript#blocks--cuddled-elses) keeps braces on the same line for conditionals:

```js
if (overviewContainer) {
  ReactDom.render//...
}

File: resources/assets/components/CampaignOverview/index.js [hunk 3]
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import CampaignTable from '../CampaignTable';
+
+class CampaignOverview extends React.Component {
+  render() {
+    const causeData = this.props;
+
+    const causeTables = Object.keys(causeData).map((cause, index) => {
```
DFurnes: This is some solid vanilla JS. 💪 We've also used Lodash's [`map`](https://lodash.com/docs/#map) helper for mapping over objects elsewhere, which is a little bit easier to read at first glance.

resources/assets/components/CampaignTable/Header.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+class Header extends React.Component {
+  render() {
+    return (
+      <thead>
+        <tr className="table__header">
```
DFurnes: Totally not necessary for this pull request, but it'd be lovely to make a general-purpose Table component for this that accepted an array of heading names & a row component and then formatted the data into it... something like this maybe:

```js
<Table headings={['Campaign Name', 'Approved', 'Pending', 'Rejected', 'Deleted']} row={Row} data={data} />
```

DFurnes:
> **questions** Is this the best way to do this? this will start a pattern of needing to look for a element on a page to know which page we are on and which components to use on it.

Yeah, that'll be a good thing to keep an eye on. We had set up a little [`rehydrate` helper](https://github.com/DoSomething/voting-app/blob/dev/resources/assets/js/utilities/rehydrate.js) in the voting app that would look for any components on the page and render them if found, which was nice (especially since we did server pre-rendering on that app).

resources/assets/app.js
sbsmith86: ack i always mess this up when switching between languages. Good 👁 . also looks like I need a 👾  there.

resources/assets/components/CampaignOverview/index.js
sbsmith86: ah yes. LoDash.

resources/assets/components/CampaignTable/Header.js
sbsmith86: oh coool. I definitely wanted a more general Table component but wasn't sure how to approach it. I will try this!

sbsmith86:
@DFurnes Thanks for approving. I made some updates based on your comments about using Lodash and creating a general Table component. I am gonna merge this, but I will be continuing to work on this view so feel free to comment more if there is anything else you see.

PR #153, opened 2017-03-07
Add test for creating a new signup

### LINKED ISSUE #109: /signups unit tests
_FEATURE OVERVIEW_
-------------------
The unit tests for the `/signups` endpoint.

katiecrane:
#### What's this PR do?
Added the first test for `/signup`! It tests creating a new signup with the bare minimum parameters (I figure this is what this endpoint will mainly be used for as things stand now). 

See if this test tests what I said it does and if you think there are any more tests that could be applied to `/signups`.

I didn't test the functionality of passing a photo with the signup because that will not be used outside of data migration. I thought about doing a test with `why_participated` and things, but I don't think that would ever all be passed at one time in the wild.

#### Relevant tickets
Fixes #109

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/tests/SignupApiTest.php
+++ b/tests/SignupApiTest.php
@@ -1,0 +1,59 @@
+<?php
+
+use Illuminate\Foundation\Testing\WithoutMiddleware;
+use Illuminate\Foundation\Testing\DatabaseMigrations;
+use Illuminate\Foundation\Testing\DatabaseTransactions;
+
+class SignupApiTest extends TestCase
+{
+    use DatabaseMigrations;
+
+    /*
+     * Base URL for the Api.
+     */
+    protected $signupsApiUrl = 'api/v2/signups';
+
+    /**
+     * Test that a POST request to /signupss creates a new signup.
+     *
+     * @group creatingAPhoto
+     * @return void
+     */
+    public function testCreatingASignup()
+    {
+    	// Create test signup
+        $signup = [
+            'northstar_id'     => str_random(24),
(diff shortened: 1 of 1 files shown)


tests/SignupApiTest.php
```diff
@@ -0,0 +1,59 @@
...
+     * @return void
+     */
+    public function testCreatingASignup()
+    {
+    	// Create test signup
+        $signup = [
+            'northstar_id'     => str_random(24),
+            'campaign_id'      => $this->faker->randomNumber(4),
+            'campaign_run_id'  => $this->faker->randomNumber(4),
+            'source'           => 'the-fox-den',
```
chloealee: :rofl:

tests/SignupApiTest.php
```diff
@@ -0,0 +1,59 @@
...
+            'do_not_forward'   => true,
+        ];
+
```
chloealee: We have a model factory for a signup if you want use [here] (https://github.com/DoSomething/rogue/blob/master/database/factories/ModelFactory.php#L30) and add anything that is missing! Then you can just call `$signup = factory(Signup::class)->create();` here.

katiecrane: @chloealee 1. thank you! I didn't realize there was a Signup one 2. I don't think I was to use that in this case because I want to create the signup via the API

chloealee: @katiecrane oh shoot - sorry! You're totally right :)

chloealee:
Looks great to me!

PR #154, opened 2017-03-09
Add migration to make a signup_id index on the events table

katiecrane:
#### What's this PR do?
Added a signup_id index on the events table!

I tested by migrating and rolling back and migrating and making sure the index appeared and disappeared and appeared. You can probably just 👀 

Somewhere deep in the guts of what happens in Rogue when we migrate, this query happens:
```
select * from `events` where `events`.`signup_id` = ? and `events`.`signup_id` is not null`
```
So with no index, as the events table grew, everything got slower. This should help with that! Plus, with how we now will want to get events per signup, we are going to want this anyway.

#### Checklist
- [ ] Tested on staging.


mshmsh5000:
🐎

PR #156, opened 2017-03-13
Seeders

### LINKED ISSUE #155: Database Seeders
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want database seeders so anyone can easily seed their local database, pull down code, and test.

chloealee:
#### What's this PR do?
- Changes the event factory model specifically to a user event model factory (in `submission_type`, it will fill as `user` versus `admin`).
- Fleshes out signup factory model to include `quantity_pending`, `why_participated`, and `source`.
- Creates a database seeder that creatse a signup and photo post, with associated event logic and relations. 

- In homestead, run `php artisan db:seed`. 
- Check your database and make sure that all events, post, photo, and signup are associated in the correct manner. 

Right now, the logic isn't entirely accurate because the signup event happens after the post event (e.g. post event_id= 3 and signup event_id = 4). I think this is ok for now though since there is no logic in the code that compares these and sees what happens first? cc @sbsmith86 @katiecrane let me know if I'm wrong on this though! 

#### Relevant tickets
Fixes #155 

#### Checklist
- [ ] Tested on staging.


chloealee:
@DFurnes so I was getting caught up on making sure the tests still pass with this PR but I went back through and realized that I don't think anything should actually affect the tests since I've only added information, not taken away in the Model Factory file. 

I think my local might be a little messed up with the tests because it's giving me an error that `Fatal error: Class 'AddSignupIdIndexToEventsTable' not found in /home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Database/Migrations/Migrator.php on line 335` but I think this is outside the scope of this PR. LMK if this is wrong though!

DFurnes:
@chloealee Strange! I just tried pulling down this branch and didn't get that error. 🕵️

Can you try running `composer dump-autoload` to see if something got borked there?

chloealee:
@DFurnes I tried that earlier! Got this: 
```
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
Generating autoload files
``` 
and then when I try to run tests again, same error :(

DFurnes:
For future reference, we fixed this error by clearing out the `migrations` table in the `rogue_test` database. We should probably do a lil' more investigation to figure out why the database isn't being rolled back between tests, since that could cause issues again in the future.

PR #158, opened 2017-03-14
Optionally embed user info in activity response

### LINKED ISSUE #157: Option to include user first name in /activities response
_FEATURE OVERVIEW_
-------------------
#### User Story
If we do this then we can display user first name in the gallery.

My understanding is that this shouldn't be included by default, but should be an option. That way we don't have to ping Northstar for users unless we really want them.

#### Why This Matters
For a cool gallery.

We do return the Northstar ID, so whatever made the request could then ask Northstar what the first name is, but that's no way to live.

katiecrane:
#### What's this PR do?
Adds the ability to include an optional parameter to bundle the user info with the request. As of now, the user info provided is just the first name. To have this included, just put `include=user` in your request.

I'm not crazy about having to pass the empty array here `return $this->transform($resource, $code, [], $include)` in `ApiController.php`, but I couldn't figure out any other way to pass the `include` info to the transformer. Is anyone else bothered by this or have any thoughts on how to avoid this weirdness?

This is to be used in the gallery 📸 !

#### Relevant tickets
Fixes #157

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/ApiController.php
+++ b/app/Http/Controllers/Api/ApiController.php
@@ -64,12 +64,16 @@ class ApiController extends BaseController
      * @param  array  $meta
      * @return \Illuminate\Http\JsonResponse
      */
-    public function transform($data, $code = 200, $meta = [])
+    public function transform($data, $code = 200, $meta = [], $include = null)
     {
         $data->setMeta($meta);
 
         $manager = new Manager;
         $manager->setSerializer(new DataArraySerializer);
+
+        if (isset($include)) {
+            $manager->parseIncludes($include);
+        }
 
         $response = $manager->createData($data)->toArray();
 
@@ -114,6 +118,8 @@ class ApiController extends BaseController
         $resource->setMeta($meta);
         $resource->setPaginator(new IlluminatePaginatorAdapter($paginator));
 
-        return $this->transform($resource, $code);
+        $include = isset($request->include) ? $request->include : null;
+
+        return $this->transform($resource, $code, [], $include);
     }
 }
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -86,7 +86,7 @@ class SignupTransformer extends TransformerAbstract
      */
     public function includeUser(Signup $signup)
     {
-        $registrar = new Registrar();
+        $registrar = app(Registrar::class);
         $northstar_id = $signup->northstar_id;
 
         return $this->item($registrar->find($northstar_id), new UserTransformer);
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -3,8 +3,8 @@
 namespace Rogue\Http\Transformers;
 
 use Rogue\Models\Event;
-use Rogue\Services\Registrar;
 use Rogue\Models\Signup;
+use Rogue\Services\Registrar;
 use League\Fractal\TransformerAbstract;
 
 class SignupTransformer extends TransformerAbstract
@@ -90,6 +90,5 @@ class SignupTransformer extends TransformerAbstract
         $northstar_id = $signup->northstar_id;
 
         return $this->item($registrar->find($northstar_id), new UserTransformer);
-
     }
 }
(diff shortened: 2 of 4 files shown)


app/Http/Transformers/SignupTransformer.php
```diff
@@ -67,4 +77,18 @@ public function includeEvents(Signup $signup)
...
+
+    /**
+     * Include the user data (optionally)
+     *
+     * @param \Rogue\Models\Signup $signup
+     * @return \League\Fractal\Resource\Item
+     */
+    public function includeUser(Signup $signup)
+    {
+        $registrar = new Registrar();
```
DFurnes: We should try to make sure all class dependencies are either fetched using the `app()` helper or injected via the constructor so that Laravel can handle resolving dependencies for us. That will allow us to easily mock this class in a test case if we ever wanted to.

app/Http/Transformers/UserTransformer.php
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Transform resource data.
+     *
+     * @param \Rogue\Models\Photo $user
+     * @return array
+     */
+    public function transform($user)
+    {
+        return [
+            'first_name' => $user->first_name,
```
DFurnes: Should we include `id` here as well?

app/Http/Controllers/Api/ApiController.php
```diff
@@ -64,13 +64,17 @@ public function item($data, $code = 200, $meta = [], $transformer = null)
      * @param  array  $meta
      * @return \Illuminate\Http\JsonResponse
      */
-    public function transform($data, $code = 200, $meta = [])
+    public function transform($data, $code = 200, $meta = [], $include = null)
```
sbsmith86: i'm not too bothered by it, but you could switch the order of the parameters of you want since you always pass `$include` and don't always need `$meta`and set a default for it - That way you could call `transfom($data, $code, $include)`

sbsmith86: actually...that might not work if you have `$meta` then you will want to pass it. hmmm. Again, I'm not too bothered by it, but will let others way in.

DFurnes:
> I'm not crazy about having to pass the empty array here `return $this->transform($resource, $code, [], $include)` in `ApiController.php`, but I couldn't figure out any other way to pass the `include` info to the transformer. Is anyone else bothered by this or have any thoughts on how to avoid this weirdness?

Ahhh yeah, that's no fun. I don't think we've ever had optional includes using Fractal before.

I could see it making sense to just pass in the `$request` to the transform methods across the board, so you don't have to worry about checking if the array exists or not case-by-case. I'm not super bothered though, so definitely something we can hold off for another PR if you want.

app/Http/Controllers/Api/ApiController.php
DFurnes: Yeah, that's a good thought too. I think the only place we regularly use the `$meta` argument is in the `paginatedCollection` method, which is safely tucked away in the base controller.

katiecrane: @sbsmith86 okay cool it seems like I am the most bothered by it. I thought about that too, but there are two functions that pass in `meta` and only this one passes in `includes`.

app/Http/Transformers/UserTransformer.php
katiecrane: @DFurnes I double checked with @ngjo and currently `first_name` is all that is requested. Northstar ID is included in the signup part of the response anyway so that will always be included.

app/Http/Transformers/SignupTransformer.php
katiecrane: @DFurnes I'm going to add this in before merging!

app/Http/Transformers/SignupTransformer.php
```diff
@@ -86,7 +85,7 @@ public function includeEvents(Signup $signup)
      */
     public function includeUser(Signup $signup)
     {
-        $registrar = new Registrar();
+        $registrar = app('Rogue\Services\Registrar');
```
DFurnes: You can still use the imported class with `app(Registrar::class)`!

PR #159, opened 2017-03-15
Campaign overview with data

### LINKED ISSUE #151: Create Campaigns Overview Page

sbsmith86:
#### What's this PR do?

Creates a `CampaignService` that first looks for campaigns in cache, if it finds them then loads them from there, if it doesn't find them then it makes a request to phoenix for the campaign info. 

Grabs a set of campaign ids for each signup we have in rogue. Gets their full campaign object, groups them by cause, and passes it to the frontend to render overview tables. 

We could probably think of better ways to grab the campaigns we are interested in. I will be looking into this when trying to figure out post counts based on status. For now, I am just grabbing them directly from the DB. 

The `CampaignService` class and the `Registrar` class are very similar and I would like to look into how to extract some of the shared logic. 

#### Relevant tickets
Fixes #151 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -7,7 +7,9 @@ use Rogue\Services\Registrar;
 
 class CampaignsController extends Controller
 {
-    public function __construct()
+    protected $phoenix;
+
+    public function __construct(Phoenix $phoenix)
     {
         $this->middleware('auth');
         $this->middleware('role:admin,staff');
@@ -19,6 +21,10 @@ class CampaignsController extends Controller
      */
     public function index()
     {
+        $allCampaigns = $this->phoenix->getAllCampaigns();
+
+        dd($allCampaigns['data'][1]);
+
         $staffPicks = collect([
             ['name' => 'Campaign 1', 'approved' => 53, 'pending' => 32, 'rejected' => 34, 'deleted' => 3],
             ['name' => 'Campaign 2', 'approved' => 54, 'pending' => 33, 'rejected' => 35, 'deleted' => 4],
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -3,7 +3,6 @@
 namespace Rogue\Http\Controllers;
 
 use Rogue\Models\Signup;
-use Rogue\Services\Phoenix;
 use Rogue\Services\Registrar;
 use Illuminate\Support\Facades\DB;
 use Rogue\Services\CampaignService;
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/CampaignsController.php
```diff
@@ -3,46 +3,63 @@
...
+     * @param \Rogue\Services\Phoenix $phoenix
+     */
+    public function __construct(Phoenix $phoenix)
     {
         $this->middleware('auth');
         $this->middleware('role:admin,staff');
+
+        $this->phoenix = $phoenix;
         $this->registrar = new Registrar();
+        $this->campaignService = new CampaignService;
```
DFurnes: Let's inject both of these dependencies like we do with the `Phoenix` client:

```php
public function __construct(Phoenix $phoenix, Registrar $registrar, CampaignService $campaignService) {
    // ...
    $this->phoenix = $phoenix;
    $this->registrar = $registrar;
    $this->campaignService = $campaignService;
```

app/Http/Controllers/CampaignsController.php
```diff
@@ -3,46 +3,63 @@
...
      */
     public function index()
     {
-        $staffPicks = collect([
-            ['name' => 'Campaign 1', 'approved' => 53, 'pending' => 32, 'rejected' => 34, 'deleted' => 3],
-            ['name' => 'Campaign 2', 'approved' => 54, 'pending' => 33, 'rejected' => 35, 'deleted' => 4],
-            ['name' => 'Campaign 3', 'approved' => 55, 'pending' => 34, 'rejected' => 36, 'deleted' => 5],
-        ]);
+        // Get campaigns ids for all signups in rogue.
+        $campaigns = DB::table('signups')->select('campaign_id')->groupBy('campaign_id')->get();
```
DFurnes: Would it make sense to put this query within the campaign service?

app/Services/CampaignService.php
```diff
@@ -0,0 +1,155 @@
...
+use Rogue\Repositories\CacheRepository;
+
+class CampaignService
+{
+    /**
+     * Create new Registrar instance.
+     */
+    public function __construct()
+    {
+        $this->phoenix = new Phoenix;
```
DFurnes: We probably want to inject this so that we can easily mock it in tests.

app/Services/CampaignService.php
```diff
@@ -0,0 +1,155 @@
...
+            if ($value === false or $value === null) {
+                $campaigns[$key] = $this->find($key);
+            }
+        }
+
+        return $campaigns;
+    }
+
+    /**
+     * Get large number of users in batches from Northstar.
```
DFurnes: `s/from Northstar/from Phoenix` ✏️

PR #162, opened 2017-03-16
Reaction test

### LINKED ISSUE #160: /reviews unit tests
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want /reaction unit tests to make sure everything is working as expected.

#### Why This Matters
So we can run tests to make sure this endpoint is always working instead of manually testing.

chloealee:
#### What's this PR do?
Creates a unit test for the `POST /reactions` endpoint that tests: 
- A single user reacting to and "un-reacting" to a photo (when a user reacts to the same photo after it has already been reacted to, it will soft delete the reaction). 
- Aggregating the total reactions for a single photo - if two different users react to the same photo, the total reactions will be 2. 

Run `vendor/bin/phpunit` (or `vendor/bin/phpunit --filter=Reaction` to only test reaction unit tests) and make sure all tests are passing. 

#### Relevant tickets
Fixes #160 

#### Checklist
- [ ] Tested on staging.


tests/ReactionsApiTest.php
```diff
@@ -0,0 +1,89 @@
...
+    public function testPostingAndSoftDeleteForReaction()
+    {
+        // Create a photo to react to.
+        $photo = factory(Photo::class)->create();
+
+        $reaction = $this->createReactionRequest(str_random(24), $photo->id, 'photo');
+
+        $this->json('POST', $this->reactionsApiUrl, $reaction);
+
+        $this->assertResponseStatus(200);
```
katiecrane: This might be out of scope for this PR, but I think we want this to be a `201` response for something being created

chloealee: @katiecrane ah for some reason I thought that `201` response codes are for updates! Since it (should be) a small update, I can do this in this PR too?

tests/TestCase.php
```diff
@@ -171,4 +171,18 @@ public function checkReportbackResponse($response)
...
+     */
+    public function createReactionRequest($northstarId, $photoId, $postType)
+    {
+        return [
+            'northstar_id' => $northstarId,
+            'reactionable_id' => $photoId,
+            'reactionable_type' => $postType,
+        ];
+    }
 }
```
chloealee: I just realized I can probably also pull more into this helper function (sending the request and asserting the response). Will do that now!

tests/ReactionsApiTest.php
katiecrane: @chloealee I thiiink it's `201` when something is created and `200` when it's otherwise successful https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

tests/ReactionsApiTest.php
```diff
@@ -0,0 +1,89 @@
...
+     * @return void
+     */
+    public function testPostingAndSoftDeleteForReaction()
+    {
+        // Create a photo to react to.
+        $photo = factory(Photo::class)->create();
+
+        $reaction = $this->createReactionRequest(str_random(24), $photo->id, 'photo');
+
+        $this->json('POST', $this->reactionsApiUrl, $reaction);
```
DFurnes: This is subjective, but I think it makes sense to inline both of these to reduce the amount of jumping around you need to do when reading the test case. In general I always try to make test code as explicit as possible so it's easy to follow exactly what steps are happening in what order.

```php
$this->json('POST', 'api/v2/reactions', [
    'northstar_id' => $this->faker->uuid,
    'reactionable_id' => $photo->id,
    'reactionable_type' => 'photo',
]);
```

tests/ReactionsApiTest.php
chloealee: @katiecrane yes you're right!

DFurnes: Yeah, good catch! Let's take some time to audit all the status codes we're using in Rogue sometime soon. I think we've tried to follow [Northstar's status codes](https://github.com/DoSomething/northstar/tree/dev/documentation#errors--status-codes) in most of our applications.

katiecrane: OOOOOO I like the idea of that

tests/ReactionsApiTest.php
chloealee: @DFurnes ah totally makes sense! I'll update in the helper function :)

tests/ReactionsApiTest.php
chloealee: oh awesome! so maybe I should hold off on updating the status code and that can be a new issue? @DFurnes @katiecrane cc @ngjo

DFurnes: That makes sense to me!

chloealee: Cool, I'll make an issue for this now too!

tests/TestCase.php
```diff
@@ -173,16 +173,16 @@ public function checkReportbackResponse($response)
...
-     * Create a reaction request.
+     * Create a reaction.
      *
      * @return array
      */
-    public function createReactionRequest($northstarId, $photoId, $postType)
+    public function createReaction($northstarId, $photoId, $postType)
     {
-        return [
+        $this->json('POST', $this->reactionsApiUrl, [
```
DFurnes: I'm still not convinced of the value of abstracting this into a method on the TestCase, especially since this would error out if called from anywhere besides `ReactionsApiTest` (since `$this->reactionsApiUrl` wouldn't be defined).

chloealee: @DFurnes updated!

PR #164, opened 2017-03-16
Total reaciton count query

### LINKED ISSUE #161: Enhance total reaction count query
_BUG_
-------------------
#### Current Behavior
- Currently, we are not using eager loading or Eloquent's `::withCount` when querying for total reaction counts in the [Post Transformer](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/PostTransformer.php#L30), which also gets called in `GET /activity`.

#### Desired Behavior
- We want to aggregate this in an efficient way. We're thinking it would be good to create a Model Function in the Post class similar to [this](https://github.com/DoSomething/rogue/pull/150#discussion_r106209658). 

#### Steps to Replicate
- Hit `GET /activity` endpoint while looking at the MySQL log and you can see that many queries are being made, which will get a little crazy for each post in the response! 
- @DFurnes also suggested to update the seed data to make many records (500K+) and then hit the `GET /activity` endpoint to see how long it takes to load. 

#### Why This Matters
- We want to run efficient queries to return data!

chloealee:
#### What's this PR do?
- Deletes `withReactionCount` model function from Post.php
- Adds `withReactionCount` model function to Photo.php 
  - Updates to use this model function in the PostTransformer to enhance the performance of the query

- `GET /activity` and `POST /reactions` endpoint should still work and return `total_reactions`

Using this new model function, you can see that less queries are being made (no separate reaction query): 
screen shot 2017-03-16 at 4 00 33 pm

compared to [before](https://cloud.githubusercontent.com/assets/9019452/23964198/13662d6e-098a-11e7-805b-4348eb25d2e7.png). 

#### Relevant tickets
Fixes #161 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -32,11 +32,24 @@ class ActivityController extends ApiController
     {
         // Create an empty Signup query, which we can either filter (below)
         // or paginate to retrieve all signup records.
-        $query = $this->newQuery(Signup::class);
+        // $query = $this->newQuery(Signup::class);
+        $query = Signup::withCount('post')
 
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Signup::$indexes);
 
         return $this->paginatedCollection($query, $request);
     }
+
+
+    /**
+     * Query for total reactions for a photo.
+     *
+     * @param int $reactionableId
+     * @return int total count
+     */
+    // public static function withReactionCount($reactionableId)
+    // {
+    //     return self::where('id', $reactionableId)->withCount('reactions')->first();
+    // }
 }
--- a/app/Http/Transformers/PostTransformer.php
+++ b/app/Http/Transformers/PostTransformer.php
@@ -3,6 +3,7 @@
 namespace Rogue\Http\Transformers;
 
 use Rogue\Models\Post;
+use Rogue\Models\Photo;
 use League\Fractal\TransformerAbstract;
 
 class PostTransformer extends TransformerAbstract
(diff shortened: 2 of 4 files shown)


chloealee:
Or is this still not quite right because it's still making a query for each post? I think I still want to try to get `::withCount` to work instead of `::where`?

DFurnes:
> Or is this still not quite right because it's still making a query for each post? I think I still want to try to get `::withCount` to work instead of `::where`?

Yeah, the downside of calling `::withReactionCount` on the individual photo is that you're still making 20 new queries if there are 20 photos in the response. So ideally you should be able to find where the initial query that returns that collection of posts is, and add the `withCount` there. 

(And that's where the callback to `withCount` that we'd played with before probably comes into play, since it might not play nicely with finding a count through a polymorphic relationship out of the box.)

chloealee:
ah yes, thank you Dave! I will keep working on this!

chloealee:
@DFurnes so we'd want to with `::withCount` [here](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/SignupTransformer.php#L63)?

DFurnes:
Ideally [here](https://github.com/DoSomething/rogue/blob/b6e16029b5dcf5b07f33ee10c0d01781b1015b68/app/Http/Controllers/Api/ActivityController.php#L35) when the initial query for the page is run!

chloealee:
oh right right we talked about that! Thank you!

chloealee:
closing this PR after today's discussion and will revisit after model relations are updated!

PR #170, opened 2017-03-20
Post table updates

### LINKED ISSUE #168: Update Posts table
Update the posts table so more general columns like `status`, `source`, `remote_addr` are stored on the post instead of the `photos` table.

sbsmith86:
#### What's this PR do?
Updates the `posts` table so that it stores `status`, `source`, and `remote_addr` instead of storing those values in the `photos` table. 

This is the first step to removing the `photos` table entirely and the polymorphic relationship between a `Post` and a `Photo`but allows use to update some of the other relationships between models and do easier aggregation in the mean time. 

Pull down branch, run migrations, post a `Post` to the api and it should store these values correctly on the `posts` table. 

Updating the db schema after a Team Bleed meeting on the DB schema. 

#### Relevant tickets
Fixes #168 

#### Checklist
- [ ] Tested on staging.


app/Http/Transformers/PhotoTransformer.php
```diff
@@ -36,8 +36,6 @@ public function transform(Photo $photo)
                     'url' => $photo->file_url,
                     'edited_url' => $photo->edited_file_url,
                 ],
-                'caption' => $photo->caption,
```
DFurnes: Is this field still stored on the photo?

sbsmith86: @DFurnes yea..oops..didn't mean to delete this but it also seems like the photo transformer isn't being used anymore. but regardless, it will be removed when we move everything to the posts table.

sbsmith86: @DFurnes added that back in. it is still stored on the photo. for now...

katiecrane:
Just making a note that we will have to wipe the Rogue `posts`, `photos`, `signups`, and `events` columns and start the migration from the beginning the next time we run it since we are dropping columns on the `photos` table! But this change is going to make life so much easier (moving `status` in particular!!)

sbsmith86:
@katiecrane yea we are gonna move forward like the DB is empty and we can move stuff around.

katiecrane:
@sbsmith86 yup I'm on board with that!

Issue #167, opened 2017-03-17
Update the Events table and it's relationship with other models


sbsmith86:
Define a polymorphic relationship between and `Event` and the other models Rogue.

ngjo:
![](https://github.trello.services/images/mini-trello-icon.png) [As a developer, I want to update the relationships between Events and other tables so that it's more clear and follows the new agreed upon model.](https://trello.com/c/XkZNKubY/330-as-a-developer-i-want-to-update-the-relationships-between-events-and-other-tables-so-that-it-s-more-clear-and-follows-the-new-ag)

PR #173, opened 2017-03-21
Polymorphic events

sbsmith86:
#### What's this PR do?

This PR updates the DB schema to so that we can set up a polymorphic relationship between events and other models. 

This change required a few steps:

* Remove the `event_id` from other tables as it had a foreign key restraint and these tables don't need to own events.
* Add unique identifiers to other tables that will be used as the `eventable_id` on the polymorphic columns on the event table
* Drop the old events table to make room for an entire new one. 
* create a new `events` table with polymorphic column to hold the relationships and a `content` column that will hold json strings of the content associated with the event. 
* define the relationship on the models. 
* update model factories and seeders.  

It should be safe to go commit by commit. 
 
⚠️  This breaks everything ⚠️ 

The PR breaks a lot and doesn't fix it yet. In order to constrain the size of this PR, it only covers updating the db schema, defining the new relationships, and creating an activity seeder to get some fake data in the system 

I will submit another PR that update the logic around creating events next. 

#### Relevant tickets
Addresses #167 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Models/Event.php
+++ b/app/Models/Event.php
@@ -13,7 +13,6 @@ class Event extends Model
      */
     protected $fillable = [];
 
-
     public function eventable()
     {
         return $this->morphTo();
--- a/app/Models/Event.php
+++ b/app/Models/Event.php
@@ -13,19 +13,9 @@ class Event extends Model
      */
     protected $fillable = ['id', 'signup_id', 'northstar_id', 'event_type', 'submission_type', 'quantity', 'quantity_pending', 'why_participated', 'caption', 'status', 'source', 'remote_addr', 'reason'];
 
-    /**
-     * An event has one signup.
-     */
-    public function signup()
-    {
-        return $this->hasOne(Signup::class);
-    }
 
-    /**
-     * An event has one post.
-     */
-    public function post()
+    public function eventable()
     {
-        return $this->hasOne(Post::class);
+        return $this->morphTo();
     }
 }
--- a/app/Models/Event.php
+++ b/app/Models/Event.php
@@ -11,7 +11,7 @@ class Event extends Model
      *
      * @var array
      */
-    protected $fillable = ['id', 'signup_id', 'northstar_id', 'event_type', 'submission_type', 'quantity', 'quantity_pending', 'why_participated', 'caption', 'status', 'source', 'remote_addr', 'reason'];
+    protected $fillable = [];
 
 
     public function eventable()
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -33,11 +33,11 @@ class Post extends Model
     }
 
     /**
-     * Each post belongs to an event.
+     * Each post has events.
      */
-    public function event()
+    public function events()
     {
-        return $this->belongsTo(Event::class);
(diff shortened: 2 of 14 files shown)


app/Models/Post.php
```diff
@@ -33,11 +33,11 @@ public function content()
...
 
     /**
-     * Each post belongs to an event.
+     * Each post has events.
      */
-    public function event()
+    public function events()
     {
-        return $this->belongsTo(Event::class);
+        $this->morphMany('Rogue\Models\Event', 'eventable');
```
DFurnes: Let's use [::class](http://php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class) for all of these relationships: `$this->morphMany(Event::class, 'eventable');`

app/Models/Event.php
```diff
@@ -13,19 +13,9 @@ class Event extends Model
      */
     protected $fillable = ['id', 'signup_id', 'northstar_id', 'event_type', 'submission_type', 'quantity', 'quantity_pending', 'why_participated', 'caption', 'status', 'source', 'remote_addr', 'reason'];
```
DFurnes: This can probably be an empty array now.

sbsmith86:
@DFurnes updated!

database/factories/ModelFactory.php
```diff
@@ -20,50 +19,33 @@
...
-        'status' => 'pending',
-        'source' => 'phoenix-web',
-        'remote_addr' => '10.0.2.2',
     ];
 });
 
 // Post Factory
 $factory->define(Post::class, function (Generator $faker) {
     return [
-        'northstar_id' => str_random(24),
```
katiecrane: I thought northstar ID still lived on Posts?

sbsmith86: Ya but it should be the same `northstar_id` that is on the signup the post relates to. If you generate it randomly in the factory, it would be different.

katiecrane: ahh smart!

database/migrations/2017_03_20_205359_create_new_events_table_schema.php
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('events', function (Blueprint $table) {
+            $table->increments('id');
+            $table->integer('eventable_id');
+            $table->string('eventable_type');
+            $table->string('content');
```
DFurnes: This might make sense as a [JSON field](https://laravel.com/docs/5.3/migrations#columns), which I think would let Laravel automatically cast arrays into this column, and also let us (not sure on performance here) make `WHERE` queries based on the JSON content.

DFurnes:
Oy, not sure why that shows as "outdated". I'm also a little unsure on whether we can use the Laravel JSON stuff… it looks like [the `where` queries don't work on MariaDB](https://mattstauffer.co/blog/new-json-column-where-and-update-syntax-in-laravel-5-3), so it might still be the right choice to define the column as `text` and then just cast it in the model later on.

PR #175, opened 2017-03-22
pending reportback page start

### LINKED ISSUE #174: Pending Posts Page: the beginning
_FEATURE OVERVIEW_
-------------------
#### User Story
Build a page that pulls in pending Posts for a given campaign run. Display the Post photo.

#### Why This Matters
Admins gotta review somewhere! This is a good first step.

katiecrane:
#### What's this PR do?
1. Creates a new route to hit the inbox for a given campaign run which hits `show` in `CampaignsController.php`. A query is run to get all the signups for that run that also have pending Posts. The relevant signups and their Posts and their Post's content is returned.

2. Passes that data to React which reformats it a bit and displays some relevant data!

image

🎨  🚨  At this stage this is going to be incomplete and have some hacky elements, but I'd love to set us up so that we don't just have to tear it all down later and do some crazy refactoring. I also have never done this before so I don't have the best idea of what that would entail, so feel free to leave lots of comments!

React is real cool.

#### Relevant tickets
Fixes #174

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -55,7 +55,7 @@ class CampaignsController extends Controller
 
         return view('pages.campaign_inbox')
             ->with('state', [
-                'Signups' => $signups,
+                'signups' => $signups,
             ]);
     }
 }
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -2,6 +2,8 @@
 
 namespace Rogue\Http\Controllers;
 
+use Rogue\Models\Signup;
+
 class CampaignsController extends Controller
 {
     public function __construct()
@@ -39,5 +41,28 @@ class CampaignsController extends Controller
                 'Environment' => $environment,
                 'Bullying' => $bullying,
             ]);
+    }
+
+    /**
+     * Show overview of campaigns.
+     */
+    public function show($campaign_run_id)
+    {
+        // pull in pending rb items for the given run
+        // 1. pull in all signups
+        $signups = Signup::where('campaign_run_id', $campaign_run_id)->get();
+
+        // 2. pull in ALL posts
+        $posts = $signups->flatMap(function ($item) {
+            return $item->posts;
+        });
+        // 3. pull in all pending photos
+
+
+        return view('pages.campaign_inbox');
+            // ->with('state', [
+            //     'Campaign' => $campaign_run_id, 
+            //     'Posts' => $posts, 
+            // ]);
     }
 }
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -44,25 +44,18 @@ class CampaignsController extends Controller
     }
 
     /**
-     * Show overview of campaigns.
+     * Show particular campaign inbox.
      */
     public function show($campaign_run_id)
     {
-        // pull in pending rb items for the given run
-        // 1. pull in all signups
-        $signups = Signup::where('campaign_run_id', $campaign_run_id)->get();
+        // Pull in all signups for the given run that have pending posts, and include their pending posts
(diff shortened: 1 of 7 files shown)


resources/views/pages/campaign_inbox.blade.php
```diff
@@ -0,0 +1,13 @@
+@extends('layouts.master')
+
+@section('main_content')
+
+    @include('layouts.header', ['header' => 'Campaign Inbox', 'subtitle' => 'all pending reportbacks for the campaign'])
+
+	    <div class="container -padded">
+	        <div id="inboxContainer" class="wrapper">
+	            shove the react here
```
sbsmith86: Nitpicky thing, but this should probably say "loading..." or something more user friendly as a user could see this briefly on the page.

resources/views/pages/campaign_inbox.blade.php
```diff
@@ -0,0 +1,13 @@
...
+	        </div>
+	    </div>
+
+@stop
```
sbsmith86: new line

resources/assets/app.js
```diff
@@ -5,11 +5,18 @@ import React from 'react';
...
 ready(() => {
   const overviewContainer = document.getElementById('overviewContainer');
+  const inboxContainer = document.getElementById('inboxContainer');
+
 
   if (overviewContainer) {
     ReactDom.render(<CampaignOverview {...window.STATE} />, overviewContainer);
   }
+
+  if (inboxContainer) {
```
sbsmith86: I'm worried about this getting crazy, always having to check if the container is on the page before injecting the js. @DFurnes  [had an idea](https://github.com/DoSomething/rogue/pull/152#issuecomment-284784711) to get around this. Not sure if this is critical at the moment but I worry this will become a pattern.

resources/views/pages/campaign_inbox.blade.php
katiecrane: whoops I meant to change this!

resources/assets/app.js
katiecrane: yeah I agree that it's not ideal. I kind of feel like it should be a separate task though? either way I'm going to make an issue for that now so we have a reference/dedicated place to discuss

app/Http/Controllers/CampaignsController.php
```diff
@@ -40,4 +42,20 @@ public function index()
...
+    public function show($campaign_run_id)
+    {
+        // Pull in all signups for the given run that have pending posts, and include their pending posts
+        $signups = Signup::whereHas('posts', function ($query) {
+            $query->where('status', 'pending');
+        })->where('campaign_run_id', $campaign_run_id)->with('posts.content')->get();
+
+        return view('pages.campaign_inbox')
+            ->with('state', [
+                'Signups' => $signups,
```
DFurnes: Variable names should be camel-cased, so `'Signups'` should be `'signups'`! 🐫

resources/assets/components/InboxItem/index.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { map } from 'lodash';
+
+class InboxItem extends React.Component {
+  displayImage(photo_url) {
+    if (photo_url == "default") {
```
DFurnes: Where does `"default"` get set here? Is that an actual string in the database? If not, it could make sense to use [default props](https://facebook.github.io/react/docs/react-without-es6.html#declaring-prop-types-and-default-props) to automatically add a placeholder when the `file_url` is empty. 🐶

katiecrane: it is an actual string in the database! that is how we decided to deal with broken images so we can deal with displaying them however we want

katiecrane: so for now I have a goofy placeholder

PR #179, opened 2017-03-23
New event logic

### LINKED ISSUE #167: Update the Events table and it's relationship with other models

sbsmith86:
#### What's this PR do?

Moves out all the manual event creation logic out of the `SignupRepository` and `PhotoRepository`  and moves it into a service provider that automatically creates events when the model is saved (which is triggered on creation and update).

Updates tests to work with this new logic/event schema. NOTE: reviews tests still fail but wanted to work on that separately to keep this contained. 

This can probably be reviewed commit by commit. It's a little hairy to read because there was a lot of deletions. 

We also need a lot of testing on this to make sure everything still works as expected. The testing will be easier when this is deployed on QA.

#### Relevant tickets
Fixes #167 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/PhotoRepository.php
+++ b/app/Repositories/PhotoRepository.php
@@ -87,6 +87,12 @@ class PhotoRepository
             $post->updated_at = $data['created_at'];
 
             $post->save(['timestamps' => false]);
+
+
+            $post->events->first()->created_at = $data['created_at'];
+            $post->events->first()->updated_at = $data['created_at'];
+
+            $post->events->first()->save(['timestamps' => false]);
         } else {
             // @TODO: keep this after the migration
             $post->content()->associate($photo);
--- a/app/Repositories/PhotoRepository.php
+++ b/app/Repositories/PhotoRepository.php
@@ -43,12 +43,6 @@ class PhotoRepository
      */
     public function create(array $data, $signupId)
     {
-        // Include signup_id when we create the Event
-        $data['signup_id'] = $signupId;
-
-        // Don't send quantity and why_participated - we don't want this to live on the post_photo event.
-        $postEvent = Event::create(array_except($data, ['quantity', 'why_participated']));
-
         if (isset($data['file'])) {
             $fileUrl = $this->aws->storeImage($data['file'], $signupId);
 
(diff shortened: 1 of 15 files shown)


app/Repositories/PhotoRepository.php
```diff
@@ -68,19 +62,14 @@ public function create(array $data, $signupId)
         if (isset($data['created_at'])) {
             $photo->created_at = $data['created_at'];
             $photo->updated_at = $data['created_at'];
-            $postEvent->created_at = $data['created_at'];
-            $postEvent->updated_at = $data['created_at'];
 
             $photo->save(['timestamps' => false]);
-            $postEvent->save(['timestamps' => false]);
```
katiecrane: This brought up a thought, will there be a way to override the timestamps in the event table if we want to? Currently everything that is being migrated gets backdated. Might be weird to have inconsistencies between signup timestamps and the timestamps of their events.

app/Repositories/SignupRepository.php
```diff
@@ -37,28 +29,17 @@ public function create($data)
...
             // Set the updated time if provided, if not, assume no updates
             if (isset($data['updated_at'])) {
                 $signup->updated_at = $data['updated_at'];
-                $event->updated_at = $data['updated_at'];
             } else {
                 $signup->updated_at = $data['created_at'];
-                $event->updated_at = $data['created_at'];
             }
 
             $signup->save(['timestamps' => false]);
```
katiecrane: When we run the migration, this will create a signup event when the signup is created (I think?) and then create another event when we save it here to change the timestamps, which I don't think we want because it's not really a real event. I couldn't figure out how to create a model without timestamps, only to save it. The workaround that immediately comes to mind is to change the event trigger from saved to created until we are done with the migration but I'm for sure up for discussion.

app/Services/PostService.php
```diff
@@ -24,18 +29,16 @@ class PostService
...
         request()->headers->set('X-Request-ID', $transactionId);
 
         // POST reportback back to Phoenix, unless told not to.
         // If request fails, record in failed_jobs table.
-        if (! isset($data['do_not_forward'])) {
-            dispatch(new SendPostToPhoenix($post));
-        }
+        // if (! isset($data['do_not_forward'])) {
+        //     dispatch(new SendPostToPhoenix($post));
+        // }
```
katiecrane: byeeeeee Phoenix!

app/Repositories/PhotoRepository.php
sbsmith86: good point. I think we can still customize those.. will dig into it.

app/Repositories/SignupRepository.php
sbsmith86: hmm interesting. I sorta want to get away from before-migration code and after-migration code cause that seems like a lot to remember to got back and clean up. I gonna see if there is a case I can put on the listener to only fire under a certain circumstance.

katiecrane: I'm on board with that, ideally we wouldn't need to change anything after we're done with migrations

DFurnes: I wonder if you can pull in the timestamps from the signup model in the `created` event, so if an "eventable" model is backdated the event is too.

katiecrane: The signup gets backdated after it gets created, so I don't think that would have the right timestamps. Unless there is a way to create a model with `'timestamps' => false` which I didn't think there was without turning off timestamps on the model altogether

katiecrane:
@sbsmith86 the timestamp updates look good to me!

DFurnes:
Post-merge 👍!

PR #180, opened 2017-03-24
Add buttons and style page more

katiecrane:
#### What's this PR do?
Styles everything closer to the wireframe. I'm not sure how to deal with matching the styles that are in the wireframe but not in Forge (gray buttons that turn either green or red on click, 3 columns, that box that the quantity is in) and so any ideas on that or if we shouldn't worry about it yet would be appreciated!

image

🎨 👀 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/CampaignInbox/index.js
+++ b/resources/assets/components/CampaignInbox/index.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import { flatMap } from 'lodash';
+import { map } from 'lodash';
 
 import InboxItem from '../InboxItem';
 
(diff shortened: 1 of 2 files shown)


resources/assets/components/CampaignInbox/index.js
```diff
@@ -15,13 +15,11 @@ class CampaignInbox extends React.Component {
...
-      <div>
-          <ul>
-            {
-              Object.keys(posts)
-                .map(key => <InboxItem key={key} details={posts[key]}/>)
-            }
-          </ul>
+      <div className="container">
+        {
+          Object.keys(posts)
```
DFurnes: This is totally subjective, but this might read nicer with Lodash's `map` function:

```js
import { map } from 'lodash';
// ...

return (
    <div className="container">
        { map(posts, (post, key) => <InboxItem key={key} details={post} /> }
    </div>
);

```

katiecrane: @DFurnes I like that a lot better!

PR #184, opened 2017-03-28
Pass user info to the front end and display it

### LINKED ISSUE #183: Include some user information on the campaign inbox
_FEATURE OVERVIEW_
-------------------
#### User Story
We want more user information (from Northstar) to appear with Posts in the campaign inbox. That also makes it easier to reach out to the member if needed.

The information we want displayed is: name, age (Northstar returns birthdate), mobile number, and email address.

There would be no way to access this information on the front end.

katiecrane:
#### What's this PR do?
Pulls the user information at the same time that we get the signup and posts for a certain campaign run. The user info is nested under each post. Then, the following are displayed on the front end:
- `first_name`
- `last_name`
- age (there is a helper to calculate this from the `birthdate` returned by Northstar)
- `email`
- `mobile` phone number

It looks like this (my account doesn't have a mobile number so there is a blank where the phone number would be):
image

👀  See if it looks okay and if I am calculating age correctly.

There's nothing being conditionally displayed right now, so if a member doesn't have a mobile number stored with us, there is still a blank space for it instead of shifting everything up.

#### Relevant tickets
Fixes #183

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/InboxItem/index.js
+++ b/resources/assets/components/InboxItem/index.js
@@ -11,6 +11,14 @@ class InboxItem extends React.Component {
     }
   }
 
+  calculateAge(date) {
+    const birthdate = new Date(date);
+    const today = Date.now();
+    const age = Math.floor((today - birthdate) / 31536000000);
+
+    return age;
+  }
+
   render() {
     const post = this.props.details;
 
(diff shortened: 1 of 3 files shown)


resources/assets/components/InboxItem/index.js
```diff
@@ -11,6 +11,14 @@ class InboxItem extends React.Component {
     }
   }
 
+  calculateAge(date) {
```
sbsmith86: interesting, we don't store age on the user? Regardless, this might be a useful helper instead of living on this specific component.

katiecrane: @sbsmith86 I guess it's possible that it's stored in Northstar, but it is not part of the response we get. I definitely think it might be useful to be more accessible, but I'm not sure where it should live then? would it go on `app.js`?

sbsmith86: we have a [`helpers.js`](https://github.com/DoSomething/rogue/blob/master/resources/assets/helpers.js)

katiecrane: wow I was looking right at it and didn't see it. thanks!

katiecrane:
@sbsmith86 updated to pull age calculation into helper function

PR #187, opened 2017-03-29
adds tertiary button to delete a post 

### LINKED ISSUE #186: Create a delete button to delete a post so it's not stored in our database.
_FEATURE OVERVIEW_
-------------------
#### User Story
As an admin, I want the option to delete the post so that it's not stored in our database (button only, no backend work). 

Sometimes a photo comes in that we need to delete completely.

Note: The wireframes doesn't have the "Delete" option, but as we talked about in a previous groomin session, we don't want a checkbox, we just want one of the statuses to be Delete. I spoke with Luke and he doesn't feel strongly with how it looks, so it could be a smaller button or whatever's easiest.

#### Why This Matters
It is illegal to keep questionable photos?

#### Definition of Done
There's the option to Delete a post.

chloealee:
#### What's this PR do?
- Adds a tertiary "Delete" button to delete a post so it's not stored in the database. 

👀 

The button should look like this (is it ok it's off center?):
screen shot 2017-03-29 at 4 09 13 pm

Another option I think could be cool and be extra obvious to not click is the `-danger` button as seen in Aurora:
screen shot 2017-03-29 at 4 07 41 pm

Would we prefer the `-danger` button instead? A couple thoughts:
- This is no longer in Forge - does this mean we don't want to use it anymore? 
- The `-danger` button has more clickable area than the `-tertiary` button. Since it's located right under the `ACCEPTED` button, it might cause more room for error, making it easier for an admin to accidentally hit `Delete` instead. 

#### Relevant tickets
Fixes #186 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/InboxItem/index.js
+++ b/resources/assets/components/InboxItem/index.js
@@ -32,6 +32,8 @@ class InboxItem extends React.Component {
           <ul className="form-actions -inline">
             <li> <input className="button" value="Accepted"/></li>
             <li> <input className="button -secondary" value="Rejected"/></li>
+            <br/>
+            <li> <input className="button -tertiary" value="Delete"/></li>
           </ul>
           <h4>Meta</h4>
           <p>


katiecrane:
I kind of like the tertiary button so it's less noticeable since I don't think it will be getting that much use. We will probably also have a popup or some type of confirmation when this is clicked so I don't think we need to worry about it getting hit by accident. Just my two cents!

chloealee:
ah good call - thanks for the input @katiecrane !

sbsmith86:
@chloealee i'm confused, isn't that what you did here?

chloealee:
@sbsmith86 yup! But offered another option to change to above :)

sbsmith86:
@chloealee yea but why not merge this?

chloealee:
@sbsmith86 oh my mistake! I thought I merged - I just closed by accident sorry for the confusion!

PR #193, opened 2017-03-31
Selectable tags buttons after Accepted is clicked 

### LINKED ISSUE #189: Add selectable tag buttons when an admin accepts a post
_FEATURE OVERVIEW_
-------------------
#### User Story
RB REVIEW: As an admin, when I Accept a post, I want selectable tags to show up under the buttons so that I can see the available tag options. See below:
screen_shot_2017-03-10_at_10 52 47_am

These tags can only be applied to photos that are 'Accepted.' The v1 tags are listed in the checklist below. I renamed 'Invisible' to 'Hide in Gallery' because I thought that'd be more clear for the campaign leads - this is different from what's in the wireframes!

#### Why This Matters
So that leads can apply additional information to the posts

#### Definition of Done
When a lead clicks the "Accepted" status, the tags listed below show up under the buttons and the lead can click on them and it looks like it's been selected.

chloealee:
#### What's this PR do?
- Hardcodes selectable tag buttons to show after the `Accepted` button has been clicked.
  - Hardcodes the state of each tag (this should come from server when we do backend work!)
- Adds custom CSS for `tag` buttons. 
- Adds ClassNames

- Click the `Accepted` button. Once clicked, the tag buttons should appear like the below:
screen shot 2017-04-03 at 12 26 05 pm

A few questions:
- Right now, the buttons are `-secondary` buttons. Is this ok or should be more stylized to be like what is in the mocks? Didn't want to break from our pattern library! 
- Do we want to add a `tags` array in the `CampaignController` and pass through as props similar to `signups` or is this hardcoding ok for now? 

**Note:** @DFurnes and I saw some funky aligning with the `Accepted`, `Rejected`, and `Delete` button. We should revisit this alignment later. 

#### Relevant tickets
Fixes #189 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/app.scss
+++ b/resources/assets/app.scss
@@ -15,17 +15,12 @@ $forge-path: "~@dosomething/forge/assets/";
   border: 0;
   margin: 10px 10px 0 0;
 
-  &:hover {
+  &:active, &.is-active {
     background: $blue;
-    color: white;
-  }
-
-  &:active {
-    background: $blue;
-    color: white;
+    color: $white;
   }
 }
 
-#left-aligned{
-  text-align: left;
+.aligned-actions li {
+  float: left;
 }
--- a/resources/assets/app.scss
+++ b/resources/assets/app.scss
@@ -2,3 +2,26 @@ $forge-path: "~@dosomething/forge/assets/";
 
 // Import Forge.
 @import "~@dosomething/forge/scss/forge";
+
+.tag {
+  list-style-type: none;
+  clear: both;
+  background: $light-gray;
+  color: $med-gray;
+  font-size: $font-regular;
+  font-weight: $weight-normal;
+  font-family: $font-proxima-nova;
+  text-align: center;
+  border: 0;
+  margin: 10px 10px 0 0;
+
+  &:hover {
+    background: $blue;
+    color: white;
+  }
+
+  &:active {
+    background: $blue;
+    color: white;
+  }
+}
--- a/resources/assets/app.scss
+++ b/resources/assets/app.scss
@@ -25,3 +25,7 @@ $forge-path: "~@dosomething/forge/assets/";
     color: white;
   }
 }
+
+#delete_post{
+  text-align: left;
+}
--- a/resources/assets/app.scss
+++ b/resources/assets/app.scss
@@ -26,6 +26,6 @@ $forge-path: "~@dosomething/forge/assets/";
   }
 }
 
-#delete_post{
+#left-aligned{
   text-align: left;
 }
--- a/resources/assets/components/InboxItem/index.js
+++ b/resources/assets/components/InboxItem/index.js
@@ -48,7 +48,7 @@ class InboxItem extends React.Component {
           <ul className="form-actions -inline">
             <li> <input className="button" value="Accepted" onClick={this.onAcceptClick}/></li>
             <li> <input className="button -secondary" value="Rejected"/></li>
-            <li> <input className="button -tertiary" id="delete_post" value="Delete"/></li>
+            <li> <input className="button -tertiary" id="left-aligned" value="Delete"/></li>
           </ul>
           {this.state.status === 'accepted' ? <Tags /> : null}
           <h4>Meta</h4>
--- a/resources/assets/components/InboxItem/index.js
+++ b/resources/assets/components/InboxItem/index.js
@@ -45,10 +45,10 @@ class InboxItem extends React.Component {
           <p>{post['postable']['caption']}</p>
           <h4>Why Statement</h4>
           <p>{post['signup']['why_participated']}</p>
-          <ul className="form-actions -inline">
-            <li> <input className="button" value="Accepted" onClick={this.onAcceptClick}/></li>
-            <li> <input className="button -secondary" value="Rejected"/></li>
-            <li> <input className="button -tertiary" id="left-aligned" value="Delete"/></li>
+          <ul>
+            <li><button className="button" onClick={this.onAcceptClick}>Accepted</button></li>
+            <li><button className="button -secondary">Rejected</button></li>
+            <li><button className="button -tertiary">Delete</button></li>
           </ul>
           {this.state.status === 'accepted' ? <Tags /> : null}
           <h4>Meta</h4>
--- a/resources/assets/components/InboxItem/index.js
+++ b/resources/assets/components/InboxItem/index.js
@@ -10,10 +10,10 @@ class InboxItem extends React.Component {
     this.state = {
       showTags: false,
     };
-    this._onAcceptClick = this._onAcceptClick.bind(this);
+    this.onAcceptClick = this.onAcceptClick.bind(this);
   }
 
-  _onAcceptClick() {
+  onAcceptClick() {
     this.setState({
       showTags: true,
     });
@@ -46,7 +46,7 @@ class InboxItem extends React.Component {
           <h4>Why Statement</h4>
           <p>{post['signup']['why_participated']}</p>
           <ul className="form-actions -inline">
-            <li> <input className="button" value="Accepted" onClick={this._onAcceptClick}/></li>
+            <li> <input className="button" value="Accepted" onClick={this.onAcceptClick}/></li>
             <li> <input className="button -secondary" value="Rejected"/></li>
             {this.state.showTags? <Tags /> : null}
             <br/>
--- a/resources/assets/components/InboxItem/index.js
+++ b/resources/assets/components/InboxItem/index.js
@@ -8,14 +8,14 @@ class InboxItem extends React.Component {
   constructor () {
     super();
     this.state = {
-      showTags: false,
+      status: 'pending',
     };
     this.onAcceptClick = this.onAcceptClick.bind(this);
   }
 
   onAcceptClick() {
     this.setState({
-      showTags: true,
+      status: 'accepted',
     });
   }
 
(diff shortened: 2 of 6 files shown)


DFurnes:
> Right now, the buttons are -secondary buttons. Is this ok or should be more stylized to be like what is in the mocks? Didn't want to break from our pattern library!

I think this is totally fine to merge, but let's make a follow-up pull request that better matches the mocks with a `.button.-tag` (or even just a standalone `.tag` if there aren't enough overlapping style rules).

> Do we want to add a tags array in the CampaignController and pass through as props similar to signups or is this hardcoding ok for now?

In my opinion, hardcoding these is totally okay for now (or maybe forever)... it all probably just depends on how often we end up changing the available tags.

resources/assets/components/InboxItem/index.js
```diff
@@ -2,7 +2,23 @@ import React from 'react';
...
 class InboxItem extends React.Component {
+  constructor () {
+    super();
+    this.state = {
+      showTags: false,
+    };
+    this._onAcceptClick = this._onAcceptClick.bind(this);
+  }
+
+  _onAcceptClick() {
```
DFurnes: I think this can just be named `onAcceptClick`.

resources/assets/components/InboxItem/index.js
```diff
@@ -2,7 +2,23 @@ import React from 'react';
...
+    this.setState({
+      showTags: true,
```
DFurnes: This doesn't need to happen in this PR, but it may eventually make more sense to save `status: 'accepted'` in the state, and then conditionally render things based on that (since the thing you're mutating is the status, and whether or not tags appear in the interface is just a side-effect of that).

chloealee:
@DFurnes cool, sounds good! For the `.button.-tag` or `.tag`, is that just custom css to add in to Rogue?

resources/assets/components/Tags/index.js
```diff
@@ -0,0 +1,21 @@
...
+import { map } from 'lodash';
+
+class Tags extends React.Component {
+  render() {
+
+    return (
+        <ul className="form-actions -inline">
+          <li><b> Tags </b></li>
+          <br/>
+          <li> <input className="button -secondary" value="Good Photo"/></li>
```
DFurnes: One thing that could make sense is having an key-value mapping of machine-name to human-readable name, and then looping over that to render this interface. That will probably help us reduce some boilerplate when rendering an existing reportback's information as well.

```js
const tags = {
  'good_photo': 'Good Photo',
  'good_quote': 'Good Quote',
  'hidden': 'Hide In Gallery',
  'sponsor': 'Good for Sponsor',
};

return (
  <ul className="form-actions -inline">
    {map(tags, (label, value) => <input className="button -secondary" value={label}/>)}
  </ul>
```

resources/assets/components/Tags/index.js
```diff
@@ -0,0 +1,21 @@
...
+          <li> <input className="button -secondary" value="Good Quote"/></li>
+          <br/>
```
DFurnes: `ul` elements should only have `li` inside them, so spacing like this should be handled with CSS if necessary.

resources/assets/components/Tags/index.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { map } from 'lodash';
+
+class Tags extends React.Component {
+  render() {
+
+    return (
+        <ul className="form-actions -inline">
+          <li><b> Tags </b></li>
```
DFurnes: This is a heading so should be pulled out from the list and given a heading tag:

```jsx
<h4>Tags</h4>
<ul className="form-actions -inline">
  {/* ... */}
```

DFurnes:
> @DFurnes cool, sounds good! For the .button.-tag or .tag, is that just custom css to add in to Rogue?

Yep! 🎨

resources/assets/components/InboxItem/index.js
chloealee: ah totally makes sense - updated that!

resources/assets/components/Tags/index.js
chloealee: @DFurnes for some reason when I do that, it breaks everything :-/ 
```
index.js:5 Uncaught Error: Cannot find module "../Tags"
    at webpackMissingModule (index.js:5)
    at Object.<anonymous> (index.js:5)
    at __webpack_require__ (bootstrap 78da61f…:19)
    at Object.defineProperty.value (index.js:5)
    at __webpack_require__ (bootstrap 78da61f…:19)
    at Object.defineProperty.value (app.js:8)
    at __webpack_require__ (bootstrap 78da61f…:19)
    at bootstrap 78da61f…:39
    at bootstrap 78da61f…:39
    at webpackUniversalModuleDefinition (universalModuleDefinition:7)
webpackMissingModule @ index.js:5
(anonymous) @ index.js:5
__webpack_require__ @ bootstrap 78da61f…:19
Object.defineProperty.value @ index.js:5
__webpack_require__ @ bootstrap 78da61f…:19
Object.defineProperty.value @ app.js:8
__webpack_require__ @ bootstrap 78da61f…:19
(anonymous) @ bootstrap 78da61f…:39
(anonymous) @ bootstrap 78da61f…:39
webpackUniversalModuleDefinition @ universalModuleDefinition:7
(anonymous) @ universalModuleDefinition:10
```
but I'm not sure why it breaks like that?

resources/assets/components/Tags/index.js
chloealee: ok will do! I'm not quite sure what you mean by the below?

>  That will probably help us reduce some boilerplate when rendering an existing reportback's information as well.

resources/assets/components/Tags/index.js
DFurnes: Are there any issues listed in the console? One thought is to make sure to wrap those in a `<div>` when returning from the `render` method if you're not. React doesn't like when you return multiple elements for a component.

chloealee: ah yes that did it - thank you! I remember that in the videos now!

resources/assets/components/Tags/index.js
DFurnes: When we end up displaying reviewed reportbacks in the admin interface, we'd need to be able to pass in values for each of these tags (e.g. `{'good_photo': true, 'good_quote': false}`).

It'd be nice to not have to manually check `values['good_photo']` and `values['good_quote']` and so on in order to render out the state of the interface.

chloealee: ooooo I see now - thank you for the explanation!

chloealee:
ooo awesome! I think I will try that custom css in this PR then if that's ok @DFurnes !

chloealee:
@DFurnes updated to add custom css styling (updating screen shot in picture of what it should look like)! Doesn't look _exactly_ like the mocks but did the best I could! Would love feedback :)

resources/assets/app.scss
```diff
@@ -2,3 +2,26 @@ $forge-path: "~@dosomething/forge/assets/";
...
+
+  &:hover {
+    background: $blue;
+    color: white;
+  }
+
+  &:active {
+    background: $blue;
+    color: white;
+  }
```
chloealee: @DFurnes this isn't working as expected :-/ When clicked, I want it to have a blue background with white font, similar to when it's hovered. Instead, it only has a blue border. See below (left is active and right is hover state). I want them to be the same! Any ideas? 

Also, it seems like only one button is able to be active at a time (when I click on another button, the border then highlights that one). Is this something we fix with React? 

screen shot 2017-03-31 at 1 01 08 pm

DFurnes: It looks like it's displaying as a text input since there isn't a `type` attribute. Try adding a `type="button"` or switching the element type to `<button>...</button>`. Not sure if you want to handle it in this PR or not, but we'll definitely want to track whether a tag is active or not within React & update the classname accordingly.

chloealee: I tried both (adding a `type="button"` and switching the element type to `<button>...</button>` and nothing changed :-/

resources/assets/components/InboxItem/index.js
```diff
@@ -30,11 +46,11 @@ class InboxItem extends React.Component {
           <h4>Why Statement</h4>
           <p>{post['signup']['why_participated']}</p>
           <ul className="form-actions -inline">
-            <li> <input className="button" value="Accepted"/></li>
+            <li> <input className="button" value="Accepted" onClick={this.onAcceptClick}/></li>
             <li> <input className="button -secondary" value="Rejected"/></li>
-            <br/>
-            <li> <input className="button -tertiary" value="Delete"/></li>
+            <li> <input className="button -tertiary" id="delete_post" value="Delete"/></li>
```
DFurnes: What's this `delete_post` ID for?

resources/assets/app.scss
DFurnes: @chloealee It still shows the gray background? Hmm... can you try forcing a rebuild by killing Webpack and running `npm start` again? Just in case something's cached.

resources/assets/components/InboxItem/index.js
chloealee: I just added that so the text is left aligned instead of center just for some styling but can totally change back! Here's the difference:
without `delete_post` ID:
screen shot 2017-03-31 at 4 32 39 pm
 with `delete_post` ID:
screen shot 2017-03-31 at 4 32 13 pm

resources/assets/app.scss
chloealee: @DFurnes I did that and even re-ran `npm run build` before doing `npm start` again :-/

resources/assets/components/InboxItem/index.js
DFurnes: Hmm, okay. I'm surprised the button is getting that extra padding. I think it's okay to merge as a quick-fix, but let's at the very least replace this with a more descriptive classname (maybe `.left-aligned`).

PR #199, opened 2017-03-31
GET /reportbacks endpoint for Phoenix-ashes gallery

### LINKED ISSUE #185: Get `/reportbacks` endpoint for Phoenix gallery to use
_FEATURE OVERVIEW_
-------------------
#### User Story
We want the `get /reportbacks` endpoint in Rogue to [mirror this endpoint in Phoenix](https://github.com/DoSomething/phoenix/blob/dev/documentation/endpoints/reportback-items.md#retrieve-all-reportback-items). The needed query parameters are:
- campaign
- count
- page

We don't need status because we can automatically only include `approved` items in the response. I don't think we will ever want to return `rejected` items. The Phoenix gallery does not display any user data, so I don't think we need the `load_user` option. It doesn't look like the Phoenix gallery uses the `random` option either so we don't need that.

#### Why This Matters
One more step to get activity out of Phoenix!

Keep Phoenix powering the gallery.

katiecrane:
#### What's this PR do?
Wrote out the `/reportbacks` `GET` endpoint to be used for the Phoenix-ashes gallery. Created a new transformer to put everything in the format that the Phoenix gallery expects, following the same format as [this endpoint](https://github.com/DoSomething/phoenix/blob/dev/documentation/endpoints/reportback-items.md#retrieve-all-reportback-items). If you are reading this I am working on the documentation but didn't want to possibly hold up review because of that!

Sample response:
```
{
  "data": [
    {
      "id": 4,
      "status": "approved",
      "caption": "Ipsam fugit magnam impedit sit quia quo.",
      "uri": "https://s3.amazonaws.com/ds-rogue-test/uploads/reportback-items/12-1484929292.jpeg",
      "media": {
        "uri": "https://s3.amazonaws.com/ds-rogue-test/uploads/reportback-items/12-1484929292.jpeg",
        "type": "image"
      },
      "created_at": "2017-03-28T14:33:32+00:00",
      "reportback": {
        "id": 2,
        "created_at": "2017-03-28T14:33:32+00:00",
        "updated_at": "2017-03-28T14:33:32+00:00",
        "quantity": null,
        "why_participated": "Vel excepturi soluta aut natus ut.",
        "flagged": "false"
      }
    },
    {
      "id": 6,
      "status": "approved",
      "caption": "Nostrum et nemo totam quia occaecati.",
      "uri": "https://s3.amazonaws.com/ds-rogue-test/uploads/reportback-items/12-1484929292.jpeg",
      "media": {
        "uri": "https://s3.amazonaws.com/ds-rogue-test/uploads/reportback-items/12-1484929292.jpeg",
        "type": "image"
      },
      "created_at": "2017-03-28T14:33:32+00:00",
      "reportback": {
        "id": 2,
        "created_at": "2017-03-28T14:33:32+00:00",
        "updated_at": "2017-03-28T14:33:32+00:00",
        "quantity": null,
        "why_participated": "Vel excepturi soluta aut natus ut.",
        "flagged": "false"
      }
    }
  ],
  "meta": {
    "pagination": {
      "total": 2,
      "count": 2,
      "per_page": 20,
      "current_page": 1,
      "total_pages": 1,
      "links": []
    }
  }
}
```
See if the logic/structure makes sense. @chloealee take a look and see what you think about the format! Or if you want to pull down the branch and test it with your code and let me know about what I should change, that works too!

#### Relevant tickets
Fixes #185

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -23,8 +23,33 @@ class ReportbackController extends ApiController
     public function __construct(ReportbackService $reportbackService)
     {
         $this->reportbackService = $reportbackService;
-        $this->transformer = new ReportbackTransformer;
-        $this->itemTransformer = new ReportbackItemTransformer;
+        $this->transformer = new PhoenixGalleryTransformer;
```
sbsmith86: I feel like this could just be `ReportbackTransformer` as we don't want name transformers based on how they are going to be used, but more like what model they are transforming.

app/Http/Transformers/PhoenixGalleryTransformer.php
```diff
@@ -0,0 +1,41 @@
...
+     * @param \Rogue\Models\Photo $photo
+     * @return array
+     */
+    public function transform(Post $post)
+    {
+        $signup = $post->signup;
+        $result = [
+            'id' => $post->postable_id,
+            'status' => $post->status,
+            'caption' => $post->content->caption,
```
sbsmith86: This lives on the `posts` table now and there is no `content()` relationship. So you should be able to just pull `$post->caption`

app/Http/Transformers/PhoenixGalleryTransformer.php
```diff
@@ -0,0 +1,41 @@
...
+            'uri' => $post->content->file_url,
```
sbsmith86: Same here. Also I don't think we need to return it twice in the response object.

sbsmith86: unless that is what is currently done in the `/reportbacks` endpoint...

katiecrane: @sbsmith86 this is what is in there:
```
      uri: "https://www.dosomething.org/api/v1/reportback-items/658",
      media: {
        uri: "https://www.dosomething.org/sites/default/files/styles/480x480/public/reportbacks/362/image42.jpeg",
        type: "image"
      }
```
they are two different URLs and I wasn't sure what they were used for so I put it in there twice. @chloealee do you know if you use both of these?

app/Http/Controllers/Api/ReportbackController.php
katiecrane: @sbsmith86 yeah I wasn't sure how to handle that since there already is a `ReportbackTransformer` but since that isn't needed anymore I could overwrite it with this?

app/Http/Transformers/PhoenixGalleryTransformer.php
sbsmith86: @katiecrane One of these points to the admin page for a repotback item. we provide a little administration link per reportback when admins are viewing the page. we don't have the equivalent in Rogue yet, so I _think_ you can leave it out with a note we need to add it in when it is created. And then we can leave the media properties the same since that points to the actual file url.

app/Http/Controllers/Api/ReportbackController.php
sbsmith86: yea i think that is fine.

app/Http/Transformers/PhoenixGalleryTransformer.php
katiecrane: @sbsmith86 ah nice! will do 👍

katiecrane:
@sbsmith86 I think all updates have been addressed now

app/Http/Transformers/ReportbackTransformer.php
```diff
@@ -2,56 +2,41 @@
...
-     */
-    protected $defaultIncludes = [
-        'reportback_items',
-    ];
-
-    /**
      * Transform resource data.
      *
-     * @param \Rogue\Models\Reportback $reportback
+     * @param \Rogue\Models\Photo $photo
```
sbsmith86: This should be a `Post` now!

katiecrane: ahh good catch! ⚾️

katiecrane:
I merged before I added in the documentation! Sorryyy there will be another PR soon for that

PR #215, opened 2017-04-11
Get the reportback inbox closer to the mocks

### LINKED ISSUE #213: Make the inbox match the mocks
_FEATURE OVERVIEW_
-------------------
#### User Story
Let's actually do what @lkpttn designed! 3 columns and the right sizes and colors!

It should look like this:
screen_shot_2017-04-05_at_12 25 45_pm

katiecrane:
#### What's this PR do?
🎨 🎨 🎨 ❗️ 

This is what it looks like now:
image

1. There are two new button options to get the green and red buttons. 
2. There is an option to have a `container__block` be a third of the page instead of a half.
3. There is a fun new option called `quantity` that makes everything bold and purple and puts it in a purple box oOoOoOo

The other thing I did was to apply these new styles where they needed to be used.

🎨 👀 Did I CSS right?

#### Relevant tickets
Fixes #213

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/app.scss
+++ b/resources/assets/app.scss
@@ -56,9 +56,8 @@ $forge-path: "~@dosomething/forge/assets/";
 }
 
 .quantity {
-  outline-color: #4e2b63;
-  padding: 2px;
-  outline-style: solid;
+  border: 3px solid #4e2b63;
+  padding: 4px 12px;
   color: #4e2b63;
   font-weight: bold;
 }
--- a/resources/assets/app.scss
+++ b/resources/assets/app.scss
@@ -24,3 +24,33 @@ $forge-path: "~@dosomething/forge/assets/";
 .aligned-actions li {
   float: left;
 }
+
+.button.-accepted {
+  background: #80E85D;
+
+  &:hover {
+    background: lighten(#80E85D, $color-tint);
+  }
+
+  &:active {
+    background: darken(#80E85D, $color-tint);
+  }
+}
+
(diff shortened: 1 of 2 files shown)


sbsmith86:
AH! Haven't reviewed the code but it look gooooood from the screenshot!

resources/assets/app.scss
```diff
@@ -24,3 +24,41 @@ $forge-path: "~@dosomething/forge/assets/";
...
+  }
+}
+
+.container__block.-third {
+  @include media($tablet) {
+    @include span(4 of 12);
+  }
+}
+
+.quantity {
```
DFurnes: I think you could give this a `display: inline-block` to get more padding on the sides.

resources/assets/app.scss
```diff
@@ -24,3 +24,41 @@ $forge-path: "~@dosomething/forge/assets/";
...
+  outline-color: #4e2b63;
```
DFurnes: I think we want to use a `border` here instead of an outline.

resources/assets/app.scss
katiecrane: @DFurnes hmmm I added it but when I uncheck and check it in dev tools I don't see any difference 🤔

DFurnes: css

katiecrane:
@DFurnes updated! also the photo has been updated to reflect the changes

PR #218, opened 2017-04-14
Overview aggregates

### LINKED ISSUE #217: Add pending, approved, and rejected counts to the campaign overview page
_FEATURE OVERVIEW_
-------------------
#### User Story

Show a count of pending, approved, and rejected posts per campaign run on the campaign overview page. 

The `pending` number should link to the inbox page.

sbsmith86:
#### What's this PR do?

Adds the number of pending, accepted, and rejected posts there are for each campaign on the campaign overview page. 

* Creates new relationships on `Signup` to define relationships to posts with a particular status.
* Adds some local query scopes to `Signup` to define some commonly used queries including one that will include the number of accepted, rejected, and pending posts on each signup retrieved.
* Adds a method on the `CampaignService` that sums up the status counts for a campaign
* Passes this info to the frontend and display it.
* Links the pending count to the inbox page
  - *Note on this* - dealing with campaign runs was becoming increasingly difficult. For one, we only get the current run from the phoenix API so we would have to do some work on that end to get all runs back. Two, when we get the current run, we get it per language and each of those could have different run ids. In order to simplify this, as a first past, we decided to show post counts _per campaign_ instead. So the inbox page has been updated to reflect that change, and we show pending posts per `campaign_id` instead of per run id. 

image

I think you should be safe to do this commit by commit. Or start looking at the `Signup` model and the relationships defined there, then look at how they are used in `CampaignService` and `CampaignController`

#### Relevant tickets
Fixes #217 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -63,12 +63,9 @@ class CampaignsController extends Controller
     /**
      * Show particular campaign inbox.
      */
-    public function show($campaign_run_id)
+    public function show($campaignId)
     {
-        // Pull in all signups for the given run that have pending posts, and include their pending posts
-        $signups = Signup::whereHas('posts', function ($query) {
-            $query->where('status', 'pending');
-        })->where('campaign_run_id', $campaign_run_id)->with('posts')->get();
+        $signups = Signup::campaign($campaignId)->has('pending')->with('pending')->get();
 
         // For each post, get and include the user
         $signups->each(function ($item) {
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -43,29 +43,19 @@ class CampaignsController extends Controller
     public function index()
     {
         $ids = $this->campaignService->getCampaignIdsFromSignups();
+
         $campaigns = $this->campaignService->findAll($ids);
-        $campaignsWithCounts = $campaigns->map(function($campaign, $key) {
+
+        $campaigns = $campaigns->map(function($campaign, $key) {
             if ($campaign) {
-
-                $campaign['accepted_count'] = 0;
-                $campaign['pending_count'] = 0;
-                $campaign['rejected_count'] = 0;
-
-                $signups = Signup::has('posts')->where('campaign_id', '=', $campaign['id'])
-                    ->withCount(['accepted', 'pending', 'rejected'])->get();
-
-                $signups->each(function($signup) use ($campaign) {
-                    $campaign['accepted_count'] += $signup['accepted_count'];
-                    $campaign['pending_count'] += $signup['pending_count'];
-                    $campaign['rejected_count'] += $signup['rejected_count'];
-                });
+                $campaign = $this->campaignService->getCampaignPostStatusCounts($campaign);
             }
 
             return $campaign;
         });
 
-        $causes = $this->campaignService->groupByCause($campaignsWithCounts);
-        dd($causes);
+        $causes = $this->campaignService->groupByCause($campaigns);
+
         return view('pages.campaign_overview')
             ->with('state', $causes);
     }
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -44,7 +44,8 @@ class CampaignsController extends Controller
     {
         $ids = $this->campaignService->getCampaignIdsFromSignups();
         $campaigns = $this->campaignService->findAll($ids);
-        $campaigns = $this->campaignService->getCampaignPostStatusCounts($campaigns);
+        $campaigns = $this->campaignService->appendStatusCountsToCampaigns($campaigns);
+
         $causes = $this->campaignService->groupByCause($campaigns);
 
         return view('pages.campaign_overview')
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -44,10 +44,30 @@ class CampaignsController extends Controller
     {
         $ids = $this->campaignService->getCampaignIdsFromSignups();
         $campaigns = $this->campaignService->findAll($ids);
-        $campaigns = $this->campaignService->groupByCause($campaigns);
+        $campaignsWithCounts = $campaigns->map(function($campaign, $key) {
+            if ($campaign) {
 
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/CampaignsController.php
```diff
@@ -43,22 +43,29 @@ public function __construct(Registrar $registrar, CampaignService $campaignServi
     public function index()
     {
         $ids = $this->campaignService->getCampaignIdsFromSignups();
+
         $campaigns = $this->campaignService->findAll($ids);
-        $campaigns = $this->campaignService->groupByCause($campaigns);
+
+        $campaigns = $campaigns->map(function($campaign, $key) {
+            if ($campaign) {
+                $campaign = $this->campaignService->getCampaignPostStatusCounts($campaign);
```
DFurnes: This seems worrisome since it'll make N database queries (one per campaign ID)... can we rephrase this query as a query to the signups table that's then grouped or filtered by the given campaign IDs?

DFurnes: I think you should be able to pass in the array of campaigns to that scope you defined on the model:

```php
$ids = array_pluck($campaign, 'id');
$campaigns = Signup::campaign($ids)
    ->includePostStatusCounts()
    ->groupBy('campaign_id')
    ->get();

$campaigns = $campaigns->map(function($campaign) {
    $campaign['accepted_count'] = $campaign->sum('accepted_count');
    $campaign['pending_count'] = $campaign->sum('pending_count');
    $campaign['rejected_count'] = $campaign->sum('rejected_count');

    return $campaign;
});
```

resources/assets/components/CampaignInbox/index.js
```diff
@@ -15,11 +15,32 @@ class CampaignInbox extends React.Component {
...
+          { map(posts, (post, key) => <InboxItem key={key} details={post} />) }
+        </div>
+      )
+    } else {
+      // @todo - make this into an actual component.
+      return (
+        <div className="container">
+          <h2 className="-padded">No Posts to review!</h2>
+          <div className="container">
+            <img src={sample(nothingHere)} />
```
DFurnes: This is awesome. 👏

app/Http/Controllers/CampaignsController.php
sbsmith86: Ah will try. great point.

sbsmith86:
@DFurnes made those updates. Can you take another look see?

app/Services/CampaignService.php
```diff
@@ -176,4 +177,33 @@ public function getCampaignIdsFromSignups()
...
+     * @return Illuminate\Database\Eloquent\Collection $campaigns
+     */
+    public function getCampaignPostStatusCounts($campaigns)
+    {
+        $ids = array_pluck($campaigns, 'id');
+
+        $campaignsWithSignups = Signup::campaign($ids)
+            ->includePostStatusCounts()
+            ->get()
+            ->groupBy('campaign_id');
```
DFurnes: I think we want this `groupBy` above the `get` so it's grouping in the SQL query (and only returning the grouped results), rather than in-memory using the returned Collection instance.

sbsmith86: hmm that wasn't working before, but i think there was something wrong with my query then.

sbsmith86: @DFurnes actually this isn't going to work. I need all of the individual signups with their post counts keyed by campaign_id so that I can then sum them. If I try to do the `groupBy` before the `get` I only get one signup per campaign_id. 

If it is more performant to do this in sql then I could try to write a sql query that will get me campaigns and counts without the collection stuff. Do you think that would be better.

sbsmith86: For example, would it be better to try and run this Raw query?

```
select signups.campaign_id,
sum(case when posts.status = 'accepted' then 1 else 0 end) as approved_count, 
sum(case when posts.status = 'pending' then 1 else 0 end) as pending_count,
sum(case when posts.status = 'rejected' then 1 else 0 end) as rejected_count
from signups
join posts
	on signups.id = posts.signup_id
where signups.campaign_id = 1334
group by signups.campaign_id;
```

DFurnes: Oh hmmm, I feel like you should be able to do that with the query builder but I'm not particularly morally opposed to raw SQL queries either (especially for a pretty unique thing like this!)

Before going too far down this route I would just double-check that `SUM()` function can use indexes and wouldn't get too slow if we had a bajillion records in the database.

app/Services/CampaignService.php
```diff
@@ -176,4 +176,54 @@ public function getCampaignIdsFromSignups()
...
+                ->leftJoin('posts', 'signups.id', '=', 'posts.signup_id')
+                ->select('signups.campaign_id',
+                    DB::raw('SUM(case when posts.status = "accepted" then 1 else 0 end) as accepted_count'),
+                    DB::raw('SUM(case when posts.status = "pending" then 1 else 0 end) as pending_count'),
+                    DB::raw('SUM(case when posts.status = "rejected" then 1 else 0 end) as rejected_count'))
+                ->wherein('campaign_id', $ids)
+                ->groupBy('signups.campaign_id')
+                ->get();
+
+        return $campaignsWithCounts ? collect($campaignsWithCounts)->keyBy('campaign_id') : null;
```
DFurnes: Should this return an empty collection instead of `null`? Particularly thinking about what happens when we call a method on this expecting that it'll be a Collection instance down below:

```php
$statusCounts = $campaignsWithCounts->get($campaign['id']);
```

app/Services/CampaignService.php
```diff
@@ -176,4 +176,54 @@ public function getCampaignIdsFromSignups()
...
+     * Gets the count of pending, accepted, and rejected stautses on each post for a campaign.
+     *
+     * @param  array $campaigns
+     * @return Illuminate\Database\Eloquent\Collection $campaigns
+     */
+    public function getCampaignPostStatusCounts($campaigns)
+    {
+        $ids = array_filter(array_pluck($campaigns, 'id'));
+
+        $campaignsWithCounts = DB::table('signups')
```
DFurnes: Nice!

app/Services/CampaignService.php
sbsmith86: yes that makes sense!

PR #221, opened 2017-04-17
Auto-Orient photo

### LINKED ISSUE #220: Orient photo based on EXIF data.
### User Story
We currently strip EXIF data when submitting to Phoenix, because Phoenix is dumb and doesn't make any use of it. Since we can be smarter in Rogue, we'd like to use the EXIF data to "actually" rotate the image so it appears nicely cross-browser.

### Additional Information (optional)
We can use Intervention Image's orientate() method for this.

### Why This Matters
We want our users' photos to appear correctly on all browsers, whether or not they support rendering orientation based on EXIF metadata.

sbsmith86:
#### What's this PR do?

Orient an uploaded file using [Intervention Image's `Orientate()`](http://image.intervention.io/api/orientate) method before storing to s3. 

Full disclosure, I am not sure this is needed. I have uploaded various photos via the `/posts` api endpoint, including a photo that was tagged as rotating incorrectly before, and it seems to working fine in rogue. 

However, this should safe guard us if something is weird. 

Also, this is very hard to test. I haven't been able to fine a photo that is uploading/displaying incorrectly so it is hard to tell if this is working as expected. 

#### Relevant tickets
Fixes #220 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -6,6 +6,7 @@ use Rogue\Models\Post;
 use Rogue\Services\AWS;
 // use Rogue\Models\Review;
 use Rogue\Services\Registrar;
+use Intervention\Image\Facades\Image;
 
 class PostRepository
 {
@@ -42,7 +43,10 @@ class PostRepository
     public function create(array $data, $signupId)
     {
         if (isset($data['file'])) {
-            $fileUrl = $this->aws->storeImage($data['file'], $signupId);
+            // Auto-orient the photo by default based on exif data.
+            $img = Image::make($data['file'])->orientate();
+
+            $fileUrl = $this->aws->storeImage((string) $img->encode('data-url'), $signupId);
 
             $editedImage = $this->crop($data, $signupId);
         } else {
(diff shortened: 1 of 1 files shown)


app/Repositories/PostRepository.php
```diff
@@ -42,7 +43,10 @@ public function __construct(AWS $aws, Registrar $registrar)
     public function create(array $data, $signupId)
     {
         if (isset($data['file'])) {
-            $fileUrl = $this->aws->storeImage($data['file'], $signupId);
+            // Auto-orient the photo by default based on exif data.
+            $img = Image::make($data['file'])->orientate();
```
DFurnes: Totally nitpicky, but I think in general we prefer spelling out full variable names: `$image` over `$img`.

DFurnes:
> Also, this is very hard to test. I haven't been able to fine a photo that is uploading/displaying incorrectly so it is hard to tell if this is working as expected.

Yeah, totally. If we do find an image, we should plop it in the repository and then set up some automated testing (comparing it against a properly "orientated" version) to prove it continues to work.

PR #226, opened 2017-04-17
Status button Component

### LINKED ISSUE #222: status buttons as their own components
_FEATURE OVERVIEW_
-------------------
#### User Story

On the inbox page, the "Accepted" and "Rejected" buttons should probably be a new React component `StatusButton` or something similar. These buttons have their own set of functionality and it would be cool if you could display them on other places in the app if you wanted. 

#### Why This Matters

Keeping things as smaller components allows us to reuse them down the line, if we need. 

The buttons are currently apart of the whole `InboxItem` component.

sbsmith86:
#### What's this PR do?

* Creates a new `StatusButton` component that can handle making status updates to the `InboxItem`
* Moves styling for this component into it's own partial.

Everything stays the same but starts to extract some of the logic. 

Right now, the `InboxItem` is in control of it's own local state and the status buttons will update that. 

If I am understanding this stuff correctly, when we move to redux we can start thinking about a global state for the app that has the status for each item on the page. The status buttons could just update the state of the item they are attached to. 

#### Relevant tickets
Fixes #222 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/StatusButton/index.js
+++ b/resources/assets/components/StatusButton/index.js
@@ -1,14 +1,12 @@
 import React from 'react';
 import './status-button.scss';
+import classnames from 'classnames';
 
 class StatusButton extends React.Component {
-  getClass() {
-    return `button -${this.props.type}`;
-  }
   // Ideal setStatus would be defined on this component and would set some sort of global state via redux
   render() {
     return (
-      <button className={this.getClass()} onClick={() => this.props.setStatus(this.props.type)}>{this.props.type}</button>
+      <button className={classnames('button', `-${this.props.type}`)} onClick={() => this.props.setStatus(this.props.type)}>{this.props.type}</button>
     )
   }
 }
--- a/resources/assets/components/StatusButton/index.js
+++ b/resources/assets/components/StatusButton/index.js
@@ -1,0 +1,18 @@
+import React from 'react';
+import './status-button.scss';
+// import { map } from 'lodash';
+
+
+class StatusButton extends React.Component {
+  getClass() {
+    return `button -${this.props.type}`;
+  }
+
+  render() {
+    return (
+      <button className={this.getClass()} onClick="">{this.props.type}</button>
+    )
+  }
+}
+
+export default StatusButton;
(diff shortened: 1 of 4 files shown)


resources/assets/components/StatusButton/index.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import './status-button.scss';
+// import { map } from 'lodash';
+
+
+class StatusButton extends React.Component {
+  getClass() {
+    return `button -${this.props.type}`;
```
DFurnes: The [`classnames`](https://www.npmjs.com/package/classnames) helper is really nice for this!

resources/assets/components/StatusButton/index.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import './status-button.scss';
+// import { map } from 'lodash';
```
DFurnes: 👾

resources/assets/components/StatusButton/index.js
```diff
@@ -0,0 +1,18 @@
...
+
+
+class StatusButton extends React.Component {
+  getClass() {
+    return `button -${this.props.type}`;
+  }
+  // Ideal setStatus would be defined on this component and would set some sort of global state via redux
+  render() {
+    return (
+      <button className={this.getClass()} onClick={() => this.props.setStatus(this.props.type)}>{this.props.type}</button>
```
DFurnes: This is totally subjective, but this looks like a great example of a functional component!

```js
import React from 'react';
import './status-button.scss';

export default (props) => (
  <button className={classnames('button', props.type} onClick={() => props.setStatus(props.type)}>
    {this.props.type}
  </button>
);
```

resources/assets/components/StatusButton/status-button.scss
```diff
@@ -0,0 +1,32 @@
+/**
+ * Table
```
DFurnes: ✏️ ❓

DFurnes:
> If I am understanding this stuff correctly, when we move to redux we can start thinking about a global state for the app that has the status for each item on the page. The status buttons could just update the state of the item they are attached to.

Yep, that's 100% it!

resources/assets/components/StatusButton/index.js
sbsmith86: ooo i like this!

PR #229, opened 2017-04-20
Display campaign info on campaign inbox page

### LINKED ISSUE #227: Add campaign information to inbox page
_FEATURE OVERVIEW_
-------------------
#### User Story
For each Post in the inbox, display `nouns verbed` after the quantity so that admins know what the members did. Also display the campaign name and run at the top of the page.

Wireframe:
screen_shot_2017-04-05_at_12 25 45_pm 1

#### Why This Matters
Context! Admins will want to know how many runs the member verbed, and also which campaign they are looking at and which run.

The campaign run id is in the url, so you could go into Phoenix or the Phoenix database and find the information you wanted which is definitely not how we want to handle that.

katiecrane:
#### What's this PR do?
Adds campaign title to the header. Adds nouns and verbs after quantity for Posts.

image

1. Does everything look right on the page
2. Does the way that the campaign data gets passed around make sense?

I don't think so!

#### Relevant tickets
Fixes #227

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -59,7 +59,6 @@ class CampaignsController extends Controller
     {
         $signups = Signup::campaign([$campaignId])->has('pending')->with('pending')->get();
 
-        // @TODO: handle inbox zero state
         // For each post, get and include the user
         $signups->each(function ($item) {
             $item->posts->each(function ($item) {
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -74,7 +74,6 @@ class CampaignsController extends Controller
             ->with('state', [
                 'signups' => $signups,
                 'campaign' => $campaign_data,
-                'title' => $campaign_data['title'],
             ]);
     }
 }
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/CampaignsController.php
```diff
@@ -67,9 +67,14 @@ public function show($campaignId)
...
         });
 
+        // Get the campaign data
+        $campaign_data = $this->campaignService->find($campaignId);
+
         return view('pages.campaign_inbox')
             ->with('state', [
                 'signups' => $signups,
+                'campaign' => $campaign_data,
+                'title' => $campaign_data['title'],
```
sbsmith86: I feel like title and could be in the `campaign` property and doesn't need to be separate.

app/Services/CampaignService.php
```diff
@@ -40,9 +40,11 @@ public function find($id)
             $campaign = $this->phoenix->getCampaign($id);
 
             $this->cache->store($campaign['data']['id'], $campaign['data']);
+
+            $campaign = $campaign['data'];
         }
 
-        return $campaign['data'];
+        return $campaign;
```
sbsmith86: Think you can just return `return $campaign['data']` without storing it in a variable.

katiecrane: @sbsmith86 I was seeing that when it was stored in cache already, the data lived in `$campaign` and not in `$campaign['data']`, so that's why I made that change

app/Http/Controllers/CampaignsController.php
katiecrane: good call!

PR #231, opened 2017-04-26
Some /posts fixes that are needed for keeper upper and also life

katiecrane:
#### What's this PR do?
1. Updated 2 validation rules:
    - `status` - fixed it to make sure that the status is any of `pending`, `accepted`, `rejected`
        - caveat: potentially in the future when posts come in, they will all be `pending`, so we may want to remove this rule and have Rogue automatically assign `pending` to incoming posts
    - `remote_addr` - updated to make sure it is a string, since it is sometimes a list of IPs 

2. When checking for an image, we can't just check if `data['file']` is empty, because broken images will come over with that value as `null`, so we get false negatives using `empty`. Instead, check if the `$data` array has a key `file` to see if the requester tried to send a file at all.

👀 

These fixes were necessary for the keeper upper script.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Requests/PostRequest.php
+++ b/app/Http/Requests/PostRequest.php
@@ -26,9 +26,9 @@ class PostRequest extends Request
             'campaign_id' => 'required|int',
             'campaign_run_id' => 'required|int',
             'caption' => 'string',
-            'status' => 'pending',
+            // 'status' => 'pending',
             'source' => 'string',
-            'remote_addr' => 'ip',
+            // 'remote_addr' => 'ip',
             // 'file' => 'string', // @TODO - should do some better validation around files.
         ];
     }
--- a/app/Http/Requests/PostRequest.php
+++ b/app/Http/Requests/PostRequest.php
@@ -26,9 +26,9 @@ class PostRequest extends Request
             'campaign_id' => 'required|int',
             'campaign_run_id' => 'required|int',
             'caption' => 'string',
-            // 'status' => 'pending',
+            'status' => 'in:pending,accepted,rejected',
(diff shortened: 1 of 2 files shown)


sbsmith86:
> caveat: potentially in the future when posts come in, they will all be pending, so we may want to remove this rule and have Rogue automatically assign pending to incoming posts

We set the default of the [`status` column to be pending ](https://github.com/DoSomething/rogue/blob/b8d1dd093901be9c90115e590ae65fb6dbe36fbe/database/migrations/2017_03_20_134838_add_columns_to_post_table.php#L16)

app/Http/Requests/PostRequest.php
```diff
@@ -26,9 +26,9 @@ public function rules()
             'campaign_id' => 'required|int',
             'campaign_run_id' => 'required|int',
             'caption' => 'string',
-            'status' => 'pending',
+            'status' => 'in:pending,accepted,rejected',
             'source' => 'string',
-            'remote_addr' => 'ip',
+            'remote_addr' => 'string',
```
sbsmith86: should it be `'ip|string'`? or is this always a string.

katiecrane: @sbsmith86 it still works with just one IP address, so I think it is always a string

katiecrane:
@sbsmith86 a-ha! I guess in the future we would still want this validation (even with setting status as `pending` by default) so no one can set crazy statuses like `pickle` or `ghost` or `flagged`

sbsmith86:
@katiecrane word! I think the validation is great :)

PR #232, opened 2017-04-27
Quantity pending

### LINKED ISSUE #177: Review quantity_pending logic
When you send a request over to the `/posts` endpoint to update quantity you have to send a `quantity_pending` param for the pending quantity to be stored. 

The logic should be that the client sends over a `quantity` param on creation/update and that gets stored in `quantity_pending` until approval. 

Would be open to discussing this further and reviewing it's necessity/how it should work.

sbsmith86:
#### What's this PR do?

Makes some updates in the code base so that `quantity` is stored in the `quantity` column on the `signups` table. 

We have two columns on the `signups` table: `quantity` and `quantity_pending`. We originally discussed storing quantity in the "pending" column until an admin reviewed it but that was proving more complicated and also wasn't inline with recent thinking. So we are only going to utilize the `quantity` column for now. Anything previously stored in `quantity_pending` will be migrated to `quantity` at a later time. 

#### Relevant tickets
Fixes #177 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Services/PostService.php
+++ b/app/Services/PostService.php
@@ -65,7 +65,7 @@ class PostService
         // @TODO: This will is only temporary and will be removed!
         // If this is a signup update, get the most recent post.
         // If there is a quantity_pending, this is a signup.
-        if ($post->quantity_pending) {
+        if ($post->quantity || $post->quantity_pending) {
             $signupId = $post->id;
             // Find the post with this signup id.
             $post = Post::where('signup_id', $signupId)->first();
--- a/app/Services/PostService.php
+++ b/app/Services/PostService.php
@@ -62,15 +62,6 @@ class PostService
     {
         $post = $this->repository->update($signup, $data);
 
-        // @TODO: This will is only temporary and will be removed!
-        // If this is a signup update, get the most recent post.
-        // If there is a quantity_pending, this is a signup.
-        if ($post->quantity || $post->quantity_pending) {
-            $signupId = $post->id;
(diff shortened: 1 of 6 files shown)


app/Services/PostService.php
```diff
@@ -65,7 +65,7 @@ public function update($signup, $data, $transactionId)
         // @TODO: This will is only temporary and will be removed!
         // If this is a signup update, get the most recent post.
         // If there is a quantity_pending, this is a signup.
-        if ($post->quantity_pending) {
+        if ($post->quantity || $post->quantity_pending) {
```
katiecrane: I'm confused about what this whole conditional is used for. It seems like `$post = $this->repository->update($signup, $data);` above would take care of everything we'd need?

sbsmith86: hmm this seems like legacy stuff that can be removed, as the @todo mentions but I wasn't sure either just wanted to make sure that `quantity` and `quantity_pending` was accounted for. I guess we can remove it now.

katiecrane: @sbsmith86 that sounds reasonable! everything seems to be working so I don't think it's breaking anything, I just couldn't figure out what it was doing. It says in the comment that if there is `quantity_pending` then it's a signup, but that's not necessarily true so that confused me. But yeah seems like we can remove, I don't have a strong preference for if that's in this PR or not.

sbsmith86:
@katiecrane removed that piece!

PR #238, opened 2017-04-28
Update /reviews endpoint and reviews test 

### LINKED ISSUE #171: Check/update reviews endpoint after posts update
We moved `status` to the `posts` table which could have broken the `/reviews` endpoint. Double check that and fix, if needed.

chloealee:
#### What's this PR do?
Updates `/reviews` endpoint and reviews tests. 

-Hit the `/reviews` endpoint
  - The post you reviewed should have the updated status in the `posts` table.
  - The information should be added in the `reviews` table with all expected info. 
  - Two events should be logged - one for the review and one for updated post. 
- Run `vendor/bin/phpunit --filter=Reviews` to make sure all tests are passing and/or `vendor/bin/phpunit` to make sure all tests are still passing! 

#### Relevant tickets
Fixes #171 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/tests/ReviewsApiTest.php
+++ b/tests/ReviewsApiTest.php
@@ -26,7 +26,7 @@ class ReviewsApiTest extends TestCase
 
         $review = [
             'post_id' => $post->id,
-            'status' => $this->faker->word(),
+            'status' => 'accepted',
             'admin_northstar_id' => str_random(24),
         ];
 
--- a/tests/ReviewsApiTest.php
+++ b/tests/ReviewsApiTest.php
@@ -37,9 +37,9 @@ class ReviewsApiTest extends TestCase
         $response = $this->decodeResponseJson();
 
         // Make sure we created a event for the review.
-        // $this->seeInDatabase('events', [
-        //     'status' => $response['data']['content']['status'],
-        // ]);
+        $this->seeInDatabase('events', [
+            'created_at' => $response['data']['created_at'],
+        ]);
 
         // Make sure a review is created.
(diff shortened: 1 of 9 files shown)


tests/ReviewsApiTest.php
```diff
@@ -13,54 +13,42 @@ class ReviewsApiTest extends TestCase
...
 
-    //     $review = [
-    //         'rogue_event_id' => $post->event_id,
-    //         'status' => $this->faker->word(),
-    //         'event_type' => 'review_photo',
-    //         'reviewer' => str_random(24),
-    //     ];
+        $review = [
+            'post_id' => $post->id,
+            'status' => $this->faker->word(),
```
katiecrane: I thought that we had validation to make sure that the Post status was `accepted` or ` rejected` or `pending`. Does this not give a validation error because it comes in through the `/reviews` endpoint instead of the `/posts` one?

chloealee: oh that is a great point @katiecrane ! I think you're right since the `ReviewsRequest` doesn't have this validation. I'll add that now!

katiecrane: @chloealee now that there is validation, I think this test will fail unless the word is one of the ones we accept. Is it bad to hardcode in one of our accepted words here?

chloealee: oh duh! I wasn't thinking, thanks for catching this :) I just updated the test - I don't think it is bad to hardcode this here!

PR #239, opened 2017-05-01
Allow backdating signup updates

katiecrane:
#### What's this PR do?
Allow capability to backdate updates to signups, meaning when the quantity or why gets updated. These updates will be coming in through the keeper upper script so we will want to preserve the original signups for this type of update. It only backdates the signup if it comes in with an `updated_at` parameter.

I made sure the events stuff all still worked properly, but that's the part I'd be most worried about I think.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -91,10 +91,22 @@ class PostRepository
      */
     public function update($signup, $data)
     {
-        $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'why_participated']));
+        if (array_key_exists('updated_at', $data)) {
+            $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'why_participated', 'updated_at']));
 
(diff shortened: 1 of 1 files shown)


chloealee:
@katiecrane I feel like I'm missing something totally obvious but wondering why we want to back date the event? If the signup already exists, an event should exist already for when the signup was created. Then, if the quantity or why gets updated, shouldn't this be a new event so we won't need to back date it?

katiecrane:
@chloealee everything you said is correct! the only thing is we will be sending over data from Phoenix regularly, but not the second it happens. So when we send over a `quantity` or `why_participated` update, we want to record the timestamp of the signup update event and the signup `updated_at` as the time when the update actually happened, not just when it got to Rogue. Am I making sense?

chloealee:
@katiecrane ah yes totally makes sense - thanks so much for explaining! This looks good to me :)

sbsmith86:
💯

PR #240, opened 2017-05-01
Soft delete posts

### LINKED ISSUE #237: Soft delete posts from DB
_FEATURE OVERVIEW_
-------------------
#### User Story

We're going to try out soft-deleting, meaning the image deletes from the DB, but we still track that it was deleted. This information could potentially be used down the line for like banning members and other things.

#### Why This Matters

So when someone deletes a post, it is deleted from the database but we have a record of this delete having happened.

sbsmith86:
#### What's this PR do?

* Sets up the `Post` model to use soft deleting when deleting a post. 
* Creates an endpoint at `DELETE /posts/{post_id}` that will trigger the deletion of a post. 
* Deletes images from s3 when the `destroy()` method is called for a post and nulls out the posts `url` field.
* Adds a test to make sure that endpoint soft deletes a post, and nulls out the url column after image has been removed from s3.
* updates documentation

#### Relevant tickets
Fixes #237 

#### Checklist
- [ ] Tested on staging.


app/Models/Post.php
```diff
@@ -4,10 +4,19 @@
 
 use Conner\Tagging\Taggable;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\SoftDeletes;
 
 class Post extends Model
 {
     use Taggable;
+    use SoftDeletes;
```
DFurnes: Totally subjective, but you can chain these together if you want: `use Taggable, SoftDeletes;`

app/Repositories/PostRepository.php
```diff
@@ -116,7 +116,32 @@ public function update($signup, $data)
...
+        $this->aws->deleteImage($post->url);
+
+        // Set the url of the post to null.
+        $post->url = null;
+        $post->save();
+
+        // Soft delete the post.
+        $post->delete();
+
+        return $post->trashed() ? true : false;
```
DFurnes: This ternary can be removed since `$post->trashed()` already returns a boolean!

app/Services/AWS.php
```diff
@@ -74,4 +75,31 @@ protected function base64StringToDataString($string)
...
+     * @return bool
+     */
+    public function deleteImage($path)
+    {
+        $bucketBaseUrl = config('filesystems.disks.s3.public_url') . '/' . config('filesystems.disks.s3.bucket');
+
+        // We need to use the relative url for the request to s3.
+        $path = str_replace($bucketBaseUrl, '', $path);
+
+        // The delete() method always returns true because it doesn't seem to do anything with
```
DFurnes: I still really want us to send a patch for this... 🐛 😄

tests/CampiagnsTest.php
```diff
@@ -0,0 +1,22 @@
...
+
+use Rogue\Services\CampaignService;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Foundation\Testing\WithoutMiddleware;
+
+class CampiagnsTest extends TestCase
+{
+    /**
+     * A basic test example.
+     *s
```
DFurnes: 👾

sbsmith86: this shouldn't even be here haha something from a different branch i accidentally added 😬

PR #245, opened 2017-05-05
pulls accepted posts instead of approved

chloealee:
#### What's this PR do?
Changes query to pull `accepted` posts instead of `approved` in the `GET /reportbacks` endpoint and updates documentation.

👀 

Needed for Rogue to power gallery work

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/ReportbackController.php
+++ b/app/Http/Controllers/Api/ReportbackController.php
@@ -43,7 +43,7 @@ class ReportbackController extends ApiController
         // 2. Only return approved Posts
         // 3. Select all the fields that we will be using
         $query = $query->join('signups', 'signups.id', '=', 'posts.signup_id')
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -43,7 +43,7 @@ public function index(Request $request)
         // 2. Only return approved Posts
         // 3. Select all the fields that we will be using
         $query = $query->join('signups', 'signups.id', '=', 'posts.signup_id')
-            ->where('status', '=', 'approved')
+            ->where('status', '=', 'accepted')
```
DFurnes: Just a thought for the future – it might be nice to make a set of constants that we could import elsewhere when needed so we don't accidentally make typos (especially for ones like this where it's so easy to do!)

Issue #242, opened 2017-05-02
When a post is reviewed, send status to database. 


chloealee:
_FEATURE OVERVIEW_
-------------------
#### User Story
As an admin, I want to be able to apply a status to a post and that status is logged in the database so that posts are reviewed.

#### Why This Matters
So we can keep track of the reviewed posts!

Issue #236, opened 2017-04-28
Add Authorization API middleware for requests to Rogue


sbsmith86:
_FEATURE OVERVIEW_
-------------------
#### User Story

Add key authorization for `POST` and `DELETE` requests to Rogue. Currently, you can send an API request to rogue without any authorization needed. 

Create an authentication middleware that will determine the request method, either `POST` or `DELETE` and check that there is a `X-DS-Rogue-API-Key` header present that matches some matching key we generate and put in the `.env` like `ROGUE_API_KEY`.

Only let the request continue if the two keys match, if not throw an error.

#### Additional Information (optional)

This is how we handle authorizing requests in Gladiator, but I am open to other ideas. @DFurnes maybe you know some other/better methods? Or does this sound ok to you?

DFurnes:
Yeah, this seems good! In the future, it'd be lovely to use Northstar's access tokens here, but it still requires a little bit of extra work. I'm hoping to revisit DoSomething/northstar#535 sometime soon though, so hopefully not a long wait on that. 😄

ngjo:
![](https://github.trello.services/images/mini-trello-icon.png) [(3) https://github.com/DoSomething/rogue/issues/236](https://trello.com/c/i9RcUtiq/397-3-https-github-com-dosomething-rogue-issues-236)

PR #248, opened 2017-05-08
Updates /reviews endpoint out of API and into Web Middleware (& updates tests)

chloealee:
#### What's this PR do?
- Updates `/reviews` endpoint out of API and into Web Middleware
- Updates tests around authorized users and makings sure only correct roles can access certain endpoints

Make sure all tests are still passing and test logic makes sense. 

Preliminary work in order to be able to update a reviewed post and send data from front end to database. 

#### Relevant tickets
Refs #242 

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/ReviewsController.php
```diff
@@ -51,13 +51,14 @@ public function reviews(ReviewsRequest $request)
...
         $reviewedPost = $this->post->reviews($review);
         $reviewedPostCode = $this->code($reviewedPost);
-
         $meta = [];
 
         if (isset($reviewedPost)) {
-            return $this->item($reviewedPost, $reviewedPostCode, $meta, $this->postTransformer);
+            return $this->item($reviewedPost, $reviewedPostCode);
         } else {
             return 404;
```
DFurnes: Let's update this to throw a `ModelNotFoundException` since I don't think this actually sets a status code.

chloealee: Ah great! I just added this but wanted to get your eyes on it one more time to make sure this is correct! All tests still passing :)

storage/debugbar/007124838e96ef988c8469481a41ea24.json
```diff
@@ -0,0 +1 @@
... \"SCRIPT_NAME\" => \"\/index.php\"\n  \"REQUEST_URI\" => \"\/api\/v1\/reportbacks?filter%5Bcampaign_id%5D=1283&as_user=5589c991a59dbfa93d8b45ae&limit=6\"\n  \"DOCUMENT_URI\" => \"\/index.php\"\n  \"DOCUMENT_ROOT\" => \"\/home\/vagrant\/DoSomething.org\/rogue\/public\"\n  \"SERVER_PROTOCOL\" => \"HTTP\/1.1\"\n  \"GATEWAY_INTERFACE\" => \"CGI\/1.1\"\n  \"SERVER_SOFTWARE\" => \"nginx\/1.8.0\"\n  \"REMOTE_ADDR\" => \"10.0.2.2\"\n  \"REMOTE_PORT\" => \"60452\"\n  \"SERVER_ADDR\" => \"10.0.2.15\"\n  \"SERVER_PORT\" => \"80\"\n  \"SERVER_NAME\" => \"rogue.dev\"\n  \"REDIRECT_STATUS\" => \"200\"\n  \"HTTP_X_DS_ROGUE_API_KEY\" => \"******\"\n  \"HTTP_USER_AGENT\" => \"GuzzleHttp\/6.2.1 curl\/7.35.0 PHP\/7.0.6-9+donate.sury.org~trusty+2\"\n  \"HTTP_HOST\" => \"rogue.dev:8000\"\n  \"HTTP_X_REQUEST_ID\" => \"1494355063.9736-0\"\n  \"PHP_SELF\" => \"\/index.php\"\n  \"REQUEST_TIME_FLOAT\" => 1494355064.0204\n  \"REQUEST_TIME\" => 1494355064\n]","request_cookies":"[]","response_headers":"array:2 [\n  \"cache-control\" => array:1 [\n    0 => \"no-cache\"\n  ]\n  \"content-type\" => array:1 [\n    0 => \"application\/json\"\n  ]\n]","path_info":"\/api\/v1\/reportbacks","session_attributes":"[]"}}
```
sbsmith86: whoooaaaa i thought this debugbar things were ignored. I will double check that

chloealee: ah thanks Shae! I'm working on not including those in this PR now :)

app/Http/Controllers/Api/ApiController.php
```diff
@@ -2,126 +2,10 @@
...
-use League\Fractal\Pagination\IlluminatePaginatorAdapter;
 
 class ApiController extends BaseController
 {
     use FiltersRequests;
-
-    /**
-     * @var League\Fractal\Manager
-     */
-    protected $manager;
```
sbsmith86: Is removing all of this going to break any of our other api routes that use this controller?

storage/debugbar/007124838e96ef988c8469481a41ea24.json
sbsmith86: @chloealee i think maybe we can just add it to the root `.gitignore` file?

sbsmith86:
@chloealee @DFurnes I still don't really understand why we need to do this. Could you run it by me one more time? I think my thought was that if we implemented  #236 the requests from the frontend of rogue could just use the auth header to confirm they can make the request. And only admins are able to log into the site.

storage/debugbar/007124838e96ef988c8469481a41ea24.json
chloealee: oh right! I thought for some reason it got added as extra commits. Do I just put `/storage/debugar` in the `.gitignore` file?

sbsmith86: yar i think that should do it.

app/Http/Controllers/Api/ApiController.php
chloealee: @sbsmith86 this shouldn't since we are now including that all in a trait that the `Controller` uses and this `ApiController` extends the `Controller`

DFurnes:
@sbsmith86 We don't want the front-end to be aware of the `X-DS-Rogue-API-Key` secret because that password is sorta the "keys to the kingdom". By letting the user's session cookie "prove" that they're a particular logged in administrator, we make sure that key stays a secret.

sbsmith86:
@DFurnes gotcha! thanks for the explanation!

PR #253, opened 2017-05-10
Add the history modal to inbox

### LINKED ISSUE #252: Add link to signup history modal and also build the modal
_FEATURE OVERVIEW_
-------------------
#### User Story
Sometimes admins want to edit the quantity.

Looks like this:
image
image

#### Why This Matters
So admins can see signup history and edit the quantity.

Nothing within reason.

katiecrane:
#### What's this PR do?
2 new components:
- `ModalContainer` - This has all the modal styles! We can wrap future modals in this so we don't have to worry about styling them.
- `HistoryModal` - This has the specific content for the quantity edit/history modal.

Also, there is added logic to decide when to open and close each modal. 

Link to the modal:
image

The modal itself:
image

Is everything okay visually? What about how the code is structured?

#### Relevant tickets
Fixes #252

#### Checklist
- [ ] Tested on staging.


resources/assets/components/HistoryModal/index.js
```diff
@@ -0,0 +1,35 @@
...
+					<h3>Reportback History</h3>
+					<p>table of all the history goes here 📖</p>
+				</div>
+				<a className="button -history">Save</a>
+			</div>
+		);
+	}
+}
+
+export default HistoryModal;
```
sbsmith86: newline 😬

resources/assets/components/ModalContainer/index.js
```diff
@@ -0,0 +1,18 @@
...
+
+	return (
+		<div className="modal-container">
+			{props.children}
+		</div>
+	);
+	
+};
+
+export default ModalContainer;
```
sbsmith86: newline 😬

resources/assets/components/ModalContainer/modal-container.scss
```diff
@@ -0,0 +1,104 @@
...
+  min-height: 100%;
+  z-index: 9999;
+  background: rgba(#000, 0.5);
+
+  @media #{$tablet} {
+    padding: 72px;
+  }
+
+  // Animations (such as `fadeOut` and `moveDown`) are defined in Neue
+  // @TODO: Is there a reasonable way to share these?
```
sbsmith86: Could we add it to the `helpers.scss` file?

resources/assets/components/ModalContainer/modal-container.scss
```diff
@@ -0,0 +1,104 @@
...
+    }
+  }
+}
+
+.button {
+  &.-history {
+    width: 100%;
+    border-radius: 0px;
+  }
+}
```
sbsmith86: another one!

resources/assets/components/HistoryModal/index.js
katiecrane: @sbsmith86 these things are one of my greatest weaknesses! thank you!!!

resources/assets/components/HistoryModal/index.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+// import history modal
```
sbsmith86: is this needed?

resources/assets/components/ModalContainer/modal-container.scss
katiecrane: @sbsmith86 this is actually copied over from DS/Modals and we aren't using this part right now anyway. I wasn't sureee if we wanted to include it anyway though.

resources/assets/components/HistoryModal/index.js
```diff
@@ -0,0 +1,35 @@
...
+						<h4>New Quantity</h4>
+						<div className="form-item">
+							<input type="text" className="text-field" placeholder="Enter # here"></input>
+						</div>
+					</div>
+
+					<h3>Reportback History</h3>
+					<p>table of all the history goes here 📖</p>
+				</div>
+				<a className="button -history">Save</a>
```
sbsmith86: i wonder if this should be disabled by default until a user changes the value in the `quantity` input. Not a merge blocker. Just a thought I was having!

resources/assets/components/InboxItem/index.js
```diff
@@ -64,6 +64,7 @@ class InboxItem extends React.Component {
               {campaign['reportback_info']['noun']} {campaign['reportback_info']['verb']}
             </div>
           </article>
+          <a href="#" onClick={e => this.props.showHistory(post['id'], e)}>Edit | Show History</a>
```
sbsmith86: should these be wrapped in there own `</a>` tags. I feel like the way it is formatted makes it look like they are two distinct links. I forgot if `edit` goes someplace different than `show history`, but if they are in fact the same, maybe we should do "edit/show history" ?

resources/assets/components/ModalContainer/modal-container.scss
sbsmith86: ahh right right

resources/assets/components/HistoryModal/index.js
katiecrane: ooooo I think that's a good idea. I'll add that to the list of stuff to do for connecting it to the backend

resources/assets/components/InboxItem/index.js
katiecrane: great question! This is the way that it's formatted in the mock-up and the whole thing appears to be just one link. They definitely go to the same place since editing the quantity and showing the history are both in this modal. this might be one for @lkpttn ?

katiecrane: @sbsmith86 do you think this part is okay for now? Once it's merged I can work on linking it to the backend and this seems like it would be quick to change later

sbsmith86: sure!

PR #256, opened 2017-05-12
Let admins update quantity

katiecrane:
#### What's this PR do?
This is a magical time. When an admin opens the history modal, they can update the quantity and hit "Save" to see the quantity update on the post (and all other posts under that same signup). Gif ⬇️ 
update_quantity

Questions:
1. @sbsmith86 mentioned that we should disable the save button until the admin has entered something into the new quantity form. I think this is a good idea, but I'm not sure if that's out of scope for this or not?
2. We should probably have some validation on the form so admins can't enter "platypus" for quantity because currently they are able to. Is that in scope?
image
3. I believe in grooming on Wednesday we will talk about how to deal with logging events of admins updating quantity and sending the admin's Northstar ID to `/posts`.
4. On the card it says to write tests for this, but as far as I can tell those would be frontend tests which I don't think we have any of yet. Did we want to start that now?

Make sure I didn't mess up any javascript/react conventions and that nothing is hacky 😬 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/CampaignInbox/index.js
+++ b/resources/assets/components/CampaignInbox/index.js
@@ -1,6 +1,5 @@
 import React from 'react';
-import { flatMap } from 'lodash';
-import { keyBy, map, sample } from 'lodash';
+import { flatMap, keyBy, map, sample, forEach } from 'lodash';
 import { RestApiClient} from '@dosomething/gateway';
 
 import InboxItem from '../InboxItem';
@@ -42,9 +41,7 @@ class CampaignInbox extends React.Component {
   }
 
   // Close the open history modal
-  hideHistory(event) {
-    event.preventDefault()
-
+  hideHistory() {
     this.setState({
       displayHistoryModal: false,
       historyModalId: null,
@@ -69,11 +66,6 @@ class CampaignInbox extends React.Component {
   }
 
   updateQuantity(post, newQuantity) {
-    // @TODO: remove these before merging
-    console.log('updating quantity');
-    console.log(post);
-    console.log(newQuantity);
-
     // Fields to send to /posts
     const fields = {
       northstar_id: post.user.id,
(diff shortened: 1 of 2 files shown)


resources/assets/components/CampaignInbox/index.js
```diff
@@ -65,6 +67,40 @@ class CampaignInbox extends React.Component {
...
+    let response = this.api.post('api/v2/posts', fields);
+
+    response.then((result) => {
+      // Update the state
+      this.setState((previousState) => {
+        const newState = {...previousState};
+        const signupChanged = post.signup_id;
+
+        // Update the quantity for each post under this signup
+        forEach (newState.posts, (value) => {
```
sbsmith86: why do you have to do this for each post, if quantity lives on the signup?

resources/assets/components/CampaignInbox/index.js
```diff
@@ -65,6 +67,40 @@ class CampaignInbox extends React.Component {
...
+          }
+        });
+
+        // Return the new state
+        return newState;
+      });
+    });
+
+    // Close the modal
+    this.hideHistory(event);
```
sbsmith86: your passing in the `event` here but never passing it into the function.

sbsmith86:
> 1. @sbsmith86 mentioned that we should disable the save button until the admin has entered something into the new quantity form. I think this is a good idea, but I'm not sure if that's out of scope for this or not?

I sorta feel like it is in scope for this work since you are doing the work of submitting the form. But up to you if you feel like it is it's own chunk!

> 2. We should probably have some validation on the form so admins can't enter "platypus" for quantity because currently they are able to. Is that in scope?

I could see an argument for this being in scope for this, but validation could be it's own beast so maybe it should be separate.

resources/assets/components/CampaignInbox/index.js
katiecrane: @sbsmith86 we end up mapping things out to be all of the posts that hold the data for their signup since we need to be able to map through the posts to pass them to the `InboxItem` component. The way this works is pretty confusing to me so I'm definitely down to discuss it

resources/assets/components/CampaignInbox/index.js
katiecrane: @sbsmith86 sorry, never passing it to which function? with this it ended up having the desired behavior of not scrolling back to the top of the page, would you expect this to not work like this?

resources/assets/components/CampaignInbox/index.js
sbsmith86: hmm interesting, yea...i feel like signup data should live in the state and posts could use the signup id associated with it to update the quantity of that signup.

resources/assets/components/CampaignInbox/index.js
DFurnes: Yeah, I'm not sure `event` is defined here (within `updateQuantity`).

katiecrane: oooo okay yeah. I removed `event` here and added a check in `hideHistory` to only use it if it's set

katiecrane:
@sbsmith86 I updated to set the state based on the API response. Currently even if you submit a word or emoji, that is what you get back in the API response and the quantity in the database gets set to 0. It seems like something with the API response in that case is incorrect.

sbsmith86:
@katiecrane Hrmmm. Can you make an issue for that as a bug?

katiecrane:
@sbsmith86 yup! I'm verifying it in postman right now. Since there's a separate card now for the button and validation do you think this is good to go?

resources/assets/components/CampaignInbox/index.js
sbsmith86: I think this should be addressed. I'm concerned about the performance hits when we have larger amounts of Posts. Is this being tracked in another card?

katiecrane: @sbsmith86 I don't think so but I think it would be a good idea to make one to clean up how this data exists on the inbox. I can make an issue for that too

sbsmith86: @katiecrane cool yea lets get that documented so we don't forget.

PR #260, opened 2017-05-17
Tags endpoint

### LINKED ISSUE #255: Create /tags endpoint and tests
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want to create the tags endpoint + tests to go along with it so that we have an endpoint for when we connect the two tag systems.

#### Why This Matters
We need that endpoint to do other tags work!

#### Definition of Done
Tags endpoint created and tested.

chloealee:
#### What's this PR do?
- Creates the `/tags` endpoint.
  - When a tag is created or deleted, information gets stored in the `tagging_tagged` table. 
  - It is also associated to the post (grab a post's tag with `$post->tagNames();`
  - An event is created when a tag is added to or deleted from a post. 
- Creates testing for this endpoint.
- Updates documentation. 

Run all tests and make sure everything is passing! 

#### Relevant tickets
Fixes #255 

#### Checklist
- [ ] Tested on staging.


app/Models/Post.php
```diff
@@ -55,4 +57,12 @@ public function reactions()
     {
         return $this->hasMany(Reaction::class);
     }
+
+    /**
+     * Get all fo the tags for the post.
+     */
+    public function tags()
```
sbsmith86: I don't think we need this? because the package comes with functionality to grab the tags associated with a post already with `tagNames()`

chloealee: oh right I forgot to take that out, thanks Shae!

app/Repositories/PostRepository.php
```diff
@@ -170,6 +170,54 @@ public function reviews($data)
...
     /**
+     * Updates a post's tags when added or deleted.
+     *
+     * @param array $data
+     *
+     * @return
+     */
+    public function tag($data)
+    {
+        $post = Post::where(['id' => $data['post_id']])->first();
```
sbsmith86: I think we can just do `Post::findOrFail($data['post_id'])` here.

app/Repositories/PostRepository.php
```diff
@@ -170,6 +170,54 @@ public function reviews($data)
...
+        if (in_array($data['tag_name'], $post->tagNames())) {
+            $post->untag($data['tag_name']);
+
+            // Create an event when the tag is deleted.
+            if (!in_array($data['tag_name'], $post->tagNames())) {
+                $post->tag_name = $data['tag_name'];
+                $post->admin_northstar_id = $data['admin_northstar_id'];
+                $post->action = 'delete';
+
+                $event = Event::create([
```
sbsmith86: Could this similar thing happen in the ModelServiceProvider?

database/migrations/2017_05_15_184344_add_soft_delete_to_tagging_tagged_table.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class AddSoftDeleteToTaggingTaggedTable extends Migration
```
sbsmith86: Why do we need this? the `tag()` and `unTag()` should handle the addition and deletion of tags.

app/Repositories/PostRepository.php
chloealee: oh yes! I tried yesterday and it didn't work but tried again and it's all good :)

chloealee: actually I don't think we can do this because we won't have access to the specific post (that we need info. from to add to `eventable_id` and `content` in the event unless I'm missing something on how to grab the post from the [documentation](https://github.com/rtconner/laravel-tagging/blob/laravel-5/docs/usage-examples.md)?

sbsmith86: In the `ModelServiceProvider`, I was thinking that you could listen to some event (`create`, `save` ?)on the `Tagged` model and log that in the events table.  That tagged model should have the `taggable_id` which could be used to pull in the post?

database/migrations/2017_05_15_184344_add_soft_delete_to_tagging_tagged_table.php
chloealee: ah ok! I'll get rid of that. For some reason, I read [this](https://dosomething.slack.com/?redir=%2Farchives%2FC0YGXUE01%2Fp1494864214528044) as adding soft deletes, I'm not sure why!

app/Repositories/PostRepository.php
chloealee: ah I see! The only two thing we'd be missing then if we did it in the ModelServiceProvider is the `admin_northstar_id` and the `action` that I manually added in the PostRepository. I can't think of a way to get this to save via this route? @sbsmith86

sbsmith86: @chloealee So I was thinking something like this. I just tried this out on my local and it seems to work but let me know if something doesn't look right here. 

```

        Tagged::saved(function($tagged) {
            $post = Post::find($tagged->taggable_id);
            $adminId = auth()->user()->northstar_id;

            $post->action = [
                'type' => 'tagged',
                'admin' => $adminId,
            ];

            Event::create([
                'eventable_id' => $post->id,
                'eventable_type' => 'Conner\Tagging\Model\Tagged',
                'content' => $post->toJson(),
            ]);
        });
```

chloealee: OHHH this is brilliant, thanks @sbsmith86 ! Totally didn't even think to use `auth()->user()->northstar_id;` here and was getting hung up thinking that I'd have to pass this to the ModelServiceProvider somehow. I gotta think outside of the box more! Thank you for this - it works great!

app/Providers/ModelServiceProvider.php
```diff
@@ -37,6 +39,22 @@ public function boot()
...
+            $adminId = auth()->user()->northstar_id;
+
+            $post->action = [
+                'type' => 'tagged',
+                'admin' => $adminId,
+            ];
+
+            Event::create([
+                'eventable_id' => $post->id,
+                'eventable_type' => 'Conner\Tagging\Model\Tagged',
```
sbsmith86: Sorry I wonder if this `eventable_type` should be `Rogue\Models\Post` because we are logging it as an event on the post even tho it's the `Tagged` model that triggers it.

chloealee: that makes sense to me! I will update that too :)

app/Providers/ModelServiceProvider.php
```diff
@@ -37,6 +39,22 @@ public function boot()
...
+            $adminId = auth()->user()->northstar_id;
+
+            $post->action = [
+                'type' => 'tagged',
+                'admin' => $adminId,
+            ];
+
+            Event::create([
+                'eventable_id' => $post->id,
+                'eventable_type' => $post->getModel(),
```
chloealee: @sbsmith86 changed it to this since it seems a little safer to do but if it's better to hardcode this, let me know!

sbsmith86: Oh cool. I like that!

app/Http/Controllers/TagsController.php
```diff
@@ -0,0 +1,51 @@
...
+     * Add or soft delete a tag to a post when reviewed.
+     *
+     * @param Rogue\Http\Requests\TagsRequest $request
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function store(TagsRequest $request)
+    {
+        $tagData = $request->all();
+        $tagData['admin_northstar_id'] = auth()->user()->northstar_id;
```
sbsmith86: Think you can remove this now!

chloealee: ah yes, sorry! :man_facepalming:

app/Http/Requests/TagsRequest.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            'post_id' => 'required',
+            'tag_name' => 'required',
```
sbsmith86: Should we also require this to be a `string`?

app/Models/Post.php
```diff
@@ -10,6 +10,8 @@ class Post extends Model
 {
     use Taggable, SoftDeletes;
 
+    protected $table = 'posts';
```
sbsmith86: hmm why do we need to set this?

app/Http/Requests/TagsRequest.php
chloealee: ah yes! Should I require the `post_id` to be an `int` while I'm at it?

app/Models/Post.php
chloealee: oh I was just following [this](https://laravel-news.com/how-to-add-tagging-to-your-laravel-app) but you're right, it is not needed :)

app/Http/Requests/TagsRequest.php
sbsmith86: Ya!

app/Http/Controllers/TagsController.php
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function store(TagsRequest $request)
+    {
+        $tagData = $request->all();
+        $taggedPost = $this->post->tag($tagData);
+
+        if (! isset($taggedPost)) {
+            throw (new ModelNotFoundException)->setModel('Post');
```
sbsmith86: So the `findOrFail` method you use to find the post will throw a `ModelNotFoundException` before we get here. Maybe we should just return `true` (or`false`) if something goes wrong from the `tags` method and we can deal with throwing a custom error response?

chloealee: that makes sense! Should we throw either a `200` or `500` (with text that says the tag wasn't successfully created/deleted) status code?

sbsmith86: ya that sounds good to me.

app/Http/Controllers/TagsController.php
```diff
@@ -0,0 +1,49 @@
...
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function store(TagsRequest $request)
+    {
+        $tagData = $request->all();
+        $taggedPost = $this->post->tag($tagData);
+
+        if (! $taggedPost) {
+            abort(500, 'Tag was not successfully created/deleted.');
```
chloealee: @sbsmith86 is it ok to use `abort()` here?

sbsmith86: @chloealee looks like you can do `response()->json(['error' => Tag was not successfully created/deleted.], 500);`

chloealee: oh ok! Updated :)

PR #263, opened 2017-05-18
Hook up the delete button to the backend

### LINKED ISSUE #262: Hook up the delete button to /posts DELETE endpoint
_FEATURE OVERVIEW_
-------------------
#### User Story
When an admin clicks "delete" the post should be soft-deleted from the database and removed from the page.

Documentation for the endpoint to use is [here](https://github.com/DoSomething/rogue/blob/master/documentation/endpoints/posts.md#posts) (scroll down). 

#### Why This Matters
In case an inappropriate photo is submitted, we need admins to be able to remove it themselves.

They would have to manually hit the endpoint, fun!

katiecrane:
#### What's this PR do?
Hooked up the "delete" button to the backend so it makes a request to delete the post. Works the same way as opening the history modal, except instead of `showHistory` it calls `deletePost` when the button is clicked.
deletepost

Click the delete button and see the post leave the page. If you refresh the page, the post should not reappear (meaning that it is really soft deleted in the db).

#### Relevant tickets
Fixes #262

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/CampaignInbox/index.js
+++ b/resources/assets/components/CampaignInbox/index.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { flatMap, keyBy, map, sample, forEach } from 'lodash';
+import { flatMap, keyBy, map, sample, forEach, reject } from 'lodash';
 import { RestApiClient} from '@dosomething/gateway';
 
 import InboxItem from '../InboxItem';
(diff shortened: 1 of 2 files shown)


resources/assets/components/CampaignInbox/index.js
```diff
@@ -103,6 +104,28 @@ class CampaignInbox extends React.Component {
     this.hideHistory();
   }
 
+  deletePost(postId, event) {
+    event.preventDefault();
+    confirm('Are you sure you want to delete this?');
+
+    console.log('bout to delete this');
```
DFurnes: 🔜 🗑

PR #266, opened 2017-05-19
Auth API middleware

### LINKED ISSUE #236: Add Authorization API middleware for requests to Rogue

sbsmith86:
#### What's this PR do?

So, apparently we already had `AuthenticatAPI` middleware but we just were not using it. This PR:

* Uses the `auth.api` middleware on all the api routes. We were using the `api` middleware but this wasn't defined and did nothing. 
* moves the posts `DELETE` endpoint to the web route group and splits out it's functionality into it's own `PostController` that authenticates the request and deletes the post
* Updates tests accordingly 

commit by commit should work!

#### Relevant tickets
Fixes #236

### DIFF
--- a/tests/PostApiTest.php
+++ b/tests/PostApiTest.php
@@ -1,12 +1,9 @@
 <?php
 
 use Rogue\Models\Post;
-use Illuminate\Foundation\Testing\WithoutMiddleware;
 
 class PostApiTest extends TestCase
 {
-    use WithoutMiddleware;
-
     /*
      * Base URL for the Api.
      */
@@ -46,7 +43,7 @@ class PostApiTest extends TestCase
         // Mock sending image to AWS.
         Storage::shouldReceive('put')->andReturn(true);
 
-        $response = $this->json('POST', $this->postsApiUrl, $post);
+        $response = $this->authed()->json('POST', $this->postsApiUrl, $post);
 
         $this->assertResponseStatus(200);
 
--- a/tests/PostApiTest.php
+++ b/tests/PostApiTest.php
@@ -60,28 +60,4 @@ class PostApiTest extends TestCase
             'quantity' => $post['quantity'],
         ]);
     }
-
-    /**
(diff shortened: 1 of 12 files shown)


chloealee:
this all looks good to me, @sbsmith86 ! I think might just need to rebase?

sbsmith86:
@chloealee whoops! didn't notice that.

tests/PostApiTest.php
```diff
@@ -5,6 +5,8 @@
 
 class PostApiTest extends TestCase
 {
+    use WithoutMiddleware;
```
DFurnes: Not necessary for merge, but I'd recommend making a helper to auth individual requests (like a `$this->authed()->json(...)` rather than disabling middleware across the board, since that allows the test environment to diverge a lot from how the real application works.

DFurnes: We made a [`withHeader` helper function](https://github.com/DoSomething/northstar/blob/49630b5f11f4f5c8d0a6ad4a63097f62f8b48548/tests/TestCase.php#L283-L296) for that in Northstar if you wanna steal the approach. 😄

sbsmith86: @DFurnes awesome! i made that change. Can you take another look?

DFurnes: Looks great!!

PR #269, opened 2017-05-23
/reviews to be used by keeper upper

### LINKED ISSUE #268: Need to be able to update status via the API
_FEATURE OVERVIEW_
-------------------
#### User Story
When we run the keeper upper script, we also want to be able to send Rogue any status updates. Currently statuses can only be altered via the web interface.

We need a new endpoint to be able to update status on the post and create a new review record.

#### Why This Matters
For a seamless transition to Rogue!

Re-review everything in Rogue that was reviewed in Phoenix between turning the keeper up on and turning Rogue on. 🙅

katiecrane:
#### What's this PR do?
1. Added a new route for this endpoint
2. Added a new controller for this endpoint (it is almost identical to [the web Reviews Controller](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/ReviewsController.php))

Are there any parts here that are specific to the web controller and should not be in the API one? I've tested it and as far as I can tell everything is working properly.

In order to keep all data on posts up to date via the keeper upper, we also need to be able to update their statuses as they are reviewed.

#### Relevant tickets
Fixes #268

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/ReviewsController.php
+++ b/app/Http/Controllers/Api/ReviewsController.php
@@ -1,0 +1,80 @@
+<?php
+
+namespace Rogue\Http\Controllers\Api;
+
+use Rogue\Models\Post;
+use Rogue\Repositories\PostRepository;
+use Rogue\Http\Requests\ReviewsRequest;
+use Rogue\Http\Transformers\PostTransformer;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+
+class ReviewsController extends ApiController
+{
+    /**
+     * The post service instance.
+     *
+     * @var Rogue\Repositories\PostRepository
+     */
+    protected $post;
+
+    /**
+     * @var \Rogue\Http\Transformers\PostTransformer
+     */
+    protected $transformer;
+
+    /**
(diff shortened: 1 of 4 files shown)


chloealee:
This looks great, @katiecrane ! One thing is to perhaps add some validation rules for what is absolutely needed from the request? I think we'll need `post_id`, `status`, and `admin_northstar_id` (only thing different is that we'll need to send the `admin_northstar_id` instead of grabbing this from the logged in user like we do from the web endpoint.

chloealee:
Oh! One more thing - do we want to add tests for this endpoint?

katiecrane:
@chloealee ahh validation! I knew I was forgetting something! I'm not suuuure about tests though, I feel like we will be removing this endpoint after the keeper upper is done so that reviews aren't accessible via the api? I didn't think we wanted anyone to be able to update reviews that way but I might be wrong

chloealee:
@katiecrane okay I'm fine to leave out tests!

katiecrane:
@chloealee validation is added!

PR #275, opened 2017-05-26
Adds updated_at param to /activity endpoint

### LINKED ISSUE #274: Add updated_at filter to /activity endpoint
_FEATURE OVERVIEW_
-------------------
#### User Story
As a data engineer, I want to be able to filter activity based on a range of the `updated_at` timestamp.

#### Why This Matters
So that data can get everything from Rogue - including updates!

#### Definition of Done
Parameter added and data team is able to get last updated posts

chloealee:
#### What's this PR do?
- Adds the `updated_at` param to the `/activity` endpoint so you can now filter results to only return signup's whose `updated_at` time is after the value of the `updated_at` param. 
- Updates documentation.
- Adds test. 

- Make sure all tests are passing! 
- You can also manually test this endpoint by itself and with other query params and expected results should be returned. 

#### Relevant tickets
Fixes #274 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Traits/FiltersRequests.php
+++ b/app/Http/Controllers/Traits/FiltersRequests.php
@@ -33,9 +33,12 @@ trait FiltersRequests
         $filters = array_intersect_key($filters, array_flip($indexes));
 
         // If there is an updated_at param, remove it from $filters and save the value.
-        if (!(array_search('updated_at', $filters))) {
+        if (array_key_exists('updated_at', $filters)) {
             $updatedAtValue = $filters['updated_at'];
             unset($filters['updated_at']);
+        }
+        else {
+            $updatedAtValue = false;
         }
 
         // You can filter by multiple values, e.g. `filter[source]=agg,cgg`
--- a/app/Http/Controllers/Traits/FiltersRequests.php
+++ b/app/Http/Controllers/Traits/FiltersRequests.php
@@ -36,8 +36,7 @@ trait FiltersRequests
         if (array_key_exists('updated_at', $filters)) {
             $updatedAtValue = $filters['updated_at'];
             unset($filters['updated_at']);
-        }
-        else {
+        } else {
             $updatedAtValue = false;
         }
 
--- a/app/Http/Controllers/Traits/FiltersRequests.php
+++ b/app/Http/Controllers/Traits/FiltersRequests.php
@@ -29,9 +29,14 @@ trait FiltersRequests
         if (! $filters) {
             return $query;
         }
-
         // Requests may be filtered by indexed fields.
         $filters = array_intersect_key($filters, array_flip($indexes));
+
+        // If there is an updated_at param, remove it from $filters and save the value.
+        if (!(array_search('updated_at', $filters))) {
+            $updatedAtValue = $filters['updated_at'];
+            unset($filters['updated_at']);
+        }
 
         // You can filter by multiple values, e.g. `filter[source]=agg,cgg`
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Traits/FiltersRequests.php
```diff
@@ -29,10 +29,18 @@ public function filter($query, $filters, $indexes)
...
-
         // Requests may be filtered by indexed fields.
         $filters = array_intersect_key($filters, array_flip($indexes));
 
+        // If there is an updated_at param, remove it from $filters and save the value.
+        if (array_key_exists('updated_at', $filters)) {
+            $updatedAtValue = $filters['updated_at'];
+            unset($filters['updated_at']);
+        }
+        else {
```
katiecrane: Style ci is complaining here I think

app/Models/Signup.php
```diff
@@ -22,7 +22,7 @@ class Signup extends Model
      * @var array
      */
     public static $indexes = [
-        'campaign_id', 'campaign_run_id',
+        'campaign_id', 'campaign_run_id', 'updated_at',
```
katiecrane: In order to make this an actual index on the DB it has to be setup in a migration file ([here's an example](https://github.com/DoSomething/rogue/blob/master/database/migrations/2017_03_09_192255_add_signup_id_index_to_events_table.php)) 😄 :pancakes:

katiecrane:
Looks great! left a couple of comments for Tuesday 🌞

app/Http/Controllers/Traits/FiltersRequests.php
chloealee: oh shoot sorry I missed that!

app/Models/Signup.php
chloealee: @katiecrane I'm not sure if I'm understanding correctly but this is already in the `signups` table! I just had to add it here so it would be able to be updated when anything else (post, reaction, review) is updated. LMK if that's not what you meant tho!

katiecrane: @chloealee ahh I meant that it had to be made an index in a migration file, not added to the table. the example I linked is a migration creating a new index on an existing column

chloealee: ooohh i see thanks @katiecrane ! And is this just to improve performance if you make it an index?

katiecrane: @chloealee yeah stuff that we query by we want to be an index so it's speedier. I'm not sure that _everything_ that we query by needs to be an index, but it seems safe to me to do that for columns that `/activity` uses. @DFurnes might have a different opinion or more to say though!

chloealee: @katiecrane that makes good sense to me - I just didn't know that before so thanks for letting me know! I added the migration :)

documentation/endpoints/activity.md
```diff
@@ -21,6 +21,9 @@ GET /api/v2/activity
 - **user** _(integer)_
   - Whether or not to include information about the user with the response.
   - e.g. `/activity?include=user`
+- **updated_at** _(timestamp)_
+  - Return records that have been updated at after the updated_at param value. 
+  - e.g. `/activity?filter[updated_at]=2017-05-25 20:14:48`
```
katiecrane: @chloealee just noticed this has a space in it, it works like that?

chloealee: @katiecrane it does! 
screen shot 2017-05-31 at 10 22 41 am

katiecrane: alright that works for me then!

Issue #276, opened 2017-05-26
Validation errors return 500 instead of 422


katiecrane:
_BUG_
-------------------
#### Current Behavior
I've tested this locally on the posts and signups endpoint. Validation errors give a `500`.

#### Desired Behavior
Validation errors should give a `422`.

#### Steps to Replicate
Hit /posts or /signups and put a string instead of an integer for the `campaign_id`, or the incorrect type for any other validated field.

#### Why This Matters
We want a `422` because it was a validation error and that is how those are signified, and we would also like a message about what went wrong like `campaign_id must be an integer`.

katiecrane:
I've done some digging on this and found that it does hit the code in Laravel to return a validation error, but somehow doesn't end up returning that.

ngjo:
![](https://github.trello.services/images/mini-trello-icon.png) [Validation error https://github.com/DoSomething/rogue/issues/276](https://trello.com/c/XnYvqbRQ/467-validation-error-https%3A%2F%2Fgithub.com%2Fdosomething%2Frogue%2Fissues%2F276)

PR #284, opened 2017-05-30
Add validation to /signups

### LINKED ISSUE #178: /signup validation
_FEATURE OVERVIEW_
-------------------
#### User Story

When requests are sent to the `/signup` endpoint we should perform some validation on the request to make sure required fields are present and correctly formatted before trying to store the signup

katiecrane:
#### What's this PR do?
Adds some validation to `/signups`! Makes the required fields enforced, and enforces the correct type on all fields.

Also, it seems that generally in Rogue when we throw a validation error it returns a `500` instead of a `422` (here included). I've done some digging on this and found that it does hit the code in Laravel to return a validation error, but somehow doesn't end up returning that (at least that is not what I am seeing in Postman). Here is the issue: https://github.com/DoSomething/rogue/issues/276

Will this let you create a signup without any of the required fields or with the wrong type for any field?

Trust no one.

#### Relevant tickets
Fixes #178

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Requests/SignupRequest.php
+++ b/app/Http/Requests/SignupRequest.php
@@ -28,7 +28,6 @@ class SignupRequest extends Request
             'quantity' => 'int',
             'why_participated' => 'string',
             'source' => 'string',
-            'remote_addr' => 'string',
         ];
     }
 }
--- a/app/Http/Requests/SignupRequest.php
+++ b/app/Http/Requests/SignupRequest.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Rogue\Http\Requests;
+
+class SignupRequest extends Request
+{
+    /**
+     * Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    public function authorize()
+    {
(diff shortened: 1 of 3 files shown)


app/Http/Requests/SignupRequest.php
```diff
@@ -0,0 +1,34 @@
...
+    public function rules()
+    {
+        return [
+            'northstar_id' => 'required|string',
+            'campaign_id' => 'required|int',
+            'campaign_run_id' => 'required|int',
+            'quantity' => 'int',
+            'why_participated' => 'string',
+            'source' => 'string',
+            'remote_addr' => 'string',
```
DFurnes: Do we want to add the `ip` validator here?

katiecrane: @DFurnes as it turns out, no, because this isn't even a column on the signup table 😬

katiecrane:
@DFurnes the latest commit is new if you wanna check it out!

PR #285, opened 2017-06-02
Disable save button when no quantity is entered

katiecrane:
#### What's this PR do?
Added `quantity` to the state of `HistoryModal` to keep an eye on if the admin has input anything in the form. Once there is something in the form, the button becomes active. If everything is removed from the form, the button becomes inactive again.

quantity

1. Go to any reportback inbox that is not at inbox zero
2. Click "Edit | See History." The "Save" button should be gray.
3. Click "Save." Nothing should happen!
4. Enter something into the quantity box. As soon as you enter it, the "Save" button should turn blue.
5. Remove what you entered from the box. The "Save" button should go back to gray.
6. Enter something into the box again and this time hit "Save". The modal should disappear.

Please note that the bug spotted by @chloealee in which the quantity does not actually update when you hit save is still at large!

#### Relevant tickets
[Trello card](https://trello.com/c/UMYf4MtA/427-(5)-rb-review-edit%2Fhistory%3A-as-a-developer%2C-i-want-to-disable-the-save-button-if-something-hasn%27t-been-entered-so-nulls-don%27t-ge)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/views/layouts/master.blade.php
+++ b/resources/views/layouts/master.blade.php
@@ -8,7 +8,7 @@
 
         <title>Rogue</title>
 
-        <link rel="icon" type="image/ico" href="http://twooter.biz/Gifs/tonguecat.png">
+        <link rel="icon" type="image/png" href="http://twooter.biz/Gifs/tonguecat.png">
         <link rel="stylesheet" href="{{ elixir('app.css', 'dist') }}">
     </head>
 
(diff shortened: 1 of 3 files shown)


resources/views/layouts/master.blade.php
```diff
@@ -8,7 +8,7 @@
 
         <title>Rogue</title>
 
-        <link rel="icon" type="image/ico" href="/favicon.ico?v1">
+        <link rel="icon" type="image/ico" href="http://twooter.biz/Gifs/tonguecat.png">
```
DFurnes: I think you would want to change `type` to `image/png` if you wanna go with this image. 👅

PR #286, opened 2017-06-06
Removes Luke's comments from status buttons 

chloealee:
#### What's this PR do?
Removes comments from status buttons. 

👀 

I know we said that we wanted to not use this method but I haven't looked at the code in awhile and changed from what I thought it was! I think this is okay since it's just the label and we want to keep `accepted` and `rejected` in the database? Or do we want conditional logic like this pseudo code 

```
if (props.type === 'accepted') {
  button label ACCEPT
} else {
  button label REJECT
}
``` 
@lkpttn @DFurnes @sbsmith86 

#### Relevant tickets
Fixes https://trello.com/c/JTrKV3Z6/468-3-%F0%9F%90%9B-the-luke-problem-%F0%9F%90%9B

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/InboxItem/index.js
+++ b/resources/assets/components/InboxItem/index.js
@@ -72,8 +72,8 @@ class InboxItem extends React.Component {
         </div>
         <div className="container__block -third">
           <ul className="form-actions -inline">
-            <li><StatusButton type="accepted" setStatus={this.setStatus}/></li>
-            <li><StatusButton type="rejected" setStatus={this.setStatus}/></li>
+            <li><StatusButton type="accepted" label="accept" setStatus={this.setStatus}/></li>
+            <li><StatusButton type="rejected" label="reject" setStatus={this.setStatus}/></li>
           </ul>
           <ul className="form-actions -inline">
             <li><button className="button -tertiary" onClick={e => this.props.deletePost(post['id'], e)}>Delete</button></li>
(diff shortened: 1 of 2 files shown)


DFurnes:
It looks like some build stuff got accidentally committed here. 📃

sbsmith86:
this is fine, but I also think we could just pass in the string we want to use here? 

https://github.com/DoSomething/rogue/blob/c09480f73c5e63965134709135f64d5d6fffccbb/resources/assets/components/InboxItem/index.js#L75-L76

Or does that not work

chloealee:
@sbsmith86 I think that would change the status in the backend/in the database though to `Accept` and `Reject`?

DFurnes:
Yeah, I think it's a good idea to separate the labels from the actual database value (like `<StatusButton type="accepted" ...>Accept</StatusButton>` , but doesn't necessarily have to happen in this PR.

sbsmith86:
we could have two props..`type` and `label` or something like that. so then you can do 

`<StatusButton type="accepted" label="accept" setStatus={this.setStatus}/>`

it shouldn't be too hard to just knock that out. But if we wanna hold off on it than 👌

chloealee:
oohh that is a great idea @sbsmith86 - thanks! Just added. Trying to figure out how to get rid of the build files now @DFurnes !

sbsmith86:
@chloealee awesome! thanks for fixing this!!!!

chloealee:
@sbsmith86 @DFurnes removed the excess files - ready for final look!

PR #290, opened 2017-06-12
Accepted posts

sbsmith86:
#### What's this PR do?
Creates a new "Single Campaign View" at `/campaigns/{id}` where admins can see and filter posts based on status. 

This required a little bit of refactoring in how we get status totals. I also had to update the `InboxItem` component to not allow reviewing. We could want reviewing on this page in the future but for now it is off. 

I also made some slight cosmetic fixes to the `InboxItem` component to make it look a little neater. 

campaignsingleview

My apologies for the size of this PR. 🤗 

It might be easiest to got commit by commit. You can also pull down the branch and make sure you can hit the single campaign page and filter out posts based on status. 

Allows admins to see approved posts.

#### Relevant tickets
Fixes https://trello.com/c/ns6aAQS0/429-8-rogue-filtering-as-a-campaign-lead-i-want-to-be-able-to-see-all-accepted-rbs-v1


app/Http/Controllers/CampaignsController.php
```diff
@@ -79,4 +79,8 @@ public function show($campaignId)
                 'campaign' => $campaign_data,
             ]);
     }
+
+    public function showCampaign($id) {
+        dd('show campaign');
```
chloealee: oops! is this meant to be here?

sbsmith86: hrrmmm I think you might be looking at an early commit?

resources/views/layouts/nav.blade.php
```diff
@@ -8,16 +8,6 @@
...
-                    <a href="#">
-                        <strong class="navigation__title">Two Fish</strong>
-                    </a>
-                </li>
-                <li>
-                    <a href="#">
-                        <strong class="navigation__title">Three Fish</strong>
-                    </a>
-                </li>
             </ul>
```
chloealee: ooo good catch!

app/Http/Controllers/CampaignsController.php
chloealee: oh you're right - sorry!

app/Services/CampaignService.php
```diff
@@ -199,33 +217,58 @@ public function getCampaignPostStatusCounts($campaigns)
...
+                ->leftJoin('posts', 'signups.id', '=', 'posts.signup_id')
+                ->select('signups.campaign_id',
+                    DB::raw('SUM(case when posts.status = "accepted" then 1 else 0 end) as accepted_count'),
+                    DB::raw('SUM(case when posts.status = "pending" then 1 else 0 end) as pending_count'),
+                    DB::raw('SUM(case when posts.status = "rejected" then 1 else 0 end) as rejected_count'))
+                ->where('campaign_id', '=', $campaign['id'])
+                ->groupBy('signups.campaign_id')
+                ->first();
+
+        return $totals ? $totals : null;
```
DFurnes: Is this ternary doing anything?

resources/assets/app.js
```diff
@@ -4,13 +4,16 @@ import './app.scss';
 import React from 'react';
 import ReactDom from 'react-dom';
 
-import CampaignOverview from './components/CampaignOverview'
-import CampaignInbox from './components/CampaignInbox'
+import CampaignOverview from './components/CampaignOverview';
+import CampaignInbox from './components/CampaignInbox';
+import CampaignSingle from './components/CampaignSingle';
+// import PostFilter from './components/PostFilter';
```
DFurnes: 👾

resources/assets/app.js
```diff
@@ -19,4 +22,10 @@ ready(() => {
   if (inboxContainer) {
     ReactDom.render(<CampaignInbox {...window.STATE} />, inboxContainer);
   }
+
+  if (singleCampaignContainer) {
+    // ReactDom.render(<StatusCounter {...window.STATE} />, document.getElementById('status-counter'));
```
DFurnes: 👾

resources/assets/components/CampaignSingle/index.js
```diff
@@ -0,0 +1,122 @@
...
+    const posts = keyBy(flatMap(props.signups, signup => {
+      return signup.posts.map(post => {
+        post.signup = signup;
+        return post;
+      });
+    }), 'id');
+
+    this.state = {
+      posts: posts,
+      filteredPosts: posts,
```
DFurnes: If you keep track of `posts` and a string `filter` value, you can calculate `filteredPosts` in the render method rather than having to store two versions of the same thing in state. [Thinking In React](https://facebook.github.io/react/docs/thinking-in-react.html) is a good read on this.

resources/assets/components/CampaignSingle/index.js
```diff
@@ -0,0 +1,122 @@
...
+      postTotals: props.post_totals,
+      displayHistoryModal: false,
+      historyModalId: null,
```
DFurnes: Could it make sense to rename this to `showHistoryForPostId` or something like that? At first glance I thought this was referring to a HTML `id` attribute.

resources/assets/components/CampaignSingle/index.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  filterPosts(status) {
+    const posts = filter(this.state.posts, {'status' : status.toLowerCase()});
+
+    this.setState({ filteredPosts: posts });
+  }
+
+  // Open the history modal of the given post
+  // @TODO - Move this into the InboxItem component so
```
DFurnes: I'm not sure InboxItem would be the best place for this, since the data that's being edited from the modal makes sense at this level for filtering and displaying it in the list, and individual inbox items would have a hard time tracking if a modal was already open.

DFurnes: It would definitely make sense to find a way to share this logic between the two pages that use it though.

resources/assets/components/CampaignSingle/index.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  // Close the open history modal
+  hideHistory(event) {
+    if (event) {
+      event.preventDefault();
+    }
+
+    this.setState({
+      displayHistoryModal: false,
```
DFurnes: I wonder if these can both be combined into `id|null` prop, and the modal is displayed as long as it's non-null.

resources/assets/components/CampaignSingle/index.js
```diff
@@ -0,0 +1,122 @@
...
+    // Fields to send to /posts
+    const fields = {
+      northstar_id: post.user.id,
+      campaign_id: post.signup.campaign_id,
+      campaign_run_id: post.signup.campaign_run_id,
+      quantity: newQuantity,
+    };
+
+    // Make API request to Rogue to update the quantity on the backend
+    let response = this.api.post('api/v2/posts', fields);
```
DFurnes: I'm a little wary of calling the promise here `response`, since that implies it's the actual HTTP response. Maybe `request` instead? Or just chain the `.then()` off the original call rather than storing it in a variable?

app/Services/CampaignService.php
sbsmith86: hmmm nope! good catch.

resources/assets/app.js
sbsmith86: omg i'm the worst.

resources/assets/components/CampaignSingle/index.js
sbsmith86: @DFurnes yea I see what you are saying. I was thinking that also, but just want to have this shared. Will be good for a future refactor. I will just update the `@TODO` for now?

resources/assets/components/CampaignSingle/index.js
sbsmith86: @DFurnes sure I see that. This is logic that was just copied over from the `CampaignInbox` because I need the modals to work on this view as well. So I don't want to introduce changes to that in this PR at the moment. And is also another good reason to get this stuff shareable between components.

resources/assets/components/CampaignSingle/index.js
sbsmith86: @DFurnes I'm not sure I understand fully. A prop on which component? And how would we push the change in the prop value so that it triggers the rendering of the modal?

sbsmith86: Also I have another card to do some work on refactoring state so I could fit this update into that work, if that is ok?

DFurnes: Oh oops, state not prop! Just noticed the `displayHistoryModal` boolean and `historyModalId` are always paired together and so wondering if it makes sense to combine them. Definitely not something that's a required change! 😄

sbsmith86: Ah gotcha!

PR #291, opened 2017-06-15
Adds check to see if signup exists before creating a new one

chloealee:
#### What's this PR do?
Adds check to see if signup exists before creating a new one. 

- Hit the `/signups` endpoint.
- Hit it again with the exact same information. 
- Check the `signups` table and make sure that there is not a duplicate signup. 

In order to keep our pattern consistent, I added a small function to the Signup Service to then forward on to the Signup Repository to check if there is an existing signup (since repository is supposed to deal with database and service is supposed to deal with business logic). I wanted to keep it consistent but if it isn't needed, LMK!). 

#### Relevant tickets
Fixes https://trello.com/c/YRabtbG4/495-2-%F0%9F%90%9B-when-rogue-is-on-and-if-you-create-a-signup-it-doesnt-check-to-see-if-it-exists-already-it-just-creates-one-every-time-so-we

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/SignupsController.php
+++ b/app/Http/Controllers/Api/SignupsController.php
@@ -50,7 +50,12 @@ class SignupsController extends ApiController
     {
         $transactionId = incrementTransactionId($request);
 
-        $signup = $this->signups->create($request->all(), $transactionId);
+        // Check to see if the signup exists before creating one.
+        $signup = $this->signups->get($request['northstar_id'], $request['campaign_id'], $request['campaign_run_id']);
+
+        if (! $signup) {
+            $signup = $this->signups->create($request->all(), $transactionId);
+        }
 
         if ($signup) {
             $code = 201;
--- a/app/Http/Controllers/Api/SignupsController.php
+++ b/app/Http/Controllers/Api/SignupsController.php
@@ -53,12 +53,10 @@ class SignupsController extends ApiController
         // Check to see if the signup exists before creating one.
         $signup = $this->signups->get($request['northstar_id'], $request['campaign_id'], $request['campaign_run_id']);
 
+        $code = $signup ? 200 : 201;
+
         if (! $signup) {
             $signup = $this->signups->create($request->all(), $transactionId);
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Api/SignupsController.php
```diff
@@ -50,7 +50,12 @@ public function store(SignupRequest $request)
...
-        $signup = $this->signups->create($request->all(), $transactionId);
+        // Check to see if the signup exists before creating one.
+        $signup = $this->signups->get($request['northstar_id'], $request['campaign_id'], $request['campaign_run_id']);
+
+        if (! $signup) {
+            $signup = $this->signups->create($request->all(), $transactionId);
+        }
 
         if ($signup) {
             $code = 201;
```
DFurnes: Let's send a `200 OK` when a signup already exists, and only send a `201 Created` if creating a new one!

chloealee: ah right, thank you @DFurnes !

app/Http/Controllers/Api/SignupsController.php
```diff
@@ -50,12 +50,18 @@ public function store(SignupRequest $request)
...
+        $signup = $this->signups->get($request['northstar_id'], $request['campaign_id'], $request['campaign_run_id']);
 
-        if ($signup) {
+        if (! $signup) {
+            $signup = $this->signups->create($request->all(), $transactionId);
             $code = 201;
         }
 
+        if ($signup) {
+            $code = 200;
```
DFurnes: Will this conditional always run, since `$signup` is created on line 57 if it didn't already exist?

It might be easier to set the response code before we try to make the new signup:

```php
$signup = $this->signups->get(...);
$code = $signup ? 200 : 201;

if (! $signup) {
    $signup = // ...
```

chloealee: oh yes, I like that better! Updated! @DFurnes

PR #292, opened 2017-06-15
Web signup api route

sbsmith86:
#### What's this PR do?

Creates a new endpoint at `POST /posts` that lives under the web middleware so we can make requests that get authenticated based on the logged in user from the frontend, instead of the API headers. 

Also - sidebar - fixes a failing CampaignService test. 

When updating quantity from the history modal in Rogue, I was getting authentication errors because we were not using the API headers. We don't want to use the API routes for requests from the frontend, we want to use web routes that authenticate based on the logged in user.

### DIFF
--- a/app/Http/Controllers/Traits/PostRequests.php
+++ b/app/Http/Controllers/Traits/PostRequests.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Rogue\Http\Controllers\Traits;
+
+use Rogue\Http\Requests\PostRequest;
+
+trait PostRequests
+{
+    /**
+     * Store a newly created resource in storage.
+     *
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Traits/PostRequests.php
```diff
@@ -0,0 +1,46 @@
+<?php
```
chloealee: smart to put his in a a trait instead of repeat code!

chloealee:
this all makes sense to me!

sbsmith86:
@chloealee awesome! can I get an official approval 🤗

chloealee:
ah shoot sorry @sbsmith86 !

PR #293, opened 2017-06-15
State refactor

### LINKED ISSUE #259: Think about how data lives in the campaign inbox/on inbox items
Currently signup information is nested within each post, so if something on the signup changes we have to filter through all the posts and see if they have the `signup_id` that changed, and then make the necessary change on their personal signup data. 

Long story short, if something changes on the signup we would like to be able to change it in just one place.

sbsmith86:
#### What's this PR do?

This updates how we are handling state on the `CampaignInbox` and `CampaignSingle` components. Before, we were extracting all of the posts from the signups passed via the `window.state` object and then looping over all of them to re-embed the signup object inside of the post so we could do things like show signup fields on a post or update the quantity.

This PR updates how we handle the signup data by adding the signups to the component's state so that when we need to do anything, like update quantity, or display `why_participated` we use the `signup_id` on the post to look up the individual signup in the `signups` object and update that. What that means is all posts related to that sign up will immediately get the update. 

This should make thing easier to work with on the frontend. 

I think you should be good to go commit-by-commit. 

There are no real changes to the experience, so pull down the branch and everything should work they way it was working. 

#### Relevant tickets
Fixes #259
https://trello.com/c/3ST2uIDd/431-10-as-a-developer-i-want-to-have-signup-data-living-in-the-state-and-post-use-the-signup-id-associate-w-it-to-update-quantity

### DIFF
--- a/resources/assets/components/CampaignInbox/index.js
+++ b/resources/assets/components/CampaignInbox/index.js
@@ -1,7 +1,8 @@
 import React from 'react';
-import { flatMap, keyBy, map, sample, forEach, reject } from 'lodash';
-import { RestApiClient} from '@dosomething/gateway';
+import { map, sample, forEach, reject } from 'lodash';
+import { RestApiClient } from '@dosomething/gateway';
 
+import { extractPostsFromSignups } from '../../helpers';
 import InboxItem from '../InboxItem';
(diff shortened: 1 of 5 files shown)


katiecrane:
yayyy!!!!

PR #295, opened 2017-06-27
Add competitions

sbsmith86:
#### What's this PR do?

* Adds a `competition` column to the `signups` table that stores if the signup is also opted into a competition
* Updates the `POST /signups` endpoint logic to update the signup if it is found under the given `northstar_id`, `campaign_id`, or `campaign_run_id`. This will only update the `quanity`, `why_participated` and `competition` fields on the signup. I added this so it is more user-friendly to update only these fields without having to hit `/posts`.
* adds logic on signup creation and updates to store the `competition` field.

Make a request that posts to `/signups`. Either create one or update an existing signup. Pass the `competition` parameter to the request. 

We are adding this to more accurately parallel how this works in phoenix. We are tabling discussions to talk about if we actually want this data in rogue, or if we want looker to pull competition data from gladiator and signup data from rogue and resolve who is signed up for competitions that way.

#### Relevant tickets

This address part 1 of this card: https://trello.com/c/srxkmp3j/480-7-as-rogue-i-want-to-have-competition-information-in-me-so-that-i-can-know-someone-is-in-a-competition-column-actively-being-sav

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -11,7 +11,7 @@ class Signup extends Model
      *
      * @var array
      */
-    protected $fillable = ['id', 'event_id', 'northstar_id', 'campaign_id', 'campaign_run_id', 'quantity', 'quantity_pending', 'why_participated', 'source', 'created_at', 'updated_at'];
+    protected $fillable = ['id', 'event_id', 'northstar_id', 'campaign_id', 'campaign_run_id', 'quantity', 'quantity_pending', 'why_participated', 'source', 'competition', 'created_at', 'updated_at'];
 
     /**
      * Attributes that can be queried when filtering.
--- a/app/Repositories/SignupRepository.php
+++ b/app/Repositories/SignupRepository.php
@@ -43,6 +43,11 @@ class SignupRepository
         return $signup;
     }
 
+    public function update($signup, $data)
+    {
+        dd('update signup');
+    }
+
     /**
      * Get a signup
      *
--- a/app/Repositories/SignupRepository.php
+++ b/app/Repositories/SignupRepository.php
@@ -43,9 +43,21 @@ class SignupRepository
         return $signup;
     }
 
+    /**
+     * Update a signup. Quanity and Why participated are the
+     * only fields that can be updated, at the moment.
+     *
+     * @param  Rogue\Models\Signup $signup
+     * @param  array $data
+     * @return \Rogue\Models\Signup|null
+     */
     public function update($signup, $data)
     {
-        dd('update signup');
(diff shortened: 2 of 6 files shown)


app/Models/Signup.php
```diff
@@ -11,7 +11,7 @@ class Signup extends Model
      *
      * @var array
      */
-    protected $fillable = ['id', 'event_id', 'northstar_id', 'campaign_id', 'campaign_run_id', 'quantity', 'quantity_pending', 'why_participated', 'source', 'created_at', 'updated_at'];
+    protected $fillable = ['id', 'event_id', 'northstar_id', 'campaign_id', 'campaign_run_id', 'quantity', 'quantity_pending', 'why_participated', 'source', 'competition', 'created_at', 'updated_at'];
```
DFurnes: The `id` and `event_id` fields should probably not be mass-assignable here.

app/Repositories/SignupRepository.php
```diff
@@ -24,6 +24,7 @@ public function create($data)
         $signup->quantity_pending = isset($data['quantity_pending']) ? $data['quantity_pending'] : null;
         $signup->why_participated = isset($data['why_participated']) ? $data['why_participated'] : null;
         $signup->source = isset($data['source']) ? $data['source'] : null;
+        $signup->competition = isset($data['competition']) ? $data['competition'] : null;
```
DFurnes: Totally subjective, but it might make sense to use mass-assignment for these fields:

```php
$signup = new Signup($data);

    // ...
```

app/Repositories/SignupRepository.php
```diff
@@ -44,6 +45,24 @@ public function create($data)
...
+     * Update a signup. Quanity and Why participated are the
+     * only fields that can be updated, at the moment.
+     *
+     * @param  Rogue\Models\Signup $signup
+     * @param  array $data
+     * @return \Rogue\Models\Signup|null
+     */
+    public function update($signup, $data)
+    {
+        $signup->quantity = isset($data['quantity']) ? $data['quantity'] : $signup->quantity;
```
DFurnes: Another potential use-case for mass-assignment:

```php
$signup->fill(array_only($data, ['quantity', 'why_participated', 'competition']));
$signup->save();

return $signup;
```

sbsmith86: oh i see! i wasn't doing mass-assigment with `update` because I wanted to make sure the model event triggered but see that `fill` is another option!

sbsmith86:
Closing this PR since we decided to no longer support this in rogue.

PR #302, opened 2017-07-13
Initial procfile

blisteringherb:
#### What's this PR do?
I've added a Procfile to support Heroku deploys for Rogue

You can view the Rogue app on Heroku here - https://dashboard.heroku.com/pipelines/289a727e-1b35-401e-8c80-4b0bf6dd4a77

Adding this file in won't make any changes to any current deploys we have. I currently have https set up on Heroku, but that seems to break the CSS, so we can turn that off for the time being as well.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,0 +1,2 @@
+web: vendor/bin/heroku-php-apache2 /public/
+


DFurnes:
This is so exciting! ☁️

PR #303, opened 2017-07-13
Return specific errors

### LINKED ISSUE #276: Validation errors return 500 instead of 422

katiecrane:
#### What's this PR do?
Wouldn't it be nice if when you tried to post something to Rogue, but had something wrong in your request, you got a helpful message in response instead of a `500`? Very soon, that will be the reality that we all live in.

This PR steals a bunch of code from Northstar to make that a reality. Taken from Northstar:
[`has_middleware` helper](https://github.com/DoSomething/northstar/blob/4a4b46482652e3e683837867af9ca32a3fcf1759/app/helpers.php#L54-L73) 

[unauthenticated function in `Handler.php`
](https://github.com/DoSomething/northstar/blob/ef9679e89e0aed0af470d9ac93f44e95b4265ca4/app/Exceptions/Handler.php#L115-L129)
[this real nice `buildJsonResponse` function
](https://github.com/DoSomething/northstar/blob/ef9679e89e0aed0af470d9ac93f44e95b4265ca4/app/Exceptions/Handler.php#L131-L157)
[and this chunk of the `render` function of `Handler.php`
](https://github.com/DoSomething/northstar/blob/ef9679e89e0aed0af470d9ac93f44e95b4265ca4/app/Exceptions/Handler.php#L70-L83)
Example response:
```
{
    "campaign_id": [
        "The campaign id must be an integer."
    ],
    "quantity": [
        "The quantity must be an integer."
    ]
}
```

Make sure that you see the correct response for various types of requests and exceptions.

All tests pass!

#### Relevant tickets
Fixes #276
[Trello card
](https://trello.com/c/XnYvqbRQ/467-5-%F0%9F%90%9B-validation-errors-return-500-instead-of-422-%F0%9F%90%9B)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -49,6 +49,9 @@ class Handler extends ExceptionHandler
     public function render($request, Exception $e)
     {
         // Re-cast specific exceptions or uniquely render them:
+        if ($e instanceof AuthorizationException) {
+            return parent::render($request, $e);
+        }
         if ($e instanceof AuthenticationException) {
             return $this->unauthenticated($request, $e);
         } elseif ($e instanceof ValidationException || $e instanceof NorthstarValidationException) {
--- a/app/Exceptions/Handler.php
+++ b/app/Exceptions/Handler.php
@@ -5,6 +5,8 @@ namespace Rogue\Exceptions;
 use Exception;
 use Illuminate\Validation\ValidationException;
 use Illuminate\Auth\Access\AuthorizationException;
+use Illuminate\Auth\AuthenticationException;
+use Illuminate\Http\Exception\HttpResponseException;
 use Symfony\Component\Debug\Exception\FlattenException;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Symfony\Component\HttpKernel\Exception\HttpException;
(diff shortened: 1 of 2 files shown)


sbsmith86:
Some style issues, but looks good! How far does this get us in terms of auditing all of our status codes?

katiecrane:
@sbsmith86 style issues are fixed!
In terms of auditing status codes, I think all this should do is change what is returned for certain exceptions. Some of them were getting flattened (`422`s in particular) and coming back as `500`s instead of something more specific. So now I think all exceptions will have the proper status code. This doesn't change when we return `200`s vs. `201`s.

sbsmith86:
gotcha

PR #305, opened 2017-07-14
Handle exceptions when cropping images.

DFurnes:
#### What's this PR do?
This pull request adds some simple error handling so we don't stop the whole dang process if a single image can't be processed. We can then look over the log from running the command for specific errors afterwards.

👀

This error occurred on Thor when it hit a post with `default` as the URL:

```
Saved edited image for post 1175
Saved edited image for post 1176
Saved edited image for post 1177
Saved edited image for post 1178
Saved edited image for post 1179
Saved edited image for post 1180

                                                       
  [Intervention\Image\Exception\NotReadableException]  
  Image source not readable          
```

#### Relevant tickets
Fixes 🐛.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Console/Commands/EditImagesCommand.php
+++ b/app/Console/Commands/EditImagesCommand.php
@@ -55,7 +55,7 @@ class EditImagesCommand extends Command
 
                     $aws->storeImageData($editedImage, 'edited_'.$post->id);
                 } catch (ImageException $e) {
-                    $this->error('Failed to edit image for post '.$post->id);
+                    $this->error('Failed to edit image for post '.$post->id.': '.$image);
                 }
 
                 $this->line('Saved edited image for post '.$post->id);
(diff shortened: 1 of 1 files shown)


app/Console/Commands/EditImagesCommand.php
```diff
@@ -45,12 +46,17 @@ public function handle(AWS $aws)
         Post::where('id', '>', $start)->chunk(100, function ($posts) use ($aws) {
             foreach ($posts as $post) {
                 $image = $post->url;
```
katiecrane: Great catch! I'm wonder if since we know that some of the images are `default` if in addition to the other things you've added if after this line we should have
`if (`$image` === "default") {
    return;
 }`
so that then we are printing out errors when we got a true error? Just a thought, what you have is totally functional!

DFurnes: That's a good thought! I wonder how many images we have with `default` as the URL, and what we should do with them. I pushed b9beb26 which prints the URL as well so we can easily see if a ton of these errors are happening because of `default` strings.

PR #311, opened 2017-07-19
Fix some reportback bugz.

DFurnes:
#### What's this PR do?
This pull request makes some changes to the `/reportbacks` endpoint in Rogue: mainly reformatting the `kudos` object so it's an array with one item in it (weird!), and switching the item's `created_at` to be a UNIX timestamp (also weird!). I'll make a Phoenix PR to address the remaining inconsistencies.

I also removed the unused `ReportbackItemTransformer` because it's confusing to have around!

👀 and :octocat: commits!

🤷‍♂️

#### Relevant tickets
References DoSomething/phoenix#7456.

#### Checklist
- [ ] Tested on staging.


app/Http/Transformers/ReportbackTransformer.php
```diff
@@ -21,8 +21,7 @@ public function transform(Post $post)
             'id' => $post->id,
             'status' => $post->status,
             'caption' => $post->caption,
-            // Add link to review reportback item in Rogue here once that page exists
-            // 'uri' => 'link_goes_here'
+            'uri' => url(config('services.phoenix.uri') . '/api/v1/reportbacks/'.$post->id, ['absolute' => true]),
```
katiecrane: will it actually live here?

DFurnes: Oh wait, I think it should be `/api/v1/reportback-items` but otherwise yes!

sbsmith86: hmm yea, it I thought this was meant to be like the RBs page in phoenix, which we don't have a parallel for yet in rogue.

katiecrane: yeah with Rogue on I don't think anything will live there (I thiiiink Phoenix would try to grab the reportback that has an rbid that matches `$post->id`??)

DFurnes: I don't think so, unless I'm misreading the [current production response](https://www.dosomething.org/api/v1/reportback-items?campaigns=7656&page=3).

katiecrane: ahhhhhhh I see what it is now 👌

PR #321, opened 2017-07-24
Paginate the single view campaign page

katiecrane:
#### What's this PR do?
1. Use `->paginate` on the query to get the signups to display on the single view campaign page. I set it at 50, but we can increase or decrease that depending on the performance we see and what works best for people.
2. Remove the 🐼  message because all posts can be accessed now
3. Add back in sorting
4. Add a new component `PagingButtons` to display the links to the adjacent pages (I figured we'll end up having pagination elsewhere too eventually)

Bottom of the page looks like this:
image

Caution: I did write some css. See if the way this is organized makes sense. It works, but I don't want to introduce any debt!

We need to paginate these pages so we aren't trying to load thousands and thousands of posts at once.

#### Relevant tickets
[Pivotal card? did I do this right?](https://www.pivotaltracker.com/story/show/149263495)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -115,8 +115,6 @@ class CampaignsController extends Controller
                 ],
                 'next_page' => $signups->nextPageUrl(),
                 'previous_page' => $signups->previousPageUrl(),
-                'next_page_copy' => $signups->nextPageUrl() ? 'next >' : null,
-                'previous_page_copy' => $signups->previousPageUrl() ? '< previous' : null,
             ]);
     }
 }
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -91,8 +91,7 @@ class CampaignsController extends Controller
      */
     public function showCampaign($id)
     {
-        // @TODO: we should paginate here instead of just showing 100
-        $signups = Signup::campaign([$id])->has('posts')->with('posts')->paginate(50);
+        $signups = Signup::campaign([$id])->has('posts')->with('posts')->orderBy('created_at', 'desc')->paginate(50);
 
         // @TODO EXTRACT AND FIGURE OUT HOW NOT TO HAVE TO DO THIS.
         $signups->each(function ($item) {
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -92,7 +92,7 @@ class CampaignsController extends Controller
     public function showCampaign($id)
     {
         // @TODO: we should paginate here instead of just showing 100
-        $signups = Signup::campaign([$id])->has('posts')->with('posts')->paginate(1);
+        $signups = Signup::campaign([$id])->has('posts')->with('posts')->paginate(50);
 
         // @TODO EXTRACT AND FIGURE OUT HOW NOT TO HAVE TO DO THIS.
         $signups->each(function ($item) {
@@ -116,6 +116,8 @@ class CampaignsController extends Controller
                 ],
                 'next_page' => $signups->nextPageUrl(),
                 'previous_page' => $signups->previousPageUrl(),
+                'next_page_copy' => $signups->nextPageUrl() ? 'next >' : null,
+                'previous_page_copy' => $signups->previousPageUrl() ? '< previous' : null,
             ]);
     }
 }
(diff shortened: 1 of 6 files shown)


resources/assets/components/PagingButtons/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+  render() {
+    const prev = this.props.prev;
+    const next = this.props.next;
+    const prevCopy = this.props.prevCopy;
+    const nextCopy = this.props.nextCopy;
+
+    return (
+        <div className="container__block">
+			<a href={prev}>{prevCopy}</a>
```
sbsmith86: indentation looks off here

resources/assets/app.scss
```diff
@@ -52,3 +52,7 @@ button.delete {
   font-weight: bold;
   margin-right: 5px;
 }
+
+.next-page {
```
sbsmith86: I feel like, as much as possible, we should keep component-level styles in the components directory. Similar to [this](https://github.com/DoSomething/rogue/tree/master/resources/assets/components/StatusButton) . Otherwise we risk this app.scss file getting bloated.

resources/assets/components/PagingButtons/index.js
```diff
@@ -0,0 +1,22 @@
...
+			<div className="next-page">
```
DFurnes: Should we check if `next` or `prev` are set here? We probably want to disable or hide the link if the user is on the first or last page.

app/Http/Controllers/CampaignsController.php
```diff
@@ -107,13 +107,17 @@ public function showCampaign($id)
...
+                'signups' => $signups->toArray()['data'],
                 'campaign' => $campaign,
                 'post_totals' => [
                     'accepted_count' => $totals->accepted_count,
                     'pending_count' => $totals->pending_count,
                     'rejected_count' => $totals->rejected_count,
                 ],
+                'next_page' => $signups->nextPageUrl(),
+                'previous_page' => $signups->previousPageUrl(),
+                'next_page_copy' => $signups->nextPageUrl() ? 'next >' : null,
```
DFurnes: Since it's not dynamic, does this need to be returned in the JSON response?

lkpttn:
@sbsmith86 @katiecrane Jumping in really quick, does this affect the sort order of the RBs? If we could do most recent RBs at the top/beginning it would solve some pain points for campaigns/impact.

resources/assets/components/PagingButtons/index.js
```diff
@@ -0,0 +1,22 @@
...
+				<a href={next}>{nextCopy}</a>
```
DFurnes: This is totally not needed for this pull request, but we should think about loading these in on the client-side so we don't need to have a full page refresh here.

resources/assets/components/PagingButtons/index.js
katiecrane: The best I could figure out was how it is now, which only shows next or previous if there is a next or previous. Disabling the link would probably be easiest, I was trying to figure out a better way to hide but this is what I came up with

resources/assets/app.scss
katiecrane: Love it, on it!

app/Http/Controllers/CampaignsController.php
katiecrane: @DFurnes this part is better now!

resources/assets/components/PagingButtons/index.js
katiecrane: @DFurnes I think this is cleaner now too, it's easier to see that the text is `null` and nothing is displayed if there isn't a page to go to

chloealee:
this looks great to me!

app/Http/Controllers/CampaignsController.php
```diff
@@ -107,13 +106,15 @@ public function showCampaign($id)
 
         return view('pages.campaign_single')
             ->with('state', [
-                'signups' => $signups,
+                'signups' => $signups->toArray()['data'],
```
DFurnes: Do you need to grab from `['data']` here? I think that comes from the transformer, not model's `toArray`.

katiecrane: @DFurnes now `$signups` is a `Illuminate\Pagination\LengthAwarePaginator` and everything that uses `signups` expects it to be an array of signups. Probably would be better to change everything else to work with the paginator, but if we want to switch it to paginate on the front end this makes sense to me for now

DFurnes: Oh right, got it! If you want, you could switch to `$signups->items()` to grab the items directly.

PR #324, opened 2017-07-26
Reportbacks Endpoint update

chloealee:
#### What's this PR do?
- `GET /reportbacks` used to only return posts that have the `status` as `accepted`. To better mimic P-A endpoint, we want to return all the reportbacks.
- The gallery in P-A only wants to show `accepted` posts. Since we're returning all reportbacks now, this PR adds a `status` filter to the endpoint.
- Removes unused `ReportbacksController`

Hit `/reportbacks?&filter[status]=accepted` and only accepted posts should be returned. 

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -68,6 +67,11 @@ public function index(Request $request)
                 $query = $query->where('signups.northstar_id', '=', $filters['northstar_id']);
             }
 
+            // Only return Posts for the given status (if there is one)
+            if (array_has($filters, 'status')) {
+                $query = $query->where('status', '=', $filters['status']);
```
DFurnes: Do we need to be able to filter by more than one status here? e.g. `status=approved,pending`?

chloealee: ooohhhh good call! I don't think we need this at the moment but in case we need to in the future, should I add this now?

DFurnes: Oh ok, got it. So the user profile will accept any post (`pending`, `accepted`, or `rejected`) and then only the gallery needs to filter (to just `accepted`), right? That seems fine to me! 👍

DFurnes: As long as we don't have any other APIs relying on the old Phoenix query string that sounds good to me to hold off on it.

chloealee: Yup - that's right @DFurnes !

Issue #325, opened 2017-07-26
Discussion: Optimize pagination query


sbsmith86:
I have been looking into some query optimizations for Rogue. New Relic is reporting [this query](https://rpm.newrelic.com/accounts/108038/applications/24971199/datastores#/overview/MySQL/trace/?id=e899ce9a-6e31-11e7-bb42-0242ac11000b_2066_3706&metric=Datastore%252Fstatement%252FMySQL%252Fposts%252Fselect) as being a pretty heavy MySQL transaction. 

I dug into this recommendation: 
![image](https://user-images.githubusercontent.com/1700409/28635758-b6a697ba-7209-11e7-82c7-fe058671e3d4.png)

It seems we should add more indexes to the `posts` table but this query in particular is querying across all rows, so i'm not sure putting an index across all rows will actually help anything (yes?)

So I was digging into why we had to count across all rows instead of just the primary key, `id`, since we are just doing a count. 

It turns out you should be able to pass the specific columns to count in the `paginate()` method [here](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Traits/TransformsRequests.php#L108)

However, this still doesn't work because there seems to be an issue in core where we don't actually pass the specified columns to [getCountForPagination()](https://github.com/laravel/framework/blob/5.2/src/Illuminate/Database/Eloquent/Builder.php#L491) Which is called in the `paginate()` method and is the method that actually creates the heavy query

So my questions for discussion are: 

* Did I find an actual issue in core that I should submit a PR for? 
* Is there something we can do on our end to make this query happy? Should I in fact, index across all columns? Should we write our own pagination query?
* Will specifying just counting `posts.id` work? `id` is a primary key and index on the `posts` table so I think it should help the queries performance, but maybe I am wrong.

sbsmith86:
Also, should be noted that this query is running when the `/reportback` and `/activity` endpoints are hit.

DFurnes:
I'm a little stumped by this one. The performance bottleneck seems to be coming from the `inner join` that we're doing on the `signups` table, but we have indexes on both `signup.id`, `posts.signup_id`, and a foreign key connecting them so not sure how much we can improve.

![screen shot 2017-07-27 at 3 34 25 pm](https://user-images.githubusercontent.com/583202/28689327-433f23ca-72e3-11e7-8a42-7d75ed1391f6.png)

[...]

PR #329, opened 2017-07-28
Use Eloquent query to improve pagination performance.

DFurnes:
#### What's this PR do?
This improves performance of this endpoint pretty dramatically on my local – from 10s to 0.6s when using the Thor database dump. The improvement comes from switching to using Eloquent's query builder rather than a plain database query (with joins and selects).

:eyes:

When using eager loading to get a relationship, Laravel makes a second query `select * from signups where signups.id in (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)` to grab the related items (rather than a `join`). Since we were doing the same thing via the "raw" database query builder and a `join` (which could affect the number of results), it wasn't able to optimize this.

The original "pagination" query:

```sql
select count(*) as aggregate from `posts` inner join `signups` on `signups`.`id` = `posts`.`signup_id` where 1 = 1 and `status` = ? and `posts`.`deleted_at` is null
```

And the one produced by Eloquent:

```sql
select count(*) as aggregate from `posts` where 1 = 1 and `status` = ? and `posts`.`deleted_at` is null
```

#### Relevant tickets
References #325.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/ReportbackController.php
+++ b/app/Http/Controllers/Api/ReportbackController.php
@@ -45,17 +45,17 @@ class ReportbackController extends ApiController
     public function index(Request $request)
     {
         // Create an empty Post query, which we can filter and paginate
-        // Only return Posts that do not have 'Hide In Gallery' tag
-        $query = $this->newQuery(Post::class)->withoutTags(['Hide In Gallery']);
-
-        // 1. Join with signups so we can access the signup data and filter by campaign
-        // 2. Only return approved Posts
-        // 3. Select all the fields that we will be using
-        $query = $query->join('signups', 'signups.id', '=', 'posts.signup_id')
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -45,42 +45,29 @@ public function __construct(ReportbackService $reportbackService)
...
-            // Only return Posts for the given northstar_id (if there is one)
-            if (array_has($filters, 'northstar_id')) {
-                $query = $query->where('signups.northstar_id', '=', $filters['northstar_id']);
-            }
-
-            // Only return Posts for the given status (if there is one)
-            if (array_has($filters, 'status')) {
-                $query = $query->where('status', '=', $filters['status']);
-            }
-
```
katiecrane: Do we not want to allow filtering by these anymore?

DFurnes: I made a goof. 🌵

PR #331, opened 2017-07-31
Turn filtering off and only show accepted posts

katiecrane:
#### What's this PR do?
This will I think let us push campaign single view pagination live. The page will only include accepted posts and does not allow filtering. I also added copy to let admins know that this page only shows accepted posts.

image

Is this the interim state that we want for this? 👀 

Currently, we can't have pagination and filtering because the pagination happens before the filtering, so if there are no `accepted` posts on the first page and the filter is set to `accepted` it will be a blank page with a `next >` button which is not what we want.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -91,8 +91,7 @@ class CampaignsController extends Controller
      */
     public function showCampaign($id)
     {
-        // @TODO: currently we do not allow filtering so we only show accepted posts, so
-        // remove the contraint on status once filtering is rockin' 'n' rollin' again
+        // Always load the page with just accepted posts
         $signups = Signup::campaign([$id])->has('posts')->with(['posts' => function ($query) {
             $query->where('status', '=', 'accepted');
         }])->orderBy('created_at', 'desc')->paginate(50);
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/CampaignsController.php
```diff
@@ -91,7 +91,11 @@ public function showInbox($campaignId)
      */
     public function showCampaign($id)
     {
-        $signups = Signup::campaign([$id])->has('posts')->with('posts')->orderBy('created_at', 'desc')->paginate(50);
+        // @TODO: currently we do not allow filtering so we only show accepted posts, so
+        // remove the contraint on status once filtering is rockin' 'n' rollin' again
+        $signups = Signup::campaign([$id])->has('posts')->with(['posts' => function ($query) {
+            $query->where('status', '=', 'accepted');
```
chloealee: @katiecrane actually this is exactly what I need for [this card](https://www.pivotaltracker.com/n/projects/2019429/stories/149809538) since we'll be dealing with filter logic with the API so this is super helpful! Maybe you can remove the comment on line 95?

katiecrane: @chloealee you won't always want exclusively accepted posts though will you?

chloealee: @katiecrane yes I think we will! So on page load, we'll only show accepted posts and then when we filter, we'll send an API request to `GET /posts` that @sbsmith86 is building out now to filter by those statuses etc. to return the posts we want to display to help with performance on initial page load.

katiecrane: @chloealee OOOOO okay! that's awesome!!!

PR #334, opened 2017-08-01
Move `campaign_id` into posts table & re-apply #329.

### LINKED ISSUE #325: Discussion: Optimize pagination query

DFurnes:
#### What's this PR do?
This pull request hopefully wraps up the performance work on the reportbacks endpoint for now:

🤔 I tried a couple approaches to avoid having to duplicate the `campaign_id` column but neither seemed to work all that well. From what I can tell, joins on big tables are just inherently not good. (2bd00f8, 4ca02f4)

🆔 The best solution, at least for now, seems to be copying the `campaign_id` into the posts table so that we can just query against the indexed column directly. This runs in ~130ms rather than the 12 or 15 seconds I was seeing with joins and subqueries. (d281c7a)

🚥 I made some updates to the `PostApiTest` since it hadn't been updated with the "edited URL" change we made pre-launch (2a06a64), and updated `PostRepository` to set `campaign_id` on new posts (1058eee). More tests are still failing, but I'll revisit in another PR.

🕰 I enabled [Clockwork](https://chrome.google.com/webstore/detail/clockwork/dmggabnehkmmfmdffgajcflpdjlnoemp?hl=en) support since that made it easier to dive into query performance on API requests, instead of die-dumping the query events like I was earlier. (b0e7903)

Let's test this thoroughly on Thor since the test suite is still pretty light! Am I missing any code paths that create posts in other ways (or any other gotchas)?

🥞 🤷‍♂️

#### Relevant tickets
Fixes #325.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/database/migrations/2017_07_31_153255_add_campaign_id_to_posts.php
+++ b/database/migrations/2017_07_31_153255_add_campaign_id_to_posts.php
@@ -1,0 +1,41 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\Schema;
+use Rogue\Models\Post;
+
+class AddCampaignIdToPosts extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('posts', function (Blueprint $table) {
+            $table->integer('campaign_id')->after('signup_id')->index()
+                ->comment('The campaign node ID for this post.');
+        });
+
+        // Copy `campaign_id` from the signups table so we can make more
+        // efficient queries against the posts table (without expensive join).
+        foreach (Post::with('signup')->cursor() as $post) {
+            $campaign_id = $post->signup->campaign_id;
+            $post->update(['campaign_id' => $campaign_id]);
+        }
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('posts', function (Blueprint $table) {
+            $table->dropColumn('campaign_id');
+        });
+    }
+}
--- a/database/migrations/2017_07_31_153255_add_campaign_id_to_posts.php
+++ b/database/migrations/2017_07_31_153255_add_campaign_id_to_posts.php
@@ -1,9 +1,9 @@
 <?php
 
+use Rogue\Models\Post;
+use Illuminate\Support\Facades\Schema;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Database\Migrations\Migration;
-use Illuminate\Support\Facades\Schema;
-use Rogue\Models\Post;
 
 class AddCampaignIdToPosts extends Migration
 {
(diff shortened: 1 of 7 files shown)


DFurnes:
Also as a heads up, this is a pretty outrageous migration – it should be quick on local and staging environments, but it'll take a good while for Thor and production.

database/migrations/2017_07_31_153255_add_campaign_id_to_posts.php
```diff
@@ -0,0 +1,41 @@
...
+    public function up()
+    {
+        Schema::table('posts', function (Blueprint $table) {
+            $table->integer('campaign_id')->after('signup_id')->index()
+                ->comment('The campaign node ID for this post.');
+        });
+
+        // Copy `campaign_id` from the signups table so we can make more
+        // efficient queries against the posts table (without expensive join).
+        foreach (Post::with('signup')->cursor() as $post) {
```
sbsmith86: Should we be doing this in this migration or in a separate script we can kick off after a deploy? If we keep it in the migration, will it block any deploys until the migration is over?

DFurnes: I think this belongs in a migration, since it lets us ensure that the schema change happens alongside the corresponding logic changes. And unless there's a misconfiguration in Jenkins, it shouldn't be possible to run more than one deploy at once.

DFurnes: It could be worthwhile to split it into two separate migrations, though, for easier rollback. 🤔

sbsmith86: ok do you have an idea about how long it will take? If we need to hold off on deploys for the next couple of days or anything, we should let people know.

DFurnes: Oh no, should be less than an hour.

sbsmith86: Ah great!

DFurnes:
Cool, let's give it a go on Thor! 😬

PR #337, opened 2017-08-01
Add POST /tags endpoint

katiecrane:
#### What's this PR do?
- Added `Api/TagsController.php` to handle posts getting tagged (or untagged) via the api
- Added a new route: `POST api/v2/tags`
- For tag or untag events via the API (or anytime there is no `auth()->user()`), store `null` for the user column on the event
- Updated documentation

👀 Hit the endpoint and make sure you get a response like one provided in the documentation. Send the same request again and the tag you just added should disappear 🎱 

We need this endpoint to migrate over the tags that live in Phoenix (excluded = Hide in Gallery, promoted = Good Photo).

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/n/projects/2019429/stories/149658211)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/TagsController.php
+++ b/app/Http/Controllers/Api/TagsController.php
@@ -2,12 +2,14 @@
 
 namespace Rogue\Http\Controllers\Api;
 
+use Rogue\Http\Controllers\Traits\TagsRequests;
 use Rogue\Http\Requests\TagsRequest;
 use Rogue\Repositories\PostRepository;
 use Rogue\Http\Transformers\PostTransformer;
 
 class TagsController extends ApiController
 {
+    use TagsRequests;
     /**
      * The post service instance.
      *
@@ -30,14 +32,5 @@ class TagsController extends ApiController
     {
         $this->post = $post;
         $this->transformer = new PostTransformer;
-    }
-
-    public function store(TagsRequest $request)
-    {
-        $post = $this->post->find($request->post_id);
-
-        $taggedPost = $this->post->tag($post, $request->tag_name);
-
-        return $this->item($taggedPost);
     }
 }
--- a/app/Http/Controllers/Api/TagsController.php
+++ b/app/Http/Controllers/Api/TagsController.php
@@ -2,21 +2,42 @@
 
 namespace Rogue\Http\Controllers\Api;
 
+use Rogue\Http\Requests\TagsRequest;
+use Rogue\Http\Transformers\PostTransformer;
+use Rogue\Repositories\PostRepository;
+
 class TagsController extends ApiController
 {
+    /**
+     * The post service instance.
+     *
+     * @var Rogue\Repositories\PostRepository
+     */
+    protected $post;
+
+    /**
+     * @var \League\Fractal\TransformerAbstract;
+     */
+    protected $transformer;
+
     /**
      * Create a controller instance.
      *
      * @param  PostContract $posts
      * @return void
      */
-    public function __construct()
+    public function __construct(PostRepository $post)
     {
-        $this->middleware('api');
+        $this->post = $post;
+        $this->transformer = new PostTransformer;
     }
 
-    public function store(TagRequest $request)
+    public function store(TagsRequest $request)
     {
-        dd('Here is what we got:', $request);
+        $post = $this->post->find($request->post_id);
+
+        $taggedPost = $this->post->tag($post, $request->tag_name);
+
+        return $this->item($taggedPost);
     }
 }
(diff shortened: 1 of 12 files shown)


app/Http/Controllers/Api/TagsController.php
```diff
@@ -0,0 +1,43 @@
...
+     * @param  PostContract $posts
+     * @return void
+     */
+    public function __construct(PostRepository $post)
+    {
+        $this->post = $post;
+        $this->transformer = new PostTransformer;
+    }
+
+    public function store(TagsRequest $request)
```
sbsmith86: One thing I did for the `PostController` store method was I created a trait `PostRequests` that had the `store()` method in it and then used that train in both `PostController.php` and `Api/PostController.php` that way the logic could be shared between the two controllers and only live in one place.

katiecrane: @sbsmith86 oooo I like that!!

DFurnes: I'm a little concerned we'll run into tiny differences we'd want to implement between `store` methods in different controllers, which we couldn't easily override with a trait. I've generally found traits are more useful for tinier bits of optional functionality.

app/Http/Controllers/Traits/TagsRequests.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(TagsRequest $request)
+    {
+        $post = $this->post->find($request->post_id);
```
DFurnes: It's a little awkward that this has an implicit dependency on the controller's `post` field. Not sure how best to improve that, but just food for thought.

resources/assets/components/CampaignInbox/index.js
```diff
@@ -81,7 +82,7 @@ class CampaignInbox extends React.Component {
         const newState = {...previousState};
         const user = newState.posts[postId].user;
 
-        newState.posts[postId] = data;
+        newState.posts[postId] = data['data'];
```
DFurnes: Should we rename the `data` parameter to `json` or `response` perhaps? `data['data']` feels awkward.

resources/assets/helpers.js
```diff
@@ -22,11 +22,21 @@ export function calculateAge(date) {
...
 
 export function getImageUrlFromProp(photoProp) {
-	const photo_url = photoProp['url'];
+	var photo_url;
 
-	if (photo_url == "default") {
-	  return "https://www.dosomething.org/sites/default/files/JenBugError.png";
-	}
+  // Sometimes we get the url right on the post and sometimes it is nested under
+  // media (in cases where it goes through the PostTransformer), so handle both cases
```
DFurnes: Let's add a `@TODO` to normalize this so our data is always passed through `PostTransformer` before getting here.

app/Http/Controllers/Api/TagsController.php
sbsmith86: Sure, but right now they are exactly the same. I can't imagine us really wanting the functionality to be different especially for tagging, and the only difference is the way we authorize the request.

katiecrane: @DFurnes Is this still the code we want to use for both controllers though? All the work I did on the front end was a result of switching to use this code in the `Controllers/TagsController`. Since it's happy with this code now, it's easy enough to move it out of the trait and back into both controllers. I find that to be less DRY but also easier to follow, so I could go either way

resources/assets/helpers.js
```diff
@@ -42,7 +52,19 @@ export function extractPostsFromSignups(signups) {
 
 export function getEditedImageUrl(photoProp) {
   const edited_file_name = `edited_${photoProp.id}.jpeg`;
-  const url_parts = photoProp['url'].split("/");
+  var url_parts;
+
+  // Sometimes we get the url right on the post and sometimes it is nested under
+  // media (in cases where it goes through the PostTransformer), so handle both cases
+  if ('url' in photoProp) {
+    url_parts = photoProp['url'].split("/");
```
DFurnes: Can we pull the rest of this condition (the code below the if/else) up into the conditional?

app/Http/Controllers/Api/TagsController.php
DFurnes: Personally I think the readability of having the code in the controller trumps DRYness here, but definitely not something that we need to change to get this merged.

app/Http/Controllers/Traits/TagsRequests.php
katiecrane: @DFurnes I'm going to add a note about that

PR #342, opened 2017-08-02
GET /posts endpoint

sbsmith86:
#### What's this PR do?

Creates `GET /api/v2/posts` (API requests) and `GET /posts` (app requests) endpoint that returns a set of posts given the query params it is passed. 

This works exactly like the `GET /reportbacks` endpoint but returns response formatted for the new rogue data schema. 

Also should note that I update the logic for the filtering to use the `FiltersRequest` trait instead of doing it separately as we are doing in the `ReportbackController`. Now that we have made some updates to the `posts` table, I think this is a little easier to use. 

👀 , commit-by-commit, Make sure the updates to `FiltersRequests` seem ok. 

Will help the app do some of the filtering, pagination and sorting it needs to do by making an api request for posts directly. 

Also make things more RESTful :) 

#### Relevant tickets
https://www.pivotaltracker.com/story/show/149811449

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Traits/FiltersRequests.php
+++ b/app/Http/Controllers/Traits/FiltersRequests.php
@@ -38,9 +38,6 @@ trait FiltersRequests
             $excludedValues = false;
         }
 
-        // Requests may be filtered by indexed fields.
-        $filters = array_intersect_key($filters, array_flip($indexes));
-
         // If there is an updated_at param, remove it from $filters and save the value.
         if (array_key_exists('updated_at', $filters)) {
             $updatedAtValue = $filters['updated_at'];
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Traits/FiltersRequests.php
```diff
@@ -29,8 +29,14 @@ public function filter($query, $filters, $indexes)
         if (! $filters) {
             return $query;
         }
-        // Requests may be filtered by indexed fields.
-        $filters = array_intersect_key($filters, array_flip($indexes));
+
+        // If there is an exclude filter, remove it from $filters and save the value.
+        if (array_key_exists('exclude', $filters)) {
```
DFurnes: Yes! 🎉 I'd love to deprecate this in the API as well.

PR #343, opened 2017-08-03
Speed up test suite & run automatically on pull requests.

DFurnes:
#### What's this PR do?
This pull request includes some improvements to make testing in Rogue easier:

📁 Mark a few failing tests as incomplete (will investigate further in another PR). Also moves tests into `Http` and `Http/Api` subfolders to make things easier to find, and also avoid confusion with non-test trait and class added for... 4439ddc

🐎 Introduces a `RefreshDatabase` trait (yanked from Laravel 5.5). This dramatically speeds up the test suite (from 1 minute to 9 seconds!!) so it'll be a lot easier to run frequently in dev. a29e9d6

🚥 Adds a Wercker config (builds only) so that we can ensure tests pass before merging! f2d7a71

👀

🍃

#### Relevant tickets

#### Checklist
- [ ] Tested on staging.


chloealee:
this will be so helpful to have them run automatically - thanks Dave!!

tests/RefreshDatabase.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+trait RefreshDatabase
+{
+    /**
+     * Refresh a conventional test database.
+     */
+    protected function refreshDatabase()
```
sbsmith86: isn't this similar to what we do in https://github.com/DoSomething/rogue/blob/master/app/Console/Commands/ClearDB.php

DFurnes: Yep, we use that command right below this! ⬇️ The difference from what we were doing before is that we only run migrations once at the start of the test suite, and then use transactions to reset it to the initial state between each method.

sbsmith86: ahhh i see. cool!

PR #345, opened 2017-08-04
Add some useful test helpers & consistently format tests.

DFurnes:
#### What's this PR do?
This pull request adds some useful test helpers & reformats tests a bit for readability.

Here's some highlights:

🔏 Adds an `actingAsAdmin` helper since this is a really common scenario. ae72f2f

🗑 Extracts a `seeSoftDeletedRecord` helper from the existing assertions in the tests. 27910c1

⏱ Adds a `mockTime` helper which allows us to manipulate the current time for tests without dropping `sleep()` calls and actually slowing things down. c702559

🏭 Adds some model state factories (for admin users & accepted/rejected posts). f29a85a, 011df7e

A few fixes to broken tests:

📊 Fixed `getPostTotals` to send expected parameter and assert against expected return value.  I think this just hadn't been updated when that method was changed at some point. 0ad5178

🔒 Updated the homepage test to not unnecessarily mock the CampaignService, and fixed an exception that would occur on an empty database. Also gave the campaign inbox a heading to assert against (in the future we should give all users a "Log Out" button, not a "Whoops"!) ec884c1

I also reformatted tests a bit for readability. These are more subjective changes:

🔢 Separated tests into "paragraphs": create models/mocks, make request, make assertions. This makes it a bit easier to read each individual test as a "story" without having to jump around to see where things are being referenced from.

🔤 Inlined URLs into the actual requests. This makes it easier to look at a test in isolation, without having to scroll to the top of the file to see what's going on.

✅ Whenever possible, made assertions against provided input rather than the response body.

🖊 Enabled StyleCI for test directory.

With great patience. 😬

I'm going to add some additional tests & look into the ones marked as incomplete in a follow-up pull request. I'm also hoping to touch up the seeders a bit for nicer local development.

#### Relevant tickets

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/tests/Http/Api/ActivityApiTest.php
+++ b/tests/Http/Api/ActivityApiTest.php
@@ -1,9 +1,8 @@
 <?php
 
+use Carbon\Carbon;
 use Rogue\Models\Signup;
 use Rogue\Models\Post;
-use Rogue\Models\Event;
-use Faker\Generator;
 
 class ActivityApiTest extends TestCase
 {
@@ -134,14 +133,16 @@ class ActivityApiTest extends TestCase
 
     public function testActivityIndexWithUpdatedAtQuery()
     {
+        Carbon::setTestNow(new Carbon('8/01/2017 15:00:00'));
+
         // Create two signups and two posts. Associate the signups and posts respectively.
         $firstSignup = factory(Signup::class)->create();
         $firstPost = factory(Post::class)->create();
         $firstPost->signup()->associate($firstSignup);
         $firstPost->save();
 
-        // Wait 1 second before making the second signup and post so they have different timestamps.
-        sleep(1);
+        // Make another signup and post later on.
+        Carbon::setTestNow(new Carbon('8/01/2017 15:01:00'));
 
         $secondSignup = factory(Signup::class)->create();
         $secondPost = factory(Post::class)->create();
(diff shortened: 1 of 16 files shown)


tests/Http/Api/ActivityApiTest.php
```diff
@@ -86,14 +78,13 @@ public function testActivityIndexWithCampaignRunIdQuery()
     {
         $signup = factory(Signup::class)->create();
 
-        $this->json('GET', $this->activityApiUrl . '?filter[campaign_run_id]=' . $signup->campaign_run_id);
+        $this->get('api/v2/activity?filter[campaign_run_id]=' . $signup->campaign_run_id);
```
sbsmith86: Looks great, Just wondering why you prefer hard coding the url instead storing in a prop?

DFurnes: Yeah, it's a little subjective! I think this may just have been a side-effect of not being as familiar with the codebase, but I found myself consistently scrolling up in each file to remember where I was as I read the tests. My general approach with testing is to be as explicit as possible so it's really easy to follow exactly what steps are being performed when looking at a test in isolation.

Definitely up for having a discussion at retro & reverting based on team opinion!

DFurnes:
Thanks for reviewing @sbsmith86! 🎉

PR #347, opened 2017-08-07
Test Cleanup, Part III

DFurnes:
#### What's this PR do?
Last general purpose cleanup pull request!

🗃 Moves non-HTTP tests (models and services) to their own directories. f8a7933, 26e4046

🐎 Adds some more tests to the activity endpoint, since before we were mostly just testing the built-in paginator (rather than actual JSON results). 555188c

🆕 Makes sure to re-fetch from database before making assertions, and swaps order of arguments so PHPUnit outputs the correct things in the "expected" and "actual" columns. 813e40c

👉 Updates `TODO`s for broken "touching". It turns out that Laravel only touches immediate parents for the `$touches` property, so a reaction would update post timestamp but _not_ the signup timestamp. We should prioritize this and fix in a future sprint. aca684d

🚥 👀 

☔️

#### Relevant tickets

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/documentation/endpoints/activity.md
+++ b/documentation/endpoints/activity.md
@@ -6,29 +6,21 @@ Retrieve all user activity data.
 GET /api/v2/activity
 ```
 ### Optional Query Parameters
-- **id** _(integer)_
-  - The signup id(s) to filter the response by.
-  - e.g. `/activity?filter[id]=121,122`
+- **filter[column]** _(integer)_
+  - Filter results by the given column: `id`, `campaign_id`, `campaign_run_id`, `northstar_id`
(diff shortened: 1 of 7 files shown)


DFurnes:
> Do you think we need tests for the other query params (`id`, `include=user`, and `northstar_id`)? 

Oh good call! I think anything that has different logic is worth testing, so `filter[id]` and `filter[northstar_id]` are probably safe to skip (since they work the same way as `filter[campaign]`), but `include` should definitely have test coverage! I'll add that now.

chloealee:
@DFurnes that makes sense to me!

DFurnes:
Updated!

PR #348, opened 2017-08-08
Filters posts powered by the API

chloealee:
#### What's this PR do?
- Changes Campaign Single Index page's initial state to be powered by API instead of sending signups from backend (`GET /posts`). 
- Powers filtering by status via the API (`GET /posts`).
- Adds relation to Post model of other related posts under the same signup (siblings).
- Updates Post Transformer to optionally include related posts (siblings) and signup object for `GET /posts`).
- Returns more information about the user in the User Transformer (user's last name, birthdate, email address, mobile number).  
- Returns the original photo url in the Post Transformer. 
- Updates `$include` param to allow for multiple values (e.g. `$include=signup,siblings`).
- Updates Post tests.  
- BONUS: 
    - Adds some missing things to README file for documentation.

- Test `campaigns/:nid` to make sure initial state looks as expected and filtering works properly. 
    - Make sure that all information about the signup and post is exposed correctly! 
    - Don't forget to check the "Original Photo" and "Edited Photo" links and that they're working correctly! 
- Make sure `/campaigns/:nid/inbox` isn't broken. 

- I think that this breaks pagination? Is this ok for now and we can make another issue to fix this? 
- @DFurnes I don't think we need to do this but should we update the activity test to make sure it includes additional information about the user? Has same functionality so not sure if this is needed! 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/149809538

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -31,7 +31,7 @@ class ActivityController extends ApiController
 
         $filters = $request->query('filter');
 
-        if (array_key_exists('status', $filters)) {
+        if (array_has($filters, 'status')) {
             // Constrain Eager Load, only returning posts with status query.
             $query = $query->with(['posts' => function ($query) use ($filters) {
                 $query->where('status', $filters['status']);
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -27,22 +27,9 @@ class ActivityController extends ApiController
     {
         // Create an empty Signup query and eager-load posts, which we
         // can either filter or paginate to retrieve all signup records.
-        $query = $this->newQuery(Signup::class);
+        $query = $this->newQuery(Signup::class)->with('posts');
 
         $filters = $request->query('filter');
-
-        if (array_has($filters, 'status')) {
-            // Constrain Eager Load, only returning posts with status query.
-            $query = $query->with(['posts' => function ($query) use ($filters) {
-                $query->where('status', $filters['status']);
-            }]);
-
-            // Remove status from $filters to query Signups.
-            unset($filters['status']);
-        } else {
-            $query = $query->with('posts');
-        }
-
         $query = $this->filter($query, $filters, Signup::$indexes);
 
         return $this->paginatedCollection($query, $request);
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -27,9 +27,22 @@ class ActivityController extends ApiController
     {
         // Create an empty Signup query and eager-load posts, which we
         // can either filter or paginate to retrieve all signup records.
-        $query = $this->newQuery(Signup::class)->with('posts');
+        $query = $this->newQuery(Signup::class);
 
         $filters = $request->query('filter');
+
+        if (array_key_exists('status', $filters)) {
+            // Constrain Eager Load, only returning posts with status query.
+            $query = $query->with(['posts' => function ($query) use ($filters) {
+                $query->where('status', $filters['status']);
+            }]);
+
+            // Remove status from $filters to query Signups.
+            unset($filters['status']);
+        } else {
+            $query->with('posts');
+        }
+
         $query = $this->filter($query, $filters, Signup::$indexes);
 
         return $this->paginatedCollection($query, $request);
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -40,7 +40,7 @@ class ActivityController extends ApiController
             // Remove status from $filters to query Signups.
             unset($filters['status']);
         } else {
-            $query->with('posts');
+            $query = $query->with('posts');
         }
 
         $query = $this->filter($query, $filters, Signup::$indexes);
--- a/resources/assets/components/CampaignSingle/index.js
+++ b/resources/assets/components/CampaignSingle/index.js
@@ -48,7 +48,8 @@ class CampaignSingle extends React.Component {
       filter: {
         status: status.toLowerCase(),
         campaign_id: this.props.campaign.id,
-      }
+      },
+      include: user,
     });
     request.then((result) => {
       var posts = keyBy(result.data, 'id');
@@ -122,7 +123,6 @@ class CampaignSingle extends React.Component {
   }
 
   render() {
-    console.log(this.state.signups);
     const posts = this.state.posts;
     const campaign = this.props.campaign;
 
--- a/resources/assets/components/CampaignSingle/index.js
+++ b/resources/assets/components/CampaignSingle/index.js
@@ -116,7 +116,7 @@ class CampaignSingle extends React.Component {
 
         <PostFilter onChange={this.filterPosts} />
 
-        { map(posts, (post, key) => post.status === this.state.filter ? <InboxItem allowReview={false} onUpdate={this.updatePost} onTag={this.updateTag} showHistory={this.showHistory} deletePost={this.deletePost} key={key} details={{post: post, campaign: campaign, signup: post.signup.data}} /> : null) }
+        { map(posts, (post, key) => post.status === this.state.filter ? <InboxItem allowReview={false} onUpdate={this.updatePost} onTag={this.updateTag} showHistory={this.showHistory} deletePost={this.deletePost} key={key} post={post} campaign={campaign} signup={post.signup.data} /> : null) }
 
         <ModalContainer>
             {this.state.displayHistoryModal ? <HistoryModal id={this.state.historyModalId} onUpdate={this.updateQuantity} onClose={e => this.hideHistory(e)} details={{post: posts[this.state.historyModalId], campaign: campaign, signups: this.state.signups}}/> : null}
--- a/resources/assets/components/CampaignSingle/index.js
+++ b/resources/assets/components/CampaignSingle/index.js
@@ -134,6 +134,7 @@ class CampaignSingle extends React.Component {
         <ModalContainer>
             {this.state.displayHistoryModal ? <HistoryModal id={this.state.historyModalId} onUpdate={this.updateQuantity} onClose={e => this.hideHistory(e)} details={{post: posts[this.state.historyModalId], campaign: campaign, signups: this.state.signups}}/> : null}
         </ModalContainer>
+
         <PagingButtons prev={this.state.prevPage} next={this.state.nextPage}></PagingButtons>
       </div>
     )
--- a/resources/assets/components/CampaignSingle/index.js
+++ b/resources/assets/components/CampaignSingle/index.js
@@ -24,16 +24,15 @@ class CampaignSingle extends React.Component {
   }
 
   componentDidMount() {
-    this.api.get('api/v2/activity', {
+    this.api.get('api/v2/posts', {
       filter: {
         status: 'accepted',
         campaign_id: this.props.campaign.id,
       },
-      include: 'user',
+      include: 'signup,siblings',
     })
     .then(json => this.setState({
-      signups: keyBy(json.data, 'signup_id'),
-      posts: extractPostsFromSignups(json.data),
+      posts: keyBy(json.data, 'id'),
       filter: 'accepted',
       postTotals: json.meta.pagination.total,
       displayHistoryModal: null,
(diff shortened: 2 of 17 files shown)


chloealee:
@DFurnes looks like the wercker tests that are failing have to do with the activity tests - maybe I'll rebase after you merge #347 to see if they're still failing and if so, will fix!

app/Http/Controllers/Api/ActivityController.php
```diff
@@ -27,9 +27,22 @@ public function index(Request $request)
...
+        if (array_key_exists('status', $filters)) {
+            // Constrain Eager Load, only returning posts with status query.
+            $query = $query->with(['posts' => function ($query) use ($filters) {
+                $query->where('status', $filters['status']);
+            }]);
+
+            // Remove status from $filters to query Signups.
+            unset($filters['status']);
+        } else {
+            $query->with('posts');
```
DFurnes: Make sure you save the updated query here: `$query = $query->with('posts')`

chloealee: oh oops! thanks for catching that!

chloealee:
@DFurnes looks like the tests are still failing :( When you have a chance, would you mind looking over with me?

app/Http/Controllers/Api/ActivityController.php
```diff
@@ -27,9 +27,22 @@ public function index(Request $request)
...
+        $query = $this->newQuery(Signup::class);
 
         $filters = $request->query('filter');
+
+        if (array_has($filters, 'status')) {
+            // Constrain Eager Load, only returning posts with status query.
+            $query = $query->with(['posts' => function ($query) use ($filters) {
+                $query->where('status', $filters['status']);
+            }]);
+
```
chloealee: @DFurnes we don't necessarily need this anymore since we are using the new `/posts` endpoint with siblings(!) but should I keep in case we need in the future? cc @sbsmith86

chloealee:
@DFurnes when I only run the `PostModelTest`s, everything passes. However, when I run all of them, I get this error: 

```
There was 1 error:

1) PostModelTest::testSiblingRelationship
ErrorException: Trying to get property of non-object

/home/vagrant/DoSomething.org/rogue/tests/Models/PostModelTest.php:49
```
have you seen this before?

app/Http/Controllers/Api/ActivityController.php
DFurnes: Hmm, I don't have strong opinions but if we don't think we'll need it, it's always fun to delete code. 🔥

resources/assets/components/InboxItem/index.js
```diff
@@ -22,8 +22,10 @@ class InboxItem extends React.Component {
     const post_id = post['id'];
     const signup = this.props.details.signup;
 
+    var posts = signup.posts ? signup.posts : post.siblings.data;
```
DFurnes: Hmm, this feels a little awkward with the implicit dependency on the `details` prop. I don't have any great suggestions off the top of my head, but perhaps something we can simplify a bit.

resources/assets/components/InboxItem/index.js
```diff
@@ -39,6 +41,22 @@ class InboxItem extends React.Component {
     const campaign = this.props.details.campaign;
     const signup = this.props.details.signup;
 
+    if (post['user']) {
```
DFurnes: Hmm, this conditional bums me out a little. What are the two different scenarios that we get here?

resources/assets/components/InboxItem/index.js
```diff
@@ -52,12 +70,12 @@ class InboxItem extends React.Component {
           </ul>
         </div>
         <div className="container__block -third">
-          {post['user'] ?
+          {post['user'] || signup.user ?
             <div>
-              <h2>{post['user']['first_name']} {post['user']['last_name']}, {calculateAge(post['user']['birthdate'])}</h2>
+              <h2>{first_name} {last_name}, {birthdate}</h2>
```
DFurnes: It might make sense to use the Northstar user's [`displayName`](https://github.com/DoSomething/gateway/blob/30565f68e35fe1051be0569c8292e643f9a8c499/src/Resources/NorthstarUser.php#L21-L36) here!

resources/assets/components/CampaignSingle/index.js
```diff
@@ -34,9 +23,48 @@ class CampaignSingle extends React.Component {
     this.filterPosts = this.filterPosts.bind(this);
   }
 
+  componentDidMount() {
+    this.api.get('api/v2/posts', {
+      filter: {
+        status: 'accepted',
+        campaign_id: this.props.campaign.id,
+      },
+      include: 'signup,siblings',
```
chloealee: You're right, it doesn't work yet :( But I agree - that looks a lot better! Should I great an issue for that in Gateway?

resources/assets/components/InboxItem/InboxTile.js
```diff
@@ -8,7 +8,7 @@ class InboxTile extends React.Component {
 
     return (
       <li>
-        <img src={getImageUrlFromProp(post)}/>
+        <img src={getImageUrlFromProp(post) ? getImageUrlFromProp(post) : post.media['url']}/>
```
chloealee: ooo yes I like that a lot better!

resources/assets/components/CampaignSingle/index.js
DFurnes: That's something we could probably fix in the `TransformsRequests` trait here!

DFurnes: Let's dig into this together sometime this week (and definitely doesn't block merging).

sbsmith86:
> I think that this breaks pagination? Is this ok for now and we can make another issue to fix this?

We have an issue for this already. I pulled it into this sprint for myself to work on because I had some bandwidth but was waiting on some of your work to proceed. Since you are pulling in paginated posts we just need to get the pagination buttons hooked up. https://www.pivotaltracker.com/story/show/149809548

chloealee:
ok awesome, thanks @sbsmith86 !

resources/assets/components/CampaignSingle/index.js
chloealee: ok awesome!

resources/assets/components/InboxItem/index.js
chloealee: @DFurnes the first scenario with `post['user']` is for populating the inbox items for the Campaign Inbox page. It looks like there is a TODO [here](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/CampaignsController.php#L64-L75) suggesting to grab the user with the signup which would actually solve this problem and then we wouldn't have to do a conditional! Should I work on that or is that outside of the scope of this PR?

resources/assets/components/CampaignSingle/index.js
```diff
@@ -34,9 +23,48 @@ class CampaignSingle extends React.Component {
...
+      postTotals: json.meta.pagination.total,
+      displayHistoryModal: null,
+      historyModalId: null,
+      nextPage: json.meta.pagination.links.next ? json.meta.pagination.links.next : null,
+      prevPage: json.meta.pagination.links.previous ? json.meta.pagination.links.previous : null,
+    }));
+  }
+
   // Filter posts based on status.
   filterPosts(status) {
```
chloealee: ah yes! updated :)

resources/assets/components/InboxItem/index.js
chloealee: ohh I like this idea! Especially since we were talking about how it's a little weird to have a space between first name and comma if there is no last name. Would I have to make a call to gateway to get this?

resources/assets/components/InboxItem/index.js
chloealee: @DFurnes would it be better to pass props like this: 
`key={key} post={post} campaign={campaign} signup={post.signup.data}` instead of `details={{post: post, campaign: campaign, signup: this.state.signups[post.signup_id]}}` [here](https://github.com/DoSomething/rogue/blob/master/resources/assets/components/CampaignSingle/index.js#L109)?

DFurnes: Yeah I think that's a great idea 👍 👍

resources/assets/components/InboxItem/index.js
DFurnes: Got it! Nah, let's add a TODO referencing that & if you're feeling up to it we can always tackle in another PR afterwards.

resources/assets/components/InboxItem/index.js
DFurnes: It should already be available on the user response that's returned from Gateway in each case, but just not automatically bundled when using `toArray()`. So you should be able to attach it as another field in the response [here](https://github.com/DoSomething/rogue/blob/ac76c91a7607afd55adee9f523270ee3c3cc9561/app/Http/Controllers/CampaignsController.php#L75) and [here](https://github.com/DoSomething/rogue/blob/ac76c91a7607afd55adee9f523270ee3c3cc9561/app/Http/Controllers/CampaignsController.php#L108):

```diff
- $item->user = $user->toArray() ?: ['first_name' => 'A Doer'];
+ $item->user = [
+   'displayName' => $user->displayName(),
+   'birthdate' => $user->birthdate,
+   // other fields you want to include...
+ ];
```

chloealee: ahh I see - so cool! For the [second example](https://github.com/DoSomething/rogue/blob/ac76c91a7607afd55adee9f523270ee3c3cc9561/app/Http/Controllers/CampaignsController.php#L108), we're not actually passing `$signups` anymore since this PR grabs it via API call. Is it ok to do for the first (in the `showInbox` function and then when rendering the user information, doing a conditional like `if (displayName)` use that info and if not get the user info another way?

DFurnes: Oh right! For the signup API call, you can add the `$user->displayName` to [the transformer](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/UserTransformer.php).

chloealee: when I tried to do the above for the Campaign Inbox page (`showInbox` function) and logged out the user, I got 
```
Object
  birthdate: "1989-04-12"
  displayName: "5978afc27f43c27d8755558c"
```
which seemed to grab the user ID instead of display name?

DFurnes: And actually it might be nice to use the transformer here too so that we only deal with formatting user responses in one place:

```
$item->user = new UserTransformer($user);
```

DFurnes: Ahhhhhhhh yeah if we don't have a last name for the user [we return their ID](https://github.com/DoSomething/gateway/blob/30565f68e35fe1051be0569c8292e643f9a8c499/src/Resources/NorthstarUser.php#L35). That's maybe something to update in Gateway. 😖

DFurnes: Let's break this out into another PR since it's getting more involved!

resources/assets/components/CampaignSingle/index.js
```diff
@@ -91,27 +84,44 @@ class CampaignSingle extends React.Component {
     this.hideHistory();
   }
 
+  // Make API call to GET /posts to get posts by filtered status.
+  getPostsByStatus(status, campaignId) {
+  this.api = new RestApiClient;
```
DFurnes: Sorry one last little thing: looks like this method isn't indented! 👾

resources/assets/components/InboxItem/index.js
chloealee: ok sounds good! I will make a tiny card for this :)

resources/assets/components/CampaignSingle/index.js
chloealee: oh oops - thank you! Updated :)

chloealee:
oof @DFurnes still getting [this error](https://github.com/DoSomething/rogue/pull/348#issuecomment-321347297) as before :-/

DFurnes:
That's frustrating! 😦 Can you try dropping a `dd()` above that line to see what's going wrong?

chloealee:
I'm getting a new error now than I was yesterday! It doesn't even seem to get to the Post Model Test: 
```
PHP Fatal error:  Class 'FakerNorthstarId' not found in /home/vagrant/DoSomething.org/rogue/database/factories/ModelFactory.php on line 47
```
I tried to include that at the top of `ModelFactory.php` but that didn't seem to do anything...

DFurnes:
@chloealee Try running `composer dump-autoload`!

chloealee:
@DFurnes thank you! I updated the tests so it should be working now! The problem was that we assumed it the Signups that were being created were with `id`s 1-5 but they were actually with `id`s in the upper 70s. Just changed this so we don't assume!

PR #349, opened 2017-08-09
Allow storage driver to be overridden & tidy up post test.

DFurnes:
#### What's this PR do?
This pull request adds some improvements to how we deal with storage:

🚗 Allows the "storage driver" to be overridden via `STORAGE_DRIVER` (still defaults to S3). This means we can easily switch to using the local filesystem for development, and everyone's dev machine doesn't need to have S3 credentials and a shared bucket!

🔗 Extracts the logic for getting a post's URL into the model... this DRYs this up a bit, and also swaps it to use `Storage::url` for generating the URL so it isn't tightly coupled to S3.

🚥 Updates the tests for the `POST api/v2/posts` for readability, and adds a test case for when a post is being added to an existing signup.

One little difference – the storage driver provides URLs using the S3 bucket subdomain, so we should test on Thor and Production to make sure that's configured correctly and publicly accessible.

```diff
- https://s3.amazonaws.com/ds-rogue-prod/uploads/reportback-items/edited_5.jpeg
+ https://ds-rogue-prod.s3.amazonaws.com/uploads/reportback-items/edited_5.jpeg
```

In the future, we can add some more logic to check for and handle non-S3 URLs on posts as well.

This should make local development less painful & also sets us up for fun things with [Glide](http://glide.thephpleague.com)!

#### Relevant tickets
[#149662237](https://www.pivotaltracker.com/story/show/149662237) / [#149661097](https://www.pivotaltracker.com/story/show/149661097)

#### Checklist
- [ ] Tested on staging.


katiecrane:
Won't it still try to save it in AWS regardless [on this line](https://github.com/DoSomething/rogue/blob/master/app/Repositories/PostRepository.php#L61)?

DFurnes:
We should really rename that class. It uses the [Storage facade](https://github.com/DoSomething/rogue/blob/9d66f70ab455662108eef3a0ec1d11917eb0ed2c/app/Services/AWS.php#L42-L43) internally, so it depends on the configured driver.

katiecrane:
@DFurnes Ah! I was tricked!

tests/Http/Api/PostApiTest.php
```diff
@@ -1,57 +1,126 @@
...
-            'quantity'         => $this->faker->numberBetween(10, 1000),
-            'why_participated' => $this->faker->paragraph(3),
+        $northstar_id = $this->faker->uuid;
+        $campaign_id = $this->faker->randomNumber(4);
+        $campaign_run_id = $this->faker->randomNumber(4);
+        $quantity = $this->faker->numberBetween(10, 1000);
+        $caption = $this->faker->sentence;
+
+        // Mock file system operations.
+        $url = 'https://s3.amazonaws.com/ds-rogue-prod/uploads/reportback-items/edited_1.jpeg';
```
katiecrane: should this URL use the subdomain?

DFurnes: Oh yeah, that's a good call.

DFurnes: Updated in 59f2715.

katiecrane: @DFurnes looks like you updated the other spot but not this one? 😬

DFurnes: Oy...! Okay, really updated in b8b4333.

app/Models/Post.php
```diff
@@ -72,4 +73,19 @@ public function reactions()
...
+     * Get the URL for the media for this post.
+     */
+    public function getMediaUrl()
+    {
+        if ($this->url === 'default') {
+            return 'default';
+        }
+
+        // Ask the storage driver for the path to the image for this post.
+        $path = Storage::url('uploads/reportback-items/edited_' . $this->id . '.jpeg');
```
sbsmith86: Should we add a `@TODO` here to update this lookup when we don't have to rely on `editied_` being in the url? could be something we forget. But totally not needed! Whatever you think is best

tests/Http/Api/PostApiTest.php
```diff
@@ -1,57 +1,126 @@
...
 
+use Rogue\Models\Signup;
+use Illuminate\Support\Facades\Storage;
+
 class PostApiTest extends TestCase
 {
     /**
-     * Test that a POST request to /posts creates a new photo post.
+     * Test that a POST request to /posts creates a new
+     * photo post and signup, if it doesn't already exist.
```
sbsmith86: I debate whether we should, in fact, create a signup on the fly if there isn't one or if we should not allow creating posts if there isn't a signup already. Perhaps, something to think about for the future.

app/Models/Post.php
DFurnes: Yeah I'll do that!

tests/Http/Api/PostApiTest.php
DFurnes: Mmm yeah, I'm totally cool with rethinking this.

PR #350, opened 2017-08-10
Spruce up database seeder & factories.

DFurnes:
#### What's this PR do?
This pull request spruces up the database seeder/model factories so local development is more pleasant! This means we'll be able to simply run `php artisan db:seed` and get a representative set of sample data, rather than having to send around dumps of the QA database.

🏭 I did a pass over the data we were seeding, and moved it all into `DatabaseSeeder` since all of our tables are interconnected (you can't really have reviews without posts, or posts without signups, etc.).

📷 I added some images to "upload" when creating models via our factories. The reason I took this approach rather than Faker's [LoremPixel](http://lorempixel.com/) generator was because it's [significantly faster](https://github.com/DoSomething/voting-app/pull/326) (and more stable!) to pull images from the local disk rather than downloading each one.

🤺 I added a little fallback in the CampaignsController when a user cannot be found in Northstar since we're seeding with random IDs. I have another idea for making this better that I'll push in a separate PR!

Give it a once-over, and if you're up for it try pulling this branch and running locally!

🌱

#### Relevant tickets

#### Checklist
- [ ] Tested on staging.


katiecrane:
The seeders look awesome but it looks like a bunch of tests are failing because `Error: Class 'FakerNorthstarId' not found`

DFurnes:
Oh no, why is that still happening?! 😭 I thought I had removed that from this PR.

DFurnes:
Alrighty, should be good after 084a8e7 finishes building.

katiecrane:
Looks like now it's getting a couple 500s in the tests 😩

DFurnes:
Ahh, I hadn't updated the `mockFile` test helper to point to the moved directory. bfc29f5

PR #355, opened 2017-08-11
Part 1: User search

sbsmith86:
#### What's this PR do?

Going to start splitting up user search into different PRs. This PR just implements basic user-search functionality and takes you to a user page that will eventually list out the campaigns that the user is signed up for. 

The search functionality is ported over from aurora/gladiator. Nothing super fancy or react-heavy happening here, we just make a call to Northstar to find the user based on the given input. Also note, this doesn't handle multiple user results well at all. So I will make a note to tackle that separately.

A gif of it: 

user-page-ex

I think you can go commit-by-commit but the order might feel a little off. The big things to note are new routes to direct folks to the right pages, new react components to handle the page display, and some little helpers to display things nicely.

Getting a head start on getting user search up and running. 

Also an aside, I think we need to really start thinking about how to addressing the `app.js` entry point and how we have to conditionally look for containers on a page to know which components to load in, it's getting noisy in that file. 

#### Relevant tickets
https://www.pivotaltracker.com/story/show/150071802

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Services/Registrar.php
+++ b/app/Services/Registrar.php
@@ -122,7 +122,7 @@ class Registrar
      * @param  int $page
      * @return collection|null
      */
-    public function search($query, $page = null)
+    public function search($query, $page = 1)
     {
         // Attempt to fetch all users.
         $users = $this->northstar->getAllUsers([
@@ -132,9 +132,9 @@ class Registrar
                 'email' => $query,
                 'mobile' => $query,
             ],
-            'page' => is_null($page) ? 1 : $page,
+            'page' => $page,
         ]);
 
-        return $users ? collect($users) : null;
+        return $users;
     }
 }
--- a/app/Services/Registrar.php
+++ b/app/Services/Registrar.php
@@ -114,4 +114,27 @@ class Registrar
 
         return collect($data);
     }
+
+    /**
+     * Search for users in Northstar.
+     *
+     * @param  string $query
+     * @param  int $page
+     * @return collection|null
+     */
+    public function search($query, $page = null)
+    {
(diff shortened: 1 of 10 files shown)


resources/assets/helpers.js
```diff
@@ -70,3 +70,34 @@ export function getEditedImageUrl(photoProp) {
...
+ */
+export function displayName(firstName, lastName) {
+  let displayName = firstName;
+
+  if (lastName) {
+    displayName = `${displayName} ${lastName}`;
+  }
+
+  return displayName;
+}
```
chloealee: @sbsmith86 I think this works great for what we need it for!

app/Services/Registrar.php
```diff
@@ -114,4 +114,27 @@ protected function getBatchedCollection($ids, $size = 50)
...
+    {
+        // Attempt to fetch all users.
+        $users = $this->northstar->getAllUsers([
+            'search' => [
+                '_id' => $query,
+                'drupal_id' => $query,
+                'email' => $query,
+                'mobile' => $query,
+            ],
+            'page' => is_null($page) ? 1 : $page,
```
DFurnes: Can we set `$page = 1` as the default parameter value above?

app/Services/Registrar.php
```diff
@@ -114,4 +114,27 @@ protected function getBatchedCollection($ids, $size = 50)
...
+        ]);
+
+        return $users ? collect($users) : null;
```
DFurnes: Is it possible to just return `$users` here?

PR #356, opened 2017-08-11
Add Glide for on-demand image processing.

DFurnes:
#### What's this PR do?
This pull request adds [Glide](http://glide.thephpleague.com) to allow us to create cropped images for posts on-demand. This removes the need for costly image processing when a user uploads an image, and means we can ask for differently formatted images in different places (perhaps small square images for a gallery, but a larger un-cropped image on a permalink page).

If you pull this branch down to your local and add `DS_ENABLE_GLIDE=true` to your `.env`, you can experiment [with the included filters](http://glide.thephpleague.com/1.0/api/quick-reference/). For example:

```php
'http://rogue.dev/images/1' // default small crop
'http://rogue.dev/images/1?w=500&h=500&fit=crop' // larger square crop
'http://rogue.dev/images/1?w=250&h=400&fit=crop&filt=sepia' // woah, crazy!
```

When we want to enable this, we can toggle the `DS_ENABLE_GLIDE` feature flag. We'll want to make sure Fastly is properly caching images & do some thorough testing on Thor first.

#### Relevant tickets
[#149661097](https://www.pivotaltracker.com/story/show/149661097)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -80,6 +80,7 @@ class Post extends Model
     public function getMediaUrl()
     {
         if ($this->url === 'default') {
+            // @TODO: We should either return a placeholder, or `null`.
             return 'default';
         }
 
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -97,11 +97,24 @@ class Post extends Model
 
     /**
      * Return the filesystem path for this post.
+     *
+     * @return string|null
      */
     public function getMediaPath()
     {
-        $path = str_replace('/storage//', '', $this->url);
+        // We store local URLs prefixed with `/storage` when using the "public" adapter.
+        if (config('filesystems.default') == 'public') {
+            return str_replace('/storage/', '', $this->url);
+        }
 
-        return $path;
+        // For S3, we store a full AWS URL, sometimes prefixed by bucket directory.
+        if (config('filesystems.default') == 's3') {
+            $path = parse_url($this->url, PHP_URL_PATH);
+            $path = str_replace('/'. config('filesystems.disks.s3.bucket') . '/', '', $path);
+
+            return $path;
+        }
+
+        return null;
     }
 }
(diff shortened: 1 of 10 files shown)


sbsmith86:
Do we need these filters on the `GET /posts` endpoint?

app/Models/Post.php
```diff
@@ -80,13 +80,41 @@ public function reactions()
...
+    /**
+     * Return the filesystem path for this post.
+     *
+     * @return string|null
+     */
+    public function getMediaPath()
+    {
+        // We store local URLs prefixed with `/storage` when using the "public" adapter.
+        if (config('filesystems.default') == 'public') {
+            return str_replace('/storage/', '', $this->url);
```
katiecrane: Won't this remove `/storage/` from the url wherever it occurs? I read this as wanting to append `/storage` to the front?

DFurnes:
This should be set up already, since the [PostTransformer](https://github.com/DoSomething/rogue/blob/c2ed52e6a3bd81fba190fe358cca460e53a56100/app/Http/Transformers/PostTransformer.php#L34) (which `GET /posts` uses) and [ReportbackTransformer](https://github.com/DoSomething/rogue/blob/c2ed52e6a3bd81fba190fe358cca460e53a56100/app/Http/Transformers/ReportbackTransformer.php#L26) use the `getMediaUrl` method to fetch the image URL.

DFurnes:
Oy, mis-clicked the "Close and comment" button. 💥

app/Models/Post.php
DFurnes: Yep, it's a little wonky! Since `/storage` is the "base path" for the storage adapter, we need to remove it when asking for a file path (but add it when giving a URL to the user).

DFurnes: One thing that could make this a little simpler would be to store local paths, but use full URLs for remote things (like a Spotify playlist or YouTube video, in the future).

katiecrane: Ahhhh I see!

PR #359, opened 2017-08-15
List user signups

sbsmith86:
#### What's this PR do?
This PR flushes out the User page showing all of a users signups. 

* Implements layout, style, and functionality for a `SignupCard` component that show info for each signup. 
  - One thing to note: we grab the campaigns for each signup after the signups have been set on the state. We don't render signups on the page until everything is available, in the meantime we display a spinner indicating we are grabbing data.
* Creates a proxy endpoint at `/api/v2/campaigns` that grabs campaign info from the phoenix api 

image

I think i messed up a lot of the commit history when rebasing since I started this branch off of a branch that wasn't merged in yet. 

Best bet is to pull down this branch and test that you can load a user page and see data on the page. 

Most of the code for the layout and functionality is in `UserOverview/index.js` and `Api/CampaignsController.php`

Styling work is done in `signup-card.scss`

Part 2 of user search work. 

#### Relevant tickets

https://www.pivotaltracker.com/story/show/150173773

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/SignupCard/signup-card.scss
+++ b/resources/assets/components/SignupCard/signup-card.scss
@@ -1,9 +1,5 @@
 @import '~@dosomething/forge/scss/toolkit';
 
-$tablet: "(min-width: 760px)" !default;
-$base-spacing: 24px !default;
-$light-gray: #ddd !default;
-
 .signup-card {
   border: 2px solid $light-gray;
 
--- a/resources/assets/components/SignupCard/signup-card.scss
+++ b/resources/assets/components/SignupCard/signup-card.scss
@@ -1,0 +1,34 @@
+@import '~@dosomething/forge/scss/toolkit';
+
+$tablet: "(min-width: 760px)" !default;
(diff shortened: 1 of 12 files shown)


resources/assets/components/SignupCard/signup-card.scss
```diff
@@ -0,0 +1,38 @@
+@import '~@dosomething/forge/scss/toolkit';
+
+$tablet: "(min-width: 760px)" !default;
+$base-spacing: 24px !default;
+$light-gray: #ddd !default;
```
DFurnes: Hmm, these should all exist in [Forge's variables](https://github.com/DoSomething/forge/blob/3828463212a96e97b154b0afa007da30b869c5ab/scss/_utilities/_variables.scss) that get imported from the toolkit.

sbsmith86: oh yea hmm not sure why i had that there. Good catch!

Issue #353, opened 2017-08-10
Tag structure


katiecrane:
_Thinkin' 'bout taggin'_
-------------------
#### Current Behavior
Right now, we are using [rtconner/laravel-tagging](https://github.com/rtconner/laravel-tagging).

#### Desired Behavior
We'd like to calm things down a bit. Perhaps using a regular ol' `hasMany` relationship.

#### Steps to Replicate
Hit this endpoint `api/v1/reportbacks`. Screenshots below are from this endpoint.

#### Why This Matters
We are running into issues where this makes some yucky sequel queries. 

For example, this is to not include posts tagged as `Hide In Gallery` in the overall results.

![image](https://user-images.githubusercontent.com/4240292/29184977-e2980f84-7dd5-11e7-90f2-d3dd01890359.png)

And this looks like it pulls the details of the tag for each post?

![image](https://user-images.githubusercontent.com/4240292/29185046-16b1751c-7dd6-11e7-84e1-d8100018569a.png)

katiecrane:
I played around locally with using a plain ol' pivot table. I tagged all the posts I have locally except one as "Hide In Gallery" and this is what runs when you hit the same endpoint as above:

![image](https://user-images.githubusercontent.com/4240292/29279756-e0832206-80e6-11e7-8182-3b1586b2e204.png)

The good news is that the big nasty query is gone!

katiecrane:
Update: this query will only ever have 20 post ids because of the pagination, so I don't think we need to worry about it.

![image](https://user-images.githubusercontent.com/4240292/29286945-d2048bf0-8101-11e7-8611-7414090c5593.png)

ngjo:
**Meeting Recap**
We met as a team and it sounds like we're all on board to move away from the tagging package and do our own thing. 

The reason we went with the package in the first place was because we (1) weren't sure what the tags would be (2) it did offer some good solutions at the time for tagging and (3) we were moving quickly to launch. Now, once we've had a chance to see how the tags system works and how we want it to work as we continue to develop new features, the package doesn't necessarily serve us well anymore.

Briefly, here are the next steps that we've broken down into cards and will be tackled during the 8/14 sprint:

1. Make new tables and relationships
2.  Update code that references old tag stuff w/ empty placeholders so the code doesn't error out + (killconner)
3. Go back to empty placeholders and fix it as you would expect it to to work and write unit tests
4. Update posts tags endpoint to work with the new tags 
[...]

PR #362, opened 2017-08-15
Add Tag model, relationship with posts, table and pivot table

katiecrane:
#### What's this PR do?
This adds new things, but doesn't change how anything works currently.

Added:
- Tag model
- Database migration to add tags table and pivot table with posts. Includes timestamps only on the Tags table and not on the pivot table. On the current tagging tables (`tagging_tagged`, `tagging_tags`) we do not have timestamps either.
- Posts relationship with Tag model

Where do we want timestamps? Nowhere? Both tables?

Background discussion can be found in this issue: https://github.com/DoSomething/rogue/issues/353.

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/n/projects/2019429/stories/150239872)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -75,6 +75,14 @@ class Post extends Model
     }
 
     /**
+     * Get the tags associated with this post.
+     */
+    public function tags()
+    {
+        return $this->belongsToMany(Tag::class);
+    }
+
+    /**
      * Get the URL for the media for this post.
      */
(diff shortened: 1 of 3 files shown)


DFurnes:
> Where do we want timestamps? Nowhere? Both tables?

I guess it doesn't hurt to add to the `post_tag` pivot table.

PR #365, opened 2017-08-15
Remove old tagging package & add new tagging functionality

katiecrane:
#### What's this PR do?
Bye bye rtconner/laravel-tagging! This PR removes that package and (I think) all references to it. So that tests (excluding the ones that specifically test tagging) don't fail, it adds placeholder functions and does not return tags with posts in some places.

PART 2:
- Adds back in all tagging functionality. Should work normally via admin review pages and the API.
- New Tags test, `testAddMultipleTagsAndDeleteOne`
- New migration to move the existing applied tags to the `post_tag` table

Are the tests passing? Any remnants of `laravel-tagging` left anywhere?

PART 2:
Does tagging work from the inbox? Does tagging work from the API? Are the right events being created when you tag? Are the right things getting their timestamps updated?

This is the first PR of 2 to keep them smaller. However, this leaves us in an intermediate state of no tags, so we can't deploy this until the second part is done. Next step is to add in tagging functionality so we can return tags with posts, and also tag posts to begin with. I'll also be re-working the tag tests so those can run too!

You will need to run `composer update` to remove the other `laravel-tagging` files locally once you pull this in.

PART 2:
Since part 2 relied on the changes in part 1, it's coming in as the second round of this PR. The old tagging tables will be removed in a future PR so we still have all the necessary data if we end up needing to revert this for some reason.

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/story/show/150239901)
[2nd Pivotal card](https://www.pivotaltracker.com/story/show/150240118)
[3rd Pivotal card](https://www.pivotaltracker.com/story/show/150240268)

#### Checklist
- [ ] Tested on staging.


katiecrane:
@chloealee omg YES! Totally missed those!

chloealee:
@katiecrane you got all the other important stuff tho!

katiecrane:
@chloealee after further work on that, I'm going to wait to drop those tables until the tagging functionality is more re-built. I need that working in order to move the `tagging_tagged` table stuff over to the `post_tag` table (I could just manually insert it all but then I have to put the current time into `created_at` and `updated_at` and it seems better to me to let Laravel handle that part)

chloealee:
that makes sense to me @katiecrane ! Sorry to derail you!

katiecrane:
@chloealee no, thank you! I forgot it was something I needed to look into!

app/Models/Tag.php
```diff
@@ -3,8 +3,11 @@
 namespace Rogue\Models;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\SoftDeletes;
 
 class Tag extends Model
 {
-    //
+    use SoftDeletes;
```
DFurnes: What's the benefit of using soft deletion here?

composer.json
```diff
@@ -9,7 +9,8 @@
     "laravel/tinker": "^1.0",
     "aws/aws-sdk-php-laravel": "^3.1",
     "barryvdh/laravel-debugbar": "^2.3",
-    "doctrine/dbal": "^2.5",
+    "doctrine/dbal": "~2.5.13",
+    "doctrine/common": "~2.6.0",
```
DFurnes: We can remove this `doctrine/common` dependency.

tests/Http/TagsTest.php
```diff
@@ -27,7 +30,7 @@ public function testAddingATagToAPost()
         $this->assertResponseStatus(200);
 
         // Make sure that the post's tags are updated.
-        $this->assertContains('Good Photo', $post->tagNames());
+        $this->assertContains('Good Photo', $post->tags()->pluck('tag_name'));
```
DFurnes: Since we do same thing so often, it might be nice to make a `tagNames()` method on the Post using the same implementation.

tests/Http/TagsTest.php
```diff
@@ -58,7 +61,39 @@ public function testDeleteTagOnAPost()
...
+
+    /**
+     * Test deleting one tag on a post only deletes that tag
+     *
+     * POST /tags
+     * @return void
+     */
+    public function testAddMultipleTagsAndDeleteOne()
+    {
+        // @TODO: Tag model event fails if no authenticated user.
```
DFurnes: Is this still true, since we've removed the tag model events?

app/Models/Post.php
```diff
@@ -157,4 +155,39 @@ public function toBlinkPayload()
...
+
+    /**
+     * Apply the given tag to this post.
+     */
+    public function tag($tagName)
+    {
+        $tag = Tag::where('tag_name', $tagName)->first();
+
+        // Create the tag if it doesn't already exist (so we don't have to create tags separately)
+        if (! $tag) {
```
katiecrane: @DFurnes OH YES this is excellent I'm changing it to this

app/Models/Post.php
```diff
@@ -157,4 +155,39 @@ public function toBlinkPayload()
...
+
+        return $this;
+    }
+
+    /**
+     * Remove the given tag from this post.
+     */
+    public function untag($tagName)
+    {
+        $tag = Tag::where('tag_name', $tagName)->first();
```
katiecrane: @DFurnes I think we eventually want to use the slug for everything, but I didn't change to that here to not change too much at once. The problem with that is when a new tag is created, could potentially be weird getting the name from the slug?

app/Models/Tag.php
katiecrane: @DFurnes For some reason I thought we used that in the old system, but turns out we didn't! I don't think there's a use for it on the tags themselves, I'm removing it.

tests/Http/TagsTest.php
katiecrane: Nope! This is a leftover comment that got copied from the previous test.

app/Models/Post.php
katiecrane: @DFurnes I made that setter but I wasn't sure how it gets called? Right now I have this:
`        $tag = Tag::firstOrCreate(['tag_name' => $tagName], ['tag_slug' => str_slug($tagName, '-')]);`
which looks for the tag with the given name, and if it's not found it creates the tag with the given name and slug. Does it make sense to do it that way?

composer.json
katiecrane: @DFurnes what changed that made us not need that?

app/Models/Post.php
DFurnes: Very fair!

composer.json
DFurnes: We added that when we were debugging the Composer weirdness. It's a dependency of `doctrine/dbal` so shouldn't be necessary here.

katiecrane: @DFurnes ahhhh right this didn't end up being part of the solution!

katiecrane: @DFurnes Removed `doctrine/common`, ran `composer update` and pushed up the new lock file, getting this in wercker again:
```
 Problem 1
    - Installation request for doctrine/annotations v1.5.0 -> satisfiable by doctrine/annotations[v1.5.0].
    - doctrine/annotations v1.5.0 requires php ^7.1 -> your PHP version (7.0.21) does not satisfy that requirement.
  Problem 2
    - Installation request for doctrine/cache v1.7.0 -> satisfiable by doctrine/cache[v1.7.0].
    - doctrine/cache v1.7.0 requires php ~7.1 -> your PHP version (7.0.21) does not satisfy that requirement.
  Problem 3
    - Installation request for doctrine/collections v1.5.0 -> satisfiable by doctrine/collections[v1.5.0].
    - doctrine/collections v1.5.0 requires php ^7.1 -> your PHP version (7.0.21) does not satisfy that requirement.
  Problem 4
    - Installation request for doctrine/instantiator 1.1.0 -> satisfiable by doctrine/instantiator[1.1.0].
    - doctrine/instantiator 1.1.0 requires php ^7.1 -> your PHP version (7.0.21) does not satisfy that requirement.
  Problem 5
    - doctrine/annotations v1.5.0 requires php ^7.1 -> your PHP version (7.0.21) does not satisfy that requirement.
    - doctrine/common v2.7.3 requires doctrine/annotations 1.* -> satisfiable by doctrine/annotations[v1.5.0].
    - Installation request for doctrine/common v2.7.3 -> satisfiable by doctrine/common[v2.7.3].
```

DFurnes: Hmm, let me swing by.

app/Http/Transformers/ReportbackTransformer.php
```diff
@@ -26,7 +26,7 @@ public function transform(Post $post)
                 'uri' => $post->getMediaUrl(),
                 'type' => 'image',
             ],
-            'tagged' => $post->tagNames(),
+            'tagged' => $post->tags()->pluck('tag_names'),
```
DFurnes: `$post->tagNames()` 🏷

tests/Http/TagsTest.php
```diff
@@ -83,4 +121,25 @@ public function testUnauthenticatedUserCantTag()
...
+    public function testPostTimestampUpdatedWhenTagAdded()
+    {
+        // Create the models that we will be using
+        $post = factory(Post::class)->create();
+
+        // Later, apply the tag to the post
+        $this->mockTime('10/21/2017 13:05:00');
+
+        $this->actingAsAdmin()->post('tags', [
+                'post_id' => $post->id,
```
DFurnes: Indentation looks a lil' wonky here! 👾

app/Http/Transformers/ReportbackTransformer.php
katiecrane: Ugh I kept trying to search all files for `tags()->pluck('tag_names')` in sublime text but it was saying there were none! Good find!!!

PR #370, opened 2017-08-17
Some little fixes to authentication middleware.

DFurnes:
#### What's this PR do?
This pull request includes a handful of little fixes made while working on the new [Gateway guard](https://github.com/DoSomething/gateway/pull/83):

🔑 Extracts `env()` call for the Rogue API key into a config value. We don't want `env()` calls outside of our config directory [because it means we can't cache those calls](https://laravel.com/docs/5.4/configuration#configuration-caching).

🍰 Uses Laravel's built-in `Authenticate` middleware (added to the framework in 5.3) instead of the one in our source, which handles multiple guards more elegantly & simplifies things a bit!

💥  Fixes a fatal error (we were previously not importing anything for `UnauthorizedAccessException` so it'd just 500), and adds protection to `DELETE` routes as well (since previously we were only protecting `POST` and `PUT`).

👀

Just trying to break things up into small PRs! 😄

#### Relevant tickets
[#150224528](https://www.pivotaltracker.com/story/show/150224528)

#### Checklist
- [ ] Tested on staging.


sbsmith86:
> adds protection to DELETE routes as well (since previously we were only protecting POST and PUT)

I wonder if this will fix the issue being reported on Kudos https://github.com/DoSomething/devops/issues/188#issuecomment-321866100

DFurnes:
Oh hmm yeah, it certainly won't hurt!

PR #371, opened 2017-08-18
Update post on campaign single page

chloealee:
#### What's this PR do?
- Allows reviewing (change quantity, tags, and status) of posts on the campaign single view page. 

- Visit campaign single view page. You should now be able to update any post on this page exactly how you're able to in the campaign inbox page. 

- Note: there are some duplicate functions from CampaignInbox `index.js` in Campaign Single `index.js` (`deletePost`, `updatePost`), and some that are almost the same but slightly tweaked because of the different states on the different pages (`updateQuantity`, `updateTag`). Once we update the CampaignInbox `index.js` to initally load all posts via API similar to the Campaign Single `index.js` view, we should update and combine these functions to live in one place for DRY code. 
- When a post is rejected and you change it to accepted, the post will immediately disappear since it's no longer under the `rejected` filter view. You won't have time to add tags to it unfortunately. Is this ok? cc @ngjo 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/150143243

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 /vendor
 /node_modules
 /public/storage
-/storage/debugar
+/storage/debugbar
 Homestead.yaml
 Homestead.json
 .env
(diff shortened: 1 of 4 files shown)


ngjo:
RE this:
> When a post is rejected and you change it to accepted, the post will immediately disappear since it's no longer under the rejected filter view. You won't have time to add tags to it unfortunately. Is this ok?

@chloealee i think as long as the post will show up if the user goes to the 'accepted' filter view, then I think it's totally fine and I can communicate it to the staff!

PR #374, opened 2017-08-21
Save campaigns in cache with prefix in ID

chloealee:
#### What's this PR do?
Store campaigns in cache with the prefix `campaign:` to fix this issue: 
```
[2017-08-21 14:48:25] thor.ERROR: ErrorException: Cannot use a scalar value as an array in /var/www/rogue/releases/20170821143532/app/Services/CampaignService.php:246
```

👀  and `/campaigns` and `/campaigns/:nid` pages should work exactly the same when hit. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/150409500

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/CacheRepository.php
+++ b/app/Repositories/CacheRepository.php
@@ -9,9 +9,9 @@ class CacheRepository
     /**
      * Instantiate a new CacheRepository.
      */
-    public function __construct()
+    public function __construct($prefix)
     {
-        $this->prefix = 'campaign';
+        $this->prefix = $prefix;
     }
 
     /**
--- a/app/Repositories/CacheRepository.php
+++ b/app/Repositories/CacheRepository.php
@@ -60,7 +60,7 @@ class CacheRepository
      * @param  string  $string
      * @return string
      */
-    public function applyPrefix($string)
+    protected function applyPrefix($string)
     {
         if (property_exists($this, 'prefix')) {
             return $this->prefix . ':' . $string;
--- a/app/Repositories/CacheRepository.php
+++ b/app/Repositories/CacheRepository.php
@@ -37,7 +37,7 @@ class CacheRepository
         $idsWithPrefix = [];
 
         foreach ($keys as $key => $id) {
-            array_push($idsWithPrefix, $this->setPrefix($id));
+            array_push($idsWithPrefix, $this->applyPrefix($id));
         }
 
         $retrieved = [];
@@ -55,12 +55,12 @@ class CacheRepository
     }
 
     /**
-     * Set a prefix on supplied string used as cache key.
+     * Apply a prefix on supplied string used as cache key.
      *
      * @param  string  $string
      * @return string
      */
-    public function setPrefix($string)
+    public function applyPrefix($string)
     {
         if (property_exists($this, 'prefix')) {
             return $this->prefix . ':' . $string;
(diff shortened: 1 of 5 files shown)


app/Repositories/CacheRepository.php
```diff
@@ -7,6 +7,14 @@
 class CacheRepository
 {
     /**
+     * Instantiate a new CacheRepository.
+     */
+    public function __construct()
+    {
+        $this->prefix = 'campaign';
```
sbsmith86: I don't think we want to hardcode this because we also [store user objects](https://github.com/DoSomething/rogue/blob/eb6d59a920611f5d022574660c0920182a55ebff/app/Services/Registrar.php#L16) in cache. The constructor could take in a `$prefix` and then set it when you instantiate it. 

So for example, here: 

Registrar.php
```
    public function __construct(Northstar $northstar, CacheRepository $cache)
    {
        $this->northstar = $northstar;
        $this->cache = $cache;
    }
```
You could do 

```    
public function __construct(Northstar $northstar)
    {
        $this->northstar = $northstar;
        $this->cache = new CacheRepository('user');
    }
```

app/Services/CampaignService.php
```diff
@@ -65,26 +65,34 @@ public function find($id)
     public function findAll($ids = [])
     {
         if ($ids) {
-            $campaigns = $this->cache->retrieveMany($ids);
+            $idsWithPrefix = [];
+
+            foreach ($ids as $key => $id) {
+                array_push($idsWithPrefix, $this->cache->setPrefix($id));
```
DFurnes: This should all be encapsulated within the `CacheRepository` (so when you use `retrieveMany` or `storeMany` it automatically prepends the prefix to the given IDs).

app/Repositories/CacheRepository.php
chloealee: ah so smart - thanks @sbsmith86 ! Updated :)

app/Services/CampaignService.php
chloealee: ah I see - so much cleaner!

app/Repositories/CacheRepository.php
```diff
@@ -26,16 +34,21 @@ public function retrieve($key)
      */
     public function retrieveMany(array $keys)
     {
+        $idsWithPrefix = [];
+
+        foreach ($keys as $key => $id) {
+            array_push($idsWithPrefix, $this->setPrefix($id));
```
DFurnes: Can you rename `setPrefix` to `applyPrefix` or something similar, and make it a protected method? At first glance, `setPrefix` reads as a setter method.

DFurnes: Actually can you also double-check this is actually replacing each ID in the array? Either way, might read a little more nicely as a `array_map`... 🗺

chloealee: yes totally!

chloealee:
@sbsmith86 @DFurnes oh wait! I just realized something - should I also update the `store` method and `retrieve` method in `CacheRepository.php` to use the same prefix logic?

DFurnes:
@chloealee Oh yes, good catch!

chloealee:
@DFurnes @sbsmith86 updated to include this logic in `store` and `retrieve` methods and also updated the tests (since some were failing!). Would you mind one more 👀 ?

PR #375, opened 2017-08-21
Transform and send signups & posts to Blink.

DFurnes:
#### What's this PR do?
This pull request adds support for persisting new signups & reportbacks in Customer.io, via Blink. This is handled using a feature flag, so that we can disable it when running Rogue locally.

I've added assertions that the relevant Blink methods are called in relevant test suites.

📮

#### Relevant tickets
[#149209459](https://www.pivotaltracker.com/story/show/149209459), [#149862583](https://www.pivotaltracker.com/story/show/149862583)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -115,7 +115,7 @@ class PostRepository
      * @param \Rogue\Models\Signup $signup
      * @param array $data
      *
-     * @return \Rogue\Models\Post
+     * @return Signup|Post
      */
     public function update($signup, $data)
     {
@@ -171,7 +171,7 @@ class PostRepository
      *
      * @param array $data
      *
-     * @return
+     * @return Post
      */
     public function reviews($data)
     {
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -7,7 +7,6 @@ use Rogue\Services\AWS;
 use Rogue\Models\Review;
 use Rogue\Models\Signup;
 use Rogue\Services\Registrar;
-use DoSomething\Gateway\Blink;
 use Intervention\Image\Facades\Image;
 
 class PostRepository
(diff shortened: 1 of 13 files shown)


app/Repositories/PostRepository.php
```diff
@@ -93,7 +113,13 @@ public function create(array $data, $signupId)
 
         // Edit the image if there is one
         if (isset($data['file'])) {
-            $editedImage = $this->crop($data, $post->id);
+            $this->crop($data, $post->id);
+        }
+
+        // Save the new post in Customer.io, via Blink.
+        if (config('features.blink')) {
```
sbsmith86: This type of logic would be perfect for the `[PostService`](https://github.com/DoSomething/rogue/blob/eb6d59a920611f5d022574660c0920182a55ebff/app/Services/PostService.php#L34) class which was meant to handle more business-type logic around creating posts so we don't conflate database concerns with service concerns. 

I know there was some concern about not needed all of this separation, but I think it makes sense now that we are adding other logic around creating posts instead of just storing in the DB.

app/Repositories/SignupRepository.php
```diff
@@ -40,6 +58,12 @@ public function create($data)
             $signup->save();
         }
 
+        // Save the new signup in Customer.io, via Blink.
+        if (config('features.blink')) {
```
sbsmith86: Same here, this could go in the `SignupService` class.

app/Repositories/PostRepository.php
DFurnes: Sure sure, I can make that switch.

DFurnes: Done in c210a92.

app/Repositories/SignupRepository.php
DFurnes: Done in c210a92.

PR #377, opened 2017-08-22
Some little test updates.

DFurnes:
#### What's this PR do?
This pull request is just a little grab-bag of test touch ups:

7️⃣ Adds a PHP version constraint to our `composer.json`. This will prevent us from accidentally running Composer with a different PHP version and getting the wrong dependencies.

🚥 Updates `PostTest` and `PostApiTest` to use the new Laravel 5.4 test syntax, including the [`UploadedFile` fake](https://laravel.com/docs/5.4/mocking#storage-fake) and [`assertSoftDeleted` assertion](https://laravel.com/docs/5.4/database-testing#available-assertions) (each replacing our own homegrown implementations).

🗄 Fakes the storage adapter for tests so we don't keep writing files to the disk every time we run the test suite (otherwise the `public/storage` directory keeps growing and growing).

👀

I'm gonna make a card to update the rest of our tests to use the new syntax, but this is a start! 😄

#### Relevant tickets
N/A 🙊

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
   "license": "MIT",
   "type": "project",
   "require": {
-    "php": ">=5.5.9",
+    "php": "~7.0.0",
     "laravel/framework": "5.4.*",
     "laravel/tinker": "^1.0",
     "aws/aws-sdk-php-laravel": "^3.1",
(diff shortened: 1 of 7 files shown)


DFurnes:
Ahhh sorry! Context: The test syntax changed in Laravel 5.4, so Taylor made this [browser-kit-testing](https://github.com/laravel/browser-kit-testing) package that gives compatibility with the old syntax. The downside of using that is we lose out on all the new goodies, and the documentation wouldn't be accurate.

katiecrane:
@DFurnes a-ha! That makes a lot of sense, thanks!

PR #381, opened 2017-08-24
Use new tags to not include excluded posts

katiecrane:
#### What's this PR do?
Instead of using the `withoutTags` scope from the old tagging package, use a Laravel method, `whereDoesntHave` .

👀 

I missed this on my first pass!

#### Relevant tickets
:cowboy_hat_face: 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/ReportbackController.php
+++ b/app/Http/Controllers/Api/ReportbackController.php
@@ -42,7 +42,9 @@ class ReportbackController extends ApiController
             // Only return posts that have been approved by a campaign lead...
             ->where('status', 'accepted')
             // ...and haven't been hidden from the gallery.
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Api/ReportbackController.php
```diff
@@ -42,7 +42,9 @@ public function index(Request $request)
             // Only return posts that have been approved by a campaign lead...
             ->where('status', 'accepted')
             // ...and haven't been hidden from the gallery.
-            ->withoutTags(['Hide In Gallery']);
+            ->whereDoesntHave('tags', function ($query) {
```
DFurnes: Is this something that'd be useful to make as a scope on the `Post` model?

PR #389, opened 2017-08-28
only updates why_participated if it is present

chloealee:
#### What's this PR do?
This first checks to see if `why_participated` is present before overriding the existing value in the database. 

Hit `POST /api/v2/posts` endpoint with `why_participated` value blank. Make sure it doesn't wipe the value out in the database and that the `why_participated` value returned in the JSON is from the original answer. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/150341552

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -120,7 +120,11 @@ class PostRepository
     public function update($signup, $data)
     {
         if (array_key_exists('updated_at', $data)) {
-            $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'why_participated', 'updated_at']));
+            if ($data['why_participated']) {
+                $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'why_participated', 'updated_at']));
+            } else {
+                $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'updated_at']));
+            }
 
             $signup->save(['timestamps' => false]);
 
@@ -129,15 +133,14 @@ class PostRepository
             $event->updated_at = $data['updated_at'];
             $event->save(['timestamps' => false]);
         } else {
-            $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'why_participated']));
+            if ($data['why_participated']) {
+                $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'why_participated']));
+            } else {
+                $signup->fill(array_only($data, ['quantity', 'quantity_pending']));
+            }
 
             // Triggers model event that logs the updated signup in the events table.
             $signup->save();
-        }
-
-        // If there is a file, create a new post.
-        if (array_key_exists('file', $data)) {
-            return $this->create($data, $signup->id);
         }
 
         return $signup;
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -121,17 +121,8 @@ class PostRepository
     {
         if (array_key_exists('updated_at', $data)) {
             // Only update if the key is set (is not null).
-            $arrayKeysToUpdate = ['why_participated', 'quantity'];
-
-            foreach ($arrayKeysToUpdate as $key => $value) {
-                if (! $data[$value]) {
-                    $key = array_search($value, $arrayKeysToUpdate);
-                    unset($arrayKeysToUpdate[$key]);
-                }
-            }
-
-            array_push($arrayKeysToUpdate, 'updated_at');
-            array_push($arrayKeysToUpdate, 'quantity_pending');
+            $nonNullArrayKeys = array_filter($data);
+            $arrayKeysToUpdate = array_keys($nonNullArrayKeys);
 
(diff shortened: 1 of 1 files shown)


app/Repositories/PostRepository.php
```diff
@@ -120,7 +120,11 @@ public function create(array $data, $signupId)
     public function update($signup, $data)
     {
         if (array_key_exists('updated_at', $data)) {
-            $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'why_participated', 'updated_at']));
+            if ($data['why_participated']) {
+                $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'why_participated', 'updated_at']));
```
chloealee: are we still using `quantity_pending`? Can we get rid of this here and delete it from the database if we're not?

app/Repositories/PostRepository.php
```diff
@@ -120,7 +120,11 @@ public function create(array $data, $signupId)
     public function update($signup, $data)
     {
         if (array_key_exists('updated_at', $data)) {
-            $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'why_participated', 'updated_at']));
+            if ($data['why_participated']) {
```
chloealee: Should we be checking that all of these values are filled in before overriding them in the database (not only `$data['why_participated']`)?

app/Repositories/PostRepository.php
katiecrane: @chloealee we can't outright delete it from the database yet because things migrated over from Phoenix used that column. We'd need to migrate all the `quantity_pending`s into the regular `quantity` column before deleting the column

katiecrane: I think you could remove it from this line though!

chloealee: ahhh that's right, thanks @katiecrane !

app/Repositories/PostRepository.php
katiecrane: We definitely need to allow `quantity` to be overwritten for quantity updates! 📈  `updated_at` might be leftover from the migration, I don't think anything else is telling us when it was updated at anymore?

chloealee: @katiecrane sorry I don't think I was clear! I mean check to see if `quantity` is filled in so we don't override this too? Or maybe this isn't a concern if `quantity` is always passed?

katiecrane: @chloealee oh oops I was thinking the exactly the wrong thing. I'm not sure exactly how gambit does it, but I feel like I always get asked how many more nouns I've verbed?

chloealee: @aaronschachter can you confirm? The only thing that isn't asked again when a user reports back for the 2+ time with another post via SMS is the `why_participated`, is that correct?

sbsmith86: i think they said they only asked `why_participated` the first time, and subsequent posts just ask for an image and caption. I think adding a safety check to the `quantity` field to make sure it is passed before using a null value makes good sense.

app/Repositories/PostRepository.php
```diff
@@ -120,7 +120,20 @@ public function create(array $data, $signupId)
...
+
+            foreach ($arrayKeysToUpdate as $key => $value) {
+                if (! $data[$value]) {
+                    $key = array_search($value, $arrayKeysToUpdate);
+                    unset($arrayKeysToUpdate[$key]);
+                }
+            }
+
+            array_push($arrayKeysToUpdate, 'updated_at');
+            array_push($arrayKeysToUpdate, 'quantity_pending');
```
katiecrane: I don't think anyone is ever sending us these values? I think updated_at was for the migration (Laravel handles that now) and I don't think anyone sends over quantity_pending.

katiecrane: Oh I see now that this is in the `updated_at` section. Nevermind!!!

app/Repositories/PostRepository.php
```diff
@@ -120,7 +120,20 @@ public function create(array $data, $signupId)
     public function update($signup, $data)
     {
         if (array_key_exists('updated_at', $data)) {
-            $signup->fill(array_only($data, ['quantity', 'quantity_pending', 'why_participated', 'updated_at']));
+            // Only update if the key is set (is not null).
+            $arrayKeysToUpdate = ['why_participated', 'quantity'];
+
+            foreach ($arrayKeysToUpdate as $key => $value) {
```
sbsmith86: I would look into php `array_filter` to handle this type of logic (removing null values from a keyed array). Furnes did something similar in this PR https://github.com/DoSomething/phoenix/pull/7493

chloealee: ah this is much cleaner, thanks @sbsmith86 ! I updated this and think this works fine! I think this is ok but before, it only updated with `why_participated` and `quantity`. Using `array_filter`, it replaces everything that is sent through if it isn't null. I think that is ok but wanted to run by you too in case I missed something!

PR #390, opened 2017-08-29
Return "tagged" not "tags"

katiecrane:
#### What's this PR do?
I goofed this when I did the new tagging stuff, but in the documentation we have `tagged` so we should return that.

👀 

This didn't break the tagging script, but it wouldn't work as expected.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Transformers/PostTransformer.php
+++ b/app/Http/Transformers/PostTransformer.php
@@ -35,8 +35,7 @@ class PostTransformer extends TransformerAbstract
                 'original_image_url' => $post->url,
                 'caption' => $post->caption,
             ],
-            // @TODO: include tags
-            'tags' => $post->tagSlugs(),
+            'tagged' => $post->tagSlugs(),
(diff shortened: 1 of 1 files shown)


app/Http/Transformers/PostTransformer.php
```diff
@@ -35,8 +35,7 @@ public function transform(Post $post)
                 'original_image_url' => $post->url,
                 'caption' => $post->caption,
             ],
-            // @TODO: include tags
-            'tags' => $post->tagSlugs(),
+            'tagged' => $post->tagSlugs(),
```
sbsmith86: I think this is going to break the front-end. I already updated it to work with the new response. but should be quick enough to fix back. https://github.com/DoSomething/rogue/pull/382/files#diff-18d307dbdc33d56a9696c2923ef00b20R24

Frontend testing soon to come so we don't run into this!

PR #392, opened 2017-08-30
Filter posts by tag

chloealee:
#### What's this PR do?
This PR allows adds the `filter[tag]=[tag_slug]` to query the `/posts` endpoint by a certain tag. 
With this, we can now filter posts by a specific tag. 

On the campaign single view page, try to filter by a tag! Make sure the responses are what you expect (posts with that specific tag). 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -203,11 +203,21 @@ class Post extends Model
     }
 
     /**
-     * Remove the given tag from this post.
+     * Returns posts without specific tag.
      */
     public function scopeWithoutTag($query, $tagSlug)
     {
         return $query->whereDoesntHave('tags', function ($query) use ($tagSlug) {
+            $query->where('tag_slug', '=', $tagSlug);
+        });
+    }
+
+    /**
+     * Returns posts with specific tag.
+     */
+    public function scopeWithTag($query, $tagSlug)
+    {
+        return $query->whereHas('tags', function ($query) use ($tagSlug) {
             $query->where('tag_slug', '=', $tagSlug);
         });
     }
--- a/resources/assets/components/CampaignSingle/index.js
+++ b/resources/assets/components/CampaignSingle/index.js
@@ -24,7 +24,6 @@ class CampaignSingle extends React.Component {
     this.deletePost = this.deletePost.bind(this);
     this.filterPosts = this.filterPosts.bind(this);
     this.getPostsByPaginatedLink = this.getPostsByPaginatedLink.bind(this);
-    this.createTagSlug = this.createTagSlug.bind(this);
     this.getPostsByTag = this.getPostsByTag.bind(this);
   }
 
@@ -35,26 +34,13 @@ class CampaignSingle extends React.Component {
   // Filter posts based on status or tag(s).
   filterPosts(filter) {
     // If the filter is a status, make API call to get posts by status.
-    if (['Pending', 'Accepted', 'Rejected'].includes(filter)) {
-      this.getPostsByStatus(filter.toLowerCase(), this.props.campaign.id);
+    if (['pending', 'accepted', 'rejected'].includes(filter)) {
+      this.getPostsByStatus(filter, this.props.campaign.id);
     } else {
       // If the filter is a tag, make the API call to get posts by tag.
-      var tagSlug = this.createTagSlug(filter);
-
-      this.getPostsByTag(tagSlug, this.props.campaign.id);
+      this.getPostsByTag(filter, this.props.campaign.id);
     }
 
-  }
-
-  // Create tag slug
-  createTagSlug(filter) {
-    var splitFilter = filter.toLowerCase().split(' ');
-
-    if (splitFilter[splitFilter.length-1] === '👻') {
-      splitFilter.pop();
-    }
-
-    return splitFilter.join('-');
   }
 
   // Open the history modal of the given post
--- a/resources/assets/components/CampaignSingle/index.js
+++ b/resources/assets/components/CampaignSingle/index.js
@@ -24,6 +24,7 @@ class CampaignSingle extends React.Component {
     this.deletePost = this.deletePost.bind(this);
     this.filterPosts = this.filterPosts.bind(this);
     this.getPostsByPaginatedLink = this.getPostsByPaginatedLink.bind(this);
+    this.createTagSlug = this.createTagSlug.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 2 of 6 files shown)


app/Models/Post.php
```diff
@@ -203,12 +203,22 @@ public function untag($tagName)
...
     {
         return $query->whereDoesntHave('tags', function ($query) use ($tagSlug) {
             $query->where('tag_slug', '=', $tagSlug);
         });
     }
+
+    /**
+     * Returns posts with specific tag.
+     */
+    public function scopeWithTag($query, $tagSlug)
```
sbsmith86: Nice!

sbsmith86: Should we have it be `withTags()` and instead of checking for equality, we check if the `tag_slug` is in a list of tags?

app/Models/Post.php
```diff
@@ -203,12 +203,22 @@ public function untag($tagName)
     }
 
     /**
-     * Remove the given tag from this post.
+     * Returns posts without specific tag.
```
katiecrane: Great catch!!!

resources/assets/components/CampaignSingle/index.js
```diff
@@ -24,15 +24,37 @@ class CampaignSingle extends React.Component {
...
     this.getPostsByStatus('accepted', this.props.campaign.id);
   }
 
-  // Filter posts based on status.
-  filterPosts(status) {
-    this.getPostsByStatus(status.toLowerCase(), this.props.campaign.id);
+  // Filter posts based on status or tag(s).
+  filterPosts(filter) {
+    // If the filter is a status, make API call to get posts by status.
+    if (filter === 'Pending' || filter === 'Accepted' || filter === 'Rejected') {
```
sbsmith86: You can use [`include()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) here to see if the filter is in an array of values, instead of multiple or statements

resources/assets/components/CampaignSingle/index.js
```diff
@@ -24,15 +24,37 @@ class CampaignSingle extends React.Component {
...
+      this.getPostsByTag(tagSlug, this.props.campaign.id);
+    }
+
+  }
+
+  // Create tag slug
+  createTagSlug(filter) {
+    var splitFilter = filter.toLowerCase().split(' ');
+
+    if (splitFilter[splitFilter.length-1] === '👻') {
```
katiecrane: Hahaha I'd be on board with removing the 👻  from the dropdown in order to not have to do this. But at the same time, I love this line.

app/Models/Post.php
chloealee: I was thinking that too and like that idea! For filtering by tag, you can only filter a tag one at a time (from the drop down menu) so we don't need it for this but could be good for the api in general to be able to send in multiple tags!

chloealee:
@katiecrane I did run `composer update` after I rebased for the tags work!

chloealee:
ah but maybe shouldn't have done that since it should have already updated with your merge? @katiecrane

resources/assets/components/CampaignSingle/index.js
```diff
@@ -24,15 +24,37 @@ class CampaignSingle extends React.Component {
...
 
-  // Filter posts based on status.
-  filterPosts(status) {
-    this.getPostsByStatus(status.toLowerCase(), this.props.campaign.id);
+  // Filter posts based on status or tag(s).
+  filterPosts(filter) {
+    // If the filter is a status, make API call to get posts by status.
+    if (filter === 'Pending' || filter === 'Accepted' || filter === 'Rejected') {
+      this.getPostsByStatus(filter.toLowerCase(), this.props.campaign.id);
+    } else {
```
sbsmith86: What happens if we want to filter by status AND tag. Could we potentially grab all the filter fields and then make one request with all of the filter values? If this feels out of scope for this PR, we could create a card to update the functionality. But we should be clear that you can only do one or the other.

katiecrane:
@chloealee that sounds fine to me!

resources/assets/components/PostFilter/index.js
```diff
@@ -21,6 +21,11 @@ class PostFilter extends React.Component {
                   <option>Accepted</option>
                   <option>Pending</option>
                   <option>Rejected</option>
+                  <option>Good Photo</option>
```
sbsmith86: I think we should do something more like `<option value="good-photo">Good Photo</option>` This way you don't have to do the `createTagSlug()` work, because you are already passing the slug as the value.

resources/assets/components/CampaignSingle/index.js
sbsmith86: Ah I see that below, we just do one large select list with filter options so we only allow one filter. In this case it is fine, but I definitely think we should update this functionality so that the `status` filter and `tags` filter are separate. We should be able to do something like give me all `Accepted` posts that have a `good-photo` tag. So can you make a note to make a new card for this work and we can prioritize that later?

resources/assets/components/CampaignSingle/index.js
chloealee: oh awesome! So much cleaner.

resources/assets/components/CampaignSingle/index.js
chloealee: Yes - I'll make a card for this now! cc @ngjo

resources/assets/components/PostFilter/index.js
chloealee: ohhhhh yes!

app/Models/Post.php
sbsmith86: Yea and it looks like we only allow you to filter out one tag. so I think both of these should be updated. Will make a card!

chloealee: awesome, thanks Shae!

resources/assets/components/PostFilter/index.js
```diff
@@ -18,9 +18,14 @@ class PostFilter extends React.Component {
           <h4>Post Filter</h4>
           <div className="select">
               <select onChange={(event) => this.change(event)}>
```
sbsmith86: Looks great! Let's make note for future filtering work that we should create a `<Filter>` component that takes in a array of value/labels` and renders the select list and handles the change event. So as we build out multiple filters, we don't have to hard code values and they can be more flexible.

chloealee: love it! I'll add a `@TODO` comment now.

PR #394, opened 2017-08-31
App Container refactor

sbsmith86:
#### What's this PR do?

Refactors how we were adding React components to the server rendered paged. 

Before we were looking for the element to add a react container component to by searching for an `id` prop on the server rendered HTML. This ended up being a lot of conditional statements to see if the element was on the page before adding the component. 

Now, in the server rendered HTML, we add a `data-container` attribute and set the value to equal the name of the component to add to the page. And then, when we load in the JS we look for the `data-container` attribute and use it's value to add the correct element to the page.

👀  commit-by-commit should be easy to follow. 

Right now, this only assumes that there one HTML element on the page to add a react container component to. This is because we really are just loading in a base container "App" per blade view and then doing all the rendering work via react so doesn't seem like we need to add multiple components to any page as of now. If this changes, we can always update this to be more dynamic. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/150273936

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/app.js
+++ b/resources/assets/app.js
@@ -17,15 +17,10 @@ ready(() => {
     CampaignOverview: CampaignOverview,
   });
 
-  const overviewContainer = document.getElementById('overviewContainer');
   const inboxContainer = document.getElementById('inboxContainer');
   const singleCampaignContainer = document.getElementById('singleCampaignContainer');
   const userOverviewContainer = document.getElementById('userOverviewContainer');
   const signupContainer = document.getElementById('signupContainer');
-
-  // if (overviewContainer) {
-  //   ReactDom.render(<CampaignOverview {...window.STATE} />, overviewContainer);
-  // }
 
   if (inboxContainer) {
     ReactDom.render(<CampaignInbox {...window.STATE} />, inboxContainer);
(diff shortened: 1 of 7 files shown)


resources/assets/app.js
```diff
@@ -3,23 +3,29 @@ import './app.scss';
...
   const singleCampaignContainer = document.getElementById('singleCampaignContainer');
   const userOverviewContainer = document.getElementById('userOverviewContainer');
   const signupContainer = document.getElementById('signupContainer');
 
-  if (overviewContainer) {
-    ReactDom.render(<CampaignOverview {...window.STATE} />, overviewContainer);
-  }
+  // if (overviewContainer) {
+  //   ReactDom.render(<CampaignOverview {...window.STATE} />, overviewContainer);
+  // }
```
chloealee: leftover comment!

chloealee: ugh sorry this didn't make it into the final PR! Disregard!

resources/assets/app.js
```diff
@@ -3,37 +3,20 @@ import './app.scss';
...
-  if (signupContainer) {
-    ReactDom.render(<Signup {...window.STATE} />, signupContainer);
-  }
+  mountContainer({
+    CampaignOverview: CampaignOverview,
+    CampaignInbox: CampaignInbox,
+    CampaignSingle: CampaignSingle,
+    UserOverview: UserOverview,
+    Signup: Signup,
+  });
```
chloealee: so much cleaner!

katiecrane:
wow this is so much nicer!!!

PR #396, opened 2017-09-06
Send payload to blink even when updating a post

sbsmith86:
#### What's this PR do?

We were only making the blink request in the post->create method and not in the post->update method. 

This surfaced a weird logic issue we have in our post creation code where we consider an "update" when a user submits a new post to an existing signup. We only consider it a "create" when we are creating both the signup and the post at the same time. So what ended up happening was we were only sending to blink when we sent over the signup and the post content at the same time (or created the signup on the fly with the post). 

This logic seems flawed and we should look at it, but in an interest to unblock the messaging team, I have added the blink integration to both update and create methods. 

👀 

See above. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/150827279

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Traits/PostRequests.php
+++ b/app/Http/Controllers/Traits/PostRequests.php
@@ -28,8 +28,6 @@ trait PostRequests
         if (! $updating) {
             $signup = $this->signups->create($request->all());
 
-            // Send the data to the PostService class which will handle determining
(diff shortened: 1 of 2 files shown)


katiecrane:
@sbsmith86 I approved but it actually looks like now there is a failing test

PR #398, opened 2017-09-08
Admin uploader 

chloealee:
#### What's this PR do?
Adds uploader modal that allows admins to upload photo(s) and caption(s) on behalf of a user. 

The commits are a little scattered with lots of work in progress to not lose work done - sorry! 

You can review by testing: 
1. Go to a user's signup page (`/signsup/:sid`). 
2. Under the "Edit | Show History" button, there is a "Upload Photo" button. Click on that.
3. A modal like this should pop up (without "Thanks for the photo! It has been automatically approved." success message):
screen shot 2017-09-08 at 10 58 34 am
4. The "SUBMIT A NEW PHOTO" button shouldn't be clickable (turn blue) until both a photo and caption are entered in the form. 
5. Once you upload a new photo and caption, a green success message should pop up in the modal (see screen shot above).  
6. If you close the modal and scroll down to the approved section, the newly uploaded photo should be added in state. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/150313667

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/documentation/endpoints/posts.md
+++ b/documentation/endpoints/posts.md
@@ -18,6 +18,8 @@ POST /api/v2/posts
     The reason why the user participated.
   - **caption**: (string).
     Corresponding caption for the post.
+  - **status**: (string).
+    Option to set status upon creation if admin uploads post for user.
   - **source**: (string).
     Where the post was submitted from.
   - **remote_addr**: (string).
--- a/resources/assets/components/Signup/index.js
+++ b/resources/assets/components/Signup/index.js
@@ -286,17 +286,16 @@ class Signup extends React.Component {
             <div className="container__row">
               <a href="#" onClick={e => this.showUploader(signup, e)}>Upload Photo</a>
 
-                <ModalContainer>
-                  {this.state.displayUploaderModal ?
-                    <UploaderModal id={this.state.historyModalId}
-                      // onUpdate={this.updateQuantity}
-                      onClose={e => this.hideUploader(e)}
-                      signup={signup}
-                      // campaign={campaign}
-                      submitReportback={this.submitReportback}
-                    />
-                  : null}
-                </ModalContainer>
+              <ModalContainer>
+                {this.state.displayUploaderModal ?
+                  <UploaderModal id={this.state.historyModalId}
+                    onClose={e => this.hideUploader(e)}
+                    signup={signup}
+                    campaign={campaign}
+                    submitReportback={this.submitReportback}
+                  />
+                : null}
+              </ModalContainer>
             </div>
 
             <MetaInformation title="Meta" details={{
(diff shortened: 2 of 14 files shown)


resources/assets/components/UploaderModal/index.js
```diff
@@ -0,0 +1,42 @@
...
+
+    const photoUploaderProps = {
+      campaignId: signup.campaign_id,
+      campaignRunId: signup.campaign_run_id,
+      northstarId: signup.northstar_id,
+      noun: {
+        plural: campaign.reportback_info.noun,
+      },
+      // @TODO: do we want this to be the admin's northstar ID so we know who submitted this?
+      source: 'Rogue admin',
```
chloealee: should we have this be more specific? Maybe something like "Rogue admin: [northstar id] or [name]"? Or is this ok and we can always check the logs for this?

resources/assets/components/Signup/index.js
```diff
@@ -184,6 +211,46 @@ class Signup extends React.Component {
...
+            message: "Thanks for the photo! It has been automatically approved.",
+          }
+        };
+
+        newState.posts[key] = post[key];
+
+        return newState;
+      });
+    })
+    .catch(error => console.log(error));
```
chloealee: Instead of `console.log(error)`, should I set an error message? Something like the below?
```
newState.successfulSubmission = {
  error: {
    message: "Something went wrong.",
  }
};
```

sbsmith86: I think the error message is a good idea! I'm not too concerned with the success message staying up there, unless we here otherwise

documentation/endpoints/posts.md
```diff
@@ -18,6 +18,8 @@ POST /api/v2/posts
     The reason why the user participated.
   - **caption**: (string).
     Corresponding caption for the post.
+  - **status**: (string).
```
sbsmith86: Did you have to do anything to make sure this got stored?

resources/assets/components/UploaderModal/index.js
sbsmith86: hmm we store the user in the events table, right? If so, then I don't think we need it here, but I think the source should be formatted `rogue-admin` for easier parsing.

documentation/endpoints/posts.md
chloealee: we actually already had that functionality! Just wasn't in the documentation so added it :)

sbsmith86: sweet!

resources/assets/components/UploaderModal/index.js
chloealee: Yes we do! Oh great call - I'll change that now!

resources/assets/components/Signup/index.js
chloealee: ok cool!

PR #401, opened 2017-09-11
Filter by multiple tags

chloealee:
#### What's this PR do?
- Adds ability to filter by multiple tags (to both include and exclude). 
  - e.g. `/api/v2/posts?filter[tag]=good-quote,good-for-storytelling`

Commit by commit and/or by testing the above endpoint with multiple variations of tags! Make sure the results are what is expected. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/150713655

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -208,7 +208,6 @@ class Post extends Model
     public function scopeWithoutTag($query, $tagSlug)
     {
         return $query->whereDoesntHave('tags', function ($query) use ($tagSlug) {
-            // $query->where('tag_slug', '=', $tagSlug);
             $values = explode(',', $tagSlug);
             if (count($values) > 1) {
                 $query->where(function ($query) use ($values) {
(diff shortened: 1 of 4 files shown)


app/Models/Post.php
```diff
@@ -203,22 +203,41 @@ public function untag($tagName)
...
-            $query->where('tag_slug', '=', $tagSlug);
+            $this->multipleValueQuery($query, $tagSlug);
         });
     }
+
+    /**
+     * Runs query when there are multiple values.
+     * e.g. `filter[tag]=good-quote,hide-in-gallery,good-photo`
+     */
+    public function multipleValueQuery($query, $queryString)
```
sbsmith86: Should this go in `helpers.php`?

chloealee: ooohh great idea!

chloealee: @sbsmith86 updated!

app/helpers.php
```diff
@@ -53,3 +53,28 @@ function has_middleware($middleware = null)
 
     return $currentRoute->middleware() ? true : false;
 }
+
+/**
+ * Runs query when there are multiple values.
```
sbsmith86: Maybe add that it runs a query where there are multiple values provided from a comma-separated list.

chloealee: ah yes that is more clear - updated!

PR #402, opened 2017-09-12
Adds Jest to Rogue

chloealee:
#### What's this PR do?
- Adds Jest and Babel Jest to Rogue
- Sets up a test to make sure it is properly set up. 

👀  and run `npm test`. Tests should work and pass: 
screen shot 2017-09-12 at 11 52 58 am

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/150971771

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
     "build": "npm run clean && NODE_ENV=production webpack",
     "build:dev": "npm run clean && NODE_ENV=development webpack",
     "clean": "rm -rf public/dist",
-    "test": "jest"
+    "test": "jest",
+    "test:ci": "NODE_ENV=test && jest"
   },
   "babel": {
     "presets": [
--- a/package.json
+++ b/package.json
@@ -24,7 +24,9 @@
   "devDependencies": {
     "@dosomething/babel-config": "^1.0.0",
     "@dosomething/webpack-config": "^2.0.4",
+    "babel-jest": "^21.0.2",
     "jest": "^21.0.2",
+    "regenerator-runtime": "^0.11.0",
     "webpack": "^2.4.0",
     "webpack-dev-server": "^2.4.0"
   }
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
     "start": "npm run clean && NODE_ENV=development webpack --watch",
     "build": "npm run clean && NODE_ENV=production webpack",
     "build:dev": "npm run clean && NODE_ENV=development webpack",
-    "clean": "rm -rf public/dist"
+    "clean": "rm -rf public/dist",
+    "test": "jest"
   },
   "babel": {
     "presets": [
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -4,7 +4,8 @@
     "start": "npm run clean && NODE_ENV=development webpack --watch",
     "build": "npm run clean && NODE_ENV=production webpack",
     "build:dev": "npm run clean && NODE_ENV=development webpack",
-    "clean": "rm -rf public/dist"
+    "clean": "rm -rf public/dist",
+    "test": "jest"
   },
   "babel": {
     "presets": [
```
chloealee: @DFurnes should I also add `flow` in the presets similar to [Phoenix Next](https://github.com/DoSomething/phoenix-next/blob/dev/package.json#L22)? 

Also, I think I can skip adding 
```
{
  "presets": ["es2015", "react"]
}
``` 
like it says [here (under Additional Configuration](https://facebook.github.io/jest/docs/en/getting-started.html#content)?

package.json
```diff
@@ -23,6 +24,9 @@
   "devDependencies": {
     "@dosomething/babel-config": "^1.0.0",
     "@dosomething/webpack-config": "^2.0.4",
+    "babel-jest": "^21.0.2",
+    "jest": "^21.0.2",
+    "regenerator-runtime": "^0.11.0",
```
chloealee: @DFurnes [This](https://facebook.github.io/jest/docs/en/getting-started.html#content) said to also install `regenerator-runtime` but I don't think PN has it - is this needed?

package.json
DFurnes: We're not using Flow in Rogue at the moment, so nope! Safe to leave out. And yep, can leave the Babel configuration as is. 😄

package.json
DFurnes: Nope, we should be using npm 4+ everywhere so no need.

package.json
chloealee: awesome, thanks @DFurnes !

package.json
chloealee: Ok cool - deleted!

DFurnes:
Let's also [add this to our Wercker builds](https://github.com/DoSomething/phoenix-next/blob/ff451405aabea585a34916daa6636fb119f2c2dc/wercker.yml#L13-L15), so we can make sure tests pass before merging PRs.

In `wercker.yml`:

```yml
- script:
    name: test javascript code
    code: npm run test:ci
```

And in `package.json`, a [CI-specific Jest task](https://github.com/DoSomething/phoenix-next/blob/ff451405aabea585a34916daa6636fb119f2c2dc/package.json#L5) (that runs tests non-interactively):

```json
"test:ci": "NODE_ENV=test && jest",
```

chloealee:
@DFurnes ah so cool!

DFurnes:
This looks great @chloealee! 🚥 🎊

PR #403, opened 2017-09-12
Accept Northstar JWT tokens, via Gateway server middleware/guard.

DFurnes:
#### What's this PR do?
This pull request adds support for [Northstar tokens](https://github.com/DoSomething/northstar/blob/dev/documentation/authentication.md) to Rogue, via [an update to Gateway](https://github.com/DoSomething/gateway/pull/83).

Authenticating API requests via the `X-DS-Rogue-API-Key` header should still work, but you should also be able to provide a valid Northstar JWT (for now, requiring the `admin` scope) as well!

In the future, it makes sense to use people's Northstar tokens when making requests to Rogue from Phoenix Ashes/Next so these requests are made on behalf of individual users - and then we can use `auth()->id()` or `token()->role` to get their real authentication details.

#### Relevant tickets
[#150224528](https://www.pivotaltracker.com/story/show/150224528)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Middleware/AuthenticateApi.php
+++ b/app/Http/Middleware/AuthenticateApi.php
@@ -34,8 +34,8 @@ class AuthenticateApi extends Authenticate
             return $next($request);
         }
 
-        // Otherwise, we can't do anything!
-        // @TODO: Instead, pass on to default auth middleware with Gateway guard.
-        throw new AuthenticationException;
+        // Otherwise, pass on to Laravel's default Authenticate
+        // middleware with the 'api' guard specified.
+        return parent::handle($request, $next, 'api');
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     "aws/aws-sdk-php-laravel": "^3.1",
     "barryvdh/laravel-debugbar": "^2.3",
     "doctrine/dbal": "~2.5.13",
-    "dosomething/gateway": "^1.5.2",
+    "dosomething/gateway": "^1.7.0",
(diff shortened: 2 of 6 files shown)


DFurnes:
One thing that's still a little awkward is getting the public key (right now you have to log into Aurora, copy-paste it from the [clients page](https://qa-aurora.dosomething.org/clients), and put it in `storage/keys/public.key`. In the future, it might make sense to expose this from a public endpoint and just make an artisan command to download it.

DFurnes:
This is ready for a look! 👀 Once DoSomething/gateway#83 is merged, I'll update this pull request to reference the tagged release & it'll be safe to merge and deploy.

DFurnes:
Yep, almost! For now, it requires a token with the `admin` scope because the API endpoints are written with the expectation that traffic is coming from "trusted" DS services. I think it makes sense to make a new set of endpoints that work with the expectation that a user is making the requests (either via the Rogue front-end, or any other DS service).

DFurnes:
> And, other apps can grab data from rogue using their tokens without needing to set a header?

Yes! This pull request would mean that any applications that request [client credentials tokens](https://github.com/DoSomething/northstar/blob/dev/documentation/endpoints/auth.md#create-token-client-credentials-grant) with the `admin` scope could make requests to Rogue, without setting the `X-Rogue-API-Key` header.

DFurnes:
Rebased this to fix conflicts & installed the [tagged Gateway release](https://github.com/DoSomething/gateway/releases/tag/v1.7.0).

PR #404, opened 2017-09-12
Include reaction information with GET /posts

katiecrane:
#### What's this PR do?
Added reaction information to the `PostTransformer`. The information included is the number of reactions on the post and and `reacted`. `Reacted` tells whether or not the `as_user` included in the request has reacted to the post. It is `false` if the user did not react or if no user was provided.

👀  This includes additional information in the response (as opposed to altering existing data) so shouldn't break anything.

This was needed to get the correct reaction data on the user profile in Phoenix.

I went a little bit down a rabbit hole because my local reaction data was funky so wasn't showing up.

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/story/show/151011784)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Transformers/PostTransformer.php
+++ b/app/Http/Transformers/PostTransformer.php
@@ -25,6 +25,11 @@ class PostTransformer extends TransformerAbstract
      */
     public function transform(Post $post)
     {
+        $reacted = false;
+        if ($post->relationLoaded('reactions')) {
+            $reacted = $post->reactions->isNotEmpty();
+        }
+
         return [
             'id' => $post->id,
             'signup_id' => $post->signup_id,
@@ -36,8 +41,8 @@ class PostTransformer extends TransformerAbstract
             ],
             'tags' => $post->tagSlugs(),
             'reactions' => [
-                'reacted' => $post->reactions->isNotEmpty(),
-                'total' => isset($post->reactions_count) ? $post->reactions_count : 'unavailable',
+                'reacted' => $reacted,
+                'total' => isset($post->reactions_count) ? $post->reactions_count : null,
             ],
             'status' => $post->status,
             'source' => $post->source,
--- a/app/Http/Transformers/PostTransformer.php
+++ b/app/Http/Transformers/PostTransformer.php
@@ -30,12 +30,15 @@ class PostTransformer extends TransformerAbstract
             'signup_id' => $post->signup_id,
             'northstar_id' => $post->northstar_id,
             'media' => [
-
                 'url' => $post->getMediaUrl(),
                 'original_image_url' => $post->url,
                 'caption' => $post->caption,
             ],
             'tags' => $post->tagSlugs(),
+            'reactions' => [
+                'reacted' => $post->reactions->isNotEmpty() ? true : false,
+                'total' => $post->reactions_count,
+            ],
             'status' => $post->status,
             'source' => $post->source,
             'remote_addr' => $post->remote_addr,
--- a/app/Http/Transformers/PostTransformer.php
+++ b/app/Http/Transformers/PostTransformer.php
@@ -36,8 +36,8 @@ class PostTransformer extends TransformerAbstract
             ],
             'tags' => $post->tagSlugs(),
             'reactions' => [
-                'reacted' => $post->reactions->isNotEmpty() ? true : false,
-                'total' => $post->reactions_count,
+                'reacted' => $post->reactions->isNotEmpty(),
+                'total' => isset($post->reactions_count) ? $post->reactions_count : 'unavailable',
             ],
             'status' => $post->status,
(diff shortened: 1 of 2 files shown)


app/Http/Transformers/PostTransformer.php
```diff
@@ -30,12 +30,15 @@ public function transform(Post $post)
...
             'northstar_id' => $post->northstar_id,
             'media' => [
-
                 'url' => $post->getMediaUrl(),
                 'original_image_url' => $post->url,
                 'caption' => $post->caption,
             ],
             'tags' => $post->tagSlugs(),
+            'reactions' => [
+                'reacted' => $post->reactions->isNotEmpty() ? true : false,
```
DFurnes: I think this ternary can probably be removed if `isNotEmpty` already returns a boolean. Also, do we eager-load reactions everywhere this transformer is used? If not, this might cause some unexpected performance issues.

app/Http/Transformers/PostTransformer.php
```diff
@@ -30,12 +30,15 @@ public function transform(Post $post)
...
+                'total' => $post->reactions_count,
```
DFurnes: Related question: do we always load a count of reactions? If not, might be good to conditionally include this field only if `reactions_count` is set on the post.

app/Http/Transformers/PostTransformer.php
sbsmith86: I think this already returns a boolean? So you probably don't need to set it with the ternary.  https://laravel.com/docs/5.4/collections#method-isnotempty

sbsmith86: @DFurnes is this using the eager-loaded object or just querying the post/reaction relationship?

DFurnes: It depends where it's being used! I'm not sure if there's a good way to check this at runtime, so we might want to just do a quick audit of everywhere `PostTransformer` gets applied and check.

katiecrane: yup great call!!!

app/Http/Transformers/PostTransformer.php
katiecrane: oooo this is a great question and I don't think that we do which would definitely break this

app/Http/Transformers/PostTransformer.php
katiecrane:
```
$query = $query->with(['reactions' => function ($query) use ($userId) {
    $query->where('northstar_id', '=', $userId);
}]);
```
conditionally adds `reactions` to the query if `as_user` is included, so we are just checking to see if that pulled up any reactions or not.

DFurnes: Oh right, got it! Hmm and just to make sure this wouldn't load the overall reactions relationship otherwise right?

app/Http/Transformers/PostTransformer.php
```diff
@@ -30,12 +30,15 @@ public function transform(Post $post)
...
             'media' => [
-
                 'url' => $post->getMediaUrl(),
                 'original_image_url' => $post->url,
                 'caption' => $post->caption,
             ],
             'tags' => $post->tagSlugs(),
+            'reactions' => [
+                'reacted' => $post->reactions->isNotEmpty(),
+                'total' => isset($post->reactions_count) ? $post->reactions_count : 'unavailable',
```
DFurnes: Hmm, can we use `null` instead of `"unavailable"` so the datatype is `int|null` rather than `int|string`?

katiecrane: @DFurnes yeah I wasn't sure what to do here because I wanted to somehow signify that "we didn't get the total number of reactions for you" vs "there are 0 reactions." I guess I could use `null` for the first case?

app/Http/Transformers/PostTransformer.php
katiecrane: A-ha yeah it was loading it here if it didn't have it already. Needed to add `$post->relationLoaded('reactions')`

app/Http/Transformers/PostTransformer.php
DFurnes: Yeah yeah, it's a lil' wonky. I think `null` is decent enough for "we didn't get this for you" since zero reactions would show up as an integer `0`.

app/Http/Transformers/PostTransformer.php
```diff
@@ -25,17 +25,25 @@ class PostTransformer extends TransformerAbstract
      */
     public function transform(Post $post)
     {
+        $reacted = false;
+        if ($post->relationLoaded('reactions')) {
```
DFurnes: Nice find!! 🕵️‍♀️

PR #406, opened 2017-09-13
Separates status and tag filters

chloealee:
#### What's this PR do?
- Separates the status and tag filters so users can filter by a combination of status and tag(s). 
- Updates some styling according to Luke's design (adding Post Filters and Posts headings):
screen shot 2017-09-13 at 3 01 02 pm

- Go to Campaign Single Page. 
- Play with a combination of filters and the results should accurately represent what you filtered by. 
- Note: If you filter by `pending` or `rejected` AND tags, there will be no results returned. I was thinking about not allowing filtering by tags if `pending` or `rejected` were chosen (not displaying tags) but decided against this in case there are some tagged rejected posts for some reason and because this still returns accurate information. 

The "FILTER" button is just a placeholder for now - waiting on @lkpttn 's final design! But thought review could be started on this PR while we wait since I'm sure I'll have to fix a lot! 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/150712305

#### Checklist
- [ ] Tested on staging.


resources/assets/components/TagsFilter/index.js
```diff
@@ -0,0 +1,55 @@
...
+  constructor() {
+    super();
+
+    this.state = {
+      'good-photo': false,
+      'good-quote': false,
+      'hide-in-gallery': false,
+      'good-for-sponsor': false,
+      'good-for-storytelling': false,
+    }
```
chloealee: should I be passing the state down  from `PostFilter` here instead of declaring again?

resources/assets/components/CampaignSingle/index.js
```diff
@@ -55,32 +56,21 @@ class CampaignSingle extends React.Component {
     });
   }
 
-  // Make API call to GET /posts to get posts by filtered status.
-  getPostsByStatus(status, campaignId) {
+  // Make API call to GET /posts to get posts by filtered status and/or tag(s).
+  getPostsByFilter(campaignId, status, activeTags) {
```
sbsmith86: Could we possibly update this to take in an object and then use the properties in the object to send in the request? If we ever wanted to filter by users, runs, etc we would have to keep adding params here. So I'm wondering if we use the filter values set by the user to create the `filter` array we send in the request.

resources/assets/components/TagsFilter/index.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import { map } from 'lodash';
+import classnames from 'classnames';
+
+import '../Tags/tags.scss';
+
+class TagsFilter extends React.Component {
```
sbsmith86: This should be a more general `MultiValueFilter` component. Then we can pass in what values it should render through the `props` we sent to it.

resources/assets/components/StatusFilter/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+class TagsFilter extends React.Component {
```
sbsmith86: It looks like you are calling this `TagsFilter` in a `StatusFilter` component?

sbsmith86: We actually should probably have this be `DropdownFilter` that takes in props that define what the `options` should be.

chloealee: oops!

resources/assets/components/PostFilter/index.js
```diff
@@ -1,35 +1,60 @@
 import React from 'react';
 
+import StatusFilter from '../StatusFilter';
+import TagsFilter from '../TagsFilter';
+
 class PostFilter extends React.Component {
```
sbsmith86: I think we should rename this to be `FilterBar` or something more general so it isn't so tightly connected to the content it is filtering.

chloealee: ah yes I like that idea! I will change that now.

sbsmith86:
This is a great start! Some of my comments might be confusing but I think we can make this a little more understandable and easier to use by doing something like this: 

My thought was we would have in the `CampaignSingle` component some JSX that looks like: 

```
<h4>Posts Filter</h4>
<FilterBar onSubmit=this.filterPosts()>
  <DropdownFilter options={accepted: 'accepted', pending: 'pending', rejected: 'rejected'} onChange=updateFilters() />
  <MultivalueFilter options={good-photo: 'Good Photo'...} onChange=updateFilters() /> 
  <button>Submit</button> 
</FilterBar>
```

The idea being that we are creating reusable components that we can use for any kind of filtering we want down the line. We can talk IRL if that doesn't make sense!

chloealee:
@sbsmith86 that makes total sense and a lot more flexible/readable! Thanks for the comments and for the code of what you're thinking too - SUPER helpful! I'll get on these changes :)

chloealee:
@sbsmith86 this is ready for another look! The only thing is that I'm getting this in the console when I load the page: 
```
warning.js:35 Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `DropdownFilter`. See https://fb.me/react-warning-keys for more information.
    in option (created by DropdownFilter)
    in DropdownFilter (created by CampaignSingle)
    in div (created by FilterBar)
    in div (created by FilterBar)
    in FilterBar (created by CampaignSingle)
    in div (created by CampaignSingle)
    in CampaignSingle (created by _class)
    in _class
```
but as far as I can tell, each of these do have unique keys since we map through these in `DropdownFilter` on line 20? 🤔

resources/assets/components/DropdownFilter/index.js
```diff
@@ -0,0 +1,30 @@
...
+    this.props.updateFilters(event.target.value);
+  }
+
+  render() {
+    return (
+        <div className="container__block -third">
+          <div className="select">
+            <select onChange={(event) => this.change(event)}>
+              {map(this.props.options, (option, key) => (
+                 <option value={key}>{option}</option>
```
sbsmith86: Ah you have to have an actual `key` prop for each `<option>` so like `<option value={key} key={key}>`

resources/assets/components/FilterBar/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+        return newState;
+      });
+    }
+  }
+
+  render() {
+    // ***Rendering the children**
+
+    // In this case DropdownFilter and MultiValueFilter are children of the FilterBar as layed out in Campaign Single.
```
chloealee: @sbsmith86 I kept your comments/explanation because I thought they were super helpful and could be good context for the future!

resources/assets/components/CampaignSingle/index.js
```diff
@@ -20,21 +22,30 @@ class CampaignSingle extends React.Component {
...
 
   // Filter posts based on status or tag(s).
-  filterPosts(filter) {
-    // If the filter is a status, make API call to get posts by status.
-    if (['pending', 'accepted', 'rejected'].includes(filter)) {
-      this.getPostsByStatus(filter, this.props.campaign.id);
-    } else {
-      // If the filter is a tag, make the API call to get posts by tag.
-      this.getPostsByTag(filter, this.props.campaign.id);
+  filterPosts(state) {
```
sbsmith86: I think you should maybe pass in the filter prop from the state instead of the whole state object. 

So if `state` looked like: 
```
{
  loadingNewPosts: false,
  filters: {
   //.....
  }
}
```
you would pass in just `state.filters` here.

resources/assets/components/CampaignSingle/index.js
```diff
@@ -20,21 +22,30 @@ class CampaignSingle extends React.Component {
...
+
+    if (state.tags) {
+      Object.keys(state.tags).forEach(function(key) {
+        if (state.tags[key] === true) {
+         activeTags.push(key);
+        }
+      });
     }
 
+    let filters = {
```
sbsmith86: Following my above comment, if `filters` is it's own object in `state` you would just be able to pass it to `getPostsByFilter()` or do whatever parsing work you need to do with the object to get it in the right format for the call.

resources/assets/components/CampaignSingle/index.js
```diff
@@ -55,32 +66,21 @@ class CampaignSingle extends React.Component {
...
+    if (filters.tags.length > 0) {
+      apiFilter['tag'] = filters.tags.toString();
+    }
 
-    this.api.get('posts', {
-      filter: {
-        tag: tagSlug,
-        campaign_id: campaignId,
-      },
+    this.api.get('api/v2/posts', {
```
sbsmith86: We have to make the call to the `GET /posts` web route so that the request is authenticated properly and we get back all the information we need. 

The reason we were seeing issues with user information not showing up was because using `api/v2/posts` was making unauthenticated requests to northstar which doesn't return full user objects

resources/assets/components/CampaignSingle/index.js
```diff
@@ -55,32 +66,21 @@ class CampaignSingle extends React.Component {
...
-        campaign_id: campaignId,
-      },
-      include: 'signup,siblings',
-    })
-    .then(json => {
-      this.setState({loadingNewPosts: false });
-      this.props.setNewPosts(json);
-    });
-  }
+    let apiFilter = {
```
sbsmith86: hmm why do you need to redefine this variable here? It looks like the same object that is passed in?

resources/assets/components/CampaignSingle/index.js
```diff
@@ -98,8 +98,13 @@ class CampaignSingle extends React.Component {
       <div className="container">
         <StatusCounter postTotals={this.props.post_totals} campaign={campaign} />
 
-        <PostFilter onChange={this.filterPosts} />
+        <h2 className="heading -emphasized">Post Filters</h2>
+        <FilterBar onSubmit={this.filterPosts}>
+          <DropdownFilter options={{accepted: 'Accepted', pending: 'Pending', rejected: 'Rejected'}} onChange={this.updateFilters} />
```
sbsmith86: I think I might have messed this up in my gist example, but you shouldn't have to pass an `onChange` prop here since you pass in when you render it in `FilterBar`.

resources/assets/components/CampaignSingle/index.js
```diff
@@ -98,8 +98,13 @@ class CampaignSingle extends React.Component {
...
+          <MultiValueFilter options={{'good-photo': 'Good Photo', 'good-quote': 'Good Quote', 'hide-in-gallery': 'Hide In Gallery 👻', 'good-for-sponsor': 'Good For Sponsor', 'good-for-storytelling': 'Good For Storytelling'}} onChange={this.updateFilters} />
```
sbsmith86: Same with this one.

resources/assets/components/FilterBar/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+import DropdownFilter from '../DropdownFilter';
+import MultiValueFilter from '../MultiValueFilter';
+
+class FilterBar extends React.Component {
+  constructor() {
+    super();
+
+    this.state = {
+      status: 'accepted',
```
sbsmith86: This probably doesn't need a default value since on component mount, it doesn't do any work to determine what we are filtering by. It's only on submit of this component, that it should do the work of updating the state of the parent component.

resources/assets/components/FilterBar/index.js
sbsmith86: Cool! maybe just keep lines 48 and 49? this one is just explaining how we are using this component in `CampaignSingle` but we want to keep this general because it could be used differently some place else.

resources/assets/components/MultiValueFilter/index.js
```diff
@@ -0,0 +1,50 @@
...
+import classnames from 'classnames';
+
+import '../Tags/tags.scss';
+
+class MultiValueFilter extends React.Component {
+  constructor() {
+    super();
+
+    this.state = {
+      'good-photo': false,
```
sbsmith86: Something for thought, we keep defining these tag key/value pairs in different places. Is there a way to use the object that is passed into the `MultivalueFilter` component and pass them up to parents as needed.

sbsmith86: And actually this `multivalueFilter` really shouldn't care what is passed into it, so defining these properties on the state sorta confines it's usage. What if when one of the filter tags was clicked, it just pushed it's key/value pair to the state.

resources/assets/components/MultiValueFilter/index.js
```diff
@@ -0,0 +1,50 @@
...
+      'good-quote': false,
+      'hide-in-gallery': false,
+      'good-for-sponsor': false,
+      'good-for-storytelling': false,
+    }
+
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick(key) {
```
sbsmith86: this could take in the `key` and `option` so you can just push that to the state.

resources/assets/components/PostFilter/index.js
```diff
@@ -1,38 +0,0 @@
-import React from 'react';
```
sbsmith86: Deleting things! yay!

resources/assets/components/MultiValueFilter/index.js
chloealee: The only reason why I defined the default state here so the button would know if it was active or not for the below (`{'is-active': this.state[key]}`) so the front end knows if it should have a background of blue if it is clicked and white if it is not clicked. Is it ok to keep this?

sbsmith86: hmm maybe the component should take in an extra property value if the key/value pair is active or not. So like `CampaignSingle` knows what it is using to grab posts and maybe it should tell the `MultivalueFilter` that along with what options to render. We want to move towards these components being as agnostic of the content they are rendering as possible. So this just can just be a component that has buttons and allows the user to select multiple of them.

chloealee: hmm I'm a little confused by this - so `CampaignSingle` would pass something like `options={{'good-photo': {label: 'Good Photo', active: false}, ...}}`?

sbsmith86: ya! that's what i'm thinking.

chloealee: ah got it - I'll try to fidget with that!

chloealee: @sbsmith86 sorry for all the questions! But just want to make sure I'm understanding correctly. We'll have to be able to change `active` to re-render showing if the tag is active or not so does that mean that this should be added to the `state` in the `CampaignSingle` page  and then passed down as `options={{'good-photo': {label: 'Good Photo', active: this.state.tags.good-photo.active}, ...}}`?

sbsmith86: yea probably. you could have a `filters` property on the state in `CampaignSingle` that the `FilterBar` updates when the user submits the filters to make the api request. It could also be used to determine what the active filters are.

chloealee: I think this would be an issue though because then, the tag buttons would only turn blue once you submit and update the state, not when you click on the actual tag. For some reason, I'm trying to update the state in the `MultiValueFilter` component but it doesn't seem to have access to the state from `CampaignSingle` :-/

sbsmith86: I updated the gist example to show what I am thinking! Check out the updates in `CampaignSingle` and If this really won't work, we can come back to it later. Basically you use the `CampaignSingle` state `tags` proper to render the `MultivalueFilter` component. This will allow React to trigger a re-render of the component when the state gets updated. https://gist.github.com/sbsmith86/f75bd68b6aada49cb35ae090a0c199dd

chloealee: Thanks so much for doing all of this, @sbsmith86 ! So I think the problem is that it still won't update the tag to be active until you hit the submit/filter button. Since the `FilterBar` has it's own state, we won't be able to update `active` in the state until `getPostsByFilter`, which is reached after the submit/filter button is pushed and when it re-renders to pass the state down to the `MultiValueFilter` button to tell if the tag is active or not. Does that make sense? I'm trying to find a way to pass down the state to `FilterBar` so we can update it there in the `updateFilters` function but not sure if you can do this?

chloealee: Maybe we can move `updateFilters` to the `CampaignSingle` page though?

chloealee: Actually I don't think that makes sense...

sbsmith86: That does make sense! but I think I don't understand why that is an issue? When `CampaignSingle` mounts, or grabs a new set of posts, it should update it's own state with the filters it used (including setting if they are active or not) whenever `CampaignSingle` updates it's state it will trigger a re-render down the tree, so `MultvalueFilter` will re-render with the new active/not active values.

chloealee: I think because it doesn't actually update the state of the tag in `updateFilters` since it doesn't have access to the state from `CampaignSingle`? 

Here's a giphy to show more what is happening (when tags are clicked, they aren't active but correct results are shown):
sep-15-2017 15-50-56

chloealee: I can also push up what I have so far if that is helpful!

chloealee: wait @sbsmith86 I think I got it working!! Thanks for your patience with me on this!

chloealee: Well I spoke too soon but getting there! I think I'll push up in a bit to walk through the problem I'm still having!

sbsmith86: ok!

resources/assets/components/FilterBar/index.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    if (['good-photo', 'good-quote', 'hide-in-gallery', 'good-for-sponsor', 'good-for-storytelling'].includes(Object.keys(value)[0])) {
+      let key = Object.keys(value)[0];
+
+      this.setState((previousState) => {
+        const newState = {...previousState};
+        newState.filters.tags[key].active = Object.values(value)[0];
+        return newState;
+      });
```
chloealee: @sbsmith86 everything is working except for this one thing! On the first page load, when I click a tag, it doesn't show up as active but the state updates correctly (saw this when I `console.log`ed the state out). However, when I click `FILTER`, the tag is highlighted as expected. When I click on another tag, it becomes active on the click. This functionality only seems to work after you click `FILTER` and I'm not sure why? Here's a gif to show what's going on: 
sep-18-2017 11-10-28

chloealee:
@sbsmith86 this is ready for another look! Also, @lkpttn I updated based on your updated design! Can you take a look at the below screen shot and let me know if you think anything is super off? (Not going to tackle the "PENDING"/"APPROVED"/"REJECTED"/"DELETED" boxes for this PR so just the "Post Filters" section!).
screen shot 2017-09-18 at 3 45 46 pm

resources/assets/components/FilterBar/index.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import './filter-bar.scss';
+
+import DropdownFilter from '../DropdownFilter';
```
sbsmith86: Don't think you need these anymore since they are used in `CampaignSingle`

chloealee: oh right!

resources/assets/components/FilterBar/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+    this.state = {
+      filters: {},
+    };
+
+    // The component provides a function the children components can use to update the state.
+    this.updateFilters = this.updateFilters.bind(this);
+  }
+
+  updateFilters(values) {
```
sbsmith86: Since you now always pass in a key/value object to this, I think you could just push the full `values` object to the state. i.e the Dropdown changes and pushes `{status: accepted}` so you don't need to check if it is a status, you just push that object to the state.

chloealee: ah yes! Just did some more work to pass the `type` to the `MultiValueFilter` so that the `tags` layer is also included in the state in the `MultiValueFilter` in order for this to work!

chloealee:
and thank you so much for all of your help with this, couldn't have done it without you @sbsmith86!

PR #409, opened 2017-09-15
Newer posts first on single view page

katiecrane:
#### What's this PR do?
All `GET /posts` endpoints return newer posts first. Anywhere using `keyBy` or `map` or anything like that on this data that doesn't have the changes made below shouldn't see any change in the order of the posts. This endpoint needed to return the posts in order because otherwise they will end up sorted per page and not overall.

Then, instead of mapping over all the posts and then creating a `<Post>` for each one, we map over an array of the correctly ordered post IDs and then grab the corresponding information from the list of full posts to build the `<Post>`.

Also, for some reason `this` was out of scope once I made these changes so I declared `props = this.props` as well. That seemed a little weird to me.

Is everything in the right order? Did that break anything else?

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/n/projects/2019429/stories/150689090)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/CampaignSingle/index.js
+++ b/resources/assets/components/CampaignSingle/index.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { map, orderBy, find } from 'lodash';
+import { map, find } from 'lodash';
 import { RestApiClient} from '@dosomething/gateway';
 import { extractSignupsFromPosts } from '../../helpers';
 
@@ -91,7 +91,6 @@ class CampaignSingle extends React.Component {
 
   render() {
     const posts = this.props.posts;
-    const postOrder = map(orderBy(posts, 'created_at', 'desc'), 'id');
     const campaign = this.props.campaign;
     const signups = this.props.signups;
 
@@ -105,7 +104,7 @@ class CampaignSingle extends React.Component {
           <div className="spinner"></div>
         :
 
-          map(postOrder, (key, value) => {
+          map(this.props.postIds, (key, value) => {
             var post = find(posts, {'id': key});
 
             return <Post key={key}
--- a/resources/assets/components/CampaignSingle/index.js
+++ b/resources/assets/components/CampaignSingle/index.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { map } from 'lodash';
+import { map, orderBy, find } from 'lodash';
 import { RestApiClient} from '@dosomething/gateway';
 import { extractSignupsFromPosts } from '../../helpers';
 
@@ -91,8 +91,10 @@ class CampaignSingle extends React.Component {
 
   render() {
     const posts = this.props.posts;
+    const postOrder = map(orderBy(posts, 'created_at', 'desc'), 'id');
     const campaign = this.props.campaign;
     const signups = this.props.signups;
+    const props = this.props;
 
     return (
       <div className="container">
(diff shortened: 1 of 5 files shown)


resources/assets/components/CampaignSingle/index.js
```diff
@@ -91,8 +91,10 @@ class CampaignSingle extends React.Component {
 
   render() {
     const posts = this.props.posts;
+    const postOrder = map(orderBy(posts, 'created_at', 'desc'), 'id');
```
DFurnes: This is fantastically simple. 🎉

resources/assets/components/CampaignSingle/index.js
```diff
@@ -103,23 +105,25 @@ class CampaignSingle extends React.Component {
         {this.props.loading || this.state.loadingNewPosts ?
           <div className="spinner"></div>
         :
-          map(posts, (post, key) =>
-            <Post key={key}
+
+          map(postOrder, function (key, value) {
```
DFurnes: Ah, JavaScript strikes again! When using `function(key, value) {` style function declarations, JavaScript creates a new `this` that points to that function (which is rather confusing). If you switch this to use an arrow function `(key, value) => {` you should still be able to access the class's `this` as expected. 😄

katiecrane: @DFurnes Ahhh I read that but I was having trouble getting it to work with an arrow function for some reason. I'll give it another shot and shoot up the bat signal if I can't get it :bat: ⚾️

katiecrane: Ah! I think it's just because I had a syntax error that I didn't figure out until after I switched it to being a regular `function`.

resources/assets/components/CampaignSingle/index.js
sbsmith86: wait I think I am still confused, why do we have to sort on the frontend again here?

sbsmith86:
I feel like we should do this on every page that displays posts. I know the card is just for the Single Campaign view so I'm ok with us just creating new cards for it, if you feel that is out of scope.

resources/assets/components/CampaignSingle/index.js
katiecrane: When we use `keyBy` here: https://github.com/DoSomething/rogue/blob/master/resources/assets/components/WithReviewing/index.js#L47
they get ordered by `post_id` ascending so we need to sort them again here in order to display them in the right order

katiecrane:
@sbsmith86 would the Inbox be the only other page we need to do this on? That one would be really similar so I think I may as well add it here if we want to do that

resources/assets/components/CampaignSingle/index.js
DFurnes: Can we create the two separate `posts` (keyed by ID) and `postIds` (ordered array) in that `WithReviewing` component? That'd also help make it more easily accessible other places where we want this behavior.

katiecrane: @DFurnes would that be adding `postIds` to the state or making a method like `getPostOrder` to return `postIds`?

sbsmith86: I think you would just have to add `postsIds` to the state in `WithReviewing` the component it wraps should have access to `postIds` as a prop.

DFurnes: Yep, what Shae said!

resources/assets/components/WithReviewing/index.js
```diff
@@ -61,6 +62,7 @@ const reviewComponent = (Component, data) => {
       this.setState({
         campaign: data.campaign,
         posts: posts,
+        postIds: map(orderBy(posts, 'created_at', 'desc'), 'id'),
```
DFurnes: Now that we're using the JSON response for this (rather than the keyed object), do we still need the `orderBy`?

katiecrane: OMG YOU ARE RIGHT WE DO NOT 🔥 !!!!

PR #410, opened 2017-09-15
Update APP_URL example

sbsmith86:
#### What's this PR do?

Updates `.env-example` with an example `APP_URL` setting so it is clear that this needs to be set. 

👀 

I am using `http://rogue.app` as the example to be consistent with Homestead documentation, but since this is based largely on how the user configures their environment we could do something like `APP_URL=YourLocalApp.url`

#### Relevant tickets

No ticket, but will help with an easier setup process.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,7 +2,7 @@ APP_ENV=local
 APP_KEY=SomeRandomString
 APP_DEBUG=true
 APP_LOG_LEVEL=debug
-APP_URL=http://localhost
+APP_URL=http://rogue.app
 ROGUE_API_KEY=abc123
 
 # Feature Flags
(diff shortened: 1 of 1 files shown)


DFurnes:
To make setup as easy as possible, I'd really like this to be consistent with the URL we suggest people use in the [Homestead setup docs](https://github.com/DoSomething/communal-docs/blob/master/Homestead/readme.md). Do we want to change those docs to suggest that people use `.app` domains?

sbsmith86:
Don't have strong opinions either way but will update to `.dev`

DFurnes:
Ok! Yeah, I'm not really sure how we ended up using something else. 👍 from me!

PR #411, opened 2017-09-15
Rotation

sbsmith86:
#### What's this PR do?
Adds a little rotate button to the `Post` component to allow admins to rotate images.

screen recording 2017-09-15 at 05 15 pm

This also updates which url we are using to display the post images. We were using the `original_image_url` which is large and mostly un-cropped. We now use the `url` which is the `edited_{postid}.jpeg` url that was generated by @katiecrane image script. 

commit-by-commit should be safe 

The edited image is the image we ultimately rotate an overwrite in storage. This might break on staging if we haven run the image script there yet because the edited image urls don't exist. But we can just run it there to get everything up to date. 
#### Relevant tickets

Fixes [#150746155](https://www.pivotaltracker.com/story/show/150746155
)

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/ImagesController.php
```diff
@@ -58,4 +61,30 @@ public function show(Post $post, Request $request)
...
+        $originalImage = $post->getMediaUrl();
+
+        // Only supports rotation, right now.
+        if ($request->input('rotate')) {
+            $value = (int) -$request->input('rotate');
+
+            $originalImage = Image::make($originalImage)->rotate($value)->encode('jpg', 75);
+        }
+
+        $image = $this->aws->storeImageData((string) $originalImage, 'edited_' . $post->id);
```
DFurnes: Is this also rotating the original image? The `edited_` files only make sense to keep around until we turn on Glide.

sbsmith86: oh it doesn't! But I think I am still confused about the difference between the two. I thought the `edited_` images were the smaller, cropped images that we should be using for display and returning in the API and the original images we shouldn't even work with.

DFurnes: The `edited_` images were just a "quick fix" to get cropped images until Glide was ready to ship, and so once Glide is enabled everywhere we can just delete them. We want to use the original source as the base image for those future Glide-powered crops, since it gives us the highest quality material in case we ever want larger formatted images.

DFurnes: So for now, let's rotate both – the "user facing" `edited_` images and the "source of truth" original versions.

sbsmith86: AH yes yes now i remember. Will do, thanks!

sbsmith86:
@DFurnes updated!

app/Http/Controllers/ImagesController.php
```diff
@@ -65,26 +65,60 @@ public function show(Post $post, Request $request)
...
+        $originalFilename = $this->getFilenameFromUrl($originalImage);
+
+        // Get the url of the processed, edited image.
+        $editedImage = $post->getMediaUrl();
 
         // Only supports rotation, right now.
         if ($request->input('rotate')) {
             $value = (int) -$request->input('rotate');
 
             $originalImage = Image::make($originalImage)->rotate($value)->encode('jpg', 75);
```
DFurnes: Let's not re-encode the original image as a JPEG, so that if we have a higher quality PNG we don't lose anything (and so that multiple rotations don't re-compress the image each time).

resources/assets/components/Post/index.js
```diff
@@ -38,6 +38,8 @@ class Post extends React.Component {
         const newState = {...prevState};
 
         newState.loading = false;
+        // Add a cache-busting string to the end of the image url
+        // so that it changes and triggers a re-render.
         newState.post.media.url = `${json.url}?time=${new Date()}`;
```
DFurnes: We'll have to think about how to handle this more generally when the [caching updates](https://github.com/DoSomething/devops/issues/321#issuecomment-329545441) kick in (since anyone else re-loading the same image URL would get the cached Fastly response then) We could either include a cache tag for each image and tell Fastly to clear it when we make a change, or maybe by always include a string like this in the image URL based on `updated_at` timestamp.

Either way, definitely not something we need to handle right now!

sbsmith86: For sure. Will make a note to test this out when the caching updates are made

sbsmith86:
@DFurnes updated without encoding the original image!

PR #417, opened 2017-09-19
Blink payload update

sbsmith86:
#### What's this PR do?
Fixes the post payload we send to blink to grab the `campaign_run_id` from the signup since we don't store this on the post. 

Also casts these ids as `Strings` because for some reason the signup payload returns them as strings so I made everything be a string for consistency. 

👀 

Fixes an issue the @sergii-tkachenko flagged where the post payload was returning`null` campaign run ids.

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/151226676

#### Checklist
- [ ] Tested on staging.


sergii-tkachenko:
Thank you!

PR #418, opened 2017-09-19
Tag loading state bug

sbsmith86:
#### What's this PR do?
Fixes bugs around tagging.

These bugs were introduced with the rotation work. 

We pass in a `post` prop to the `Post` component and everytime a tag gets clicked we re-render the post based on the `post` property. 

The issue was after the rotation work, I set a `state.props` property so that we can update the media url when the rotation was finished. 

Now, when the `post` prop was updated, the state never got the change, so the tags did not update their active/inactive styles.

This uses `componentWillReceiveProps` to update `state.post` when we send in new props. 

Also fixing some styling things that seemed weird with the loading/not-loading states.

commit-by-commit 👀 

It might be worth a little refactor of the `Post` and `Tags` components so this is a little bit more understandable, but this helps us fix the immediate bug. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/151230061

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/Post/index.js
+++ b/resources/assets/components/Post/index.js
@@ -128,7 +128,7 @@ class Post extends React.Component {
         {/* Review block and meta data */}
         <div className="container__block -third">
           <div className="container__row">
-            <ReviewBlock post={post} onUpdate={this.props.onUpdate} onTag={this.props.onTag} deletePost={this.props.deletePost} />
+            <ReviewBlock post={this.props.post} onUpdate={this.props.onUpdate} onTag={this.props.onTag} deletePost={this.props.deletePost} />
           </div>
           <div className="container__row">
             <MetaInformation title="Meta" details={{
--- a/resources/assets/components/Post/index.js
+++ b/resources/assets/components/Post/index.js
@@ -75,7 +75,7 @@ class Post extends React.Component {
     return (
       <div className="post container__row">
         {/* Post Images */}
-        <div className="container__block -third">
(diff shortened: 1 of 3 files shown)


resources/assets/components/Tags/tags.scss
```diff
@@ -15,12 +15,7 @@
   margin: 10px 10px 0 0;
   border-radius: 5px;
 
-  &:hover {
-    background: $blue;
-    color: $white;
-  }
-
-  &:active, &.is-active {
+  &:active, &.is-active, &is-loading {
```
DFurnes: `&is-loading` typo!

resources/assets/components/Post/index.js
```diff
@@ -20,12 +20,18 @@ class Post extends React.Component {
...
-      post: this.props.post
+      post: this.props.post,
     };
 
     this.api = new RestApiClient;
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.post !== this.state.post) {
+      this.setState({ post: nextProps.post });
```
DFurnes: This makes me think that `post` should just be a prop, and not duplicated in the component's state... instead, let the parent handle making the API request to rotate it and send back down the updated data via props.

sbsmith86: yea i think that makes more sense. This definitely should be refactored. Will tackle that this week since I have some bandwidth this sprint!

PR #420, opened 2017-09-21
MultiValueFilter Jest Tests

chloealee:
#### What's this PR do?
- Writes a snapshot test for the `MultiValueFilter` component to make sure the front end renders what we expect. 
- Writes a test to make sure the button becomes active when clicked. 

- Run `npm test` and all tests should pass!

#### Relevant tickets
Writes `MultiValueFilter` test in [this Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/150341644).

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,12 @@
       "@dosomething/babel-config"
     ]
   },
+  "jest": {
+    "moduleNameMapper": {
+      "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/resources/assets/__mocks__/fileMock.js",
+      "\\.(css|scss)$": "identity-obj-proxy"
+    }
+  },
   "dependencies": {
     "@dosomething/forge": "^6.7.4",
     "@dosomething/gateway": "^1.1.5",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -5,28 +5,40 @@
...
   },
   "devDependencies": {
     "@dosomething/babel-config": "^1.0.0",
     "@dosomething/webpack-config": "^2.0.4",
     "babel-jest": "^21.0.2",
+    "enzyme": "^2.9.1",
+    "enzyme-to-json": "^2.0.0",
+    "identity-obj-proxy": "^3.0.0",
     "jest": "^21.0.2",
+    "react-dom": "^15.6.1",
```
DFurnes: Looks like this got duplicated from `dependencies`. Let's remove this line.

chloealee: oops! Thanks for catching that - deleted! :)

chloealee:
@DFurnes hmm weird - it passes on my local but doesn't in the wercker build 🤔

chloealee:
Seems like it is mad because it says `TypeError: Object.values is not a function`. That comes from Javascript thought so not sure why it is saying that?

DFurnes:
Ahh, a surprise! I don't think the version of Node we're using on Wercker supports `Object.values`, which is a [relatively new JavaScript feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values#Browser_compatibility). Install [babel-polyfill](https://babeljs.io/docs/usage/polyfill/) and import it near the top of the [main entry point](https://github.com/DoSomething/rogue/blob/master/resources/assets/app.js) for the application & that should fix things up!

PR #421, opened 2017-09-22
ReviewBlock Jest tests

chloealee:
#### What's this PR do?
Adds Jest tests for the `ReviewBlock` component: 
- Snapshot test to make sure the review block renders "Accept", "Reject", and "Delete" buttons. 
- Snapshot test to make sure the review block renders the Tag buttons for an accepted post.
- Test to make sure that "Accept" button is active when clicked. 
- Test to make sure that "Reject" button is active when clicked. 
- Test to makes ure that tag buttons are active when clicked. 
- Updates button classes to make more sense (i.e. instead of `-accepted accepted` it is now `-accepted is-selected`). Also consolidates some repeptitive css code into one class. 

Make sure all tests are passing and that the login in the tests make sense! 

#### Relevant tickets
Fixes `ReviewBlock` tests in https://www.pivotaltracker.com/n/projects/2019429/stories/150341644

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/ReviewBlock/test.js
+++ b/resources/assets/components/ReviewBlock/test.js
@@ -60,9 +60,6 @@ test('it renders an active Reject button when clicked', () => {
   expect(component.find('.rejected').first().hasClass('button -rejected rejected'));
 });
 
-//@TODO: how to do the delete button?
-// By testing the popup?
-
 test('it renders an active tag button when clicked', () => {
   const post = {
     status: 'pending',
--- a/resources/assets/components/ReviewBlock/test.js
+++ b/resources/assets/components/ReviewBlock/test.js
@@ -4,7 +4,19 @@ import toJson from 'enzyme-to-json';
 
 import ReviewBlock from './index';
 
-test('it renders the review block', () => {
+test('it renders accepted, rejected, and delete buttons', () => {
+    const post = {
+      status: 'pending',
+    };
+
+  const component = shallow(
+    <ReviewBlock post={post} />
+  );
+
+  expect(toJson(component)).toMatchSnapshot();
+});
+
+test('it renders the review block with tags', () => {
     const post = {
       status: 'accepted',
     };
--- a/resources/assets/components/ReviewBlock/test.js
+++ b/resources/assets/components/ReviewBlock/test.js
@@ -27,3 +27,43 @@ test('it renders the review block with tags', () => {
 
   expect(toJson(component)).toMatchSnapshot();
 });
+
+test('it renders an active Accept button when clicked', () => {
+  const post = {
+    status: 'pending',
+  };
+
+  const component = shallow(
+    <ReviewBlock post={post} />
+  );
+
+    // Click the "Accept" button.
+    component.find('.accepted').first().simulate('click');
+
+    // It should now show the active "Accept" button.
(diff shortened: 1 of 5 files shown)


resources/assets/components/ReviewBlock/test.js
```diff
@@ -0,0 +1,78 @@
...
+    <ReviewBlock post={post} />
+  );
+
+  // Click a tag button.
+  component.find('.tag').first().simulate('click');
+
+  // It should now show the active "Reject" button.
+  expect(component.find('.tag').first().hasClass('is-active'));
+});
+
```
chloealee: @DFurnes is it possible to do a test that checks if the `confirm` pop up happens when you click on the `Delete` button?

DFurnes: Hmm, I think you can probably create a mock of that method on the `window`, kinda like we did with the parent component's `updateFilters` callback in `MultiValueFilter`:


```js
const confirmStub = sinon.stub('confirm', window);
confirmStub.returns(true);

// @TODO: simulate the click on the thing that asks for confirmation

expect(confirmStub.calledOnce);
```

resources/assets/components/ReviewBlock/test.js
```diff
@@ -0,0 +1,78 @@
...
+test('it renders an active Accept button when clicked', () => {
+  const post = {
+    status: 'pending',
+  };
+
+  const component = shallow(
+    <ReviewBlock post={post} />
+  );
+
+    // Click the "Accept" button.
```
DFurnes: Indentation is wonky here!

resources/assets/components/ReviewBlock/test.js
```diff
@@ -0,0 +1,78 @@
...
+    component.find('.accepted').first().simulate('click');
+
+    // It should now show the active "Accept" button.
+    expect(component.find('.accepted').first().hasClass('button -accepted accepted'));
```
DFurnes: Oh hmm, this whole `-accepted accepted` and `-rejected rejected` thing is pretty weird. Maybe we can switch those `accepted` and `rejected` classes to `is-selected` or something more clear?

I also noticed most of the [styles](https://github.com/DoSomething/rogue/blob/31cf6bf3a8d882527dde1135d757a464756ffc0f/resources/assets/components/StatusButton/status-button.scss) for those classes are the same, can we perhaps join them into a `outlined-button` class (and only have specific overrides for the hover state)?

resources/assets/components/ReviewBlock/test.js
DFurnes: It looks like that might also be something to test on [`WithReviewing`](https://github.com/DoSomething/rogue/blob/31cf6bf3a8d882527dde1135d757a464756ffc0f/resources/assets/components/WithReviewing/index.js) instead (since that's where the actual `window.confirm` call is made), and just assert that the callback to that method is used here.

chloealee: @DFurnes ok that makes sense! Should I add to the list in [this story](https://www.pivotaltracker.com/n/projects/2019429/stories/150341644) to tackle on a separate PR?

DFurnes: I think that makes sense! 👍

PR #422, opened 2017-09-22
Add query string to posts

sbsmith86:
#### What's this PR do?
This PR does two things: 

1) [Adds a cache-busting timestamp](https://github.com/DoSomething/rogue/commit/bf2b0196054cfe6390a34566f35922e6461e64b4)
 to image urls in the response we return in the Post transformer so that we always get the new image on page reloads. 
    * I'm not sure if this is the best approach since we do really want the cached image for better performance. 
      - One option could be to only return the image with the query string if it is updated. This would require us to actually update the post record to know if it has been updated or not. We currently don't do that. Also that brings up another issue where any updates we make to Post records will cause an event log in the `events` table. So we can talk through what we really want. 

2) A little refactoring of the `Post` component. We define the `rotate` function on the parent component and pass it in as a prop so that it doesn't have to maintain it's own state and can get updated images. 

3) Refactoring of some of the image URL helpers. We had two function to grab the original image url and the edited image url. And both of these functions had to take into account that on some pages we were grabbing posts via the API and getting one response structure, and on other pages we were getting posts directly from the DB and using another response structure. 

Things have normalized a bit and we are always grabbing images via posts return in the API. 

👀 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/150746155

There was an issue where after you rotated and image, if you reloaded the page, you couldn't see your updated image because of caching issues. 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Transformers/PostTransformer.php
+++ b/app/Http/Transformers/PostTransformer.php
@@ -35,9 +35,12 @@ class PostTransformer extends TransformerAbstract
             'id' => $post->id,
             'signup_id' => $post->signup_id,
             'northstar_id' => $post->northstar_id,
+            // Add cache-busting query string to urls to make sure we get the
+            // most recent version of the image.
+            // @NOTE - Remove if we get rid of rotation.
             'media' => [
-                'url' => $post->getMediaUrl() . '?time='. Carbon::now(),
-                'original_image_url' => $post->url . '?time='. Carbon::now(),
+                'url' => $post->getMediaUrl() . '?time='. Carbon::now()->timestamp,
+                'original_image_url' => $post->url . '?time='. Carbon::now()->timestamp,
                 'caption' => $post->caption,
             ],
             'tags' => $post->tagSlugs(),
--- a/app/Http/Transformers/PostTransformer.php
+++ b/app/Http/Transformers/PostTransformer.php
@@ -35,8 +35,8 @@ class PostTransformer extends TransformerAbstract
             'signup_id' => $post->signup_id,
             'northstar_id' => $post->northstar_id,
             'media' => [
-                'url' => $post->getMediaUrl(),
-                'original_image_url' => $post->url,
+                'url' => $post->getMediaUrl() . '?time='. Carbon::now(),
+                'original_image_url' => $post->url . '?time='. Carbon::now(),
                 'caption' => $post->caption,
             ],
             'tags' => $post->tagSlugs(),
--- a/app/Http/Transformers/PostTransformer.php
+++ b/app/Http/Transformers/PostTransformer.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Http\Transformers;
 
+use Carbon\Carbon;
 use Rogue\Models\Post;
 use League\Fractal\TransformerAbstract;
 
--- a/resources/assets/components/Post/index.js
+++ b/resources/assets/components/Post/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { remove, map, clone } from 'lodash';
 import { RestApiClient } from '@dosomething/gateway';
-import { getImageUrlFromProp, getEditedImageUrl, displayCaption } from '../../helpers';
+import { getImageUrlFromPost, getEditedImageUrl, displayCaption } from '../../helpers';
 
 import './post.scss';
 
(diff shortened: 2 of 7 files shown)


app/Http/Transformers/PostTransformer.php
```diff
@@ -34,9 +35,12 @@ public function transform(Post $post)
             'id' => $post->id,
             'signup_id' => $post->signup_id,
             'northstar_id' => $post->northstar_id,
+            // Add cache-busting query string to urls to make sure we get the
+            // most recent version of the image.
+            // @NOTE - Remove if we get rid of rotation.
             'media' => [
-                'url' => $post->getMediaUrl(),
-                'original_image_url' => $post->url,
+                'url' => $post->getMediaUrl() . '?time='. Carbon::now()->timestamp,
```
DFurnes: Can we base this off the post's `updated_at` timestamp, and then `->touch()` it when we rotate a photo?

resources/assets/components/Post/index.js
```diff
@@ -82,22 +66,24 @@ class Post extends React.Component {
...
-            <div className="admin-tools__rotate">
-              <a className="button -tertiary rotate" onClick={(event) => this.rotate(event) }></a>
+              <a href={getImageUrlFromPost(post, 'edited')} target="_blank">Edited Photo</a>
             </div>
+            {this.props.rotate ?
+              <div className="admin-tools__rotate">
+                {this.state.loading ?
+                  <div className="spinner"></div>
+                :
+                  <a className="button -tertiary rotate" onClick={(event) => this.handleClick(event)}></a>
```
DFurnes: If you want, this could be simplified to `onClick={this.handleClick}` 🐭

resources/assets/components/Post/post.scss
```diff
@@ -20,25 +20,47 @@
...
+        background-size: 20px;
+      }
     }
   }
 
   .post__image {
-    width: 300px;
-    height: 300px;
-  }
+    // .spinner {
```
DFurnes: 👾

resources/assets/components/WithReviewing/index.js
```diff
@@ -190,6 +191,26 @@ const reviewComponent = (Component, data) => {
...
+
+      let response = this.api.post(`images/${postId}?rotate=90`);
+
+      return response.then((json) => {
+        this.setState((prevState) => {
+          const newState = {...prevState};
+
+          // Add a cache-busting string to the end of the image url
+          // so that it changes and triggers a re-render.
+          newState.posts[postId].media.url = `${json.url}?time=${new Date()}`;
```
DFurnes: Can this query string be removed now that it's being returned in the response?

resources/assets/helpers.js
```diff
@@ -298,3 +256,77 @@ function stripExifData(image, dv = null) {
   // We can just create a blob with the original data.
   return new Blob([dataView], { type: 'image/jpeg' });
 }
+
+// DEPRECIATED: Uses getImageUrlFromPost() instead.
```
DFurnes: This is the absolute nitpickiest comment, but this should be "deprecated". 🤓

resources/assets/helpers.js
```diff
@@ -298,3 +256,77 @@ function stripExifData(image, dv = null) {
...
+export function getImageUrlFromProp(photoProp) {
+  let photo_url;
+
+  // Sometimes we get the url right on the post and sometimes it is nested under
+  // media (in cases where it goes through the PostTransformer), so handle both cases
+  // @TODO: make sure everything goes through a transformer so we don't need this
```
DFurnes: 👍 👍

app/Http/Transformers/PostTransformer.php
sbsmith86: i'd rather not because I think that would create another event record for an update made to the post, which I guess it is, but i'm not sure that is information we need to record

resources/assets/components/Post/index.js
sbsmith86: :) !!!!

resources/assets/components/WithReviewing/index.js
sbsmith86: This ultimately hits a `image/{post_id}` endpoint that just rotates the image and returns the url we already had for it. So we don't actually update the url. But what I could do is have that return a new url with the time stamp on it, but again we aren't really updating the record in the DB. I went back and forth on whether we should do that, but since we use the model service provider to create an event record for every update, I have to be a little bit more careful that we only log meaningful events.

resources/assets/helpers.js
sbsmith86: whoops! good catch haha

resources/assets/components/WithReviewing/index.js
DFurnes: Mmm gotcha. I like the cleanliness of always returning the URL from one place, but I do get the data trade-off of whether or not we want events for stuff like this.

resources/assets/components/Post/post.scss
DFurnes: Ah haha sorry – can we remove these lines since they're commented out?

app/Http/Transformers/PostTransformer.php
DFurnes: Hmm, that's fair. One downside of using the current timestamp here is that people will never get cached resources, so they'll have to re-download all photos every time they load the page. But maybe that's an okay trade-off for now if we're thinking of removing this functionality once we ship Glide image URLs?

sbsmith86: Yea I was thinking that this would probably relatively temporary. Ultimately, I would like us to re-think some of the event behavior in general so we can have more control of when we fire things off or not. So an update doesn't always trigger a record in the events table. For example, because we touch a signup every time a post is updated, we run into issues with getting event records on signups when nothing has changed, and it is hard to tell what has happened.

sbsmith86:
@DFurnes I know you already approved this, but I made some updates based on your comments, if you want to give it a final look.

PR #424, opened 2017-09-26
Accept opt-out/details param when creating a signup

katiecrane:
#### What's this PR do?
- Added `signup_details` table to hold the opt-out information. I named this table so it can be adapted to store other details in the future, but if we want to make it more affiliate or opt-out specific for now I can swap it!
- When we get a request to create a new signup, check to see if a `'details'` parameter is included. If it is, add the `signup_id` to the list in the `signup_details` table.

Do the names make sense? Is the Signup Service the correct place for this logic? Does this function how we want it to?

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/story/show/151225098)

#### Checklist
- [X] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.


chloealee:
This looks great so far, @katiecrane ! A few questions: 
- Should we write tests for this to make sure that it is storing as expected if the endpoint has that param? 
- Is the idea here to only store the signups that are opted out in order to download the CSV? Do we want any other info. besides the `signup_id`? 
- [The Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151225098) mentioned also including a new column on the signup table. Do we still want that or is it outdated?

katiecrane:
@chloealee great questions!!!
1. Tests are a GREAT idea! I will definitely start working on that now to add to this PR.
2. We want to store signups that are opted-in (or not opted-out I guess). We will want to include: 
`Campaign Id | Campaign Run Id | Northstar Id | First Name | Email | Mobile | Age`
The first 3 we can get from the Signup, and the last 4 we can get from Northstar using the Northstar ID. So that's why I'm thinking we only need to store the `signup_id` because from that we can get everything else we need.
3. I saw that too and the card mostly talked about the new table, so I thought that note was outdated since we would only need one or the other.

sbsmith86:
Great questions @chloealee ! Yea want to +1 the question of why we are adding a new table instead of just a column.

katiecrane:
@sbsmith86 I thought what we said in either grooming or planning was to make this a separate table, am I remembering that wrong? The card also mostly talks about a table and not a column so I went with that. I can definitely switch it to a column and I think it would be less code and more straightforward to do it that way. The downside that I see is that we might have this column on all signups that only ever gets used for 1 or 2 campaigns (possibly).

sbsmith86:
@katiecrane yea when I wrote the card, I was under the assumption that it was a separate table, but after talking with the team at grooming I think we decided to just make it a column on the signup. The idea was that we could store any additional signup option on this table not just opt-in/opt-out information. so you could have a `details` column on the `signups` table that stored a value like `truth-opt-out` and then to create the file they need of opt-INs we would just grab all the signups for the campaign that didn't have `truth-opt-out` in the details column. 

I think the idea to this approach was that we could use it for more than just one or two campaigns. If we wanted to signify something else significant on the signup for a different campaign, we could use this column.

katiecrane:
@sbsmith86 Ah got it!! That makes perfect sense to me and sound super familiar, I think I was confused! Thank you for the explanation!!

app/Models/Signup.php
```diff
@@ -11,7 +11,7 @@ class Signup extends Model
      *
      * @var array
      */
-    protected $fillable = ['id', 'event_id', 'northstar_id', 'campaign_id', 'campaign_run_id', 'quantity', 'quantity_pending', 'why_participated', 'source', 'created_at', 'updated_at'];
```
chloealee: sorry if I missed this! Why are we getting rid of the `event_id` in this table?

katiecrane: ah! it's not actually still in the table so I just removed it since I noticed.

sbsmith86: ooo i think `event_id` was left over here from a past iteration of the DB schema when we had everything tied to an event

chloealee: ah right right - thank you both!

PR #425, opened 2017-09-26
Individual signup page order change and added information 

chloealee:
#### What's this PR do?
- On the user page, the `campaign_id`, `campaign_run_id`, and campaign run start date (if we have it) is exposed to provide the admin with more information: 
screen shot 2017-09-26 at 5 38 31 pm
- The signups are now returned by signups with the most recently created posts at the top in descending order. 
  - The card wanted to sort by more recent campaigns first, and then of the more recent campaigns, sort the ones that the member has RB on first (campaign leads said that they just want to see these first as it's most important to see what they've RB for over everything they've signed up for). However, I was unable to do this because the `/activity` endpoint returns signups (not campaigns). I started to return by most recent signups in descending order but of those results, couldn't figure out a way to then filter by signups that had posts first. I figured this would be helpful to the campaigns team to see the most recent created post first, since that should correlate to the most recent signup. @ngjo is this ok? I'm sure my explanation here is very confusion so happy to chat more IRL with anyone to figure out! 

- Go to a user page (`/users/:northstar_id`)
- The signups should be returned in the order of the most recently created post. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/150849332

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -27,7 +27,7 @@ class ActivityController extends ApiController
     {
         // Create an empty Signup query and eager-load posts, which we
         // can either filter or paginate to retrieve all signup records.
-        if ($request->query('orderByPost') === 'desc') {
+        if ($request->query('orderBy') === 'desc') {
             $query = $this->newQuery(Signup::class)->with('posts')->orderBy('created_at', 'desc');
         } else {
             $query = $this->newQuery(Signup::class)->with('posts');
@@ -37,12 +37,13 @@ class ActivityController extends ApiController
 
         $query = $this->filter($query, $filters, Signup::$indexes);
 
-        // if ($request->query('limit') === 'all') {
+        if (($request->query('limit') === 'all')) {
+            $signups = Signup::where([
+                'northstar_id' => $filters['northstar_id'],
+            ])->with('posts')->orderBy('created_at', 'desc')->get();
 
-        //     dd($request->query['parameters']);
-        //     // unset($request->query('limit'));
-        //     return $this->collection($query, $request);
-        // }
+            return $this->collection($signups);
+        }
 
         return $this->paginatedCollection($query, $request);
     }
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -39,6 +39,7 @@ class ActivityController extends ApiController
 
         if (($request->query('limit') === 'all')) {
             $signups = $query->get();
+
             return $this->collection($signups);
         }
 
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -38,10 +38,7 @@ class ActivityController extends ApiController
         $query = $this->filter($query, $filters, Signup::$indexes);
 
         if (($request->query('limit') === 'all')) {
-            $signups = Signup::where([
-                'northstar_id' => $filters['northstar_id'],
-            ])->with('posts')->orderBy('created_at', 'desc')->get();
-
+            $signups = $query->get();
             return $this->collection($signups);
         }
 
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -28,7 +28,10 @@ class ActivityController extends ApiController
         // Create an empty Signup query and eager-load posts, which we
         // can either filter or paginate to retrieve all signup records.
         if ($request->query('orderBy') === 'desc') {
-            $query = $this->newQuery(Signup::class)->with('posts')->orderBy('created_at', 'desc');
+            // $query = $this->newQuery(Signup::class)->with('posts')->orderBy('posts.created_at', 'desc');
+            $query = Signup::with(['posts' => function ($q) {
+                $q->orderBy('created_at', 'desc');
+            }]);
         } else {
             $query = $this->newQuery(Signup::class)->with('posts');
         }
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -27,7 +27,7 @@ class ActivityController extends ApiController
     {
         // Create an empty Signup query and eager-load posts, which we
         // can either filter or paginate to retrieve all signup records.
-        if ($request->query('orderBy') === 'desc') {
+        if ($request->query('orderByPost') === 'desc') {
             // $query = $this->newQuery(Signup::class)->with('posts')->orderBy('posts.created_at', 'desc');
             $query = Signup::with(['posts' => function ($q) {
                 $q->orderBy('created_at', 'desc');
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -27,10 +27,10 @@ class ActivityController extends ApiController
     {
         // Create an empty Signup query and eager-load posts, which we
         // can either filter or paginate to retrieve all signup records.
+        $query = $this->newQuery(Signup::class)->with('posts');
+
         if ($request->query('orderBy') === 'desc') {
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Api/ActivityController.php
```diff
@@ -27,9 +27,17 @@ public function index(Request $request)
     {
         // Create an empty Signup query and eager-load posts, which we
         // can either filter or paginate to retrieve all signup records.
-        $query = $this->newQuery(Signup::class)->with('posts');
+        if ($request->query('orderByPost') === 'desc') {
+            // $query = $this->newQuery(Signup::class)->with('posts')->orderBy('posts.created_at', 'desc');
```
chloealee: will get rid of this, just wanted to keep for now in case we want to add this back in and go with this order instead!

resources/assets/components/UserOverview/index.js
```diff
@@ -28,7 +28,9 @@ class UserOverview extends React.Component {
     this.api.get('api/v2/activity', {
       filter: {
         northstar_id: id,
-      }
+      },
+      orderByPosts: 'desc',
+      limit: 100,
```
chloealee: is 100 enough for now to expose all signups for a user? I don't think we have any users that have more than 100 signups...

app/Http/Controllers/Api/ActivityController.php
katiecrane: I'm trying to figure out the difference between these two queries. Does the first one just order the posts under the signups in reverse chronological order, and the second (uncommented-out one) actually order the _signups_ in order of which ones own the most recent posts?

resources/assets/components/UserOverview/index.js
katiecrane: This made me really curious so I dug into it! It looks like there are 23 users in Thor with over 100 signups, with the most having 243.

app/Http/Controllers/Api/ActivityController.php
chloealee: oh sorry @katiecrane ! So the first one was just supposed to be  $query = $this->newQuery(Signup::class)->with('posts')->orderBy('created_at', 'desc');` (I was trying to see if `posts.created_at` worked but it didn't!). In that case, that just orders all signups with posts in descending order.

And yes, you're exactly right for the uncommented-out query!

resources/assets/components/UserOverview/index.js
katiecrane: But one of them is Ben. And one of them used to work here. And 2 of them used to be an intern. And one of them is Diane. And, of course, one of them is Nami.

katiecrane: And it's 32 not 23 whoops!

chloealee: oooo @katiecrane smart! I totally should have done that! Wondering if I should do a limit of `500`?

resources/assets/components/SignupCard/index.js
```diff
@@ -19,12 +19,29 @@ class SignupCard extends React.Component {
 
     const signupUrl = `/signups/${signup.signup_id}`;
 
+    var campaign_run_start_date = null;
+
+    for (var key in campaign.campaign_runs.current) {
```
sbsmith86: let's do `for (let key in campaign.campaign_runs.current)` using `let` instead of `var` since it acts as an index that will change throughout the loop

resources/assets/components/UserOverview/index.js
sbsmith86: I wonder if it would be a better approach to allow an `all` option for the `limit` param. So instead of using the default or trying to figure out the highest number we just return everything if `limit=all`

sbsmith86:
> The card wanted to sort by more recent campaigns first, and then of the more recent campaigns, sort the ones that the member has RB on first (campaign leads said that they just want to see these first as it's most important to see what they've RB for over everything they've signed up for). However, I was unable to do this because the /activity endpoint returns signups (not campaigns).

* we do have campaign data in cache that we could pass to the `state` of this page in the `UsersController@show` method and those campaign object could show campaign dates if we still want to sort by most recent campaigns first. So you would sort those in decending order, and use the `campaign_id` to figure out which signups to show first. 

>  I started to return by most recent signups in descending order but of those results, couldn't figure out a way to then filter by signups that had posts first.

* Could you create an array of all signups that do not have an empty `posts` array embedded, and then sort that array and show it first, then have another array of signups with empty `posts` and show those next?

app/Http/Controllers/Api/ActivityController.php
katiecrane: Okay I think the uncommented-out one makes sense to me!

chloealee:
thanks for that suggestions @sbsmith86 ! @ngjo and I synced IRL and for now, we decided to just return most recent signup in descending order for now and then revisit sorting by most recent campaign/with posts if the campaigns team is still adamant about this!

sbsmith86:
That works!

resources/assets/components/UserOverview/index.js
chloealee: I love this idea @sbsmith86 ! I just updated but it's a WIP bc I want to extract a little more. I'll let you know when it's ready for another 👀 !

sbsmith86: Or maybe because that parameter is expecting a number, if it is explicitly set to `null` we give all results?

app/Http/Controllers/Api/ActivityController.php
```diff
@@ -27,11 +27,22 @@ public function index(Request $request)
     {
         // Create an empty Signup query and eager-load posts, which we
         // can either filter or paginate to retrieve all signup records.
-        $query = $this->newQuery(Signup::class)->with('posts');
+        if ($request->query('orderBy') === 'desc') {
+            $query = $this->newQuery(Signup::class)->with('posts')->orderBy('created_at', 'desc');
+        } else {
+            $query = $this->newQuery(Signup::class)->with('posts');
+        }
```
sbsmith86: I think you can clean up these conditionals a bit, since you only need to orderby, if the param is passed

```php
$query = $this->newQuery(Signup::class)->with('posts');

if ($request->query('orderBy') === 'desc') {
  $query = $query->orderBy('created_at', 'desc');
}
```

app/Http/Controllers/Api/ActivityController.php
```diff
@@ -27,11 +27,22 @@ public function index(Request $request)
...
 
         $filters = $request->query('filter');
+
         $query = $this->filter($query, $filters, Signup::$indexes);
 
+        if (($request->query('limit') === 'all')) {
+            $signups = $query->get();
+
+            return $this->collection($signups);
```
sbsmith86: Should this not be paginated?

resources/assets/components/SignupCard/index.js
```diff
@@ -19,12 +19,29 @@ class SignupCard extends React.Component {
 
     const signupUrl = `/signups/${signup.signup_id}`;
 
+    var campaign_run_start_date = null;
+
+    for (let key in campaign.campaign_runs.current) {
```
sbsmith86: I think it might be nice to have this in a function defined on the component and then in the rendering you could just do `const campaignRunStartDate = this.getCampaignStartDate(campaign)`

app/Http/Controllers/Api/ActivityController.php
chloealee: I don't think we want it to be paginated since we want to show all the signups for a user! The issue right now is that it only returns 20 since that is the default on pagination and some users have more than 20 signups.

app/Http/Controllers/Api/ActivityController.php
chloealee: oh yeah duh! I don't know why I did it the other way.

app/Http/Controllers/Api/ActivityController.php
sbsmith86: gotcha

chloealee:
@sbsmith86 would love one final look - just cerated the `getCampaignRunStartDate` per your suggestion!

PR #426, opened 2017-09-27
Add signup details to factory

katiecrane:
#### What's this PR do?
In the Signup factory we want to add some `details` to the signups. Each signup gets `details` assigned at random from the following list:
`[null, null, 'fun-affiliate-stuff', 'i-say-the-tails']`

I put `null` twice because I figure it is closer to reality to have at least 50% `null`, but if we don't care that much about that I can remove it so it is only there once.

When you seed your database, does this fill some values (or some `null`s) into your `signups` database for `details`?

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/story/show/151225172)

#### Checklist
- [ ] Tested on staging.

027c45e2e5e90e91da83b585348f4a21bfc9d4395c2cfc8e87754e41a06853eb

### DIFF
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -55,7 +55,7 @@ $factory->define(Signup::class, function (Generator $faker) {
         'quantity_pending' => $faker->randomNumber(4),
         'why_participated' => $faker->sentence(),
         'source' => 'phoenix-web',
-        'details' => $details_options[rand(0,3)],
+        'details' => $details_options[rand(0, 3)],
     ];
 });
 
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -46,7 +46,6 @@ $factory->defineAs(Post::class, 'rejected', function () use ($factory) {
 $factory->define(Signup::class, function (Generator $faker) {
     $faker->addProvider(new FakerNorthstarId($faker));
     $faker->addProvider(new FakerCampaignId($faker));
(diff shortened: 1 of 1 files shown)


database/factories/ModelFactory.php
```diff
@@ -46,6 +46,7 @@
 $factory->define(Signup::class, function (Generator $faker) {
     $faker->addProvider(new FakerNorthstarId($faker));
     $faker->addProvider(new FakerCampaignId($faker));
+    $details_options = [null, null, 'fun-affiliate-stuff', 'i-say-the-tails'];
```
sbsmith86: Might want to add a note about the extra `null` values. I tried to figure out if there was a nice way of making sure something is picked more than others, but this seems fine to me.

database/factories/ModelFactory.php
```diff
@@ -54,6 +55,7 @@
         'quantity_pending' => $faker->randomNumber(4),
         'why_participated' => $faker->sentence(),
         'source' => 'phoenix-web',
+        'details' => $details_options[rand(0, 3)],
```
sbsmith86: I think you can use `faker->randomElement($details_options)`;

katiecrane: GREAT FIND!!!!!

database/factories/ModelFactory.php
katiecrane: I could also remove it, I don't think the proportions will matter so much for this? And it does look pretty funny.

sbsmith86: yea it's probably not that important!

PR #427, opened 2017-09-27
Include param as array

sbsmith86:
#### What's this PR do?

Allows the API to accept an `include` parameter in a `GET` request as an array _and_ a comma-separated set of strings. 

Gateway JS will parse the passed array and add `&include[]=signup&include[]=siblings` to the query string. 

Apparently, when grabbing query string values, laravel will parse that as an array so we end up with `['signup','siblings']` on the server. 

So all this does is check if we got an array in the query param, if so just use that, else parse the comma separated list. 

commit-by-commit?

I feel like I am lacking a little bit of context on this one, so if there was something else that needed to be done, let me know!

Also, i decided to leave support for the comma-separated version of the parameter just in case anything was using it. 

#### Relevant tickets
https://www.pivotaltracker.com/story/show/150224221

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Traits/TransformsRequests.php
+++ b/app/Http/Controllers/Traits/TransformsRequests.php
@@ -116,14 +116,10 @@ trait TransformsRequests
 
         $resource->setPaginator(new IlluminatePaginatorAdapter($paginator));
 
-        $includes = null;
+        $includes = $request->query('include');
 
-        if ($request->query('include')) {
-            if (is_array($request->query('include'))) {
-                $includes = $request->query('include');
-            } else {
-                $includes = explode(',', $request->query('include'));
-            }
+        if (is_string($includes)) {
+            $includes = explode(',', $request->query('include'));
         }
 
         return $this->transform($resource, $code, [], $includes);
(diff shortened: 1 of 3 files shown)


chloealee:
This looks good to me and I like that you have the check in case if the old way is passed! Would just love @DFurnes to double check in case there was something else he also meant!

app/Http/Controllers/Traits/TransformsRequests.php
```diff
@@ -117,8 +117,13 @@ public function paginatedCollection($query, $request, $code = 200, $meta = [], $
         $resource->setPaginator(new IlluminatePaginatorAdapter($paginator));
 
         $includes = null;
+
         if ($request->query('include')) {
-            $includes = explode(',', $request->query('include'));
+            if (is_array($request->query('include'))) {
```
DFurnes: We can simplify this conditional a bit, since the "special" case is if the query is a CSV. You should also be able to remove the wrapping `if ($request->query('include')) {` since it should return null if not provided.

```php
$includes = $request->query('include');

if (is_string($includes)) {
    $includes = explode(',', $request->query('include'));
}
```

sbsmith86: ah yep haha this is much better.

PR #430, opened 2017-09-28
Signup details file exports

katiecrane:
#### What's this PR do?
- Added new `ExportController` with a `show` function to trigger a download of the signup details csv for the given Campaign. Also includes corresponding route.
- New helper on the backend `getAgeFromBirthdate` to calculate ages based on birthdays.
- In `CampaignsController.php`, check to see if any signups on the given campaign have `details`. This has to be on the backend because we need to check ALL signups, not just ones on the current page. Then, conditionally display the export button (only show if there is something to export).

The button:

image

The part that I'm least sure about is how the front-end decides how to display the button each time. Seems a little wild to run the 
```
$export = Signup::whereNotNull('details')->where('campaign_id', $id)->get();
``` 
query every time a single view page gets hit, and then re-run it if the admin decides to download. Does it make sense to cache that?

For now there are no plans to have multiple types of details per campaign, so we assume any signups under a campaign with details all have the same details.

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/story/show/151225385)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -79,13 +79,11 @@ class CampaignsController extends Controller
     {
         $campaign = $this->campaignService->find($id);
         $totals = $this->campaignService->getPostTotals($campaign);
-        $export = Signup::whereNotNull('details')->where('campaign_id', $id)->get();
 
         return view('pages.campaign_single')
             ->with('state', [
                 'campaign' => $campaign,
                 'initial_posts' => 'accepted',
-                'export' => ! $export->isEmpty(),
                 'post_totals' => [
                     'accepted_count' => $totals->accepted_count,
                     'pending_count' => $totals->pending_count,
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -85,7 +85,7 @@ class CampaignsController extends Controller
             ->with('state', [
                 'campaign' => $campaign,
                 'initial_posts' => 'accepted',
-                'export' => !$export->isEmpty(),
+                'export' => ! $export->isEmpty(),
                 'post_totals' => [
                     'accepted_count' => $totals->accepted_count,
                     'pending_count' => $totals->pending_count,
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -2,7 +2,6 @@
 
 namespace Rogue\Http\Controllers;
 
-use Rogue\Models\Signup;
 use Rogue\Services\Registrar;
 use Rogue\Services\CampaignService;
 
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Http\Controllers;
 
+use Rogue\Models\Signup;
 use Rogue\Services\Registrar;
 use Rogue\Services\CampaignService;
 
@@ -78,11 +79,13 @@ class CampaignsController extends Controller
     {
         $campaign = $this->campaignService->find($id);
         $totals = $this->campaignService->getPostTotals($campaign);
+        $export = Signup::whereNotNull('details')->where('campaign_id', $id)->get();
 
         return view('pages.campaign_single')
             ->with('state', [
                 'campaign' => $campaign,
                 'initial_posts' => 'accepted',
+                'export' => !$export->isEmpty(),
                 'post_totals' => [
                     'accepted_count' => $totals->accepted_count,
                     'pending_count' => $totals->pending_count,
--- a/app/Http/Controllers/ExportController.php
+++ b/app/Http/Controllers/ExportController.php
@@ -35,9 +35,7 @@ class ExportController extends Controller
         // Run the query
         $signups = Signup::whereNotNull('details')->where('campaign_id', $campaignId)->get();
 
-        // Compile the data
-        $results = $this->export->exportSignups($signups, $campaignId);
-
-        return response($results['output'], 200, $results['headers']);
+        // Compile the data and trigger the CSV download
+        return $this->export->exportSignups($signups, $campaignId);
     }
 }
--- a/app/Http/Controllers/ExportController.php
+++ b/app/Http/Controllers/ExportController.php
@@ -36,21 +36,8 @@ class ExportController extends Controller
         $signups = Signup::whereNotNull('details')->where('campaign_id', $campaignId)->get();
 
         // Compile the data
-        $final_results = $this->export->exportSignups($signups);
+        $results = $this->export->exportSignups($signups, $campaignId);
 
-        // Format as csv
-        $output = '';
-        foreach ($final_results as $row) {
-            $output .= implode(',', array_values($row)) . "\n";
-        }
-
-        // Build and return the file
-        $filename = 'export_' . $campaignId . '.csv';
-        $headers = [
-          'Content-Type'        => 'text/csv',
-          'Content-Disposition' => 'attachment; filename="'.$filename.'"',
-        ];
-
-        return response($output, 200, $headers);
+        return response($results['output'], 200, $results['headers']);
     }
 }
--- a/app/Http/Controllers/ExportController.php
+++ b/app/Http/Controllers/ExportController.php
@@ -3,25 +3,26 @@
 namespace Rogue\Http\Controllers;
 
 use Rogue\Models\Signup;
-use Rogue\Services\Registrar;
+use Rogue\Services\ExportService;
 
 class ExportController extends Controller
 {
     /**
-     * Registrar instance
+     * ExportService instance
      *
-     * @var Rogue\Services\Registrar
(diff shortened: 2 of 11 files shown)


sbsmith86:
Just a thought, we should put more context around what is being exported. I'm not sure this needs to be a large button as much as some text about downloading a signup details report or something and a link to download the csv.

katiecrane:
@sbsmith86 I'm on board with that! I tagged @lkpttn in `#rogue` so I expect there will be some design changes/context additions

resources/assets/components/StatusCounter/index.js
```diff
@@ -7,7 +7,15 @@ export default (props) => (
...
               <span className="count">{props.postTotals.pending_count}</span>
               <span className="status">Pending</span>
-              <a className="button -secondary" href={`/campaigns/${props.campaign.id}/inbox`}>Review</a>
+              <div>
+                <a className="button -secondary" href={`/campaigns/${props.campaign.id}/inbox`}>Review</a>
+              </div>
+              {props.download ?
+                <div>
+                  <a className="button -secondary" href={`/exports/${props.campaign.id}`}>📩 Export</a>
+                </div>
```
chloealee: @katiecrane your emoji game is always 🔛 ☝️ !

app/Http/Controllers/ExportController.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        $this->registrar = $registrar;
+    }
+
+    /**
+     * Download the export of signup details for the specified campaign.
+     *
+     * @param int $campaignId
+     */
+    public function show($campaignId)
```
sbsmith86: I would love for us to extract the logic for this particular exported file into an `ExportService` If we wanted to export a file differently for something else we couldn't use this `show@ExportController` method, so if you could call something here like `$this->exportService->exportSignups(data)` we could abstract some of the business logic tied to the truth campaign needs from this route.

resources/assets/components/CampaignSingle/index.js
```diff
@@ -134,7 +135,7 @@ class CampaignSingle extends React.Component {
 
     return (
       <div className="container">
-        <StatusCounter postTotals={this.props.post_totals} campaign={campaign} />
+        <StatusCounter postTotals={this.props.post_totals} campaign={campaign} download={downloads}/>
```
sbsmith86: I don't think the download button should be coupled to the `StatusCounter` component. It probably should be it's own component so it can be displayed separated from the `StatusCounter` whenever we need it.

app/Http/Controllers/ExportController.php
katiecrane: @sbsmith86 I'm into that idea! Would I be extracting the "Compile the data" part out of this method to get the data and then still build the file in here? That was we can make the data be whatever we want each time but keep the file building more centralized?

app/Http/Controllers/CampaignsController.php
```diff
@@ -78,11 +79,13 @@ public function showCampaign($id)
     {
         $campaign = $this->campaignService->find($id);
         $totals = $this->campaignService->getPostTotals($campaign);
+        $export = Signup::whereNotNull('details')->where('campaign_id', $id)->get();
 
         return view('pages.campaign_single')
             ->with('state', [
                 'campaign' => $campaign,
                 'initial_posts' => 'accepted',
+                'export' => ! $export->isEmpty(),
```
sbsmith86: I wonder if we need more control here re: whether to show the export button or not. Since we might want to use the `details` column for things that won't be exported. So, for example, what if we only decide to show the `export` button if `truth-opt-out` is in the `details column for any of the signups. It's a little tied to the business logic, but we can contain that piece in the controller and extract it easily, if we need to.

app/Http/Controllers/ExportController.php
sbsmith86: Yea I think the actual work of what is in the file, columns and such should be extracted. So all you have to do is pass an array of signups, and get the file.

app/Http/Controllers/CampaignsController.php
katiecrane: @sbsmith86 I definitely like the idea of that but I'm wondering if we should wait until a case like that come up to support that? Or are we pretty sure that that is going to happen?

lkpttn:
@katiecrane I really like @sbsmith86's suggestion for context, I'll whip that up today.

app/Http/Controllers/CampaignsController.php
DFurnes: I feel like it can't hurt to show the export button regardless, especially if coupled with your suggestion above about visually de-emphasizing this & providing more context to the admin.

app/Http/Controllers/ExportController.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        // Run the query
+        $signups = Signup::whereNotNull('details')->where('campaign_id', $campaignId)->get();
+
+        // Compile the data
+        $final_results = $this->export->exportSignups($signups);
+
+        // Format as csv
```
DFurnes: Not at all necessary for merge, but you might want to look into using [`league/csv`](http://csv.thephpleague.com) to create this CSV file since I think we might run into some edge-cases here when doing this ourselves (like if a user's name included a `,` for example).

app/Http/Controllers/CampaignsController.php
sbsmith86: I guess i was just concerned with us making the assumption that anytime something is in the `details` field, we want to export something. But we can leave it there if you want. Would definitely push for it being de-emphasized, then.

app/Services/ExportService.php
```diff
@@ -0,0 +1,48 @@
...
+                'mobile' => $northstar_user->mobile,
+                'age' => getAgeFromBirthdate($northstar_user->birthdate),
+            ];
+
+            array_push($final_results, $next_row);
+        }
+
+        return $final_results;
+    }
+}
```
sbsmith86: New line 😬

app/Http/Controllers/ExportController.php
sbsmith86: Sorry i feel like the CSV logic should also be pulled out. I don't think the controller should care about how to build a CSV, especially if we go with @DFurnes suggestion of using a third-party package for it.

katiecrane: ooooo I think I misunderstood your other response! so the `return response($output, 200, $headers);` part stays the same but the file gets built in the `ExportService`?

app/Services/ExportService.php
katiecrane: argh I am so bad at that when creating new files and you are so good at catching it! thanks @sbsmith86 !!

app/Http/Controllers/ExportController.php
sbsmith86: yea. Just having a service that has all the logic around building an export file, would be nice. You could have a method in the service class that just takes in an array of data and spits out a file. I might be nitpicking, but just trying to keep things a little more de-coupled.

app/Http/Controllers/ExportController.php
katiecrane: @DFurnes I feel like I should add this now rather than later so we don't end up in a case where the the file is all mucked up and then we have to hurry to add in that package to handle it better

DFurnes: I think that's a good call! 👍

app/Services/ExportService.php
```diff
@@ -0,0 +1,74 @@
...
+     *
+     * @param array $signups
+     * @param int $campaignId
+     */
+    public function makeCSV($data, $campaignId)
+    {
+        // Format as CSV
+        $output = '';
+        foreach ($data as $row) {
+            $output .= implode(',', array_values($row)) . "\n";
```
DFurnes: Let's handle this with the CSV writer:

```php
$writer = Writer::createFromFileObject(new SplFileObject());
$writer->insertAll($data);

return $writer->output('export_' . $campaignId . '.csv');
```

app/helpers.php
```diff
@@ -78,3 +78,16 @@ function multipleValueQuery($query, $queryString, $filter)
...
+{
+    // Make DateTime for right now
+    $now = new DateTime('now');
+
+    // Make DateTime for birthday
+    $birthday = new DateTime();
+    $birthday->setTimestamp(strtotime($birthdate));
+
+    // Calculate the difference and return the years
+    return $now->diff($birthday)->y;
```
DFurnes: This could be simplified a little with Carbon's [difference methods](http://carbon.nesbot.com/docs/#api-difference):

```php
$birthdate = new Carbon($birthdate);
$now = new Carbon();

return $birthdate->diffInYears($now);
```

resources/assets/components/UserExport/user-export.scss
```diff
@@ -0,0 +1,14 @@
+/**
+ * User Export
+ */
+@import '~@dosomething/forge/scss/toolkit';
+
+.user-export {
+	margin-bottom: $base-spacing * 2;
```
DFurnes: Indentation looks a lil' weird here.

PR #432, opened 2017-09-29
Pagination performance

sbsmith86:
#### What's this PR do?

Changes the pagination method on the `GET /activity`endpoint to default to use Laravel's `simplePaginate` and Fractal's `Cursor`. 

For the most part, this just steals from how [Northstar deals with pagination](https://github.com/DoSomething/northstar/blob/1ce09c3ed300bc7166d4df5446544bbdfad2381d/app/Http/Controllers/Traits/TransformsResponses.php#L93-L118). h/t @DFurnes 

The only thing I changed a bit, was I updated the `paginateCollection` method to take in new `$pagination` param that we set to `cursor` in the `ActivityController`. I did this so that the client requesting data could set an optional `pagination` query param OR we can default to this method in code. I did it this way so that we could update this endpoint's performance without needed other teams to update how they are making the request to get this data.

:eyes:

The hope is that this will fix a lot of the [New Relic Alerts](https://rpm.newrelic.com/accounts/108038/incidents/27182371) that have been popping up around this endpoint. The alert seems to signify that there is a performance issue with running the `select count(*) AS aggregate FROM signups` that was used to include total counts in the meta data. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/151378040

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/ActivityController.php
+++ b/app/Http/Controllers/Api/ActivityController.php
@@ -43,6 +43,6 @@ class ActivityController extends ApiController
             return $this->collection($signups);
         }
 
-        return $this->paginatedCollection($query, $request);
+        return $this->paginatedCollection($query, $request, 200, [], null, 'cursor');
     }
 }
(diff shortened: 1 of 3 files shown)


chloealee:
This is so cool, @sbsmith86 ! Do we need to update any documentation for this?

sbsmith86:
@chloealee YEP good catch. I am going to add the new pagination option to the docs.

PR #433, opened 2017-09-29
Adds GET /events endpoint

chloealee:
#### What's this PR do?
- Adds `GET /events` endpoint 
- Returns either paginated results of all events or all events per signup (`/events?filter[signup_id]=:sid`
- Updates documentation

- Hit `/api/v2/events` and you should get paginated list of all events 
- hit `/api/v2/events?filter[signup_id]=18`and you should get non-paginated results of all events that are signup events for signup 18 (or any :sid)

Not sure if this is the best way to do this or not for the future (when we want to grab all events associated by a post or a different type of post in the future) but figured we can start with this? 

#### Relevant tickets
First part in this [Pivotal card](https://www.pivotaltracker.com/n/projects/2019429/stories/150986550). Wanted to break this up so it didn't turn into a monster PR! 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/EventController.php
+++ b/app/Http/Controllers/Api/EventController.php
@@ -26,13 +26,13 @@ class EventController extends ApiController
     public function index(Request $request)
     {
         $query = $this->newQuery(Event::class);
-
         $filters = $request->query('filter');
 
         if ($filters['signup_id']) {
             $events = $query->where('eventable_id', $filters['signup_id'])->orderBy('created_at', 'desc')->get();
 
             $signupEvents = [];
+
             foreach ($events as $event) {
                 if ($event->eventable_type === "Rogue\Models\Signup") {
                     array_push($signupEvents, $event);
@@ -41,6 +41,7 @@ class EventController extends ApiController
 
             return $this->collection($signupEvents);
         }
+
         return $this->paginatedCollection($query, $request);
     }
 }
--- a/app/Http/Controllers/Api/EventController.php
+++ b/app/Http/Controllers/Api/EventController.php
@@ -26,12 +26,13 @@ class EventController extends ApiController
     public function index(Request $request)
     {
         $query = $this->newQuery(Event::class);
+
         $filters = $request->query('filter');
 
-        if ($filters['signup_id']) {
-            $events = $query->forSignup($filters['signup_id'])->orderBy('created_at', 'desc')->get();
+        $query = $this->filter($query, $filters, Event::$indexes);
 
-            return $this->collection($events);
+        if ($filters['signup_id']) {
+            $query = $query->forSignup($filters['signup_id'])->orderBy('created_at', 'desc');
         }
 
         return $this->paginatedCollection($query, $request);
--- a/app/Http/Controllers/Api/EventController.php
+++ b/app/Http/Controllers/Api/EventController.php
@@ -29,17 +29,9 @@ class EventController extends ApiController
         $filters = $request->query('filter');
 
         if ($filters['signup_id']) {
-            $events = $query->where('eventable_id', $filters['signup_id'])->orderBy('created_at', 'desc')->get();
+            $events = $query->forSignup($filters['signup_id'])->orderBy('created_at', 'desc')->get();
 
-            $signupEvents = [];
-
-            foreach ($events as $event) {
-                if ($event->eventable_type === "Rogue\Models\Signup") {
-                    array_push($signupEvents, $event);
-                }
-            }
-
-            return $this->collection($signupEvents);
+            return $this->collection($events);
         }
 
         return $this->paginatedCollection($query, $request);
--- a/app/Http/Controllers/Api/EventController.php
+++ b/app/Http/Controllers/Api/EventController.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Rogue\Http\Controllers\Api;
+
+use Rogue\Models\Event;
+use Illuminate\Http\Request;
+use Rogue\Http\transformers\EventTransformer;
+
+class EventController extends ApiController
+{
+    /**
+     * Create a controller instance.
+     */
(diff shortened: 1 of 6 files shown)


app/Http/Transformers/EventTransformer.php
```diff
@@ -15,18 +15,22 @@ class EventTransformer extends TransformerAbstract
...
     {
+        $content = [];
+        $contentString = substr($event->content, 1, -1);
+        $contentStringWithNoQuotations = str_replace('"', '', $contentString);
+        $contentArray = explode(',', $contentStringWithNoQuotations);
+
+        foreach ($contentArray as $attribute) {
+            $attributeParts = explode(':', $attribute);
+            $content[$attributeParts[0]] = $attributeParts[1];
+        }
```
sbsmith86: Could we use Attribute Casting to turn this JSON field into an array? https://laravel.com/docs/5.5/eloquent-mutators#array-and-json-casting

chloealee: omg yes thank you @sbsmith86 ! This is so much better I didn't know about this!!

app/Http/Controllers/Api/EventController.php
```diff
@@ -0,0 +1,47 @@
...
+
+            $signupEvents = [];
+
+            foreach ($events as $event) {
+                if ($event->eventable_type === "Rogue\Models\Signup") {
+                    array_push($signupEvents, $event);
+                }
+            }
+
+            return $this->collection($signupEvents);
```
DFurnes: We probably still want to paginate this, since otherwise a signup with a lot of events could break the query (even if we decide not to expose a pagination interface in the admin interface for now).

app/Http/Transformers/EventTransformer.php
```diff
@@ -15,18 +15,22 @@ class EventTransformer extends TransformerAbstract
      */
     public function transform(Event $event)
     {
+        $content = [];
+        $contentString = substr($event->content, 1, -1);
```
DFurnes: This field should be saved as JSON in the database, so we should be able to use `json_decode($event->content)` here to prepare it for the response. In the future, we should look into using either Eloquent's built-in `array` casting or an accessor & mutator to handle this within the `Event` model.

app/Http/Transformers/EventTransformer.php
DFurnes: Nice, glad this worked! 👍

app/Http/Controllers/Api/EventController.php
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public function index(Request $request)
+    {
+        $query = $this->newQuery(Event::class);
+        $filters = $request->query('filter');
+
+        if ($filters['signup_id']) {
+            $events = $query->where('eventable_id', $filters['signup_id'])->orderBy('created_at', 'desc')->get();
+
+            $signupEvents = [];
```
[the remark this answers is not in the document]
chloealee: oooo this is awesome! Thank you @DFurnes !

app/Http/Controllers/Api/EventController.php
chloealee: @DFurnes how do we grab all of the results if the pagination is on?

DFurnes: We don't want to! (Or if we do, we'd want to add pagination on the front-end).

app/Http/Transformers/EventTransformer.php
chloealee: updated this!

app/Http/Controllers/Api/EventController.php
chloealee: ooohhhh I see! So is this bad then too? https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Api/ActivityController.php#L43 (we wanted to show all user activity instead of just the first 20 that were returned - we probably want to paginate this on the front end?

DFurnes: Yeah, we always want to make sure we have a reasonable upper-bound on response size. Otherwise edge-cases with particularly "busy" posts (or a malicious user who just requested `/v2/activity?limit=all` 😱) could hammer our server.

DFurnes: And for a MVP, I don't think it's unreasonable to say "here are the latest 20" history events, or something of that sort. We can always add pagination later if people request it.

chloealee: Makes total sense - thanks for explaining! Made a card for it here: https://www.pivotaltracker.com/n/projects/2019429/stories/151566633 :)

sbsmith86: Totally my fault for suggesting this “all” option. @dfunes thanks!

DFurnes: Not a big deal at all! 😄

chloealee: I'm glad I learned this @sbsmith86 so it was a good mistake!

chloealee: @dfurnes updated!

app/Http/Controllers/Api/EventController.php
```diff
@@ -0,0 +1,39 @@
...
+     */
+    public function index(Request $request)
+    {
+        $query = $this->newQuery(Event::class);
+        $filters = $request->query('filter');
+
+        if ($filters['signup_id']) {
+            $events = $query->forSignup($filters['signup_id'])->orderBy('created_at', 'desc')->get();
+
+            return $this->collection($events);
```
DFurnes: Let's switch this to use `paginatedCollection` before merging!

chloealee: sorry I did that but forgot to push up! 😬

PR #436, opened 2017-10-02
User page error

sbsmith86:
#### What's this PR do?

Fixes two issues related to updates made to the `/activity` endpoint and to showing campaign run information on signup cards. 

* The `/activity` endpoint was updated to use fractal's collection class as `FractalCollection` but I did not update how we were instantiating that class so it took into account the new class name. Not sure how this worked at all in my testing. 

* We were unable to render some `SignupCard`s because there are cases were we can not grab the campaign information. So we need to make sure we have a `campaign` object before trying to use any of it's expected values. 

Go to a user page, and it should load. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/151604916

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Traits/TransformsRequests.php
+++ b/app/Http/Controllers/Traits/TransformsRequests.php
@@ -32,7 +32,7 @@ trait TransformsRequests
      */
     public function collection($data, $code = 200, $meta = [], $transformer = null)
     {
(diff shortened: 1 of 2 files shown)


chloealee:
ah thanks for catching the campaign date error @sbsmith86 !

PR #437, opened 2017-10-02
No results

sbsmith86:
#### What's this PR do?
EMPTY STATES EVERYWHERE. 

This PR adds an "empty state" to 

### Campaign Inboxes https://github.com/DoSomething/rogue/commit/3b82075c94e1dca2029ab0f8243e7e7d4e59440a
image

### Campaign Single pages https://github.com/DoSomething/rogue/commit/42cf292ea332361a2094d1023668ed306d739abc
image

### User pages https://github.com/DoSomething/rogue/commit/121adc2b169d34b7f61fec2ae2be03634abf9fdc

image

### Signup pages https://github.com/DoSomething/rogue/commit/838321951dc3a7a3c541072afd4a4cbb05e52719
image

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/150985503

#### Checklist
- [ ] Tested on staging.


resources/assets/components/CampaignSingle/index.js
```diff
@@ -145,24 +146,29 @@ class CampaignSingle extends React.Component {
...
+                deletePost={this.props.deletePost}
+                showHistory={this.props.showHistory}
+                rotate={this.props.rotate}
+                showSiblings={true}
+                showQuantity={true}
+                allowHistory={true} />;
+            })
+          :
+            <Empty details={{
+              "header":  "There are no results!",
```
DFurnes: It'd be nice to have these each as individual `header` and `copy` props so we can use types to validate their input in the future or set default values if nothing is provided. In general, I'm somewhat wary of sending object props unless giving an API response to a "data" component.

resources/assets/components/Empty/empty.scss
```diff
@@ -0,0 +1,19 @@
+@import '~@dosomething/forge/scss/toolkit';
+
+.empty {
+  text-align: center;
+
+  .image {
```
DFurnes: It's probably a good idea to scope this `.empty__image` so we don't accidentally have a naming conflict if we were ever to add an `.image` class elsewhere.

sbsmith86:
@DFurnes updated! will merge.

PR #439, opened 2017-10-05
History log of events

chloealee:
#### What's this PR do?
Adds a log of up to 20 of the most recent events for a signup. 

- Adds to `EventTransformer` to normalize data. Right now, some `content` objects in the `events` table are returned with `\` (more recent events) while others are not. Because of this, the content column wasn't always being returned an an object but as a string. We want it to always be returned in the same format. 
- When the `HistoryModal` is opened, an api call to the `/events` endpoint is made to get the events data to display in the log. 
- Creates a `parseEventData` function in the `HistoryModal` component to parse through events and make sure we don't return those that nothing actually got updated on the signup (`quantity` or `why_participated`. This is because of the way our events are set up. Right now, if a new post is added, a signup event is created since the post touches the signup but nothing really has changed on the signup properties. 
- Updates `Table` component to be compatible with both events data and campaign data to be used to display both tables (event log and campaign table). 
- Adds  `CampaignRow` and `EventRow` to create different types of rows for the table. 
- Adds `history-modal.scss` so that you can scroll through the modal if there are many events. 
- Screenshot of what the history log looks like: 
screen shot 2017-10-06 at 10 38 54 am

- Make sure you can see the events log on the signup page (`/signups/:sid`), the campaign single page (`/campaigns/:nid`), and the campaign inbox page ('/campaigns/:nid/inbox`). 
- Make sure the events are parsed correctly. 
- For the signup page: 
  - Signup `index.js`'s `showHistory` function make API call to `/events` and passes the signup events to `HistoryModal`.
  - `HistoryModal` parses the events in the `parseEventData` function and sends to the `Table`, along with what `Type` of table to create. 
  - The `Table` sends the data to the `Row` to create the specific rows. 
- For the campaign single and inbox pages: 
  - `WithReviewing` component makes API call `/events` in the `showHistory` function and passes to `CampaignSingle` and `CampaignInbox` pages. Those both pass to `HistoryModal` through a prop. 
  - The rest is the same as above. 

#### Relevant tickets
Fixes [this Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/150986550)

#### Checklist
- [ ] Tested on staging.


resources/assets/components/HistoryModal/index.js
```diff
@@ -36,7 +64,10 @@ class HistoryModal extends React.Component {
           </div>
 
           <h3>Reportback History</h3>
-          <p>table of all the history goes here 📖</p>
+          <p>A log of up to 20 of the most recent signup events! 📖</p>
```
chloealee: @ngjo does this copy make sense? We're going to hold on front end pagination to get more than 20 results (since via the API only returns 20 per page) so wanted to make that note here!

ngjo: hiyah @chloealee ! Can we change it to this:

chloealee: Yes that is so much better, thanks @ngjo ! Updated :) Looks like the below (still need to add user's name instead of NS ID which I'm assuming we want instead?):
screen shot 2017-10-06 at 10 07 10 am

ngjo: @chloealee looking great! And @lkpttn what were you thinking for the User column? Name seems useful for scanning purposes but I could also see admin vs member indication and northstar ID being useful for searching too. 

Either way, we can tackle whatever Luke comes up with in a future (possibly next!) sprint as an enhancement!

chloealee: ok sounds good, thanks for the feedback, @ngjo ! So just leave as northstar id for now?

resources/assets/components/HistoryModal/index.js
```diff
@@ -35,8 +64,11 @@ class HistoryModal extends React.Component {
             </div>
           </div>
 
-          <h3>Reportback History</h3>
-          <p>table of all the history goes here 📖</p>
+          <h3>📖 History 📖</h3>
+          <p>Below shows the 20 most recent changes to the member's signup. This includes changes to the quantity or why. If you need changes beyond the 20 listed here, please reach out to Team Bleed!</p>
+          <div className="container">
```
chloealee: yikes! What does this red mean?

DFurnes: Ahhh yeah, this is just Github being a mess. Their syntax highlighting breaks whenever it encounters a `'` or some other random things in JSX blocks.

resources/assets/components/Signup/index.js
```diff
@@ -78,12 +78,23 @@ class Signup extends React.Component {
   }
 
   // Open the history modal of the given post
-  showHistory(postId, event) {
+  showHistory(postId, event, signupId) {
```
DFurnes: It looks like this component has a `signup_id` prop, is that safe to use here? Or if not, can we get the signup ID from the given `postId` argument?

resources/assets/components/Table/Row.js
```diff
@@ -1,16 +1,80 @@
...
+    }
+
+    return <td className="table__cell">{cell.title}</td>
+  }
+
   render() {
-    const { campaign } = this.props;
-    const campaignUrl = this.props.campaign_id ? `/campaigns/${this.props.campaign_id}` : '/campaigns';
-    const inboxUrl = this.props.campaign_id ? `/campaigns/${this.props.campaign_id}/inbox` : '/campaigns';
+    const content = this.props.type === 'campaigns' ? this.createCampaignRow(this.props.data) : this.createEventRow(this.props.data);
```
DFurnes: This line makes me think we should split this up into separate `CampaignRow` and `EventRow` components, and `Row` should just be responsible for generally formatting tabular data.

resources/assets/components/WithReviewing/index.js
```diff
@@ -76,12 +76,22 @@ const reviewComponent = (Component, data) => {
     }
 
     // Open the history modal of the given post
-    showHistory(postId, event) {
-      event.preventDefault()
+    showHistory(postId, event, signupId) {
```
DFurnes: Not needed for this pull request, but we should see if we can avoid duplicating this in the `Signup` as well.

resources/assets/components/Signup/index.js
chloealee: so right!! thanks for catching that !

resources/assets/components/Table/Row.js
chloealee: this is a great call! I just updated :)

PR #440, opened 2017-10-06
Loading spinner on review buttons

katiecrane:
#### What's this PR do?
spinner

Makes it look like this! I copied how we do this for Tags!

Does this look okay for now? Any JS edits?

I haven't yet figured out how to fix the buttons changing size, but it's the end of the sprint so I wanted to go ahead and push up what I had in case we decide this is okay for now.

#### Relevant tickets
[Pivotal Card](https://www.pivotaltracker.com/story/show/150986287)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/ReviewBlock/index.js
+++ b/resources/assets/components/ReviewBlock/index.js
@@ -4,8 +4,14 @@ import Tags from '../Tags';
 import StatusButton from '../StatusButton';
 
 class ReviewBlock extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { loadingButtons: false };
+  }
+
   setStatus(status) {
-    this.props.onUpdate(this.props.post.id, { status: status });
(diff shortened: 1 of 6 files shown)


ngjo:
@katiecrane Yah, we can fix the button changing thing in a future sprint! I'll make a card for it and link to this. Thanks!

DFurnes:
I think the buttons changing size is probably related to adding & removing that 2px border.

resources/assets/components/ReviewBlock/index.js
```diff
@@ -4,8 +4,14 @@ import Tags from '../Tags';
 import StatusButton from '../StatusButton';
 
 class ReviewBlock extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { loadingButtons: false };
```
DFurnes: Is this state ever used here?

PR #442, opened 2017-10-10
Always return signup when updating a signup or creating another post for a signup

chloealee:
#### What's this PR do?
- Always return a signup when updating a signup or creating another post for a signup so it is consistent. 

👀 

#### Relevant tickets
Fixes this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151852854)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -115,7 +115,7 @@ class PostRepository
      * @param \Rogue\Models\Signup $signup
      * @param array $data
      *
-     * @return Signup|Post
+     * @return Signup
      */
     public function update($signup, $data)
     {
@@ -145,7 +145,9 @@ class PostRepository
 
         // If there is a file, create a new post.
         if (array_key_exists('file', $data)) {
-            return $this->create($data, $signup->id);
+            $post = $this->create($data, $signup->id);
+
+            return $post->signup;
         }
 
         return $signup;
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -115,7 +115,7 @@ class PostRepository
      * @param \Rogue\Models\Signup $signup
      * @param array $data
      *
-     * @return Signup
+     * @return Signup|Post
      */
     public function update($signup, $data)
     {
@@ -145,9 +145,7 @@ class PostRepository
 
         // If there is a file, create a new post.
         if (array_key_exists('file', $data)) {
-            $post = $this->create($data, $signup->id);
-
-            return $post->signup;
+            return $this->create($data, $signup->id);
         }
 
         return $signup;
--- a/app/Services/PostService.php
+++ b/app/Services/PostService.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Services;
 
+use Rogue\Models\Post;
 use DoSomething\Gateway\Blink;
 use Rogue\Repositories\PostRepository;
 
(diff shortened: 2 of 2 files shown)


sbsmith86:
@DFurnes are you sure we don't need to check with sergii on this? I feel like if we always send the `signup`  blink will not register RBs that get created when a user updates their quantity or why participated.

chloealee:
(working on the tests and figuring out why wercker is failing!)

DFurnes:
> @DFurnes are you sure we don't need to check with sergii on this? I feel like if we always send the `signup`  blink will not register RBs that get created when a user updates their quantity or why participated. 

Ahhh @sbsmith86 you're right, this would definitely cause Blink issues. Maybe the quick fix is to just add an `&& $postOrSignup instanceof Post` to the feature flag conditional here, and we can refactor this in the repository later on.

chloealee:
hmm @DFurnes when I do that check to see if it is an `instanceof Post`, it says false because this is what `$data` is at that point: 
```

.array:15 [
  "northstar_id" => "56d9f70da59dbf176b8b45d2"
  "campaign_id" => 1283
  "campaign_run_id" => 1343
  "quantity" => 108
  "why_participated" => "Inventore voluptates adipisci iure aliquam doloremque magni minima. Voluptas cum sit qui. Eaque dolores dolor est laborum accusantium nisi dolorem."
  "num_participants" => null
  "caption" => "Consequatur voluptate laudantium fugiat soluta eos possimus sed."
  "source" => "phpunit"
  "remote_addr" => "143.141.209.63"
  "crop_x" => 0
  "crop_y" => 0
  "crop_width" => 100
  "crop_height" => 100
  "crop_rotate" => 90
  "file" => Illuminate\Http\Testing\File {#1187
    +name: "photo.jpg"
    +tempFile: stream resource {@930
      timed_out: false
      blocked: true
      eof: false
      wrapper_type: "plainfile"
      stream_type: "STDIO"
      mode: "r+b"
      unread_bytes: 0
      seekable: true
      uri: "/tmp/phpOkHSzP"
      options: []
    }
    +sizeToReport: null
    -test: true
    -originalName: "photo.jpg"
    -mimeType: "image/jpeg"
    -size: 91
    -error: 0
    #hashName: null
    path: "/tmp"
    filename: "phpOkHSzP"
    basename: "phpOkHSzP"
    pathname: "/tmp/phpOkHSzP"
    extension: ""
    realPath: "/tmp/phpOkHSzP"
    aTime: 2017-10-10 20:48:27
    mTime: 2017-10-10 20:48:27
    cTime: 2017-10-10 20:48:27
    inode: 922388
    size: 91
    perms: 0100600
    owner: 1000
    group: 1000
    type: "file"
    writable: true
    readable: true
    executable: false
    file: true
    dir: false
    link: false
  }
]
```

chloealee:
One thing that I'm still confused by is why it's just saying it received an `id` instead of `signup_id` when we dumped things out in the terminal @DFurnes ? The [Signup Transformer](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/SignupTransformer.php#L49) (and `https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/PostTransformer.php#L36`) should both return `signup_id`...

DFurnes:
I think the `id` was coming [from the `toBlinkPayload`](https://github.com/DoSomething/rogue/blob/67b3cf8be475c9ca2a09adb0cb61de6de9c13bb8/app/Models/Signup.php#L92-L110) "transformer" on the model.

chloealee:
oo thanks for clarifying, @DFurnes !

chloealee:
thank you for all the help @DFurnes and @sbsmith86 ! This is now passing all the tests! 🎉

PR #443, opened 2017-10-11
Export signups with null details

sbsmith86:
#### What's this PR do?

For this first pass, We only want to export signups that have no details. PN only sends over details if the user _opts out_ of third party messaging. Otherwise we assume they are opted-in. For the pass at thisexport, we only want to see opt-INs not opt-OUTs.

👁 

If we do use this export functionality for more things, we should really think about refactoring this so that we can pass in options about what to actually export. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/151767351

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/ExportController.php
+++ b/app/Http/Controllers/ExportController.php
@@ -33,7 +33,7 @@ class ExportController extends Controller
     public function show($campaignId)
     {
         // Run the query
-        $signups = Signup::whereNotNull('details')->where('campaign_id', $campaignId)->get();
+        $signups = Signup::whereNull('details')->where('campaign_id', $campaignId)->get();
 
         // Compile the data and trigger the CSV download
         return $this->export->exportSignups($signups, $campaignId);


sbsmith86:
@katiecrane `affiliate-opt-out`

app/Http/Controllers/ExportController.php
```diff
@@ -33,7 +33,7 @@ public function __construct(ExportService $export)
     public function show($campaignId)
     {
         // Run the query
-        $signups = Signup::whereNotNull('details')->where('campaign_id', $campaignId)->get();
+        $signups = Signup::whereNull('details')->where('campaign_id', $campaignId)->get();
```
katiecrane: For most campaigns, this will cause the export to include all signups for the campaign. Should we consider grabbing specific details here for now?

katiecrane: oh wait that still wouldn't help because the details we want are `null` hmmmm

sbsmith86: hmm well it's weird cause for this campaign, we only want to export the `NULL` values. But we only want to allow that option if it was offered. I considered trying to update PN to send all opt-ins with some details but that seemed like a bit of a larger lift since we already worked the logic around opting-out. So i'm fine for now, if we just export all signups everywhere, but I would love if we only offered this option on the campaigns where that makes sense.

katiecrane: @sbsmith86 Yeah I started confusing myself with double negatives haha. I'm on board with this for now and eventually only showing the exports where relevant!

sbsmith86: cool. Can I get the final approval? I will add a card for re-working this

katiecrane: @sbsmith86 whoops sorry! of course!

PR #444, opened 2017-10-11
GET /api/v3/signups endpoint

chloealee:
#### What's this PR do?
Creates the `GET /api/v3/signups` endpoint and adds documentation. 

Hit `GET /api/v3/signups` and you should receive all signups in a paginated response. 

#### Relevant tickets
Fixes first bullet point in this [Pivotal Ticket](https://www.pivotaltracker.com/n/projects/2019429/stories/151852986)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Api/SignupsController.php
+++ b/app/Http/Controllers/Api/SignupsController.php
@@ -5,9 +5,7 @@ namespace Rogue\Http\Controllers\Api;
 use Rogue\Services\PostService;
 use Rogue\Services\SignupService;
 use Rogue\Http\Requests\SignupRequest;
-use Illuminate\Http\Request;
 use Rogue\Http\Transformers\SignupTransformer;
-use Rogue\Models\Signup;
 
 class SignupsController extends ApiController
 {
--- a/app/Http/Controllers/Api/SignupsController.php
+++ b/app/Http/Controllers/Api/SignupsController.php
@@ -5,15 +5,12 @@ namespace Rogue\Http\Controllers\Api;
 use Rogue\Services\PostService;
 use Rogue\Services\SignupService;
 use Rogue\Http\Requests\SignupRequest;
+use Illuminate\Http\Request;
 use Rogue\Http\Transformers\SignupTransformer;
+use Rogue\Models\Signup;
 
 class SignupsController extends ApiController
 {
-    /**
-     * @var \League\Fractal\TransformerAbstract;
-     */
-    protected $transformer;
-
     /**
      * The signup service instance.
      *
@@ -38,6 +35,23 @@ class SignupsController extends ApiController
     {
         $this->signups = $signups;
         $this->posts = $posts;
+
+        $this->transformer = new SignupTransformer;
+    }
+
+
+    /**
+     * Returns signups.
+     * GET /signups
+     *
+     * @param Request $request
+     * @return ]Illuminate\Http\Response
+     */
+    public function index(Request $request)
+    {
+        $query = $this->newQuery(Signup::class);
+
+        return $this->paginatedCollection($query, $request);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/Api/SignupsController.php
```diff
@@ -38,6 +35,22 @@ public function __construct(SignupService $signups, PostService $posts)
...
+    }
+
+    /**
+     * Returns signups.
+     * GET /signups
+     *
+     * @param Request $request
+     * @return ]Illuminate\Http\Response
+     */
+    public function index(Request $request)
```
DFurnes: I think it might be a good idea to move this into a new controller (maybe namespaced in `\Rogue\Http\Controllers\Three`?) since it's a "v3" route.

app/Http/Controllers/Api/SignupsController.php
```diff
@@ -10,11 +12,6 @@
 class SignupsController extends ApiController
 {
     /**
-     * @var \League\Fractal\TransformerAbstract;
-     */
-    protected $transformer;
```
DFurnes: Let's make sure we keep this so we're not declaring the field dynamically.

documentation/endpoints/signups.md
```diff
@@ -1,6 +1,105 @@
 ## Signups
 
-Create a Signup
+## Retrieve all Signups
+
+```
+GET /api/v2/signups
```
DFurnes: This should be `GET /api/v3/signups`.

app/Http/Controllers/Api/SignupsController.php
```diff
@@ -38,6 +35,22 @@ public function __construct(SignupService $signups, PostService $posts)
...
+     * GET /signups
+     *
+     * @param Request $request
+     * @return ]Illuminate\Http\Response
+     */
+    public function index(Request $request)
+    {
+        $query = $this->newQuery(Signup::class);
+
+        return $this->paginatedCollection($query, $request);
```
DFurnes: We might want to use a separate transformer for our v3 routes as well, since that would allow us to change the behavior of how this is displayed: for example, resetting default includes (only returning signups, unless `?include=posts` is provided) & changing response structure (`signup_id` → `id`, `signup_source` → `source`).

chloealee: ah that makes a lot of sense, thanks @DFurnes !

app/Http/Transformers/v3SignupTransformer.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace Rogue\Http\Transformers;
+
+use Rogue\Models\Signup;
+use Rogue\Services\Registrar;
+use League\Fractal\TransformerAbstract;
+
+class v3SignupTransformer extends TransformerAbstract
```
chloealee: @DFurnes is this name ok? I'm horrible at naming things! 🙈

DFurnes: Let's put this in a `Rogue\Http\Transformers\Three` namespace like the controller!

app/Http/Transformers/v3SignupTransformer.php
```diff
@@ -0,0 +1,42 @@
...
+     * @return array
+     */
+    public function transform(Signup $signup)
+    {
+        // @TODO - This is temporary. We have migrated data that has stored quanity in the
+        // quanity_pending column on the signup. However, since then we updated the business
+        // logic to store everything in the quanity column and not use the quanity_pending
+        // column at all. We only want to return what is in the quanity_pending column
+        // if is the only place quanity is stored.
+        if (! is_null($signup->quantity_pending) && is_null($signup->quantity)) {
```
DFurnes: One thing that'd be nice is to move this into the model (perhaps as an [accessor](https://laravel.com/docs/5.5/eloquent-mutators#defining-an-accessor) on the `quantity` field) so that we don't need this logic repeated in each of the transformers.

chloealee: great idea! I just updated this :)

PR #446, opened 2017-10-13
Use a cursor to grab signups to lessen the load on memory

sbsmith86:
#### What's this PR do?

Uses eloquent [`cursor`](https://laravel.com/docs/5.5/eloquent#chunking-results) method to iterate through all signups more efficiently. 

👀 

#### Relevant tickets
Fixes 🐛 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/ExportController.php
+++ b/app/Http/Controllers/ExportController.php
@@ -32,10 +32,7 @@ class ExportController extends Controller
      */
     public function show($campaignId)
     {
-        // Run the query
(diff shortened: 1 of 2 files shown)


DFurnes:
I'm a little concerned that we're still going to run into issues by building up an array in-memory. Can we try passing the cursor to `makeCSV` so it can write the rows one-by-one?

sbsmith86:
@DFurnes yes! I wasn't sure if that is where I should focus or not. Thank you for that suggestion.

sbsmith86:
@DFurnes updated!

PR #448, opened 2017-10-13
Add snapshot test for Posts

katiecrane:
#### What's this PR do?
Added a snapshot test for Posts. I only passed the attributes necessary for the tests to pass.

Is there anything else on the Post on the frontend that we should be testing? The only other thing I can think of is rotating the image but I'm not sure that that's something we want to test with this. Should I add tests for different values of `showQuantity` and `allowHistory`?

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/story/show/151647046)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     ]
   },
   "jest": {
+    "testURL": "http://rogue.dev",
     "moduleNameMapper": {
       "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/resources/assets/__mocks__/fileMock.js",
       "\\.(css|scss)$": "identity-obj-proxy"
--- a/resources/assets/components/Post/test.js
+++ b/resources/assets/components/Post/test.js
@@ -1,0 +1,19 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import Post from './index';
+
+it('renders correctly', () => {
+  const post = {
+    caption: 'Here is my awesome caption!',
+  };
+
+  const signup = {
+    signup_id: 7,
+  }
+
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -19,6 +19,7 @@
     ]
   },
   "jest": {
+    "testURL": "http://rogue.dev",
```
chloealee: 🕵️‍♀️

chloealee:
@katiecrane this is awesome!! I think this is good as is since the card just says snapshot test! We might want to add some functionality though for the rotation tools as a different card?

resources/assets/components/Post/test.js
```diff
@@ -0,0 +1,19 @@
...
+  };
+
+  const signup = {
+    signup_id: 7,
+  }
+
+  const tree = renderer.create(
+    <Post post={post} signup={signup}></Post>
+  ).toJSON();
+  expect(tree).toMatchSnapshot();
```
chloealee: This looks great to me, @katiecrane ! @DFurnes just out of curiosity, is there much difference between using `react-test-renderer` vs. `shallow` from `enzyme`?

DFurnes: Mmm, good question! Plain ol' `react-test-renderer` renders the full tree (including any child components), whereas `react-test-renderer/shallow` would work more like Enyzme's `shallow` helper.

I think Enyzme just gives some extra helpers on top of `react-test-renderer/shallow` to make it easier to make assertions on the rendered component, so it's probably not a big deal here (since we're just doing a snapshot test).

chloealee: ah that is good to know! thanks so much, @DFurnes !

katiecrane: most interesting!!

PR #449, opened 2017-10-16
DELETE /api/v3/signups/:signup_id endpoint & tests 

chloealee:
#### What's this PR do?
- Create the `DELETE /api/v3/signups/:signup_id` endpoint and tests
- Adds soft deletes to signup model 
- Adds `deleted_at` column to the `signups` table
- Adds documentation 

- Hit the `DELETE /api/v3/signups/:signup_id` endpoint. 
- You should get the following response:
```
{
    "code": 200,
    "message": "Signup deleted."
}
```
- Check the database to make sure that the `deleted_at` column is filled in.
- All tests should pass! 

#### Relevant tickets
Fixes third task in this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151852986)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -53,4 +53,21 @@ class SignupsController extends ApiController
 
         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
     }
+
+    /**
+     * Delete a signup.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function destroy($id)
+    {
+        $signupDeleted = Signup::findOrFail($id)->delete();
+
+        if ($signupDeleted) {
+            return response()->json(['code' => 200, 'message' => 'Signup deleted.']);
+        } else {
+            return response()->json(['code' => 500, 'message' => 'There was an error deleting the signup']);
+        }
+    }
 }
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -26,6 +26,8 @@ class SignupsController extends ApiController
     public function __construct()
     {
         $this->transformer = new SignupTransformer;
+
+        $this->middleware('auth.api');
     }
 
     /**
@@ -47,20 +49,19 @@ class SignupsController extends ApiController
      * GET /signups/:id
      *
      * @param Request $request
-     * @param int $id
+     * @param Signup $signup
      * @return Illuminate\Http\Response
      */
-    public function show(Request $request, $id)
+    public function show(Request $request, Signup $signup)
     {
-        $signup = Signup::findOrFail($id);
-
         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
     }
 
     /**
      * Delete a signup.
+     * DELETE /signups/:id
      *
-     * @param  Object $signup
+     * @param  \Rogue\Models\Signup $signup
      * @return \Illuminate\Http\Response
      */
     public function destroy(Signup $signup)
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -6,9 +6,12 @@ use Rogue\Models\Signup;
 use Illuminate\Http\Request;
 use Rogue\Http\Controllers\Api\ApiController;
 use Rogue\Http\Transformers\Three\SignupTransformer;
+use Northstar\Http\Controllers\Traits\TransformsResponses;
 
 class SignupsController extends ApiController
 {
+    use TransformsResponses;
+
     /**
      * @var \League\Fractal\TransformerAbstract;
      */
@@ -60,14 +63,10 @@ class SignupsController extends ApiController
      * @param  \Illuminate\Http\Request $request
      * @return \Illuminate\Http\Response
      */
-    public function destroy($id)
+    public function destroy(Signup $signup)
     {
-        $signupDeleted = Signup::findOrFail($id)->delete();
+        $signup->delete();
 
-        if ($signupDeleted) {
-            return response()->json(['code' => 200, 'message' => 'Signup deleted.']);
-        } else {
-            return response()->json(['code' => 500, 'message' => 'There was an error deleting the signup']);
-        }
+        return $this->respond('Signup deleted.', 200);
     }
 }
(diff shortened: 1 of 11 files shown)


tests/Http/Three/SignupTest.php
```diff
@@ -105,4 +105,39 @@ public function testSignupWithIncludedPosts()
...
+     *
+     * @return void
+     */
+    public function testUnauthenticatedUserDeletingASignup()
+    {
+        $signup = factory(Signup::class)->create();
+
+        $response = $this->delete('api/v3/signups/' . $signup->id);
+
+        $response->assertResponseStatus(401);
```
chloealee: When I test this in postman, it throws a `401` response if the `X-DS-Rogue-API-Key` is not present in the header. However, here, it throws a `302` error code and I'm not sure why. I'm wondering if I need to add a function to `BrowserKitTestCase` of `withoutRogueApiKey` similar to `withRogueApiKey` but if it doesn't have this function, it shouldn't have that header and should throw the `401` anyway?

chloealee: I think that this is throwing a `302` because it is trying to re-route the page to the login page when the test runs - but really would like this response like in Postman:
```
{
    "error": {
        "code": 401,
        "message": "Unauthorized."
    },
    "debug": {
        "file": "/home/vagrant/DoSomething.org/rogue/app/Exceptions/Handler.php",
        "line": 85
    }
}
```

DFurnes: I think the problem is that the `delete` test method doesn't send along `Accept: application/json` so we get the "web" response (whereas you're probably including that when you test in Postman). Let's use `deleteJson` instead to make sure we get that header:

```php
$response = $this->deleteJson('api/v3/signups/' . $signup->id);
```

chloealee: ooo yes, thank you so much and for the explanation too, @DFurnes ! That worked!

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -53,4 +53,21 @@ public function show(Request $request, $id)
...
+
+    /**
+     * Delete a signup.
+     *
+     * @param  \Illuminate\Http\Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function destroy($id)
+    {
+        $signupDeleted = Signup::findOrFail($id)->delete();
```
DFurnes: I think we can assume this is always successful (it'll throw an exception otherwise, unless we were using custom model events). I'd also like to start using [route model binding](https://laravel.com/docs/5.5/routing#route-model-binding) for grabbing the relevant resource. Let's add the [respond](https://github.com/DoSomething/northstar/blob/755c26b51f72b36bf205585d447d7abbb6170a36/app/Http/Controllers/Traits/TransformsResponses.php#L123-L141) method from Northstar's `TransformRequests` in Rogue & switch this to:

```php
public function destroy(Signup $signup)
{
    $signup->delete();

    return $this->respond('Signup deleted.', 200);
}
```

app/Models/Signup.php
```diff
@@ -3,9 +3,18 @@
 namespace Rogue\Models;
 
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\SoftDeletes;
 
 class Signup extends Model
 {
+    use SoftDeletes;
```
DFurnes: Let's add a newline after this so it's not touching the following docblock. 👾

app/Http/Controllers/Three/SignupsController.php
chloealee: oh cool! That is a lot better to read!

chloealee: @DFurnes I _think_ I am doing everything correctly to use Northstar's `TransformRequest` but it looks like it can't find it:
```
{
    "error": {
        "code": 500,
        "message": "Trait 'Northstar\\Http\\Controllers\\Traits\\TransformsResponses' not found"
    },
    "debug": {
        "file": "/home/vagrant/DoSomething.org/rogue/app/Http/Controllers/Three/SignupsController.php",
        "line": 13
    }
}
```
I have to probably include that somehow into Rogue, right?

app/Models/Signup.php
chloealee: added!

chloealee:
oops sorry @DFurnes - this comment got hidden! https://github.com/DoSomething/rogue/pull/449#discussion_r145481233

app/Http/Controllers/Three/SignupsController.php
DFurnes: We want to import Rogue's [`TransformRequests` trait](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Traits/TransformsRequests.php) here, you can't require classes across apps! We didn't copy `respond` when we made our version of that class here, so you'll need to copy-paste that from the Northstar source into the corresponding Rogue file.

chloealee: ah right! That's what I was thinking but then thought I was having a brain fart 🤦‍♂️ Also just realized I didn't read your comment correctly that explicitly states that - sorry!!

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -53,4 +56,17 @@ public function show(Request $request, $id)
...
+
+    /**
+     * Delete a signup.
+     *
+     * @param  Object $signup
+     * @return \Illuminate\Http\Response
+     */
+    public function destroy(Signup $signup)
+    {
+        $signup->delete();
```
chloealee: @DFurnes I followed [this tutorial](https://laracasts.com/series/laravel-from-scratch-2017/episodes/9) and I think did everything needed for Route Model Binding however `$signup` here is just an empty signup model, so it's not actually deleting the record. Since we're on 5.4, I think that's all that is needed (in addition to the change made in the `routes` file but not sure why it's not working. Is there something I'm missing

app/Http/routes.php
```diff
@@ -97,4 +97,5 @@
     // signups
     Route::get('signups', 'Three\SignupsController@index');
     Route::get('signups/{id}', 'Three\SignupsController@show');
+    Route::delete('signups/{signup}', 'Three\SignupsController@destroy');
```
chloealee: @DFurnes also changed here from `signups/{id}` to `signups/{signup}`

app/Http/Controllers/Three/SignupsController.php
DFurnes: Hmmm weird! Let's look at this together IRL tomorrow morning.

chloealee: @DFurnes ok! and thanks for that @sbsmith86 ! I saw that somewhere too to add that but that didn't seem to work either :-/

chloealee: @sbsmith86 just in case you were wondering, the reason it wasn't working is because we weren't applying the `api` middleware to these routes which is where we were calling `\Illuminate\Routing\Middleware\SubstituteBindings::class,`. The PR is now updated to use the `api` middleware!

PR #450, opened 2017-10-17
Support Contentful Campaign Ids

sbsmith86:
#### What's this PR do?

Allows Rogue to support Contentful Campaigns directly. Contentful Campaign IDs are strings in the vain of `6LQzMvDNQcYQYwso8qSkQ8`. So this PR updates the DB and some logic to support that. Also Contentful campaigns do not have runs, so this PR removes the requirements around `campaign_run_id`. Also adds tests for sending a request from Contentful. 

I think you can go commit-by-commit. 

@DFurnes would love for you to check out the tests to make sure I am making the correct assumptions about what the request would look like. Mainly that PN would only send over a `campaign_id` with no run. 

This will help PN talk directly to Rogue without needing a PA proxy endpoint. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/151853719

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Requests/PostRequest.php
+++ b/app/Http/Requests/PostRequest.php
@@ -23,7 +23,7 @@ class PostRequest extends Request
     {
         return [
             'northstar_id' => 'required|string',
-            'campaign_id' => 'required|int',
+            'campaign_id' => 'required',
             'campaign_run_id' => 'required|int',
             'caption' => 'string',
             'status' => 'in:pending,accepted,rejected',
--- a/app/Http/Requests/PostRequest.php
+++ b/app/Http/Requests/PostRequest.php
@@ -24,7 +24,7 @@ class PostRequest extends Request
         return [
             'northstar_id' => 'required|string',
             'campaign_id' => 'required',
-            'campaign_run_id' => 'required|int',
+            'campaign_run_id' => 'int',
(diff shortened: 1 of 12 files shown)


app/Http/Requests/PostRequest.php
```diff
@@ -23,8 +23,8 @@ public function rules()
     {
         return [
             'northstar_id' => 'required|string',
-            'campaign_id' => 'required|int',
-            'campaign_run_id' => 'required|int',
+            'campaign_id' => 'required',
+            'campaign_run_id' => 'int',
```
katiecrane: Do you think it's worth it to add [Complex Conditional Validation](https://laravel.com/docs/5.5/validation#conditionally-adding-rules) here to make `campaign_run_id` required when the `campaign_id` is an integer or if that's something we should look at handling elsewhere?

sbsmith86: @katiecrane sorry i missed this! Since we actually want to deprecate `campaign_run_id` entirely so i'm not too concerned with adding a bunch of validation around it.

katiecrane: @sbsmith86 cool!!!! 🔥 🏃

PR #451, opened 2017-10-17
Fixes History Modal & Update Quantity Bug

chloealee:
#### What's this PR do?
While parsing through content column of Events tables, checks if value is set first before trimming to fix 🐛 

Make sure you can open and update history modal (usually with more recent entries). 

For some reason, the content column of the Events table is not getting stored in the same manner as older events (some have spaces of `\` in the response). I'll make another card for this to check out to make sure all of our data is being stored in a normalized manner. 

#### Relevant tickets
Fixes this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/152039043).

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Transformers/EventTransformer.php
+++ b/app/Http/Transformers/EventTransformer.php
@@ -29,8 +29,8 @@ class EventTransformer extends TransformerAbstract
             foreach ($array as $key => $value) {
                 $keyValues = explode(':', $value);
 
-                $key = trim($keyValues[0], '"');
-                $value = trim($keyValues[1], '"');
+                $key = isset($keyValues[0]) ? trim($keyValues[0], '"') : '';
+                $value = isset($keyValues[1]) ? trim($keyValues[1], '"') : '';
                 $content[$key] = $value;
             }
 
--- a/app/Http/Transformers/EventTransformer.php
+++ b/app/Http/Transformers/EventTransformer.php
@@ -29,7 +29,7 @@ class EventTransformer extends TransformerAbstract
             foreach ($array as $key => $value) {
                 $keyValues = explode(':', $value);
 
-                $key = isset($keyValues[0]) ? trim($keyValues[0], '"') : '';
+                $key = trim($keyValues[0], '"');
                 $value = isset($keyValues[1]) ? trim($keyValues[1], '"') : '';
                 $content[$key] = $value;
             }


katiecrane:
> For some reason, the content column of the Events table is not getting stored in the same manner as older events (some have spaces of \ in the response).

Just wanted to make sure that this particular thing is a separate issue not fixed in this PR? Making sure I understand the situation 😃

chloealee:
@katiecrane yes! Sorry for the confusion! We have a separate card for that [here](https://www.pivotaltracker.com/n/projects/2019429/stories/151720494
) - this PR is just a quick fix.

app/Http/Transformers/EventTransformer.php
```diff
@@ -29,8 +29,8 @@ public function transform(Event $event)
             foreach ($array as $key => $value) {
                 $keyValues = explode(':', $value);
 
-                $key = trim($keyValues[0], '"');
-                $value = trim($keyValues[1], '"');
+                $key = isset($keyValues[0]) ? trim($keyValues[0], '"') : '';
+                $value = isset($keyValues[1]) ? trim($keyValues[1], '"') : '';
                 $content[$key] = $value;
```
katiecrane: @chloealee does this error out when `$key` is `''`? can that be an array key?

chloealee: @katiecrane great q! So this line is basically just getting rid of double `""`. Some values were being returned as `"59cff102a0bfad0e3c322e57""` so the `trim` gets rid of the extra `"`. However, if it isn't set, it throws the error `(1/1) ErrorException
Undefined offset: 1`

katiecrane: @chloealee right! what I'm wondering is in the case where `$keyValues[0]` isn't set, then `$key` gets set as `''` right? I'm wondering what happens when you do `$content[$key] = $value;` with `$key` set as `''` and if that works because it seems a little funky to me to have `''` as an array key

chloealee: @katiecrane ah I see what you mean now! Makes total sense! So in this case, I think the key will always be set so I removed the `isset` check for the key and just kept on `$value` - thanks for flagging that!

katiecrane: @chloealee woo! sorry to confuse you at first!

chloealee: not at all, @katiecrane ! I just wasn't reading carefully! 👓

PR #452, opened 2017-10-17
PATCH /api/v3/signups/:signup_id

chloealee:
#### What's this PR do?
- Adds the `PATCH /api/v3/signups/:signup_id` endpoint.
  - Adds a custom validator for this endpoint that either `quantity` or `why_participated` must be present. 
- Updates documentation.
- Adds tests to make sure that successful request to this endpoint works and the validation for this endpoint works. 

- Hit the `PATCH /api/v3/signups/:signup_id` endpoint and make sure that the signup was updated as expected. 
- Also make sure that an event was created for this (in the database and in the history modal log)
- Make sure all tests pass! 

#### Relevant tickets
Fixes the last task in this[ Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151852986)

#### Checklist
- [X ] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -4,6 +4,7 @@ namespace Rogue\Http\Controllers\Three;
 
 use Rogue\Models\Signup;
 use Illuminate\Http\Request;
+use Rogue\Http\Requests\Three\SignupRequest;
 use Rogue\Http\Controllers\Api\ApiController;
 use Rogue\Http\Transformers\Three\SignupTransformer;
 use Rogue\Http\Controllers\Traits\TransformsRequests;
@@ -20,7 +21,7 @@ class SignupsController extends ApiController
     /**
      * Create a controller instance.
      *
-     * @param  PostContract  $posts
+     * @param PostContract  $posts
      * @return void
      */
     public function __construct()
@@ -35,7 +36,7 @@ class SignupsController extends ApiController
      * GET /signups
      *
      * @param Request $request
-     * @return ]Illuminate\Http\Response
+     * @return Illuminate\Http\Response
      */
     public function index(Request $request)
     {
@@ -49,7 +50,7 @@ class SignupsController extends ApiController
      * GET /signups/:id
      *
      * @param Request $request
-     * @param Signup $signup
+     * @param \Rogue\Models\Signup $signup
      * @return Illuminate\Http\Response
      */
     public function show(Request $request, Signup $signup)
@@ -61,7 +62,7 @@ class SignupsController extends ApiController
      * Delete a signup.
      * DELETE /signups/:id
      *
-     * @param  \Rogue\Models\Signup $signup
+     * @param \Rogue\Models\Signup $signup
      * @return \Illuminate\Http\Response
      */
     public function destroy(Signup $signup)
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Three/SignupsController.php
```diff
@@ -53,4 +54,33 @@ public function show(Request $request, $id)
...
+     * Updates a specific signup.
+     * PATCH /signups/:id
+     *
+     * @param SignupRequest $request
+     * @param int $id
+     * @return Illuminate\Http\Response
+     */
+    public function update(SignupRequest $request, $id)
+    {
+        $signup = Signup::findOrFail($id);
```
DFurnes: Let's use route model binding here as well:

```php
public function update(SignupRequest $request, Signup $signup)
{
        // ...
```

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -53,4 +54,33 @@ public function show(Request $request, $id)
...
+
+        if (isset($request->all()['quantity'])) {
```
DFurnes: We can do this with mass-assignment:

```php
// @TODO: Remove `array_filter` with 'only' changes in Laravel 5.5.
$fields = array_filter($request->only('quantity', 'why_participated'));
$signup->update($fields);
```

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -53,4 +54,33 @@ public function show(Request $request, $id)
...
+            $signup->quantity = $request->all()['quantity'];
+        }
+
+        if (isset($request->all()['why_participated'])) {
+            $signup->why_participated = $request->all()['why_participated'];
+        }
+
+        $signup->save();
+
+        if ($signup) {
```
DFurnes: We should return the signup here:

```php
return $this->item($signup);
```

documentation/endpoints/signups.md
```diff
@@ -234,3 +234,29 @@ Example response:
...
+  - **quantity**: (int) required if why_participated is not present.
+    The number of reportback nouns verbed.
+  - **why_participated**: (string) required if quantity is not present.
+    The reason why the user participated.
+  
+Example request body:
+```
+[
+  "quantity" => "200"
+  "why_participated" => "bcuz I luv endpointz"
```
DFurnes: 😆

chloealee: I can't take credit for this - I saw it in another file and copy and pasted and got a giggle out of it so kept it here :) I'm glad you got a laugh out of it too!

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -20,7 +21,7 @@ class SignupsController extends ApiController
     /**
      * Create a controller instance.
      *
-     * @param  PostContract  $posts
+     * @param PostContract  $posts
```
DFurnes: Oh can you kill this `@param` here since we're not actually injecting it?

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -70,4 +71,21 @@ public function destroy(Signup $signup)
...
+
+    /**
+     * Updates a specific signup.
+     * PATCH /signups/:id
+     *
+     * @param SignupRequest $request
+     * @param \Rogue\Models\Signup $signup
+     * @return \Illuminate\Http\Response
+     */
+    public function update(SignupRequest $request, Signup $signup)
```
DFurnes: Let's move this up so it goes `index`, `show`, `edit`, `update`, `destroy` & all the controllers are consistent!

app/Http/Controllers/Three/SignupsController.php
chloealee: oh yes - oops!

app/Http/Controllers/Three/SignupsController.php
chloealee: yes!

PR #454, opened 2017-10-18
Queue exports

sbsmith86:
#### What's this PR do?

This PR moves CSV processing to a queue job that sends the generated file to the authenticated user when done via email.

Example:

image

To follow the logic flow:

The request is sent to `ExportController.php`
The controller dispatches the queue job, `ExportSignups.php`
That job uses the `ExportService.php` to build the actual CSV file
Then the job queues a mailer defined in `ExportDone` which send the email.
`ExportDone` uses the `export_done.blade.php` template to customize the actual output of the email.

* This should help with the memory bottlenecks we are seeing on prod when exporting large sets of data. There are two big wins here: We process the csv in the background so the page is not delayed, and we limit the scope of signups we export to the current run.

* This requires some configuration to use our mandril smtp server for sending the email. We also have to set our queue to `redis`

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/152039163

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/ExportController.php
```diff
@@ -32,7 +42,16 @@ public function __construct(ExportService $export)
      */
     public function show($campaignId)
     {
-        // Compile the data and trigger the CSV download
-        return $this->export->exportSignups($campaignId);
+        // Get the full campaign object.
+        $campaign = $this->campaigns->find($campaignId);
+
+        // Get the email of the authenticated user that made the request.
+        $user = $this->registrar->find(auth()->user()->northstar_id);
```
DFurnes: Does `auth()->id()` work here? (I think the answer might be "no", but curious to double-check.)

app/Services/ExportService.php
```diff
@@ -58,6 +65,6 @@ public function exportSignups($campaignId)
             $writer->insertOne($nextRow);
         }
 
-        return $writer->output('export_' . $campaignId . '.csv');
+        return $writer->__toString();
```
DFurnes: Let's return `(string) $writer` rather than explicitly calling the magic method! 🎩

app/Http/Controllers/ExportController.php
sbsmith86: i did work. Grabbing the authorized user in the controller seemed to be the only place I could get it.

DFurnes: Oh nice!! I was thinking it might return the local numerical ID. Let's switch it! 🆔

sbsmith86: Oh wait! I am sorry, i miss understood the question. I will check if `auth()->id()` works for real. I just was confirming that what I had does work 🤦‍♂️

sbsmith86: Ah ok double checked, and yes, `auth()->id()` returns the local primary key of the user record .

DFurnes: Ok cool, I'll make a note for myself to look into making that more useful here.

PR #456, opened 2017-10-23
Move restart task

sbsmith86:
#### What's this PR do?

Fixes a deploy bug I introduced in #454 where I defined the task to restart the queue in the wrong place. 

👀 

We're using capistrano 3 for Rogue so this fix updates our configuration to work with that. 

#### Relevant tickets
Fixes 🐛 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/capistrano/tasks/laravel.rake
+++ b/lib/capistrano/tasks/laravel.rake
@@ -16,9 +16,17 @@ namespace :laravel do
       end
     end
   end
+
+  desc 'Restart queue'
+  task :restart_queue_worker, :on_error => :continue do
+    on roles(:all) do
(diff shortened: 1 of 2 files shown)


lib/capistrano/tasks/laravel.rake
```diff
@@ -16,9 +16,17 @@ namespace :laravel do
       end
     end
   end
+
+  desc 'Restart queue'
+  task :restart_queue_worker, :on_error => :continue do
+    on roles(:all) do
+      run "ps -ef | grep 'queue:work' | awk '{print $2}' | xargs sudo kill -9"
```
DFurnes: Not super important, but can we do this with [`php artisan queue:restart`](https://laravel.com/docs/5.5/queues#queue-workers-and-deployment) so we don't kill any in-process jobs?

PR #463, opened 2017-10-24
JS Linting

sbsmith86:
#### What's this PR do?
This PR adds JS linting. For linting we are using [eslint](https://eslint.org/) and uses DoSomething.org's shared [eslint configuration](https://github.com/DoSomething/eslint-config) package to define our style standards. 

* We are adding this late into the life of the project so there are many errors. This adds the eslint package and dependencies. It also adds the `eslint-loader` to our webpack config so that it will check for errors when compiling a build. 

However this **will not fail** the build. There will be many errors as we start to clean up these files, so we want to surface them, but we want to continue to be able to work in the meantime. 

👁 

I suggest installing [SublimeLinter](http://www.sublimelinter.com/en/latest/) so that your files can be linted in the editor for a better workflow. 

Just to reiterate, **there will be many errors and warnings** the idea is that we will clean up these files as we are working in them instead of trying to do it in bulk. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/151127764

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,24 @@
+module.exports = {
+  // Use Babel parser when running linter so
+  // we can use newer language features.
+  parser: 'babel-eslint',
+
+  // Extend our DoSomething.org code style.
+  // https://github.com/dosomething/eslint-config
+  extends: '@dosomething/eslint-config',
+
+  env: {
+    // Register Jest global variables.
+    jest: true,
+  },
+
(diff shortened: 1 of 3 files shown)


sbsmith86:
hmm @DFurnes seems to still be failing the wercker build but on my local it doesn't... 🤔

PR #465, opened 2017-10-24
Adds Quantity and Why Participated to Uploader

chloealee:
#### What's this PR do?
- Adds optional quantity and why participated fields in the uploader for admins to populate a user's first post or also change these fields when uploading a post on their behalf. 
- Adds language that warns admins that the post will automatically be approved, that they'll have to add tags in Rogue, and that the user will not receive an email if they upload a photo. These don't look like the wireframes in the [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151067785) but figured we can revisit this when @lkpttn @ngjo are back since the uploader is a little different from these anyway! 

- Go to a user's signup page (`/signups/:signup_id`). 
- Click on "Upload Photo" and you should see the below: 
screen shot 2017-10-24 at 2 33 56 pm
- Try to upload a photo with caption and it should go through. 
- Try to upload another photo with caption, why, and quantity and also should go through and update on the signup page (and database). 

- A few changes from the [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151067785):
  - The last task says "Adding in text to explain that quantity + why are optional if they've provided it before - @luke what's the best language for this!?" Instead of adding this text since @ngjo and @lkpttn are out this week, I added language the placeholders saying that these are optional. We can change these later according to what @lkpttn thinks is best! 
  - Same with the first task - not exactly like the wireframe but we can change these as well later (more info. above).  
  - The [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151067785) also mentions that sometimes, a user will ask admins/campaign leads to fix their "why participated" statement. This PR doesn't allow you to do this - you _must_ upload a photo in order to update the "quantity" or "why participated" in this modal. I think updating the "why" belongs in the "Edit | Show History" modal similar (since you can also update the "quantity" here without adding a new photo). Here's a [placeholder card](https://www.pivotaltracker.com/n/projects/2019429/stories/152252363) to add this and discuss when @ngjo and @lkpttn are back! 

#### Relevant tickets
Fixes this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151067785).

#### Checklist
- [ ] Tested on staging.


resources/assets/components/ReportbackUploader/index.js
```diff
@@ -72,9 +91,17 @@ class ReportbackUploader extends React.Component {
 
             <div className="wrapper">
               <div className="form-item">
-                <label className="field-label" htmlFor="caption">Add a caption for the photo.</label>
+                <label className="field-label" htmlFor="caption">Add a caption for the photo:</label>
                 <input className="text-field" id="caption" name="caption" type="text" onChange={this.handleOnCaptionUpdate} placeholder="60 characters or less" ref={input => (this.caption = input)} />
               </div>
+              <div className="form-item">
+                <label className="field-label" htmlFor="quantity">Update quantity:</label>
+                <input className="text-field" id="quantity" name="quantity" type="text" onChange={this.handleOnQuantityUpdate} placeholder="Optional if it already exists" ref={input => (this.quantity = input)} />
```
chloealee: do we need any validation around this to make sure this is an integer? If any text is submitted here, it changes the quantity to `0`

katiecrane: @chloealee looks like that's the same thing that happens using the history modal too, so I could see validating both of those as a separate task

chloealee: @katiecrane that makes sense - great call! I created a new card for that here: https://www.pivotaltracker.com/n/projects/2019429/stories/152253258

katiecrane: @chloealee one more question for you! I'm a little unfamiliar with how `onChange` works here. It will only call that function if they enter something into that field and then submit the form?

chloealee: yes if the value there changes @katiecrane !

PR #466, opened 2017-10-25
Northstar auth updates

sbsmith86:
#### What's this PR do?

There was a bug in the signup export where we were not pull full user objects from northstar, so we were not populating the CSV with email, mobile, DOB, etc. 

It was happening because we were not authenticating with Northstar correctly. We now have to make calls to northstar `asClient()` using client credentials. This makes the call as if it is Rogue, not the authenticated user. 

I updated this change in all of our northstar calls in the `Registrar` class, updated tests to reflect the change. I also found a real edge case where we don't have user information for real, and the CSV was failing. So in this PR i update to check we have user fields before trying to put them in the CSV. 

:eyes: , test should pass. 

#### Relevant tickets
Fixes 🐛 

#### Checklist
- [ ] Tested on staging.


.env.example
```diff
@@ -39,6 +39,8 @@ S3_URL=yourS3Url
 NORTHSTAR_URL=https://northstar-qa.dosomething.org
 NORTHSTAR_AUTH_ID=
 NORTHSTAR_AUTH_SECRET=
+NORTHSTAR_CLIENT_SECRET=
+NORTHSTAR_CLIENT_ID=
```
katiecrane: @sbsmith86 We used to have this but it was removed [here.](https://github.com/DoSomething/rogue/pull/351/files) Do you know why we need to bring it back?

sbsmith86: Yea after this PR https://github.com/DoSomething/rogue/pull/403, we now should use client authentication to specify that the user that is making the request is Rogue (the application, not the user logged into the application) so we actually need these now.

sbsmith86: And this parallel Gateway PR https://github.com/DoSomething/gateway/pull/83

app/Services/ExportService.php
```diff
@@ -55,13 +55,13 @@ public function exportSignups($campaignId, $campaignRunId)
...
                 'campaign_run_id' => $signup->campaign_run_id,
                 'northstar_id' => $signup->northstar_id,
-                'first_name' => $northstarUser->first_name,
-                'last_name' => $northstarUser->last_name,
-                'email' => $northstarUser->email,
-                'mobile' => $northstarUser->mobile,
-                'zip_code' => $northstarUser->addr_zip,
-                'birthdate' => $northstarUser->birthdate,
-                'age' => getAgeFromBirthdate($northstarUser->birthdate),
+                'first_name' => $northstarUser->first_name ?? 'N/A',
```
chloealee: @sbsmith86 I've never seen `??` before - so cool! Curious how it works? (I tried to google php ?? and all I got were articles about what php is haha)

.env.example
katiecrane: @sbsmith86 thanks for the context! makes sense to me 😄

app/Services/ExportService.php
katiecrane: I had the same thought this shortcut is awesome!!!

katiecrane: I found something that uses `?:` in a similar way, do you know if the two differ at all?

sbsmith86: ah yes yes https://en.wikipedia.org/wiki/Null_coalescing_operator#PHP

sbsmith86: PHP 7 specific tho..

sbsmith86: @katiecrane the main difference is that `??` checks `isset()` instead of just the truthy-ness of a thing.

sbsmith86:
@katiecrane @chloealee if you are all good with this, can I get an official approval?

app/Services/ExportService.php
chloealee: so cool!

PR #467, opened 2017-10-25
Refactors Event Logic 

chloealee:
#### What's this PR do?
- Only creates a signup event if `why_participated`, `quantity`, `quantity_pending`, or `deleted_at` has changed. 
- This prevents creating unnecessary/redundant signup events in the events table when a post is updated but nothing about the actual signup is updated.
- Adds tests to make sure:
  - An event gets created when a signup is created
  - An event gets created when a signup is updated
  - An event gets created when a post is created
  - An event gets created when a post is updated, but a signup event does not get created when this happens.

- Go to a user's signup page. 
- Upload a new post for the user. 
- You should only see one event added in the events table for the new post. 
- Update the quantity of the signup. 
- You should then see a new signup event in the events table. 
- If you reject or approve of a post for this signup, you should only see a review and post event. 

I based this logic off of the fact that we want to create an event for the following: 
- When a signup is created.
- When a signup's `quantity`/`quantity_pending`, `why_participated`, and `deleted_at` are updated/changed. 
- When a post is created. 
- When a post is updated.
- When a review is made.
Is there anything else I'm missing? 

- Adds a `@TODO` in the code for when we move `quantity` to a post instead of signup, we'll want to remove the check in this file. Should I add this on a card somewhere so it doesn't get lost? 
- I think we still want [this code](https://github.com/DoSomething/rogue/blob/master/resources/assets/components/HistoryModal/index.js#L23) to partse the events table data until the data in the `content` column is all standardized. 

#### Relevant tickets
Fixes this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151720823).

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Providers/ModelServiceProvider.php
+++ b/app/Providers/ModelServiceProvider.php
@@ -19,7 +19,7 @@ class ModelServiceProvider extends ServiceProvider
     {
         // When a Signup's why_participated, quantity, quantity_pending, or deleted_at are changed, create an event.
         // @TODO: when we move quantity on the post, we'll want to remove this check below.
-        Signup::saved(function ($signup) {
+        Signup::saving(function ($signup) {
             if ($signup->isDirty('why_participated') || $signup->isDirty('quantity') || $signup->isDirty('quantity_pending') || $signup->isDirty('deleted_at')) {
                 $signup->events()->create([
                     'content' => $signup->toJson(),
--- a/app/Providers/ModelServiceProvider.php
+++ b/app/Providers/ModelServiceProvider.php
@@ -18,6 +18,7 @@ class ModelServiceProvider extends ServiceProvider
     public function boot()
     {
         // When a Signup's why_participated, quantity, quantity_pending, or deleted_at are changed, create an event.
+        // @TODO: when we move quantity on the post, we'll want to remove this check below.
         Signup::saved(function ($signup) {
             if ($signup->isDirty('why_participated') || $signup->isDirty('quantity') || $signup->isDirty('quantity_pending') || $signup->isDirty('deleted_at')) {
                 $signup->events()->create([
--- a/app/Providers/ModelServiceProvider.php
+++ b/app/Providers/ModelServiceProvider.php
@@ -19,7 +19,7 @@ class ModelServiceProvider extends ServiceProvider
     {
         // When a Signup's why_participated, quantity, quantity_pending, or deleted_at are changed, create an event.
         // @TODO: when we move quantity on the post, we'll want to remove this check below.
-        Signup::saving(function ($signup) {
+        Signup::saved(function ($signup) {
             if ($signup->isDirty('why_participated') || $signup->isDirty('quantity') || $signup->isDirty('quantity_pending') || $signup->isDirty('deleted_at')) {
                 $signup->events()->create([
                     'content' => $signup->toJson(),
--- a/app/Providers/ModelServiceProvider.php
+++ b/app/Providers/ModelServiceProvider.php
@@ -17,14 +17,16 @@ class ModelServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-        // When Signups are saved create an event for them.
+        // When a Signup's why_participated, quantity, quantity_pending, or deleted_at are changed, create an event.
         Signup::saved(function ($signup) {
-            $signup->events()->create([
-                'content' => $signup->toJson(),
-                // Use the authenticated user if coming from the web,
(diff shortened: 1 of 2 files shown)


app/Providers/ModelServiceProvider.php
```diff
@@ -17,14 +17,17 @@ class ModelServiceProvider extends ServiceProvider
...
+        // When a Signup's why_participated, quantity, quantity_pending, or deleted_at are changed, create an event.
+        // @TODO: when we move quantity on the post, we'll want to remove this check below.
         Signup::saved(function ($signup) {
-            $signup->events()->create([
-                'content' => $signup->toJson(),
-                // Use the authenticated user if coming from the web,
-                // otherwise use the id of the user in the request.
-                'user' => auth()->user() ? auth()->user()->northstar_id : $signup->northstar_id,
-            ]);
+            if ($signup->isDirty('why_participated') || $signup->isDirty('quantity') || $signup->isDirty('quantity_pending') || $signup->isDirty('deleted_at')) {
```
sbsmith86: does this work? i thought that `isDirty()` only checks if something has been changed before it is `Model::saved()` maybe change that to `Signup::saving()`

sbsmith86:
> Is there anything else I'm missing?

I think we also want to track Post deletions.

So this PR only updates the event logic for signup events but Post events will stay the same?

app/Providers/ModelServiceProvider.php
chloealee: it did work but `Signup::saving()` works too! I changed it to that since it is more readable.

chloealee:
Oh sorry @sbsmith86 ! I just checked and an event is created when a Post is deleted as well! Something weird though is that it's a little hard to tell that it was a delete action. Here's what is in the content column for a delete post event:
```
"{\"id\":307,\"signup_id\":289,\"campaign_id\":\"18\",\"northstar_id\":\"5589c991a59dbfa93d8b45ae\",\"url\":null,\"caption\":\"yo\",\"status\":\"accepted\",\"source\":\"rogue-admin\",\"remote_addr\":\"\",\"created_at\":\"2017-10-26 17:41:04\",\"updated_at\":\"2017-10-27 14:25:51\",\"deleted_at\":null}"
```
Here the `deleted_at` is `null`...

app/Providers/ModelServiceProvider.php
chloealee: hmm actually it seems like `Signup::saving()` breaks the tests bc 
```
21) Tests\Http\Three\SignupTest::testDeletingASignup
Illuminate\Database\QueryException: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'eventable_id' cannot be null (SQL: insert into `events` (`content`, `user`, `eventable_id`, `eventable_type`, `updated_at`, `created_at`) values ("{\"northstar_id\":\"5570b6cea59dbf3b7a8b4567\",\"campaign_id\":1173,\"campaign_run_id\":4611,\"quantity_pending\":7921,\"why_participated\":\"Ut ut hic omnis et ut sint cum perspiciatis.\",\"source\":\"phoenix-web\",\"details\":\"fun-affiliate-stuff\"}", 5570b6cea59dbf3b7a8b4567, , Rogue\Models\Signup, 2017-10-27 14:45:20, 2017-10-27 14:45:20))
```

chloealee: I double checked with testing manually and tests are passing so am going to change back to `Signup::saved()`

sbsmith86:
@chloealee ok!

chloealee:
@sbsmith86 actually found something interesting! When you delete a post, it does create an event, but only because `save` is being called on [this line](https://github.com/DoSomething/rogue/blob/master/app/Repositories/PostRepository.php#L169). Do we want to add more explicitly to create an event when a post is destroyed, if it doesn't go through the API to delete the post? 

I was trying to write an additional test to make sure an event gets created when a post is deleted but it was failing which led me to this conclusion. I think that I should write the test to make sure an event is created when a post is deleted via the API instead of when a post is deleted ($post->deleted();` since this is how our logic works now. I can explicitly write this in the doc block. 

I don't want to create add the creation of an event when a post is deleted (in the `ModelServiceProvider.php` to use `Post::deleted` because when a post is deleted, it will then trigger two events. 

Does that work for you to write the test to make sure an event is created when a post is deleted via the API?

sbsmith86:
@chloealee that works for me!

chloealee:
@sbsmith86 great! Just added that test :)

PR #468, opened 2017-10-26
Allow requests to include flag to not send data to Blink

katiecrane:
#### What's this PR do?
Added the ability to send `dont_send_to_blink` flag with Post and Signup requests. If the flag is true, we do not send anything to Blink. If the flag is false or not included, the data will be set to Blink as long as the other already existing conditions are true (ex. Blink feature flag is true).

Is the logic right?

We don't want to send a bunch of events to Blink when we run the upcoming Signupless Reportback migration.

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/152320855)

#### Checklist
- [X] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.

### DIFF
--- a/app/Services/PostService.php
+++ b/app/Services/PostService.php
@@ -47,7 +47,7 @@ class PostService
         $post = $this->repository->create($data, $signupId);
 
         // Send to Blink unless 'dont_send_to_blink' is TRUE
-        $should_send_to_blink = !(array_key_exists('dont_send_to_blink', $data) && $data['dont_send_to_blink']);
+        $should_send_to_blink = ! (array_key_exists('dont_send_to_blink', $data) && $data['dont_send_to_blink']);
 
         // Save the new post in Customer.io, via Blink.
         if (config('features.blink') && $should_send_to_blink) {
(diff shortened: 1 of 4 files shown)


chloealee:
@katiecrane looks like there are some style ci errors but besides that looks good!

app/Services/PostService.php
```diff
@@ -70,8 +73,11 @@ public function update($signup, $data, $transactionId)
     {
         $postOrSignup = $this->repository->update($signup, $data);
 
+        // Send to Blink unless 'dont_send_to_blink' is TRUE
+        $should_send_to_blink = ! (array_key_exists('dont_send_to_blink', $data) && $data['dont_send_to_blink']);
+
         // Save the new post in Customer.io, via Blink.
-        if (config('features.blink') && $postOrSignup instanceof Post) {
+        if (config('features.blink') && $postOrSignup instanceof Post && isset($data['dont_send_to_blink']) && $should_send_to_blink) {
```
sbsmith86: why do you need this check: `isset($data['dont_send_to_blink'])` since you also have `$should_send_to_blink` that should be enough.

katiecrane: meant to remove that one!

PR #469, opened 2017-10-26
Refresh Bug on Uploader Modal 

chloealee:
#### What's this PR do?
- Adds `updateQuantityOrWhyState` function so when we update the quantity or why participated when uploading a post on behalf of the user, the state on the signup page also updates.
  - This function is needed because we only need to update the state, not make a new call since the submit button in the modal already makes that API call. 
- This fixes the bug that we have right now of when you upload a post with a quantity and/or why, the signup page doesn't reflect this information until the page is refreshed. 

- Go to a user's signup page `/signups/:signup_id`
- Upload a photo with a quantity and/or why participated
- You should see these reflected at the top of the signup page when you close the modal without refreshing the page
- No additional events should be created in the events table

#### Relevant tickets
Fixes this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/152292775).

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/ReportbackUploader/index.js
+++ b/resources/assets/components/ReportbackUploader/index.js
@@ -55,12 +55,16 @@ class ReportbackUploader extends React.Component {
       status: 'accepted',
     };
 
+    var signup = this.props.signup;
+
     if (this.quantity.value) {
       reportback.impact = this.quantity.value;
+      signup.quantity = reportback.impact;
     }
 
     if (this.why_participated.value) {
       reportback.whyParticipated = this.why_participated.value;
+      signup.why_participated = reportback.whyParticipated;
     }
 
     const fileType = reportback.media.file ? reportback.media.file.type : null;
@@ -75,7 +79,7 @@ class ReportbackUploader extends React.Component {
       media: this.defaultMediaState,
     });
 
-    this.props.updateQuantityOrWhyState(reportback.impact ? reportback.impact : null, reportback.whyParticipated ? reportback.whyParticipated : null);
+    this.props.updateSignup(signup);
   }
 
   render() {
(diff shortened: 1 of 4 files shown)


resources/assets/components/Signup/index.js
```diff
@@ -217,6 +217,18 @@ class Signup extends React.Component {
     this.hideHistory();
   }
 
+  // Set newstate with updated quantity and why_participated.
+  updateQuantityOrWhyState(quantity, why) {
```
sbsmith86: it is possible for this to just be `updateSignup(signup)` that just takes in a whole signup object and updates the state with that?

resources/assets/components/ReportbackUploader/index.js
```diff
@@ -74,6 +74,8 @@ class ReportbackUploader extends React.Component {
     this.setState({
       media: this.defaultMediaState,
     });
+
+    this.props.updateQuantityOrWhyState(reportback.impact ? reportback.impact : null, reportback.whyParticipated ? reportback.whyParticipated : null);
```
sbsmith86: Oh not really needed for this PR but this `ReportbackUploader` needs to be renamed to `PostUploader` and should be refactored to just use rogue terminology.

resources/assets/components/Signup/index.js
chloealee: oooo I like that idea! Just updated that!

resources/assets/components/ReportbackUploader/index.js
chloealee: ok! I made a card for that here: https://www.pivotaltracker.com/n/projects/2019429/stories/152350532

PR #470, opened 2017-10-27
Creates GET /api/v3/posts and GET /api/v3/posts/:post_id endpoints 

chloealee:
#### What's this PR do?
- Creates `GET /api/v3/posts` and `GET /api/v3/posts/:post_id` endpoints 
- Creates tests to make sure those both work
- Updates documentation 

- Hit both of those endpoints and they should work! 
- Make sure all tests pass 

#### Relevant tickets
Fixes the first task in this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151852968)

#### Checklist
- [X ] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Rogue\Http\Controllers\Three;
+
+use Rogue\Services\PostService;
+use Rogue\Repositories\SignupRepository;
+use Rogue\Http\Transformers\PostTransformer;
+use Rogue\Http\Controllers\Traits\PostRequests;
+use Rogue\Http\Controllers\Api\ApiController;
+
+class PostsController extends ApiController
+{
+    use PostRequests;
+
+    /**
+     * The post service instance.
+     *
+     * @var \Rogue\Services\PostService
+     */
+    protected $posts;
+
+    /**
+     * The signup repository instance.
+     *
+     * @var \Rogue\Repositories\SignupRepository
+     */
+    protected $signups;
+
+    /**
+     * @var \League\Fractal\TransformerAbstract;
+     */
+    protected $transformer;
+
+    /**
+     * Create a controller instance.
+     *
+     * @param  PostContract  $posts
+     * @return void
+     */
+    public function __construct(PostService $posts, SignupRepository $signups)
+    {
+        $this->posts = $posts;
+        $this->signups = $signups;
+
+        $this->transformer = new PostTransformer;
+    }
+}
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -2,13 +2,13 @@
 
 namespace Rogue\Http\Controllers\Three;
 
+use Rogue\Models\Post;
+use Illuminate\Http\Request;
 use Rogue\Services\PostService;
 use Rogue\Repositories\SignupRepository;
 use Rogue\Http\Transformers\PostTransformer;
-use Rogue\Http\Controllers\Traits\PostRequests;
 use Rogue\Http\Controllers\Api\ApiController;
-use Illuminate\Http\Request;
-use Rogue\Models\Post;
+use Rogue\Http\Controllers\Traits\PostRequests;
 
 class PostsController extends ApiController
 {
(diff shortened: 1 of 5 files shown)


documentation/endpoints/posts.md
```diff
@@ -1,91 +1,9 @@
 ## Posts
 
-## Create a Post and/or Create/Update a Signup
```
katiecrane: I think we still want to document how to create a post for this endpoint because we do still use it

documentation/endpoints/posts.md
```diff
@@ -178,3 +96,118 @@ Example Response:
...
+    "status": "accepted",
+    "source": "rogue-admin",
+    "remote_addr": "",
+    "created_at": "2017-10-27T14:50:22+00:00",
+    "updated_at": "2017-10-27T14:50:22+00:00"
+  }
+}
+```
+
+## Create a Post and/or Create/Update a Signup
```
chloealee: @katiecrane here it is!

documentation/endpoints/posts.md
chloealee: @katiecrane sorry this is a little misleading! It is still here but I just moved it to under the GET endpoint documentation. I pinged you the line where it was moved to! :)

documentation/endpoints/posts.md
katiecrane: @chloealee thanks!!

app/Http/Controllers/Three/PostsController.php
```diff
@@ -0,0 +1,62 @@
...
+use Illuminate\Http\Request;
+use Rogue\Services\PostService;
+use Rogue\Repositories\SignupRepository;
+use Rogue\Http\Transformers\PostTransformer;
+use Rogue\Http\Controllers\Api\ApiController;
+use Rogue\Http\Controllers\Traits\PostRequests;
+
+class PostsController extends ApiController
+{
+    use PostRequests;
```
katiecrane: I know we use `PostRequests` to DRY up this code so we don't have the same thing in a bunch of different places, but I've found working with it confusing. I think we should all re-visit this and see if we still want it. Definitely not needed for this PR since Shae is out and of course we want to hear from her too!

app/Http/Controllers/Three/PostsController.php
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @param  PostContract  $posts
+     * @return void
+     */
+    public function __construct(PostService $posts, SignupRepository $signups)
+    {
+        $this->posts = $posts;
+        $this->signups = $signups;
+
+        $this->transformer = new PostTransformer;
```
katiecrane: Can we create the transformer in the same way as the `PostService` and `SignupRepository` or is it different?

tests/Http/Three/PostTest.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public function testPostShow()
+    {
+        $post = factory(Post::class)->create();
+        $this->get('api/v3/posts/' . $post->id);
+
+        $this->assertResponseStatus(200);
+        $this->seeJsonStructure([
+            'data' => [
+                'id',
```
katiecrane: Would it be worth checking if the `id` returned matches the `id` of the created `post`?

app/Http/Controllers/Three/PostsController.php
chloealee: @katiecrane great q! Now that I'm looking back on this code, I'm having trouble seeing how the transformer gets created in `PostService` and `SignupRepository` :-/ Do you know where this happens? If I comment out this line (47), it breaks though.

katiecrane: @chloealee ah I meant can we create it in the same way that we create the `PostService` in this particular block, not in the `PostService.php` file itself

tests/Http/Three/PostTest.php
chloealee: I don't think it could hurt! Just added :)

app/Http/Controllers/Three/PostsController.php
chloealee: @katiecrane ohohoh I see now! Yes it does! Changed :)

PR #471, opened 2017-10-30
PATCH /api/v3/posts/:post_id

chloealee:
#### What's this PR do?
- Adds the `PATCH /api/v3/posts/:post_id` endpoint 
- Adds corresponding tests for this endpoint 

- Make sure you can update a post's `status` and/or `caption` when hitting this endpoint
- Make sure you can't change a post's `status` to anything but `pending`, `accepted`, or `rejected`
- Make sure tests pass! 

Is there anything else about a post we want to be able to update? 

#### Relevant tickets
Fixes the second task in this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151852968)

#### Checklist
- [ X ] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -71,7 +71,9 @@ class PostsController extends ApiController
      */
     public function update(PostRequest $request, Post $post)
     {
+        // @TODO: Remove `array_filter` with 'only' changes in Laravel 5.5.
         $fields = array_filter($request->only('status', 'caption'));
+
         $post->update($fields);
 
         return $this->item($post);
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -8,6 +8,7 @@ use Rogue\Services\PostService;
 use Rogue\Repositories\SignupRepository;
 use Rogue\Http\Transformers\PostTransformer;
 use Rogue\Http\Controllers\Api\ApiController;
+use Rogue\Http\Requests\Three\PostRequest;
 use Rogue\Http\Controllers\Traits\PostRequests;
 
 class PostsController extends ApiController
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Three/PostsController.php
```diff
@@ -59,4 +60,20 @@ public function show(Request $request, Post $post)
...
+     * Updates a specific post.
+     * PATCH /posts/:id
+     *
+     * @param Request $request
+     * @param \Rogue\Models\Post $post
+     * @return \Illuminate\Http\Response
+     */
+    public function update(PostRequest $request, Post $post)
+    {
+        $fields = array_filter($request->only('status', 'caption'));
```
DFurnes: Can we add a TODO above this line?

```php
// @TODO: Remove `array_filter` with 'only' changes in Laravel 5.5.
$fields = array_filter($request->only('status', 'caption'));
```

app/Http/Requests/Three/PostRequest.php
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Get the validation rules that apply to the request.
+     *
+     * @return array
+     */
+    public function rules()
+    {
+        return [
+            'status' => 'in:pending,accepted,rejected',
```
DFurnes: Should we add any validation to the `caption` field? Perhaps `string|max:140`?

chloealee:
@DFurnes @katiecrane oh I just thought of one more thing! We don't want this update to be sent to Blink, right? And similarly, do we want an update when a signup is update for that restful api endpoint? I feel like we might want it for the signup update but maybe not the post update?

app/Http/Controllers/Three/PostsController.php
chloealee: ah yes! Updated :)

DFurnes:
Yeah, I don't think we want to send events to Customer.io for updates – only on new signups or posts. This should be fine as now since we make that API call in `PostService` and `SignupService`, which aren't used for this endpoint.

app/Http/Requests/Three/PostRequest.php
chloealee: great call, @DFurnes ! I updated this and also the `testValidationUpdatingAPost` test as well to incorporate this.

PR #472, opened 2017-10-31
DELETE /api/v3/posts/:post_id

chloealee:
#### What's this PR do?
- Creates the `DELETE /api/v3/posts/:post_id` endpoint
- Creates tests for this endpoint

- Hit the `DELETE /api/v3/posts/:post_id` endpoint. 
- Make sure the post was soft deleted.
- Hit that same endpoint without the `X-DS-Rogue-API-Key` and you shouldn't be able to delete the post. 
- Make sure all tests pass!

@katiecrane @DFurnes I just realized that when you hit this endpoint (and also the `DELETE /api/v3/signups/:signup_id` endpoint), it doesn't create an event. Do we want it to create an event for these?

#### Relevant tickets
Fixes the last task in this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151852968).

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -71,7 +71,9 @@ class PostsController extends ApiController
      */
     public function update(PostRequest $request, Post $post)
     {
+        // @TODO: Remove `array_filter` with 'only' changes in Laravel 5.5.
         $fields = array_filter($request->only('status', 'caption'));
+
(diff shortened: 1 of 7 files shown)


chloealee:
great point, thanks @DFurnes ! Here's the card: https://www.pivotaltracker.com/n/projects/2019429/stories/152445437

PR #473, opened 2017-11-01
POST /api/v3/signups

chloealee:
#### What's this PR do?
- Creates the `POST /api/v3/signups` endpoint. 
- Creates a test for this functionality. 

- Hit this endpoint with new data and make sure a `201` response code is returned and the data is in the database. 
- Hit this endpoint again with same data and make sure the `200` response code is returned and the signup isn't duplicated. 
- Make sure you can't make this request without the `X-DS-Rogue-API-Key`
- Make sure all tests pass! 

#### Relevant tickets
Fixes this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/152482999).

#### Checklist
- [ ] Tested on staging.


app/Http/Requests/Three/SignupUpdateRequest.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
chloealee: is this weird to have separate requests for the update and the creation of a signup? Should this validation just live in the controller?

tests/Http/Three/SignupTest.php
```diff
@@ -9,6 +9,78 @@
 class SignupTest extends BrowserKitTestCase
 {
     /**
+     * Test that a POST request to /signups creates a new signup.
+     *
+     * POST /api/v3/signups
+     * @return void
+     */
+    public function testCreatingASignup()
```
chloealee: should I also add tests to make sure you can't do any of these things unauthorized? We have one for the delete endpoint but should each endpoint have a test for this?

chloealee:
I think the wercker build is failing because it is actually trying to send to blink. It seems like it is running [this code](https://github.com/DoSomething/rogue/blob/master/app/Services/SignupService.php#L51-L55) but isn't `$this->mock(Blink::class)->shouldReceive('userSignup');` supposed to mock this?

chloealee:
ah I'm an idiot - it's because I wasn't using Blink in the first place in the test! All fixed now :)

DFurnes:
Phew! Good catch 🎣

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -19,18 +21,48 @@ class SignupsController extends ApiController
...
 
     /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(SignupRequest $request)
+    {
+        $transactionId = incrementTransactionId($request);
```
DFurnes: Hmm, I'd really love to move this into a middleware. Let's consider pulling in the ticket for this next sprint.

app/Http/Requests/Three/SignupUpdateRequest.php
DFurnes: I'm personally a fan, since it's a little easier to reason about (especially if you want to drop a `dd()` before the validation). And it's also nice if you want to share a base set of rules between controller methods. 👍

tests/Http/Three/SignupTest.php
```diff
@@ -5,10 +5,83 @@
...
+    {
+        $northstarId = '54fa272b469c64d7068b456a';
```
DFurnes: Let's use `$this->faker->northstar_id` and `$this->faker->campaign_id` for these two!

tests/Http/Three/SignupTest.php
DFurnes: I think that's a good idea, just to be safe! 🔒

chloealee: ok cool! Do we want to have `GET /signups` and `GET /signups/:signup_id` be authorized requests as well? Right now they are not.

DFurnes: I think we probably want that for the index... maybe not for the `/signups/:id` one?

chloealee: wait actually they all should be (https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Three/SignupsController.php#L30) - but for some reason I am able to without this in Postman for these and also the PATCH method 🤔

DFurnes: We [somewhat awkwardly](https://github.com/DoSomething/rogue/blob/1fe5bd4319dd3edcef73d53bb5b3a8016acac41c/app/Http/Middleware/AuthenticateApi.php#L23-L26) bypass this for all `GET` endpoints in the middleware, which we should probably remove in the future.

chloealee: ahhh! Is it ok to add `UPDATE` there for now?

app/Http/Controllers/Three/SignupsController.php
chloealee: Great call! Here's the ticket for it: https://www.pivotaltracker.com/n/projects/2019429/stories/152511949

tests/Http/Three/SignupTest.php
chloealee: added the `UPDATE` for now but didn't for `GET` endpoints for now! @DFurnes

app/Http/Requests/Three/SignupUpdateRequest.php
chloealee: awesome, updated @DFurnes !

PR #474, opened 2017-11-03
Creates POST /api/v3/posts

chloealee:
#### What's this PR do?
- Creates the `POST /api/v3/posts` endpoint
- Creates tests for this endpoint
- Adds tests to make sure an unauthenticated user/app can't update posts

- Hit the `POST /api/v3/posts` endpoint and make sure a signup and post are created if a signup didn't exist yet or a post is created if the signup already exists
- Make sure all tests pass! 

#### Relevant tickets
Fixes this [Pivotal Ticket](https://www.pivotaltracker.com/n/projects/2019429/stories/152482999)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -51,6 +51,42 @@ class PostsController extends ApiController
     }
 
     /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(PostRequest $request)
+    {
+        $transactionId = incrementTransactionId($request);
+
+        $signup = $this->signups->get($request['northstar_id'], $request['campaign_id'], $request['campaign_run_id']);
+
+        $updating = ! is_null($signup);
+
+        // @TODO - should we eventually throw an error if a signup doesn't exist before a post is created? I create one here because we haven't implemented sending signups to rogue yet, so it will have to create a signup record for all posts.
+        if (! $updating) {
+            $signup = $this->signups->create($request->all());
+
+            $post = $this->posts->create($request->all(), $signup->id, $transactionId);
+
+            $code = 200;
+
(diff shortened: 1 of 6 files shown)


app/Http/routes.php
```diff
@@ -102,8 +102,9 @@
     Route::delete('signups/{signup}', 'Three\SignupsController@destroy');
 
     // posts
+    Route::post('posts', 'Three\PostsController@store');
```
DFurnes: Waaaaaait, does this method exist? 👀

chloealee: omg that is so weird I don't know where it went!!

chloealee: oh wait wait ok that's right - it goes through the `PostRequests` trait so actually goes to the `store` method here: https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Traits/PostRequests.php#L19.

sbsmith86: if this `PostRequests` trait is confusing people then we can kill it, I just thought it would be easier than having the same method in a bunch of different places.

DFurnes: OH! Tricky, I totally missed that. I'm cool with keeping it for the v2 endpoints, but I think we should have a separate `store` method here for the v3 version of this route (since we'll want to customize some of the implementation details, like where we get `northstar_id` from).

app/Http/Controllers/Three/PostsController.php
```diff
@@ -51,6 +51,42 @@ public function __construct(PostService $posts, SignupRepository $signups, PostT
     }
 
     /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(PostRequest $request)
```
chloealee: @DFurnes added the same code here but in its own `store` method! Figured this will soon be updated though with the customized northstar updates we want to make.

PR #475, opened 2017-11-03
POST /api/v3/post/:post_id/reactions

chloealee:
#### What's this PR do?
- Adds the `POST /api/v3/post/:post_id/reactions` endpoint
- Adds tests that mimic the v2 tests for reactions 
- Adds additional test to make sure an unauthenticated user/app cannot post/soft-delete a reaction

- Hit the `POST /api/v3/post/:post_id/reactions` endpoint
- Make sure the record is in the database
- Hit it again and make sure it is soft deleted
- Hit the endpoint again with a different `northstar_id`. You should see a new record in the `reactions` database.
- Make sure all tests pass!

#### Relevant tickets
Adds the same endpoint as v2 in v3 (apart of this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151853171)).

#### Checklist
- [ ] Tested on staging.


app/Http/Transformers/Three/ReactionTransformer.php
```diff
@@ -0,0 +1,31 @@
...
+     * @return array
+     */
+    public function transform(Reaction $reaction)
+    {
+        return [
+            'northstar_id' => $reaction->northstar_id,
+            'post_id' => (string) $reaction->post_id,
+            'term' => [
+                'name' => 'heart',
+                'id' => 641,
```
sbsmith86: what is the `641` id again?

app/Http/Controllers/Three/ReactionController.php
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Get the number of reactions a post has.
+     *
+     * @param int $postId
+     * @return int
+     */
+    private function getTotalReactions($postId)
+    {
+        return count(Post::find($postId)->reactions);
```
DFurnes: Let's move this onto the `Post` model (and make sure that it's counted with a `COUNT` query on the database, not by loading all the actual reactions into memory.

app/Http/Transformers/Three/ReactionTransformer.php
chloealee: @sbsmith86 great q! I think that was the id for in P-A land for the heart "reaction" - should we cut this out since it's not applicable anymore?

sbsmith86: unless anything depends on it, I feel like we can kill it since for now we only have one reaction type.

app/Http/Controllers/Three/ReactionController.php
```diff
@@ -0,0 +1,86 @@
...
+                'northstar_id' => $userId,
+                'post_id' => $post->id,
+            ]);
+
+            $code = 200;
+            $action = 'liked';
+        } else {
+            // If the post was previously "liked" by this user, soft delete the reaction. Otherwise, restore the reaction.
+            $code = 201;
+            if (is_null($reaction->deleted_at)) {
```
DFurnes: Let's replace this with `$reaction->trashed()` so we're not relying on database implementation details.

app/Http/Transformers/Three/ReactionTransformer.php
chloealee: I also killed the `total` line under `term` bc I don't think it's relevant either! Not sure if we even want the `term` section now that I think of it?

app/Http/Controllers/Three/ReactionController.php
```diff
@@ -0,0 +1,86 @@
...
+        $this->validate($request, [
+            'northstar_id' => 'required|string',
+        ]);
+
+        $userId = $request['northstar_id'];
+        // Check to see if the post has a reaction from this particular user with id of northstar_id.
+        $reaction = Reaction::withTrashed()->where(['northstar_id' => $userId, 'post_id' => $post->id])->first();
+
+        // If a post does not have a reaction from this user, create a reaction.
+        if (is_null($reaction)) {
```
DFurnes: This feels like a pretty gnarly set of conditionals, I wonder if we can simplify it at all. Could we extract the `create` below here into a `findOrCreate` and then use `$reaction->wasRecentlyCreated` to determine if we just made it?

```php
$reaction = Reaction::withTrashed()->firstOrCreate(['northstar_id' => $userId, 'post_id' => $post->id]);

if ($reaction->wasRecentlyCreated || $reaction->trashed) {
  // we're adding a new reaction in these cases
} else {
  // otherwise we must be removing
}
```

app/Http/Transformers/Three/ReactionTransformer.php
DFurnes: Yeah, we can remove all this old Phoenix stuff! 🔥

app/Http/Controllers/Three/ReactionController.php
chloealee: ooo these are so cool @DFurnes ! I never knew about `wasRecentlyCreated` or `firstOrCreate` before! For the conditionals, we would want:
```
if (! $reaction->wasRecentlyCreated || $reaction->trashed()){
```
for that first line though, right? Because if it was just created in line 45, we don't want to create a new reaction? Unless I'm not understanding correctly?

DFurnes: Yeah, the logic still gets a little weird since we're soft-deleting these.

My thought was we'd always call `$reaction->untrash()` in that part of the conditional (and it just wouldn't do anything if it's a newly created reaction), since in both these cases we want our end-state (the values for `$code` and `$action`) to be the same.

app/Http/Controllers/Three/ReactionController.php
```diff
@@ -0,0 +1,70 @@
...
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request, Post $post)
+    {
+        $this->validate($request, [
+            'northstar_id' => 'required|string',
+        ]);
+
+        $userId = $request['northstar_id'];
+        // Check to see if the post has a reaction from this particular user with id of northstar_id. If not, create one.
```
chloealee: ah I see! I updated it and tested and seems to work! Is this what you meant @DFurnes ?

documentation/README.md
```diff
@@ -70,4 +70,9 @@ Endpoint                                       | Functionality
 `GET /api/v3/posts`                            | [Get posts](endpoints/posts.md#retrieve-all-posts)
 `GET /api/v3/posts/:post_id`                   | [Get a post](endpoints/signups.md#retrieve-a-specific-post)
 `DELETE /api/v3/posts/:post_id`                | [Delete a post](endpoints/signups.md#delete-a-post)
-`PATCH /api/v3/posts/:post_id`                 | [Update a post](endpoints/signups.md#update-a-post)   
+`PATCH /api/v3/posts/:post_id`                 | [Update a post](endpoints/signups.md#update-a-post)
```
katiecrane: Should this use `endpoints/posts` instead of `endpoints/signups`?

chloealee: ah yes great catch!

app/Http/Controllers/Three/ReactionController.php
DFurnes: Yeah, this seems great!

chloealee:
hmm I'm getting this error:
```
{"error":{"code":404,"message":""},"debug":{"file":"\/home\/vagrant\/DoSomething.org\/rogue\/vendor\/laravel\/framework\/src\/Illuminate\/Routing\/RouteCollection.php","line":178}}
```
 and when I go to the `store` method and `dd` something out in the first line, it doesn't get there...
 for some reason it is no longer registering this route? 🤔

DFurnes:
@chloealee Routes moved to `routes/web.php` and `routes/api.php` in #481 for Laravel 5.5. Looks like you ran into some merge conflicts and got the old location for the file.

chloealee:
oooohh thanks @DFurnes !

app/Http/Transformers/Three/ReactionTransformer.php
sbsmith86: @chloealee I don't think this was removed?

documentation/endpoints/reactions.md
```diff
@@ -31,3 +32,40 @@ Example Response
...
+  --X-DS-Rogue-API-Key "API_KEY"
+```
+
+Example Response
+```
+{
+    "data": {
+        "northstar_id": "5589c991a59dbfa93d8b45ae",
+        "post_id": "61",
+        "term": {
```
sbsmith86: Think you can remove this from the docs too.

app/Http/Transformers/Three/ReactionTransformer.php
chloealee: @sbsmith86 oops I removed from the other `ReactionTransformer` by accident - thanks for flagging!

documentation/endpoints/reactions.md
chloealee: ah yes - thanks @sbsmith86 !

PR #476, opened 2017-11-03
Adds GET /api/v3/post/:post_id/reactions endpoint

chloealee:
#### What's this PR do?
tell us, what have you changed?
- Adds the `GET /api/v3/post/:post_id/reactions` endpoint
- Adds tests for this 

- Hit this endpoint and make sure all the associated reactions to the post are returned. 
- Make sure the tests pass! 

#### Relevant tickets
Fixes the first task in this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151853171).

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/Three/ReactionController.php
```diff
@@ -0,0 +1,100 @@
...
+     * GET /post/:post_id/reactions
+     *
+     * @param \Rogue\Models\Post $post
+     * @return Illuminate\Http\Response
+     */
+    public function index(Post $post)
+    {
+        $reactions = Reaction::withTrashed()->where(['post_id' => $post->id])->get();
+
+        return $this->collection($reactions);
```
chloealee: @DFurnes I know we want this to be paginated but I don't think this will work with `paginatedCollection` since it expects the `$query` and `$request` to be sent in with it. The `$request` won't have anything in it since the specific post we want is in the URL. Should I write a custom paginator for this case?

DFurnes: You should still be able to do this with the method's `$request` parameter. It just needs that to know what URL to append the pagination query strings to in the response.

chloealee: oh you're right! That's awesome - thank you!

app/Http/Controllers/Three/ReactionController.php
```diff
@@ -0,0 +1,104 @@
...
+     * Returns all reactions for a post.
+     * GET /post/:post_id/reactions
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \Rogue\Models\Post $post
+     * @return Illuminate\Http\Response
+     */
+    public function index(Request $request, Post $post)
+    {
+        $query = $this->newQuery(Reaction::class);
```
DFurnes: This will return an index of _all_ reactions, which I don't think is what we want for this endpoint. Instead, let's send along the reaction query you'd made below (minus the `get()` that fetches the whole thing):

```diff
- $query = $this->newQuery(Reaction::class);
+ $query = Reaction::withTrashed()->where(['post_id' => $post->id]);
```

chloealee: ohohoh you're right ! Updated :)

app/Http/Controllers/Three/ReactionController.php
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \Rogue\Models\Post $post
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request, Post $post)
```
sbsmith86: hmm shouldn't this `store` method be on the `POST` PR?

app/Http/Transformers/Three/ReactionTransformer.php
```diff
@@ -0,0 +1,31 @@
...
+     *
+     * @param \Rogue\Models\Reaction $reaction
+     * @return array
+     */
+    public function transform(Reaction $reaction)
+    {
+        return [
+            'northstar_id' => $reaction->northstar_id,
+            'post_id' => (string) $reaction->post_id,
+            'term' => [
```
sbsmith86: We can remove `term` here

documentation/endpoints/reactions.md
```diff
@@ -31,3 +32,90 @@ Example Response
...
+  --X-DS-Rogue-API-Key "API_KEY"
+```
+
+Example Response
+```
+{
+    "data": {
+        "northstar_id": "5589c991a59dbfa93d8b45ae",
+        "post_id": "61",
+        "term": {
```
sbsmith86: And here!

sbsmith86:
did we decide to kill this endpoint for now?

If not, I'm a little confused about the changes in this PR because it seems to include the `POST` endpoint stuff too.

chloealee:
oof you're right @sbsmith86 - I'm sorry! I rebased against my post branch and that's probably why it grabbed everything. Once [this](https://github.com/DoSomething/rogue/pull/475/files) is merged, I'll rebase again to make this more clear and then ping you - sorry about that!

sbsmith86:
sounds like a plan!

chloealee:
@sbsmith86 this is rebased and ready for review!

app/Http/Controllers/Three/ReactionController.php
```diff
@@ -67,4 +67,21 @@ public function store(Request $request, Post $post)
...
+     *
+     * @param \Illuminate\Http\Request $request
+     * @param \Rogue\Models\Post $post
+     * @return Illuminate\Http\Response
+     */
+    public function index(Request $request, Post $post)
+    {
+        $query = Reaction::withTrashed()->where(['post_id' => $post->id]);
+
+        $reactions = Reaction::withTrashed()->where(['post_id' => $post->id])->get();
```
sbsmith86: hmm looks like you are doing this query twice?

chloealee: ah sorry for not catching that!

PR #477, opened 2017-11-03
Cleanup part1

sbsmith86:
#### What's this PR do?

First round of JS clean up. There are a lot of files touched, but this round of cleanup is mostly focused on [PropType validation](https://reactjs.org/docs/typechecking-with-proptypes.html) which is the heaviest and newest thing we are requiring after adding some linting to our JS files. 

The idea is to specify what props a component should expect and what of what type those props should be. This is actually really good to have because if we run into issues were we don't pass in a prop for some reason we can handle that error upfront. 

giphy-downsized 7

:eyes: 

Running `npm run build` should not fail. 

More to come. 

#### Relevant tickets
Addresses https://www.pivotaltracker.com/story/show/151127764

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/CampaignInbox/index.js
+++ b/resources/assets/components/CampaignInbox/index.js
@@ -69,7 +69,7 @@ CampaignInbox.propTypes = {
   hideHistory: PropTypes.func.isRequired,
   historyModalId: PropTypes.number,
   loading: PropTypes.bool.isRequired,
-  postIds: PropTypes.array, // eslint-disable-line react/forbid-prop-types
+  postIds: PropTypes.arrayOf(PropTypes.number),
   posts: PropTypes.object, // eslint-disable-line react/forbid-prop-types
   rotate: PropTypes.func.isRequired,
   showHistory: PropTypes.func.isRequired,
--- a/resources/assets/components/CampaignInbox/index.js
+++ b/resources/assets/components/CampaignInbox/index.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { map, find, isEmpty } from 'lodash';
 import { RestApiClient } from '@dosomething/gateway';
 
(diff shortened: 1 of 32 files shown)


resources/assets/components/HistoryModal/index.js
```diff
@@ -76,4 +77,12 @@ class HistoryModal extends React.Component {
   }
 }
 
+HistoryModal.propTypes = {
```
sbsmith86: I have no idea why this is red. If anyone has an idea, let me know!

DFurnes: Ahhh, a classic! Github's JSX syntax highlighting [is broken](https://git.io/vF8fk).

sbsmith86: oh fun!

resources/assets/components/CampaignInbox/index.js
```diff
@@ -61,4 +62,31 @@ class CampaignInbox extends React.Component {
   }
 }
 
+CampaignInbox.propTypes = {
+  campaign: PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
```
DFurnes: This totally isn't necessary for this pull request, but we should try to avoid using `PropTypes.object` where we can, since it doesn't give us any assurances about the contents of the object. One good solution might be to create a `types.js` file somewhere and export `PropType.shape` definitions from there...

```js
// types.js
export const CampaignType = PropTypes.shape({
  id: PropTypes.string.isRequired,
  // ...
});


// CampaignInbox/index.js
import { CampaignType } from '../../../types.js';

CampaignInbox.PropTypes = {
  campaign: CampaignType,
  // ...
};

```

resources/assets/components/CampaignInbox/index.js
```diff
@@ -61,4 +62,31 @@ class CampaignInbox extends React.Component {
...
+  deletePost: PropTypes.func.isRequired,
+  displayHistoryModal: PropTypes.bool,
+  hideHistory: PropTypes.func.isRequired,
+  historyModalId: PropTypes.number,
+  loading: PropTypes.bool.isRequired,
+  postIds: PropTypes.array, // eslint-disable-line react/forbid-prop-types
```
DFurnes: This could be `PropTypes.arrayOf(PropTypes.string)` to silence this linter warning.

resources/assets/components/ReportbackUploader/index.js
```diff
@@ -117,10 +119,21 @@ class ReportbackUploader extends React.Component {
   }
 }
 
-ReportbackUploader.defaultProps = {
-  noun: {
-    plural: 'items',
-  },
+ReportbackUploader.propTypes = {
+  // campaignId: PropTypes.int.isRequired,
+  // campaignRunId: PropTypes.int.isRequired,
```
DFurnes: 👾

DFurnes:
Also 🏆 for atomic commits, so nice & tidy!

screen shot 2017-11-06 at 10 09 42 am

resources/assets/components/CampaignInbox/index.js
sbsmith86: oh interesting. I really like that! The only reason Ignored it was because I didn't want to have to shape out all of these large objects since we are doing things like passing in the entire campaign object, user object, etc. But this is a really nice solution to this. I will maybe do that in a separate PR so I can keep the work chunked.

PR #479, opened 2017-11-06
Artisan command to load in pre-Gambit (or any!) signups from CSV

katiecrane:
#### What's this PR do?
Includes a script to import signups from a csv! The csv will include ALL pre-Gambit signups. The script checks if a signup already exists for the given user, campaign, and campaign run before creating a new one so there are no duplicates.

Will this run right?

This is gonna make Rogue the source of truth!!

#### Relevant tickets
[Pivotal](https://www.pivotaltracker.com/story/show/151064896)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Console/Commands/ImportSignupsCommand.php
+++ b/app/Console/Commands/ImportSignupsCommand.php
@@ -17,6 +17,13 @@ class ImportSignupsCommand extends Command
     protected $signature = 'rogue:signupimport {path}';
 
     /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Create signups based on the data in the given CSV file.';
+
+    /**
      * Create a new command instance.
      *
      * @return void
--- a/app/Console/Commands/ImportSignupsCommand.php
+++ b/app/Console/Commands/ImportSignupsCommand.php
@@ -2,7 +2,6 @@
 
 namespace Rogue\Console\Commands;
 
-use Carbon\Carbon;
 use League\Csv\Reader;
 use Rogue\Models\Signup;
 use Rogue\Services\Registrar;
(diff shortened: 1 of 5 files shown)


app/Console/Commands/ImportSignupsCommand.php
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct(Registrar $registrar)
+    {
+        parent::__construct();
+
+        $this->registrar = $registrar;
```
DFurnes: It looks like this might not be used!

app/Console/Commands/ImportSignupsCommand.php
```diff
@@ -0,0 +1,74 @@
...
+            // Create a signup if there isn't on already
+            if (! $existing_signup) {
+                $signup = new Signup;
+                $signup->northstar_id = $missing_signup['northstar_id'];
+                $signup->campaign_id = $missing_signup['campaign_node_id'];
+                $signup->campaign_run_id = $missing_signup['campaign_run_id'];
+                $signup->source = 'sms';
+                $signup->created_at = $missing_signup['signup_created_at_timestamp'];
+                $signup->updated_at = Carbon::now();
+                $signup->save(['timstamps' => false]);
```
DFurnes: `timstamps` → `timestamps` 👀

katiecrane: hahahhahahha

katiecrane: the tests still pass though it looks like it still works 😂

app/Console/Commands/ImportSignupsCommand.php
katiecrane: oh it sure isn't. old plans.

PR #485, opened 2017-11-09
User Search Failure Page

chloealee:
#### What's this PR do?
- If a name is searched or an invalid email address (either spelled incorrectly or we don't have a member with that email address), the page is redirected to the search page with a "No user found!" message at the top: 
screen shot 2017-11-09 at 2 23 19 pm

👀 

@ngjo I actually think that this will work for the other case you mentioned in the [Pivotal Card](https://www.pivotaltracker.com/story/show/151136337/comments/181993342) too that Pili flagged! 

#### Relevant tickets
Fixes this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151136337).

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -67,10 +67,10 @@ class UsersController extends Controller
         // Attempt to fetch all users.
         $users = $this->registrar->search($query);
 
-        if ($users->count() === 1) {
-            return redirect()->route('users.show', [$users->first()->id]);
+        if ($users->count() === 0){
+            return redirect()->route('users.index')->with('status', 'No user found!');
         }
 
-        return redirect()->route('users.index')->with('status', 'No user found!');
+        return redirect()->route('users.show', [$users->first()->id]);
     }
 }
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -67,12 +67,10 @@ class UsersController extends Controller
         // Attempt to fetch all users.
         $users = $this->registrar->search($query);
 
-        if (! $users) {
-            return redirect()->route('users.index')->with('status', 'No user found!');
-        } elseif ($users->count() === 1) {
+        if ($users->count() === 1) {
             return redirect()->route('users.show', [$users->first()->id]);
         }
 
-        return view('users.search')->with(compact('users', 'query'));
+        return redirect()->route('users.index')->with('status', 'No user found!');
     }
 }
--- a/app/Http/Controllers/UsersController.php
+++ b/app/Http/Controllers/UsersController.php
@@ -67,7 +67,7 @@ class UsersController extends Controller
         // Attempt to fetch all users.
         $users = $this->registrar->search($query);
 
-        if ($users->count() === 0){
+        if ($users->count() === 0) {
             return redirect()->route('users.index')->with('status', 'No user found!');
         }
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/UsersController.php
```diff
@@ -67,12 +67,10 @@ public function search(Request $request)
         // Attempt to fetch all users.
         $users = $this->registrar->search($query);
 
-        if (! $users) {
```
chloealee: `! $users` should never be the case because if there are not results, `$users` looks like this: 
```
NorthstarUserCollection {#424 ▼
  #items: []
  #total: 0
  #perPage: 20
  #currentPage: 1
  #hasMore: null
  #paginator: null
}
```

app/Http/Controllers/UsersController.php
```diff
@@ -67,12 +67,10 @@ public function search(Request $request)
...
-            return redirect()->route('users.index')->with('status', 'No user found!');
-        } elseif ($users->count() === 1) {
+        if ($users->count() === 1) {
             return redirect()->route('users.show', [$users->first()->id]);
         }
 
-        return view('users.search')->with(compact('users', 'query'));
+        return redirect()->route('users.index')->with('status', 'No user found!');
```
DFurnes: Can we pass the empty `$users` collection to the view instead, and display an empty state there? I'm not sure if there are any scenarios where we would right now, but it'd be nice if we could display more than one result if there was one.

chloealee: @DFurnes right now, if a user is found, it automatically redirects to that user page (`/users/:northstar_id`). Should I change it to return the results in the grid format and then have the user who is searching click on the user they want?

DFurnes: I feel like that's a kinda nice shortcut, unless it's been causing trouble for people.

chloealee: ok I think that should be fine! @ngjo is this ok if we change this?

DFurnes: If we want to keep the current behavior, we could always just check for `$users->count() === 0` rather than `! $users` too. I'd just rather not rule out the possibility of having more than one search result.

chloealee: cool - chatted with @ngjo and since only results are returned that are an exact match, we'll stick with the user flow right now! I updated the above, @DFurnes !

sbsmith86:
@ngjo might know this better, but if we don't find a user, do we usually try and find them in aurora as well? Should we add a note that an admin can also try and find the user in aurora?

ngjo:
@sbsmith86 we haven't fully onboarded staff onto aurora, but we can do that and then yes! this makes a lot of sense

sbsmith86:
@ngjo Gotcha. I def don't want to block this while that onboarding happens. But that could be something to keep in mind for the FAQ page or something down the line

ngjo:
@sbsmith86 cool! sounds good to me - ill make a note of it!

app/Http/Controllers/UsersController.php
```diff
@@ -67,12 +67,10 @@ public function search(Request $request)
...
 
-        if (! $users) {
+        if ($users->count() === 0) {
             return redirect()->route('users.index')->with('status', 'No user found!');
-        } elseif ($users->count() === 1) {
-            return redirect()->route('users.show', [$users->first()->id]);
         }
 
-        return view('users.search')->with(compact('users', 'query'));
+        return redirect()->route('users.show', [$users->first()->id]);
```
katiecrane: getting rid of not strictly necessary `else` statements is my favorite 🎉 🎉 🎉

PR #487, opened 2017-11-09
Update test suite for Laravel 5.4+.

DFurnes:
#### What's this PR do?
This pull request updates Rogue's remaining tests to use Laravel 5.4+ syntax, rather than the [Browser Kit compatibility package](https://github.com/laravel/browser-kit-testing). This means our test syntax will match the docs, and we get some extra goodies added in the newer release.

I removed our custom `ClearDB` and `RefreshDatabase` implementations (and the example `Inspire` command), and updated any test methods that changed.

One surprise that took a while to debug: we were accidentally [disabling foreign key constraints](https://github.com/DoSomething/rogue/blob/057dcb9340b3ac89842013ed70d5d2b6570f2626/app/Console/Commands/ClearDB.php#L44) for our whole test suite (since we never re-enabled them after the `clear:db` command ran). Since Laravel's `migrate:fresh` command doesn't have this bug, we had a _toooon_ of failures at first since we were creating posts without signups in a lot of tests. Luckily, this could be easily fixed [in the model factory](https://github.com/DoSomething/rogue/pull/487/commits/cbee182023288b5a2c90f8979d21a19e1349f965).

I removed a few browser tests in this pull request, and will add them back in a follow-up PR. 🔜

#### Relevant tickets
[#152601905](https://www.pivotaltracker.com/story/show/152601905)

#### Checklist
- [ ] Tested on staging.


database/factories/ModelFactory.php
```diff
@@ -26,6 +24,9 @@
     $url = app(AWS::class)->storeImage($upload, $faker->unique()->randomNumber(5));
 
     return [
+        'signup_id' => function () {
+            return factory(Signup::class)->create()->id;
+        },
         'url' => $url,
```
chloealee: do we need to associate the signup and post here or is this just so it doesn't get mad that posts are created without signups?

tests/Http/Api/ActivityApiTest.php
```diff
@@ -208,20 +208,18 @@ public function testActivityIndexWithUpdatedAtQuery()
 
         // Create two signups and two posts. Associate the signups and posts respectively.
         $firstSignup = factory(Signup::class)->create();
-        $firstPost = factory(Post::class)->create();
-        $firstPost->signup()->associate($firstSignup);
+        $firstSignup->posts()->save(factory(Post::class)->make());
```
chloealee: ooo this is so nice!

chloealee: since the Post model factory now returns a signup id, will this conflict here when we associate these? I think the Post model factory also creates a signup (returns a newly created `signup_id`) so I'm wondering if this will clash?

tests/Http/Api/ReactionsApiTest.php
```diff
@@ -20,37 +19,39 @@ class ReactionsApiTest extends BrowserKitTestCase
     public function testPostingAndSoftDeleteForReaction()
     {
         // Create a post to react to.
-        $post = factory(Post::class)->create();
+        $signup = factory(Signup::class)->create();
+        $signup->posts()->save(factory(Post::class)->make());
+        $post = $signup->posts->first();
```
katiecrane: Doesn't the post factory make a signup for you?

tests/Http/Api/ReactionsApiTest.php
```diff
@@ -65,62 +66,36 @@ public function testPostingAndSoftDeleteForReaction()
...
     public function testAggregateReactions()
     {
-        $post = factory(Post::class)->create();
+        $signup = factory(Signup::class)->create();
+        $signup->posts()->save(factory(Post::class)->make());
+        $post = $signup->posts->first();
 
         // Create a reaction.
-        $this->withRogueApiKey()->post('api/v2/reactions', [
+        $response = $this->withRogueApiKey()->post('api/v2/reactions', [
```
chloealee: should this be `postJson`?

database/factories/ModelFactory.php
DFurnes: Yep, this is because we were causing (and ignoring) foreign key constraints when we used to create posts w/o signups in the test suite. This makes sure that we always create and associate a corresponding signup.

tests/Http/ReviewsTest.php
```diff
@@ -70,9 +69,7 @@ public function testUnauthenticatedUserCantReviewPosts()
     public function testUpdatedPostAndSignupWithReview()
```
chloealee: do we need this test anymore?

tests/Http/Api/ActivityApiTest.php
DFurnes: I believe it should only evaluate the closure in the Post factory if a value isn't already provided for `signup_id`, which it should be here. Lemme double-check.

tests/Http/TagsTest.php
```diff
@@ -22,18 +21,18 @@ public function testAddingATagToAPost()
...
-                'post_id' => $post->id,
-                'tag_name' => 'Good Photo',
-            ]);
+        $response = $this->actingAsAdmin()->post('tags', [
+            'post_id' => $post->id,
+            'tag_name' => 'Good Photo',
+        ]);
 
-        $this->assertResponseStatus(200);
+        $response->assertSuccessful();
```
chloealee: this is so cool @DFurnes ! When should we use this vs `$response->assertStatus(200);`?

tests/Http/Api/ReactionsApiTest.php
DFurnes: Ah yes! I had thought I'd need to fix each of these manually at first, will revert. 🔙

tests/Http/Api/ReactionsApiTest.php
DFurnes: Oh, yep probably!

tests/Http/Three/ReactionTest.php
```diff
@@ -25,30 +25,30 @@ public function testPostingAndSoftDeleteForReaction()
...
-        $this->seeJsonSubset([
+        $response->assertJson([
             'meta' => [
                 'total_reactions' => 1,
             ],
         ]);
 
         // React (unlike) again to the same post with the same user.
-        $this->withRogueApiKey()->post('api/v3/post/' . $post->id . '/reactions', [
+        $response = $this->withRogueApiKey()->post('api/v3/post/' . $post->id . '/reactions', [
```
chloealee: should we use `postJson`?

tests/Http/Three/ReactionTest.php
```diff
@@ -66,26 +66,26 @@ public function testAggregateReactions()
         $post = factory(Post::class)->create();
 
         // Create a reaction.
-        $this->withRogueApiKey()->post('api/v3/post/' . $post->id . '/reactions', [
+        $response = $this->withRogueApiKey()->post('api/v3/post/' . $post->id . '/reactions', [
```
chloealee: `postJson`?

tests/Http/TagsTest.php
DFurnes: It's subjective I think. This will accept any non-error code which might be nice for those cases where we used to return the wrong `201` or `200` status codes.

tests/Http/Three/ReactionTest.php
```diff
@@ -66,26 +66,26 @@ public function testAggregateReactions()
...
+        $response->assertStatus(200);
+        $response->assertJson([
             'meta' => [
                 'total_reactions' => 1,
             ],
         ]);
 
         // A second user reacts to the same post..
-        $this->withRogueApiKey()->post('api/v3/post/' . $post->id . '/reactions', [
+        $response = $this->withRogueApiKey()->post('api/v3/post/' . $post->id . '/reactions', [
```
chloealee: again here (`postJson`)?

tests/Http/Three/SignupTest.php
```diff
@@ -113,10 +112,10 @@ public function testSignupsIndex()
     {
         factory(Signup::class, 10)->create();
 
-        $this->get('api/v3/signups');
+        $response = $this->get('api/v3/signups');
```
chloealee: how do you choose to use `get` vs `getJson`?

tests/Http/Api/ReportbackApiTest.php
```diff
@@ -2,30 +2,34 @@
...
     public function testGetReportbacks()
     {
-        $signups = factory(Signup::class, 10)->create();
+        factory(Signup::class, 10)->create()
+            ->each(function (Signup $signup) {
+                // Give each signup a post and give it status of "accepted"
+                $post = factory(Post::class)->make();
+                $post->campaign_id = $signup->campaign_id;
+                $post->northstar_id = $signup->northstar_id;
+                $post->status = 'accepted';
```
katiecrane: Probably out of scope but I think we can just create an accepted post here

tests/Http/TagsTest.php
chloealee: cool!

tests/Http/UserTest.php
```diff
@@ -1,67 +0,0 @@
-<?php
-
-namespace Tests\Http;
-
-use Rogue\Models\User;
-use Rogue\Services\Phoenix;
-use Tests\BrowserKitTestCase;
-use DoSomething\Gateway\Northstar;
-
-class UserTest extends BrowserKitTestCase
```
katiecrane: Would it make sense to mark these as incomplete instead of removing them?

tests/Http/Three/SignupTest.php
DFurnes: Mmm, good catch. We should probably always use `getJson` since it sets the right `Accept` and `Content-Type` headers (although a lot of cases _do_ work without specifying those too).

chloealee: oh that is neat that it does that for you!

tests/Http/Api/ReportbackApiTest.php
DFurnes: Oh yeah! Done in 073fc5c3c4f9261d677de62e77326a7cd2160cc9.

tests/Http/Api/ReactionsApiTest.php
DFurnes: Done in 2f711e8.

tests/Http/UserTest.php
DFurnes: I've got a PR ready to push up after merging this, but otherwise I definitely would!

tests/Http/ReviewsTest.php
DFurnes: Hmm, not sure! I _thiiiink_ we still want these until we can move to a "push" model for getting updates to the data team but then we can probs remove?

DFurnes:
@katiecrane @chloealee I think I addressed all your comments, ready for another look! 👀

tests/Http/Three/SignupTest.php
```diff
@@ -88,19 +91,15 @@ public function testNotCreatingDuplicateSignups()
      */
     public function testUnauthenticatedUserCreatingASignup()
     {
-        $northstarId = '54fa272b469c64d7068b456a';
-        $campaignId = '6LQzMvDNQcYQYwso8qSkQ8';
-        $campaignRunId = $this->faker->randomNumber(4);
-
         $response = $this->json('POST', 'api/v3/signups', [
```
chloealee: oh @DFurnes found another! (sorry for being so nitpicky not sure if this matters too much!)

DFurnes: OMG HOW DID I MISS THAT ONE

database/factories/ModelFactory.php
katiecrane: Looking at this again and now I'm confused about how the signup and post get associated?

DFurnes: They get associated by setting `signup_id` on the post (since it's a "has many" relationship, which is defined on the `signup_id` column in the posts table).

DFurnes: Here's some more [examples](https://laravel.com/docs/5.5/database-testing#relationships) from the Laravel docs. The nice thing about doing it this way is we don't have to remember to manually build the relationship in each test case.

katiecrane: oh nice! You can just set the column! amazing!

PR #488, opened 2017-11-10
Mobile Link to Signup Page

chloealee:
#### What's this PR do?
If a user doesn't have a first name, their mobile number will link to their signup page.

👀 

The mobile number will only link to their signup page if there is no user name. If there is a user name, the mobile number will not be a link. Do we want the first name AND the mobile number to always link to the signup page, even if the user has both for more consistency? 

#### Relevant tickets
Fixes this [Pivotal Ticket](https://www.pivotaltracker.com/n/projects/2019429/stories/151462735).

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/Users/UserInformation.js
+++ b/resources/assets/components/Users/UserInformation.js
@@ -20,9 +20,9 @@ const UserInformation = props => (
               if (props.user.mobile) {
                 if (props.linkSignup && isEmpty(props.user.first_name)) {
                   return (<span><a href={`/signups/${props.linkSignup}`}>{props.user.mobile}</a><br /></span>);
-                } else {
-                  return (<span>{props.user.mobile}<br /></span>);
                 }
+
+                return (<span>{props.user.mobile}<br /></span>);
               }
               return (null);
             })()
--- a/resources/assets/components/Users/UserInformation.js
+++ b/resources/assets/components/Users/UserInformation.js
@@ -19,11 +19,13 @@ const UserInformation = props => (
             (() => {
               if (props.user.mobile) {
                 if (props.linkSignup && isEmpty(props.user.first_name)) {
+
                   return (<span><a href={`/signups/${props.linkSignup}`}>{props.user.mobile}</a><br /></span>);
                 }
 
                 return (<span>{props.user.mobile}<br /></span>);
               }
+
               return (null);
             })()
           }
--- a/resources/assets/components/Users/UserInformation.js
+++ b/resources/assets/components/Users/UserInformation.js
@@ -14,7 +14,16 @@ const UserInformation = props => (
         }
         <p>
           {props.user.email ? <span>{props.user.email}<br /></span> : null}
-          {props.user.mobile ? <span>{props.user.mobile}<br /></span> : null }
+
+          {props.user.mobile ?
+              {props.linkSignup && isEmpty(props.user.first_name) ?
+                <span><a href={`/signups/${props.linkSignup}`}>{props.user.mobile}</a><br /></span>;
+                :
+                <span>{props.user.mobile}<br /></span>;
+              }
+            : null
+          }
+
           {displayCityState(props.user.addr_city, props.user.addr_state) ?
             <span>{displayCityState(props.user.addr_city, props.user.addr_state) }<br /></span>
             :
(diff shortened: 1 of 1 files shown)


resources/assets/components/Users/UserInformation.js
```diff
@@ -14,7 +14,16 @@ const UserInformation = props => (
...
+          {props.user.mobile ?
+              {props.linkSignup && isEmpty(props.user.first_name) ?
+                <span><a href={`/signups/${props.linkSignup}`}>{props.user.mobile}</a><br /></span>;
+                :
+                <span>{props.user.mobile}<br /></span>;
+              }
+            : null
+          }
+
           {displayCityState(props.user.addr_city, props.user.addr_state) ?
```
chloealee: @sbsmith86 this is giving me an error when I do the nested ternary. I think this is ok to do but when I don't do a nested ternary it works. Is this because of PropTypes rules perhaps?

sbsmith86: It's an `eslint` thing. Is the `no-nested-ternary` rule. So you probably need an actual if-statement here instead.

chloealee: ah I see - thanks for clarifying @sbsmith86 !

resources/assets/components/Users/UserInformation.js
```diff
@@ -14,7 +14,20 @@ const UserInformation = props => (
...
         <p>
           {props.user.email ? <span>{props.user.email}<br /></span> : null}
-          {props.user.mobile ? <span>{props.user.mobile}<br /></span> : null }
+
+          {
+            (() => {
+              if (props.user.mobile) {
+                if (props.linkSignup && isEmpty(props.user.first_name)) {
+                  return (<span><a href={`/signups/${props.linkSignup}`}>{props.user.mobile}</a><br /></span>);
+                } else {
```
katiecrane: I think you might not need this else and can just return `(<span>{props.user.mobile}<br /></span>)` here since the `if` returns something as well

resources/assets/components/Users/UserInformation.js
```diff
@@ -14,7 +14,20 @@ const UserInformation = props => (
...
+                  return (<span>{props.user.mobile}<br /></span>);
+                }
+              }
+              return (null);
+            })()
```
katiecrane: These `()` look funny can you explain them to me?

chloealee: @katiecrane great q! I was trying to find a way to do an `if else` statement within JSX and found [this](https://stackoverflow.com/questions/45429836/immediately-invoked-function-expression-inside-jsx) on Stack overflow, which is where I learned how to do this!

resources/assets/components/Users/UserInformation.js
chloealee: smart!

resources/assets/components/Users/UserInformation.js
katiecrane: @chloealee a-ha! thanks!

resources/assets/components/Users/UserInformation.js
```diff
@@ -14,7 +14,20 @@ const UserInformation = props => (
...
           {props.user.email ? <span>{props.user.email}<br /></span> : null}
-          {props.user.mobile ? <span>{props.user.mobile}<br /></span> : null }
+
+          {
+            (() => {
+              if (props.user.mobile) {
+                if (props.linkSignup && isEmpty(props.user.first_name)) {
+                  return (<span><a href={`/signups/${props.linkSignup}`}>{props.user.mobile}</a><br /></span>);
+                }
+
```
katiecrane: I truly don't know if there should be a newline here or not, but I think this and the next return on line 27 should be consistent in that regard?

DFurnes: I forget if this in our code style, but I'd vote for a newline in both! 🍃

chloealee: thanks @katiecrane and @DFurnes ! A newline above the `return`? Just want to make sure I understand correctly! If above the return, should I also put a newline above the `return` in line 22?

PR #490, opened 2017-11-13
Add Dusk test suite.

DFurnes:
#### What's this PR do?
This pull request adds [Laravel Dusk](https://laravel.com/docs/5.5/dusk) for automating functional tests with a real browser. This is cool because it lets us test real user flows in a real browser, while still making database assertions ~or mocking external API requests~*.

*oof, kidding – apparently you can't set mocks with Laravel Dusk, which is a bummer. I'm gonna keep thinking on this & see if there's any good solutions here (since we definitely will need to mock external requests for most of our flows). 🤔

Tests can be run locally on Homestead with `dusk`, and will be run automatically on PRs with Wercker.

I'm still working out some issues on my local, but it's ready for a first look! 👀 🚧

Laravel's browser assertions were moved into Dusk in Laravel 5.4, so this was necessary to maintain test coverage while removing that compatibility package. It's also nice because it means we can finally test some full user flows! 🌊

#### Relevant tickets
[#152601905](https://www.pivotaltracker.com/story/show/152601905)

#### Checklist
- [ ] Tested on staging.


tests/Browser/pages/HomePage.php
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Get the element shortcuts for the page.
+     *
+     * @return array
+     */
+    public function elements()
+    {
+        return [
+            '@title' => '.header__title',
```
chloealee: I love how you can do this - so cool!

tests/BrowserKitTestCase.php
```diff
@@ -1,117 +0,0 @@
...
-     */
-    public function withRogueApiKey()
-    {
-        $header = $this->transformHeadersToServerVars(['X-DS-Rogue-API-Key' => env('ROGUE_API_KEY')]);
-
-        $this->serverVariables = array_merge($this->serverVariables, $header);
-
-        return $this;
-    }
-}
```
chloealee: 🔪

tests/Browser/pages/Page.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    /**
+     * Get the global element shortcuts for the site.
+     *
+     * @return array
+     */
+    public static function siteElements()
+    {
+        return [
+            '@element' => '#selector',
```
chloealee: I was reading about this in the docs and other posts and still a little confused how it knows what the `selector` is - would you mind explaining this?

chloealee:
> *oof, kidding – apparently you can't set mocks with Laravel Dusk, which is a bummer. I'm gonna keep thinking on this & see if there's any good solutions here (since we definitely will need to mock external requests for most of our flows). 🤔

@DFurnes so does this mean we can't do the functional tests in [this card](https://www.pivotaltracker.com/n/projects/2019429/stories/151961644) using Dusk?

DFurnes:
This is ready for review, with some caveats:

🎭 It turns out you _can't_ set mocks for Dusk requests... which is going to cause issues since we make requests to Phoenix/Contentful/Northstar in a lot of places. In hindsight, that makes sense since they're separate processes but it's still a real bummer for our needs. I'm gonna try to do more tinkering on this before the end of the sprint.

🥞 It seems like the `RefreshDatabase` trait isn't compatible with Dusk, since the "server" transactions aren't committed and therefore aren't available to the "browser" request. Another one that I'll try to do some more thinking on, but for now it makes the idea of writing a lot of Dusk tests seem a little fragile.

DFurnes:
> @DFurnes so does this mean we can't do the functional tests in [this card](https://www.pivotaltracker.com/n/projects/2019429/stories/151961644) using Dusk? 

Yeah, I think we may run into issues since these views make requests to Northstar. 😖 One idea would be to configure Dusk to connect to Northstar QA (like our normal local dev environments) for now, but that's not really ideal either.

chloealee:
@DFurnes ah I see! Also, since a lot of these flows making API requests, would that be a problem too? Or does Dusk allow internal API requests but not external?

tests/Browser/pages/Page.php
DFurnes: These are basically just a "table of contents" for CSS selectors on a page.

So for example, I could put a selector like `.header__title` (or something really hairy like `#campaign-inbox .table-row:first-of-type .table-cell:first-of-type`) and reference it by a friendlier name like `@title` or `@first-result` in my tests.

chloealee: ohhh we have an `id` that is `selector`? I see! For some reason I thought that was a placeholder word or something so was confused how it knew what the `selector` is on the page. Thanks, @DFurnes !

DFurnes: Oh sorry, no! This was an example included as part of the default Dusk scaffolding. I'll remove it so there isn't any confusion. 💥

chloealee: ohhhh got it! Thanks @DFurnes !

tests/DuskTestCase.php
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Setup the test environment.
+     *
+     * @return void
+     */
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->configureMocks();
```
katiecrane: Are we able to use certain types of mocks in Dusk? Is this needed?

DFurnes:
> @DFurnes ah I see! Also, since a lot of these flows making API requests, would that be a problem too? Or does Dusk allow internal API requests but not external? 

Dusk is running a real browser so it's basically the same as clicking through the UI on your machine (or on QA/Thor). Any internal or external API requests will work just like they would normally!

The problem is that un-mocked external API requests make our tests more fragile since they rely on the state of an external system that could have changed data or be offline. Even so, I think we can probably start writing tests against Northstar QA/Phoenix Staging for now (since that's no worse than the issues we sometimes run into on Dr00ps deploys with Ghost Inspector).

tests/DuskTestCase.php
DFurnes: Oooof, yeah noooope! This was from my earlier more naïve days.

katiecrane: 😆

chloealee:
ahhh that makes so much sense @DFurnes ! Thanks for taking the time to explain more!

DFurnes:
@chloealee @sbsmith86 @katiecrane Alright, this is ready for one last once-over! ✌️ I'll revisit some ideas for mocking & database refreshes after getting some other things closed out for the sprint.

PR #492, opened 2017-11-13
Configure trusted proxies for Heroku.

DFurnes:
#### What's this PR do?
Since we have Heroku running behind a load balancer, we need to configure [Trusted Proxies](https://laravel.com/docs/5.5/requests#configuring-trusted-proxies) so that Heroku knows that we're running over HTTPS. This will allow it to create the proper `https://` URLs.

This should make no difference on AWS or local environments, where we'll leave the `TRUSTED_PROXY_IP_ADDRESSES` environment variable empty.

This reflects the same implementation in [Northstar](https://github.com/DoSomething/northstar/pull/572) & [Phoenix Next](https://github.com/DoSomething/phoenix-next/pull/35).

#### Relevant tickets
DoSomething/devops#341

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -15,6 +15,7 @@ class Kernel extends HttpKernel
      */
     protected $middleware = [
         \Illuminate\Foundation\Http\Middleware\CheckForMaintenanceMode::class,
+        \Rogue\Http\Middleware\TrustProxies::class,
     ];
 
     /**
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -15,7 +15,10 @@ class Kernel extends HttpKernel
      */
     protected $middleware = [
         \Illuminate\Foundation\Http\Middleware\CheckForMaintenanceMode::class,
+        \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
+        \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
         \Rogue\Http\Middleware\TrustProxies::class,
+        \Rogue\Http\Middleware\TrimStrings::class,
     ];
 
     /**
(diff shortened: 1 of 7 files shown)


app/Http/Kernel.php
```diff
@@ -15,6 +15,10 @@ class Kernel extends HttpKernel
      */
     protected $middleware = [
         \Illuminate\Foundation\Http\Middleware\CheckForMaintenanceMode::class,
+        \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
```
katiecrane: I don't think this file exists 🔍

katiecrane: Oh wait it's from laravel I take it back!

app/Http/Middleware/TrimStrings.php
```diff
@@ -0,0 +1,17 @@
...
+{
+    /**
+     * The names of the attributes that should not be trimmed.
+     *
+     * @var array
+     */
+    protected $except = [
+        //
+    ];
+}
```
katiecrane: I'm confused about what this does, can you explain it to me?

DFurnes: Yes! This property allows us to skip certain fields when trimming extra spaces from the end of strings (for example, Laravel normally includes `password` and `password_confirmation` here, since users might want to have passwords that end in a space character). Since we don't handle passwords in Rogue, I've left it blank for now.

Here's a [little blurb](https://laravel.com/docs/5.5/requests#input-trimming-and-normalization) from the official docs.

app/Http/Kernel.php
DFurnes: A trick! 😈

app/Http/Middleware/TrimStrings.php
katiecrane: Oh got it! Laravel has all the inner workings and this is just where we set the ones not to trim 👌

DFurnes: Yep!

PR #494, opened 2017-11-14
Updates to Trusted Proxy settings.

DFurnes:
#### What's this PR do?
Whoops – two updates to our trusted proxy settings:

🗃 Only parse the `TRUSTED_PROXY_IP_ADDRESSES` environment variable into an array if it has commas, otherwise use as a string. This is required for the `*` option [that Heroku requires](https://devcenter.heroku.com/articles/getting-started-with-laravel#trusting-the-load-balancer).

🔥 Discards the `Forwarded` and `X-Forwarded-Host` headers, as instructed in [Heroku's documentation](https://devcenter.heroku.com/articles/getting-started-with-laravel#trusting-the-load-balancer) because they are not sent by (or sanitized) by Heroku or AWS.

👀

I tested #492 after deploying it and it didn't work, turns out this is why!

#### Relevant tickets
DoSomething/devops#341

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/config/trustedproxy.php
+++ b/config/trustedproxy.php
@@ -1,5 +1,11 @@
 <?php
 
+// Pre-process any comma-separated values into an array.
+$trustedProxies = env('TRUSTED_PROXY_IP_ADDRESSES');
+if (str_contains($trustedProxies, ',')) {
+    $trustedProxies = explode(',', $trustedProxies);
+}
+
 return [
     /*
      * Set trusted proxy IP addresses.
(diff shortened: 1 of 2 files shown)


config/trustedproxy.php
```diff
@@ -1,5 +1,11 @@
 <?php
 
+// Pre-process any comma-separated values into an array.
+$trustedProxies = env('TRUSTED_PROXY_IP_ADDRESSES');
```
sbsmith86: is this a required var now?

DFurnes: Nope, if not set it'll just be null!

PR #495, opened 2017-11-14
User Search Dusk Test

chloealee:
#### What's this PR do?
- Adds two Dusk tests! 
  - Check the flow works for logging in, navigating to the user search page, and searching by just first name throws an error "No user found!"
  - Check the flow to navigate to a user page by a valid email address in our system and it redirects you to their user page. 

- Make sure logic works. 
- All tests are passing running `php artisan dusk`! 

- Have some ❓ s about specific user when testing - more comments in the PR! 
- @DFurnes @sbsmith86 @katiecrane to get the tests to pass, you'll have to create your own `.env.dusk.local` file by copying the `.env.dusk.local.example` file and entering your own email address and pw for `NORTHSTAR_EMAIL` and `NORTHSTAR_PASSWORD` (and other values!) 

#### Relevant tickets
Fixes the third task in this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151961644).

#### Checklist
- [ ] Tested on staging.


tests/Browser/UserSearchTest.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Test visiting the User Search Page.
+     * Searching for a user by name returns "No user found!" message.
+     */
+    public function testIncorrectUserSearchPage()
+    {
+        $this->browse(function (Browser $browser) {
+            $browser->loginAs(User::find(1))
+                ->visit(new UserSearchPage)
+                ->assertSeeIn('@title', 'Users');
```
chloealee: @DFurnes when I run the Dusk test just to find the title of the page matches "Users", I get this error: 
```
1) Tests\Browser\UserSearchTest::testIncorrectUserSearchPage
Facebook\WebDriver\Exception\NoSuchElementException: no such element: Unable to locate element: {"method":"css selector","selector":"body .header__title"}
  (Session info: headless chrome=59.0.3071.109)
  (Driver info: chromedriver=2.32.498513 (2c63aa53b2c658de596ed550eb5267ec5967b351),platform=Linux 4.4.0-81-generic x86_64)

/home/vagrant/DoSomething.org/rogue/vendor/facebook/webdriver/lib/Exception/WebDriverException.php:102
/home/vagrant/DoSomething.org/rogue/vendor/facebook/webdriver/lib/Remote/HttpCommandExecutor.php:320
/home/vagrant/DoSomething.org/rogue/vendor/facebook/webdriver/lib/Remote/RemoteWebDriver.php:535
/home/vagrant/DoSomething.org/rogue/vendor/facebook/webdriver/lib/Remote/RemoteWebDriver.php:175
/home/vagrant/DoSomething.org/rogue/vendor/laravel/dusk/src/ElementResolver.php:332
/home/vagrant/DoSomething.org/rogue/vendor/laravel/dusk/src/Concerns/MakesAssertions.php:270
/home/vagrant/DoSomething.org/rogue/tests/Browser/UserSearchTest.php:21
/home/vagrant/DoSomething.org/rogue/vendor/laravel/dusk/src/TestCase.php:92
/home/vagrant/DoSomething.org/rogue/tests/Browser/UserSearchTest.php:25
```

DFurnes: There isn't a user with an `id` of `1` when this test is run (so it ends up trying to log in as `null` and moving on)! You can either create a user from a factory, or use the `$browser->login()` method to automatically log in as the [default user](https://github.com/DoSomething/rogue/blob/6177293cb2346313c6c49c826ed0ad0c3ac01d74/tests/DuskTestCase.php#L15-L23).

DFurnes: Two handy tools for debugging Dusk tests:

1. Adding a `$browser->dump();` which will dump the HTML contents in the console (just like `$this->dump()` in an API test). 📄 
2. Adding a `$browser->screenshot('name');` which will generate a screenshot of the page in the `tests/screenshots` directory. 📸

chloealee: @DFurnes OHH I forgot it grabs from the `rogue_test` database and just grabbed from whatever database was hooked up (local/qa/prod)! Thank you!

chloealee: those are great tricks to know, thank you @DFurnes !

DFurnes: All that said, it looks like switching to use `$browser->login()` creates it's own issue, since we don't have a valid Northstar token here when the search page makes calls to Northstar QA. Oy!! 

I think your best bet for now might be to actually step through the real auth flow on Northstar QA (using env vars to keep secrets out of the repo) and then we can figure out a better way to mock these external dependencies later on.

chloealee: sounds good, thanks @DFurnes !

tests/Browser/UserSearchTest.php
```diff
@@ -0,0 +1,48 @@
...
+                // ->type('@search', 'taylor')
+                // ->press('Submit')
+                // ->assertSeeIn('@messages', 'No user found!');
+        // });
+
+        $this->browse(function (Browser $browser) {
+            $browser->visit(new HomePage)
+                    ->click('@login-button')
+                    ->dump();
+                    // ->assertPathIs('https://northstar-qa.dosomething.org/register');
```
chloealee: sorry @DFurnes - another q! When I dump here, it gives me this error: 
```
."<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body><pre style="word-wrap: break-word; white-space: pre-wrap;">{"error":"unsupported_grant_type","message":"The authorization grant type is not supported by the authorization server.","hint":"Check the `grant_type` parameter"}</pre></body></html>"
```
I don't think I need to be sending a token or anything in the header at this point since I should just be redirected to https://northstar-qa.dosomething.org/register but maybe I'm wrong?

DFurnes: We need to make sure we have the correct `NORTHSTAR_AUTH_ID` and `NORTHSTAR_AUTH_SECRET` set here (in our `.env.dusk.*` files) so Northstar knows we're allowed to log users in.

We'll also need to figure out what redirect URL this is setting when making the `/authorize` request, and add that to Northstar QA as a valid `redirect_uri`... you can get that from `$browser->driver->getCurrentURL()`.

chloealee: ah ok I will play around with that - thank you!

tests/Browser/UserSearchTest.php
```diff
@@ -0,0 +1,54 @@
...
+
+        $this->browse(function (Browser $browser) {
+            $browser->visit(new HomePage)
+                    ->click('@login-button')
+                    ->assertPathIs('/register')
+                    ->clickLink('Log In')
+                    ->type('username', 'clee@dosomething.org')
+                    ->type('password', env('NORTHSTAR_PASSWORD'))
+                    ->press('Log In')
+                    ->assertPathIs('/campaigns');
```
chloealee: @DFurnes the redirect path is `/campaigns` and I can successfully get to that page! However, whenever I try to do anything next, it doesn't register. Will this be fixed by registering this as a `redirect_uri` in Northstar QA like you mentioned before?

.env.dusk.ci
```diff
@@ -1,52 +0,0 @@
-APP_ENV=ci
```
chloealee: @DFurnes To make sure that things in `.env.dusk.ci` and `.env.dusk.local` got ignored in github, I tried `git rm -rf --cached .` and then `git add .` It then just deleted those files all together here - is this ok?

DFurnes: We probably don't want to delete the files altogether since they set other useful test environment settings (like using the file cache driver), but we can provide specific overrides by exporting environment variables to the shell (from protected environment variable in the [Wercker config](https://app.wercker.com/dosomething/rogue/environment)).

storage/debugbar/.gitignore
```diff
@@ -1,2 +0,0 @@
-*
-!.gitignore
```
DFurnes: This shouldn't be deleted (since we don't want to accidentally commit any files generated by the Laravel DebugBar package that are placed in here).

.env.dusk.ci
DFurnes: Although that still leaves us with trouble on local dev, doesn't it. 😖 Hmm, maybe we do gitignore these files but then use a `.env.dusk` template to create them?

storage/debugbar/.gitignore
chloealee: @DFurnes oops! Seems like this got removed when I did the `git rm -rf --cached .` and then `git add .`. When I tried to re-add this, I got the message "If you decide to go ahead and use a name which begins with a dot the file will be hidden." Do you know how I can re-add this?

.env.dusk.ci
chloealee: would this just to be creating `.env.dusk.ci.example` and .env.dusk.local.example` files?

DFurnes: Yeah, that's what I was thinking. Still a little awkward, but better than nothing!

storage/debugbar/.gitignore
DFurnes: Are you trying to add the `.gitignore` file in the Finder or macOS save dialog? I think you might have to make dotfiles from the command line.

chloealee: ooohhh I was trying in Sublime @DFurnes !

tests/Browser/UserSearchTest.php
chloealee: @DFurnes this was the last q - not sure if you saw! :)

chloealee: oh I just saw that I am able to add that as a `redirect_uri` actually on Aurora! But not sure if that fixes the problem so maybe I'm misunderstanding?

DFurnes: Ah sorry, totally missed this! Yeah, you should be able to paste in the full URL for this into the client's "edit" form on Aurora. I can swing by if you want to walk through it IRL!

chloealee: @DFurnes if you wouldn't mind swinging by, that would be so helpful! I added it but am getting something returned that I'm not expecting (confirms I'm on the `/campaigns` page but when I dump it out, it is a lot of code with `new TypeError("cannot call a class as a function"` etc.)

DFurnes: Ooooh weird! Let's definitely dig into that tomorrow.

tests/Browser/UserSearchTest.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        // Create a signup so that the campaign over page will load.
+        $signup = factory(Signup::class)->create();
+
+        $this->browse(function (Browser $browser) {
+            $browser->visit(new HomePage)
+                    ->click('@login-button')
+                    ->assertPathIs('/register')
+                    ->clickLink('Log In')
+                    ->type('username', 'clee@dosomething.org')
```
chloealee: @DFurnes right now, I'm testing with my own account and have my password in the `NORTHSTAR_PASSWORD=` env variable. I'm not sure how to do this since the tests won't pass without a valid user in Northstar QA right now so I don't think a factory will work. 

I'm thinking for being, I can also create a `NORTHSTAR_EMAIL` env variable in `.env.dusk.local` that devs can input their own credentials to get tests to pass? And for the wercker tests, we can also add a test one? 

OR (this might be better) to just create a test account that is an admin that we can all use across the board? 

Would love your thoughts on this!

tests/Browser/UserSearchTest.php
```diff
@@ -0,0 +1,67 @@
...
+                    // We're already logged in from the first test.
+                    ->assertPathIs('/campaigns')
+                    ->clickLink('User Search')
+                    ->assertPathIs('/users')
+                    ->on(new UserSearchPage)
+                    ->assertSeeIn('@title', 'Users')
+                    ->type('@search', 'clee@dosomething.org')
+                    ->press('Submit')
+                    ->on(new UserPage)
+                    ->assertSeeIn('@title', 'User');
```
chloealee: If we decide to use a test user, we will know their NS id and then we can maybe do an assertion like `->assertPathIs'/users/:northstar_id`?

DFurnes:
@chloealee Can you rebase this against master to fix Wercker builds?

tests/Browser/UserSearchTest.php
DFurnes: I think having people enter their own account is a good idea! 👍

tests/Browser/UserSearchTest.php
DFurnes: I think asserting the page content is probably enough here – maybe look for the user's name in the heading?

chloealee:
@DFurnes rebased but still looks like it is failing :-/

DFurnes:
> @DFurnes rebased but still looks like it is failing :-/

OH, because we still need to merge #498. 🤦‍♂️

tests/Browser/UserSearchTest.php
chloealee: ok! I guess the only way we can tell what the user's name is without making a new call is to also add this in the `.env.dusk.local` file as a variable? Or, I can just make sure that the "User Info" heading is there? 

screen shot 2017-11-16 at 10 54 23 am

chloealee: oh wait, no we have the email address as an env variable so I can just use that!

DFurnes: Ah yeah, smart! 👍

chloealee:
@DFurnes getting [this error](https://app.wercker.com/dosomething/rogue/runs/build/5a0dbb40f591740001285f83?step=5a0dbb574267f3000182869e) for the wercker build. Do I have to add `.env.dusk.local` and `.env.dusk.ci` in wercker?

chloealee:
ahh @DFurnes I see now. I think in this line, it tries to copy the `.env.dusk.ci` file but since it's just `.env.dusk.ci.example` now, it can't find that. A couple questions:

1. Should I also add the `NORTHSTAR_EMAIL` and `NORTHSTAR_PASSWORD` env keys to this doc for when it copies? 
2. When I switch branches, my `.env.dusk.ci` and `.env.dusk.local` files disappear and I have to create them again every time. Did I do something wrong? In the past working with `.env` files, I just changed the `.env.example` file to remove the `.example` part and edit. I can't do that here because I don't want to change adding the template files in the PR. Not sure what to do in this case...

DFurnes:
Oh, I lied on the `.env.dusk.ci.example` rename, since we're already using that as a template! Let's change that back to `.env.dusk.ci` & re-add to the repo, and we can add any _seeeecret_ environment variables to the Wercker config.

chloealee:
ok @DFurnes ! The only thing is when I rename it to `.env.dusk.ci` and `.env.dusk.local`, it won't push back up in the repo since it's ignored?

chloealee:
sorry for all the questions! Do we add the environment variables [here](https://app.wercker.com/dosomething/rogue/environment) in Wercker? In this case, I can just add my username and pw !

DFurnes:
> ok @DFurnes ! The only thing is when I rename it to `.env.dusk.ci` and `.env.dusk.local`, it won't push back up in the repo since it's ignored? 

Right, we'd want to un-ignore the `.env.dusk.ci` file.

> sorry for all the questions! Do we add the environment variables [here](https://app.wercker.com/dosomething/rogue/environment) in Wercker? In this case, I can just add my username and pw ! 

Yep, just make sure to mark them as protected variables so they don't show up in logs.

.env.dusk.ci
```diff
@@ -43,6 +43,7 @@ NORTHSTAR_AUTH_ID=
 NORTHSTAR_AUTH_SECRET=
 NORTHSTAR_CLIENT_SECRET=
 NORTHSTAR_CLIENT_ID=
+NORTHSTAR_EMAIL=
```
DFurnes: I think you might need to leave these lines commented out for the Wercker values to be used.

chloealee:
hmm @DFurnes getting `Facebook\WebDriver\Exception\NoSuchElementException: no such element: Unable to locate element: {"method":"css selector","selector":"body .header__title"}` which I'm guessing because the page isn't loading properly. This is just for the homepage though so not sure why this would be?

DFurnes:
I think you might need to comment out the other `NORTHSTAR_...` env vars in the `.env.dusk.ci` so the values from Wercker get applied. Let me swing by and we can look into it together!

chloealee:
@DFurnes yay I updated with `.env.dusk.local.example` and the note your wrote and everything is passing!

PR #496, opened 2017-11-14
Small bug fixes

sbsmith86:
#### What's this PR do?
I was testing some stuff to make sure I was writing the correct tests 😂 

I found two small issues: 

* The `Quantity` component seems to be getting the `quantity` prop as a number and sometimes as a string. So I updated the validation to allow both with a `@TODO` to figure out why. 

* Delete functionality was broken because we were removing the post from the state but not the `postId` so when the page tried to re-render and sort, we were was using the wrong set of postIds and crashing. This PR removes the deleted post's id as well. 

matilda_eyes

#### Relevant tickets
Fixes 🐛 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/Quantity/index.js
+++ b/resources/assets/components/Quantity/index.js
@@ -20,6 +20,7 @@ class Quantity extends React.Component {
 
 Quantity.propTypes = {
   noun: PropTypes.string,
+  // @TODO - figure out why this come in two different ways.
   quantity: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.number,
--- a/resources/assets/components/Quantity/index.js
+++ b/resources/assets/components/Quantity/index.js
@@ -20,7 +20,10 @@ class Quantity extends React.Component {
 
 Quantity.propTypes = {
   noun: PropTypes.string,
-  quantity: PropTypes.string,
+  quantity: PropTypes.oneOfType([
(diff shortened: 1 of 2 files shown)


resources/assets/components/Quantity/index.js
```diff
@@ -20,7 +20,11 @@ class Quantity extends React.Component {
 
 Quantity.propTypes = {
   noun: PropTypes.string,
-  quantity: PropTypes.string,
+  // @TODO - figure out why this comes in two different ways.
```
DFurnes: Oy! I wonder if this is another case where we sometimes send it through the transformer & sometimes don't.

resources/assets/components/WithReviewing/index.js
```diff
@@ -196,6 +196,13 @@ const reviewComponent = (Component, data) => {
             // Remove the deleted post from the state
             delete(newState.posts[postId]);
 
+            // Remove the postId from the state.
+            const postIdIndex = newState.postIds.indexOf(postId);
```
DFurnes: We could use `newState.postIds.includes(postId)` for a lil' better readability. 🔢

sbsmith86: @DFurnes I think `includes()` just returns a boolean and I was using the index to actually remove the value from the array.

PR #498, opened 2017-11-15
Use updated Wercker box.

DFurnes:
#### What's this PR do?
This pull request uses our [updated Docker box](https://github.com/DoSomething/ds-docker-php/pull/2), which includes ChromeDriver, Node 8, and npm 5.3. This gets us deterministic npm installs using the `package.lock` file & faster builds (2 minutes 22 seconds, down from 4 minutes on the old box).

🚥 👀 

~We can remove the `:box-updates` tag if we decide to merge this in to `latest`.~ We did!

#### Relevant tickets
References #490.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/wercker.yml
+++ b/wercker.yml
@@ -1,16 +1,12 @@
-box: dosomething/ds-docker-php
+box: dosomething/ds-docker-php:build-updates
 
 build:
     steps:
       - script:
-          name: install chromedriver
-          code: |-
-            sudo apt-get update
-            sudo apt-get -y install libxpm4 libxrender1 libgtk2.0-0 libnss3 libgconf-2-4
-            sudo apt-get -y install chromium-browser
-      - script:
           name: start mysql
-          code: sudo service mysql start
+          code: |-
+            chown -R mysql:mysql /var/lib/mysql /var/run/mysqld
+            sudo service mysql start
       - script:
           name: configure environment
           code: |-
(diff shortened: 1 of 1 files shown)


wercker.yml
```diff
@@ -1,16 +1,12 @@
...
-          name: install chromedriver
-          code: |-
-            sudo apt-get update
-            sudo apt-get -y install libxpm4 libxrender1 libgtk2.0-0 libnss3 libgconf-2-4
-            sudo apt-get -y install chromium-browser
-      - script:
           name: start mysql
-          code: sudo service mysql start
+          code: |-
+            chown -R mysql:mysql /var/lib/mysql /var/run/mysqld
```
DFurnes: This seems to be necessary to [fix a permission when starting MySQL](https://serverfault.com/a/872576) (which seems to be a consequence of a newer MySQL release being included in the new build). I tried applying the fix in the Dockerfile & that didn't work, so I think this is our best bet.

DFurnes:
One heads up – this will mean you'll _need_ to be using npm 5+ in order for your changes to be reflected on staging/production, since installs are done from the `package-lock.json` file. If you're using one of the newer Homestead releases, you should be 100% fine.

DFurnes:
@DoSomething/team-bleed Heads up that you'll need to rebase any existing branches against `master` to get builds passing (since otherwise you won't have the included MySQL fix).

sbsmith86:
oooooooo 👀  `@DoSomething/team-bleed`!!!

PR #500, opened 2017-11-15
Fix inconsistencies between up & down migrations.

DFurnes:
#### What's this PR do?
I forgot I'd done this! This was from sorting through a bunch of fatal errors that'd occur when running `php artisan migrate:refresh` (back when I thought that was causing issues with the Laravel 5.4+ test runner). It turns out that wasn't the issue, but it's also nice to be able to rollback migrations again. 😅 

If you pull down this branch, you should be able to safely run `php artisan migrate` and then `php artisan rollback` without getting errors!

🥞 🤷‍♂️ 

#### Relevant tickets
N/A

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/database/migrations/2016_07_27_172344_create_kudos_table.php
+++ b/database/migrations/2016_07_27_172344_create_kudos_table.php
@@ -20,7 +20,7 @@ class CreateKudosTable extends Migration
             $table->integer('taxonomy_id')->index()->comment('The taxonomy_term.tid that this kudos belongs to.');
             $table->timestamps();
 
-            $table->unique(['file_id', 'drupal_id', 'northstar_id', 'taxonomy_id']);
+            $table->unique(['file_id', 'drupal_id', 'northstar_id', 'taxonomy_id'], 'file_drupal_ns_taxonomy_unique');
         });
     }
 
--- a/database/migrations/2016_10_11_200952_drop_file_id_from_reactions.php
+++ b/database/migrations/2016_10_11_200952_drop_file_id_from_reactions.php
@@ -14,6 +14,8 @@ class DropFileIdFromReactions extends Migration
     {
         Schema::table('reactions', function (Blueprint $table) {
             $table->dropColumn('file_id');
+
+            $table->dropUnique('file_drupal_ns_taxonomy_unique');
         });
     }
 
@@ -25,6 +27,8 @@ class DropFileIdFromReactions extends Migration
     public function down()
     {
         Schema::table('reactions', function (Blueprint $table) {
+            $table->unique(['file_id', 'drupal_id', 'northstar_id', 'taxonomy_id'], 'file_drupal_ns_taxonomy_unique');
+
             $table->integer('file_id')->index()->comment('The reportback_file.fid that this kudos applies to.');
         });
     }
--- a/database/migrations/2016_10_11_200952_drop_file_id_from_reactions.php
+++ b/database/migrations/2016_10_11_200952_drop_file_id_from_reactions.php
@@ -27,9 +27,9 @@ class DropFileIdFromReactions extends Migration
     public function down()
     {
         Schema::table('reactions', function (Blueprint $table) {
-            $table->unique(['file_id', 'drupal_id', 'northstar_id', 'taxonomy_id'], 'file_drupal_ns_taxonomy_unique');
-
             $table->integer('file_id')->index()->comment('The reportback_file.fid that this kudos applies to.');
+
+            $table->unique(['file_id', 'drupal_id', 'northstar_id', 'taxonomy_id'], 'file_drupal_ns_taxonomy_unique');
         });
     }
 }
--- a/database/migrations/2017_04_27_183942_drop_unique_northstar_id_index_reactions_table.php
+++ b/database/migrations/2017_04_27_183942_drop_unique_northstar_id_index_reactions_table.php
@@ -13,7 +13,7 @@ class DropUniqueNorthstarIdIndexReactionsTable extends Migration
     public function up()
     {
         Schema::table('reactions', function (Blueprint $table) {
-            $table->dropUnique('kudos_file_id_drupal_id_northstar_id_taxonomy_id_unique');
+            //
         });
     }
 
@@ -25,7 +25,7 @@ class DropUniqueNorthstarIdIndexReactionsTable extends Migration
     public function down()
     {
         Schema::table('reactions', function (Blueprint $table) {
-            $table->unique('kudos_file_id_drupal_id_northstar_id_taxonomy_id_unique');
+            //
         });
     }
 }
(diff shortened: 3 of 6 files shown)


DFurnes:
I also just wanted a chance to test out automatic review requests!

screen shot 2017-11-15 at 1 59 05 pm

database/migrations/2016_07_27_172344_create_kudos_table.php
```diff
@@ -20,7 +20,7 @@ public function up()
             $table->integer('taxonomy_id')->index()->comment('The taxonomy_term.tid that this kudos belongs to.');
             $table->timestamps();
 
-            $table->unique(['file_id', 'drupal_id', 'northstar_id', 'taxonomy_id']);
+            $table->unique(['file_id', 'drupal_id', 'northstar_id', 'taxonomy_id'], 'file_drupal_ns_taxonomy_unique');
```
katiecrane: Is this just so that index isn't named something with `id` in it 4 times?

database/migrations/2016_10_11_200952_drop_file_id_from_reactions.php
```diff
@@ -25,6 +27,8 @@ public function up()
     public function down()
     {
         Schema::table('reactions', function (Blueprint $table) {
+            $table->unique(['file_id', 'drupal_id', 'northstar_id', 'taxonomy_id'], 'file_drupal_ns_taxonomy_unique');
+
             $table->integer('file_id')->index()->comment('The reportback_file.fid that this kudos applies to.');
```
katiecrane: ooo can you do it in this order? this looks like it makes the index using a column that doesn't exist yet?

database/migrations/2017_04_27_183942_drop_unique_northstar_id_index_reactions_table.php
```diff
@@ -13,7 +13,7 @@ class DropUniqueNorthstarIdIndexReactionsTable extends Migration
     public function up()
     {
         Schema::table('reactions', function (Blueprint $table) {
-            $table->dropUnique('kudos_file_id_drupal_id_northstar_id_taxonomy_id_unique');
+            //
```
katiecrane: Why is this removed?

database/migrations/2016_07_27_172344_create_kudos_table.php
DFurnes: Yeah sorry, should probably have added comments for these. By default, Laravel makes an index with the combined column names `file_id_drupal_id_northstar_id_taxonomy_id_unique` and that was giving an "index name too long" error.

database/migrations/2017_04_27_183942_drop_unique_northstar_id_index_reactions_table.php
DFurnes: The index didn't exist, so it was unable to drop it.

database/migrations/2016_07_27_172344_create_kudos_table.php
katiecrane: I approve of that error that is too long! makes sense!

database/migrations/2016_10_11_200952_drop_file_id_from_reactions.php
DFurnes: This is just the inverse of the `up` migration above it! The actual index is created [a few migrations earlier](https://github.com/DoSomething/rogue/blob/24d7f508159293a8c7ec0e0a3fa73fe1cf75f40e/database/migrations/2016_07_27_172344_create_kudos_table.php#L23).

DFurnes: Oh, although I see what you mean now. 🙃 Yeah I'm surprised that works!

database/migrations/2017_04_27_183942_drop_unique_northstar_id_index_reactions_table.php
katiecrane: Sorry this is like time traveling and breaks my brain. Isn't it created in `2016_07_27_172344_create_kudos_table.php`?

database/migrations/2016_10_11_200952_drop_file_id_from_reactions.php
katiecrane: Yeah the `up` one seems in the opposite order that I'd expect too? I'm surprised it works too!

katiecrane: although @DFurnes if you're dropping a column and and index that uses that column, does the order matter?

sbsmith86:
@DFurnes wondering what the status of this PR is.

DFurnes:
Oh, yeah whooops. Wanted to answer @katiecrane's question before merging but never got around to tracking it down. I can try to figure that out this week.

database/migrations/2017_04_27_183942_drop_unique_northstar_id_index_reactions_table.php
DFurnes: Alright yeah – it's created there. The problem that came up was that we deleted `file_id`, a part of that compound index, in [this migration](https://github.com/DoSomething/rogue/pull/500/files#diff-3d5477b189c52ceb8a7b12d8639cf21e) in 2016. When we originally were trying to recreate the index in the `down` migration here, we couldn't:

1. We're trying to just set an index by name... but not specifying what columns it should be tracking.
2. We wouldn't be able to recreate the index here since the `kudos_id` column was deleted back in 2016, and so wouldn't have been recreated yet (we're still "working back" from 2018).

Deleting the index [back then](https://github.com/DoSomething/rogue/pull/500/files#diff-3d5477b189c52ceb8a7b12d8639cf21e) seemed to be the most sensible call at this point.

DFurnes: At least I think that's what's happening.

katiecrane: @DFurnes Yes okay I'm onboard with that I think that is correct. You have merged the operations of `2017_04_27_183942_drop_unique_northstar_id_index_reactions_table.php` and `2016_10_11_200952_drop_file_id_from_reactions.php` into `2016_10_11_200952_drop_file_id_from_reactions.php`.

In that case should `2017_04_27_183942_drop_unique_northstar_id_index_reactions_table.php` be deleted since it is not doing anything?

DFurnes: I'm nervous about deleting a migration, since our existing instances will have thought they'd run it (so if you ran `php artisan migrate:rollback` on your machine after pulling this, it'd explode if it couldn't find the file it thought it had run).

katiecrane: Oh right right right yeah it's in the `migrations` table already. Okay!

DFurnes:
We survived!

PR #502, opened 2017-11-16
Prevent Fastly from caching Glide images.

DFurnes:
#### What's this PR do?
Currently, the glide `images/{post}` route is run through the [`web` middleware group](https://github.com/DoSomething/rogue/blob/8994e436aba82a1ba9008ae99510297c70f3e2a7/app/Http/Kernel.php#L30-L37). Since this adds a session cookie to the response, Fastly will not cache the content and gives it a `x-cache: MISS`. This pull request moves the route into the `routes/api.php` middleware group, which is stateless.

I also had to make some tweaks to get images to properly load from posts that I had seeded on my local machine... I'm still not 100% sure why that started working differently (maybe a change in the filesystem driver in one of the Laravel upgrades?)

👀

This is an important part of getting Glide shipped, since we don't want Rogue to have to re-process images from scratch every time someone tries to view a gallery!

#### Relevant tickets
[#152475940](https://www.pivotaltracker.com/story/show/152475940)

#### Checklist
- [ ] Tested on staging.


routes/api.php
```diff
@@ -9,8 +9,11 @@
  * @see \Rogue\Providers\RouteServiceProvider
  */
 
+// Assets
+$router->get('images/{post}', 'ImagesController@show');
```
DFurnes: I had originally considered moving this into a separate `routes/assets.php` group that was stateless & unprefixed (see 2a27ba6) but I think that's probably overkill. Instead, I'm just removing the `api` prefix on this file & applying it for each version below. Let me know what y'all think.

DFurnes:
Another thing we should look into before shipping this to Production is marking these responses from `images/{post}` with a [cache key](https://docs.fastly.com/guides/purging/getting-started-with-surrogate-keys) so we can destroy all the generated image URLs if we ever delete a post.

routes/api.php
```diff
@@ -20,7 +23,7 @@
 });
 
 // v2 routes
-$router->group(['prefix' => 'v2', 'middleware' => ['auth.api']], function () {
+$router->group(['prefix' => 'api/v2', 'middleware' => ['auth.api']], function () {
```
katiecrane: How is this working without this?

katiecrane: Just read your above comment on why it is being added, but I'm confused about how the `api` part of the path was applied before?

routes/api.php
katiecrane: That makes sense to me!

routes/api.php
DFurnes: It was previously applied [in the RouteServiceProvider](https://github.com/DoSomething/rogue/blob/73c28d575e9647254e59e32a9d0a5f0273822166/app/Providers/RouteServiceProvider.php#L66), but since we need some stateless routes _without_ that prefix it seemed to make sense to pull it out. (Plus it was kinda hard to find!)

katiecrane: Ahh I see! Should this line still be in there then? `->group(base_path('routes/api.php'));` in `mapApiRoutes()`?

DFurnes: Yes, that's what loads this file into the app!

katiecrane: OH I see the prefix part now jeez!! That WAS hard to find!

PR #505, opened 2017-11-20
Blank state if there are no events to show in the History Modal

katiecrane:
#### What's this PR do?
If there are no events under a signup, instead of not opening the history modal at all (which means there would be no way to update quantity either), show a blank state!

Blank:
image

Not blank:
image

Does it look okay? Is the copy okay?

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/152851215)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/HistoryModal/index.js
+++ b/resources/assets/components/HistoryModal/index.js
@@ -70,8 +70,10 @@ class HistoryModal extends React.Component {
           <h3>📖 History 📖</h3>
           <div className="container">
             { ! isEmpty(parsedEvents) ?
-              <p>Below shows the 20 most recent changes to the member's signup. This includes changes to the quantity or why. If you need changes beyond the 20 listed here, please reach out to Team Bleed!</p>
-              <Table headings={['Quantity', 'Why Participated', 'Updated At', 'User']} data={parsedEvents} type="events" />
+              <div>
+                <p>Below shows the 20 most recent changes to the member's signup. This includes changes to the quantity or why. If you need changes beyond the 20 listed here, please reach out to Team Bleed!</p>
+                <Table headings={['Quantity', 'Why Participated', 'Updated At', 'User']} data={parsedEvents} type="events" />
+              </div>
               :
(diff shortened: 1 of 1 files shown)


chloealee:
@katiecrane This looks good! I _thought_ that we always took the first event and put it in the history modal but now looking at the code, we might have changed that? Do we still want that logic to always include the first even in the history log? If not, this looks great! And maybe we should just remove this in the [comment](https://github.com/DoSomething/rogue/blob/master/resources/assets/components/HistoryModal/index.js#L39-L40) so it's not confusing?

katiecrane:
@chloealee oooo I didn't change anything about which events were included. This PR is to handle the modal not crashing when there are NO events so in that case we can't take the first one. It looks like the chunk that you linked to is original to [The History Log PR ](https://github.com/DoSomething/rogue/pull/439) (with some spaces added with a linting fix) so I don't think any of the logic has actually changed?

chloealee:
@katiecrane ah okay! Yeah I was trying to see if there was a fix but maybe I was just off when I wrote that bc looking at that code, it doesn't actually look like we're grabbing the first one if there is only one... 

Another thing I'm trying to think through though - isn't there always an event with a signup? When a signup is created, doesn't it trigger an event?

katiecrane:
@chloealee Sorry I'm confused what you were trying to find a fix for?

I think there was some case on Thor where the events didn't exist? This is just to protect us in edge cases like that and hopefully won't ever see the light of day! But in those cases I don't think we want to block updating quantity/have the modal completely not open due to an error.

chloealee:
Sry! I meant a fix / change to the logic to always grab the first event and add it to the history modal. Maybe I just didn't write it correctly to do that in the first place though bc now that code is confusing me! 

That makes sense to me for the edge cases!

katiecrane:
@chloealee ahh I get it now! Do you think we should add a card to look into that?

chloealee:
@katiecrane I'm not sure if it is _that_ big of a deal, especially since we're eventually going to re-do the events system/table?

PR #506, opened 2017-11-20
Add quantity column to posts table

katiecrane:
#### What's this PR do?
Added a `quantity` column to the `posts` table.
- column is after `northstar_id`
- defaults to `null`
- is nullable
- is an integer

Is that where we want the column to be? Are there different qualities we want it to have?

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/152900013)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/database/migrations/2017_11_20_202525_add_quantity_column_to_posts_table.php
+++ b/database/migrations/2017_11_20_202525_add_quantity_column_to_posts_table.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class AddQuantityColumnToPostsTable extends Migration
+{
+    /**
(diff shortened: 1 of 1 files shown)


database/migrations/2017_11_20_202525_add_quantity_column_to_posts_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('posts', function (Blueprint $table) {
+            $table->integer('quantity')->after('northstar_id')->default(null)->nullable();
```
sbsmith86: I don't think you need to set the `default(null)` explicitly because the default of `nullable()` columns is null

katiecrane: @sbsmith86 ah yeah you're right!

PR #507, opened 2017-11-20
Filtering Dusk Test

chloealee:
#### What's this PR do?
- Adds Dusk tests for filtering post
- Casts `campaign_id` on a signup to a string (before it was an integer)

- Make sure logic seems correct
- Make sure tests pass when running `php artisan dusk` in your vagrant box!

@DFurnes I tried to add a `Login` component but I kept getting this error:
```
Failed asserting that '/_dusk/login/20' matches PCRE pattern "/^\/campaigns/u".
```
Reading the documentation more closely, since this has to do with some urls, I'm wondering if maybe a component isn't the best thing for the login functionality since it says this in the documentation? `As such, components are not bound to specific URLs.`

#### Relevant tickets
Fixes second task in this [Pivotal Ticket](https://www.pivotaltracker.com/n/projects/2019429/stories/151961644)

#### Checklist
- [ ] Tested on staging.


tests/Browser/FilterPostsTest.php
```diff
@@ -0,0 +1,78 @@
...
+                    ->assertPathIs('/campaigns')
+                    ->visit('/campaigns/' . $signup->campaign_id)
+                    ->on(new CampaignSinglePage($signup->campaign_id))
+                    ->assertSee('Post Filters')
+                    ->select('select', 'pending')
+                    ->assertSelected('select', 'pending')
+                    ->press('Apply Filters')
+                    ->pause(5000)
+                    // ->screenshot('hello');
+                    ->dump();
```
chloealee: @DFurnes this is still WIP but not sure why this is happening - when I dump out the HTML here, it says that there are no results for this search. However, in line 24, I create a post that has a status of "pending" so this should pop up in the results. Are there any gotchas that I'm not thinking of?

tests/Browser/pages/CampaignSinglePage.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /**
+     * Get the URL for the page.
+     *
+     * @return string
+     */
+    public function url()
+    {
+        return '/campaigns' . $campaignId;
```
DFurnes: I think this needs to be `$this->campaignId`!

tests/Browser/FilterPostsTest.php
DFurnes: I left a comment below, I think the campaign ID wasn't being added to the URL!

tests/Browser/pages/CampaignSinglePage.php
chloealee: ah! I changed that to `$this->campaignId` and it's still working the same as when it was just `$campaignId` 🤔

chloealee: also realized I was missing another `/` after `'/campaigns` so I'm not sure how it was getting to the page! But I updated to `return '/campaigns/' . $this->campaignId` and still getting the same error that there are no results :-/ It's weird too because the whole time when I was doing the screenshot, it was in fact on the campaign single page (again not sure how it got there with those two errors)

tests/Browser/FilterPostsTest.php
```diff
@@ -0,0 +1,78 @@
...
+                    //     $browser->login($browser);
+                    // })
+                    ->click('@login-button')
+                    ->assertPathIs('/register')
+                    ->clickLink('Log In')
+                    ->type('username', env('NORTHSTAR_EMAIL'))
+                    ->type('password', env('NORTHSTAR_PASSWORD'))
+                    ->press('Log In')
+                    ->assertPathIs('/campaigns')
+                    ->visit('/campaigns/' . $signup->campaign_id)
```
chloealee: oh it probably was ok since it was correct here and then the below line (`->on(new CampaignSinglePage($signup->campaign_id))` probably doesn't need to be there.

tests/Browser/FilterPostsTest.php
chloealee: ah @DFurnes I figured it out! When creating the post, for some reason, the post's `campaign_id` is a string, not an integer, so when it does the query it doesn't find anything. Hmm this is tricky for long term testing bc sometimes the `campaign_id` will be an integer (if coming from P-A) and sometimes it will be a string (if coming from P-N). For now, I think I will cast it to whatever the type the `$signup->campaign_id` is as a workaround.

DFurnes: Oooof! That's frustrating! Ideally we want these to always be strings (even if it's a string of numbers), so maybe something we can fix in the model with a [cast](https://laravel.com/docs/5.5/eloquent-mutators#attribute-casting)?

chloealee: oh I love that - thanks @DFurnes !

tests/Browser/FilterPostsTest.php
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * Helper function to create a post with a specific status and associate it with a signup.
+     */
+    private function createAssociatedPostWithStatus($signup, $status)
+    {
+        $post = $signup->posts()->save(factory(Post::class)->make(['status' => $status]));
+        $post->campaign_id = $signup->campaign_id;
+        $post->save();
+
```
chloealee: this was a little strange but for some reason, the post wasn't getting the `campaign_id` when it was created. I had to fill this in and save it manually here.

chloealee: I tried to update the model factory by doing:
```
$factory->define(Post::class, function (Generator $faker) {
    $uploadPath = $faker->file(storage_path('fixtures'));
    $upload = new UploadedFile($uploadPath, basename($uploadPath), 'image/jpeg');
    $url = app(AWS::class)->storeImage($upload, $faker->unique()->randomNumber(5));
    return [
        'signup_id' => $signup->campaign_id,
        'url' => $url,
        'caption' => $faker->sentence(),
        'source' => $faker->randomElement(['phoenix-oauth', 'phoenix-next']),
        'remote_addr' => $faker->ipv4,
        'status' => 'pending',
    ];
});
```

but for some reason in the test, when I `dd` the `$signup->campaign_id` and the `$post->campaign_id` they were different. So reverting back tot his helper method.

PR #513, opened 2017-11-21
Artisan command to move quantity from Signup to Post

katiecrane:
#### What's this PR do?
Here it is!
What it does:
1. Takes in an argument of the Signup ID to start with. If none given, starts at the very beginning (ID=1).
2. Grabs all the signups (with their posts their posts) in chunks of 100.
3. Check if the signup has posts. If not, move on to the next signup.
4. See if any of the posts have quantity. If they do, see if it's ALL of them. If it's all of them, move on to the next signup. If it's only some of them, subtract the quantity on the posts from the quantity on the signup. If some of the posts are `accepted`, grab the most recently submitted `accepted` one and give it the difference in quantity. If there are no accepted posts, grab the most recently submitted post and give it the quantity from the signup. Give all other posts with no quantity `0`.
5. If none of the posts have a quantity, do the same thing except just using the raw quantity from the signup.
6. Move on to the next signup.

Important notes:
- as it stands now, WILL update `updated_at` on Signups
- can be re-run, re-started at any point without duplicating or otherwise mucking anything up (except re-running will again update `updated_at` on posts and signups)
- includes a progress bar
- all output goes straight to the logs prefixed with `rogue:postquantity:`
- does NOT send messages to Blink because it does everything directly to the models and does not hit the `PostService` or `SignupService` which hold the Blink sending logic

Question: Do we want to save the Posts without timestamps? Or do we want running this to update their `updated_at`? Do we want to touch the signup?

Does this do what it should based on [this doc](https://docs.google.com/document/d/1ZhwLH7LxzWfWuX5RMfFlLwJYdR2UOArBFXIePtL3tlQ/edit)?

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/152900315)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Console/Commands/PostQuantity.php
+++ b/app/Console/Commands/PostQuantity.php
@@ -66,21 +66,21 @@ class PostQuantity extends Command
                 if ($acceptedPosts->isNotEmpty()) {
                     $mostRecentAcceptedPost = $acceptedPosts->first();
                     $mostRecentAcceptedPost->quantity = $quantity;
-                    $mostRecentAcceptedPost->save(['touches' => FALSE]);
+                    $mostRecentAcceptedPost->save(['touches' => false]);
                     info('rogue:postquantity: Put quantity ' . $quantity . ' on post ' . $mostRecentAcceptedPost->id);
                 }
                 // If no accepted posts, put quantity on most recent post (based on creation)
                 else {
                     $mostRecentPost = $posts->sortByDesc('created_at')->first();
                     $mostRecentPost->quantity = $quantity;
-                    $mostRecentPost->save(['touches' => FALSE]);
+                    $mostRecentPost->save(['touches' => false]);
                     info('rogue:postquantity: Put quantity ' . $quantity . ' on post ' . $mostRecentPost->id);
                 }
                 // Put quantity of 0 on all other posts under this signup
                 foreach ($posts as $post) {
                     if (is_null($post->quantity)) {
                         $post->quantity = 0;
-                        $post->save(['touches' => FALSE]);
+                        $post->save(['touches' => false]);
                         info('rogue:postquantity: Put quantity 0 on post ' . $post->id);
                     }
                 }
--- a/app/Console/Commands/PostQuantity.php
+++ b/app/Console/Commands/PostQuantity.php
@@ -1,0 +1,98 @@
+<?php
+
+namespace Rogue\Console\Commands;
+
+use Rogue\Models\Signup;
+use Illuminate\Console\Command;
+
+class PostQuantity extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'rogue:postquantity {start=1}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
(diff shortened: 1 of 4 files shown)


ngjo:
@katiecrane Quasar needs the signup `updated_at` to be updated in order to ingest it on their end. This will update all the signups and James + Sena are coming to groom on Wed to talk about the rollout plan to make sure they're able to ingest on their end. cc @jamjensen

app/Console/Commands/PostQuantity.php
```diff
@@ -0,0 +1,96 @@
...
+        // Get all signups starting from $start in order of ID
+        Signup::where('id', '>=', $start)->orderBy('id')->with('posts')->chunk(100, function ($signups) use ($bar) {
+            foreach ($signups as $signup) {
+                // Get the posts for the signup
+                $posts = $signup->posts;
+
+                // If no posts, move on
+                if ($posts->isEmpty()) {
+                    info('rogue:postquantity: No posts for signup ' . $signup->id);
+                    $bar->advance();
```
chloealee: this signup bar is so cool!

chloealee:
this looks great to me @katiecrane ! Just had a question about your question: 

> Question: Do we want to save the Posts without timestamps? Or do we want running this to update their updated_at? 

What do you mean by the Posts without timestamps?

katiecrane:
@chloealee I have an answer to your question about my question!

Laravel allows you to save a model without updated its `updated_at` timestamp. We definitely need to update the `updated_at` on the Signup and I think because of that it makes sense to do the Post too but @ngjo I'm not sure if we have an answer from data on this particular detail or if it even matters.

chloealee:
ahhh I see @katiecrane thanks for explaining that! I was thinking the `created_at` so got confused. I think you're right if updating the signup based on the changes to the post, it makes sense to also update the post's `updated_at`!

ngjo:
@katiecrane we're gonna talk about it at groom tomorrow! Sena and James are joining us for the first 10-15 min!

katiecrane:
@ngjo ack I even went back in our conversation and I somehow missed that crucial bit! thanks Jen!

app/Console/Commands/PostQuantity.php
```diff
@@ -0,0 +1,123 @@
...
+                if ($postQuantityTotal) {
+                    $postQuantities = $posts->pluck('quantity')->toArray();
+
+                    // If none have null quant, continue we are done here
+                    if (! in_array(null, $postQuantities, true)) {
+                        continue;
+                    }
+
+                    // If some have null quant we need to fill them in
+                    $missingQuantity = $signup->getQuantity() - $postQuantityTotal;
```
sbsmith86: Just making sure I got this, we need this logic in there so we can handle posts that come in while this change is being made, right?

katiecrane: @sbsmith86 This came out of the impromptu quantity discussion we had about doing the subtraction on the Rogue side. Regardless of what quantity we are returning, if we are doing any storing of quantity on the posts before the script runs, we need this logic to work with signups that have some posts with quantity and some without.

sbsmith86: right right ok!

PR #514, opened 2017-11-28
Return sum of post quantities or signup quantity

katiecrane:
#### What's this PR do?
When hitting `v1/reportbacks`, if the particular post we are returning has a quantity of `null` (indicating that we have not yet added the quantity values into the posts table), we return the `quantity` from the signup that the post is associated with. Otherwise, if the post _does_ have a quantity (including `0`), sum up the quantities of that post and all the other posts under the same signup and return that.

Does any documentation need to be updated with this?

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/153073306)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -233,4 +233,29 @@ class Post extends Model
     {
         return Reaction::where('post_id', $postId)->count();
     }
+
+    /**
+     * Get the total quantity for all posts under the signup that this post is associated with.
+     *
+     * @return int
+     */
+    public function getTotalQuantity()
+    {
+        $signup = $this->signup;
+
+        // If the post has NULL quantity, it means that the quantity is still on the signup
+        if (is_null($this->quantity)) {
+            return $signup->quantity;
+        }
+
+        // Loop through all the posts and sum their quantities
+        $posts = $signup->posts;
+        $quantity = 0;
+
+        foreach ($posts as $post) {
+            $quantity += $post->quantity;
+        }
+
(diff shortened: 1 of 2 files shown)


app/Models/Post.php
```diff
@@ -233,4 +233,29 @@ public function getTotalReactions($postId)
...
+     * Get the total quantity for all posts under the signup that this post is associated with.
+     *
+     * @return int
+     */
+    public function getTotalQuantity()
+    {
+        $signup = $this->signup;
+
+        // If the post has NULL quantity, it means that the quantity is still on the signup
+        if (is_null($this->quantity)) {
```
chloealee: oh wait - one question @katiecrane ! What happens if there is `quantity_pending` on the signup?

chloealee: also, do we want to return the `quantity` if it isn't `null`?

katiecrane: ahh shoot! I gotta add a check for that!

app/Models/Post.php
```diff
@@ -233,4 +233,29 @@ public function getTotalReactions($postId)
...
+            return $signup->getQuantity();
```
chloealee: got it - this makes sense to me now!

katiecrane: @chloealee good to merge now?

chloealee: @katiecrane yes!

PR #515, opened 2017-11-28
Quantity update to GET /api/v2/activity

chloealee:
#### What's this PR do?
- Updates `SignupTransformer` to return quantity as a summed total of quantity across all posts under a signup, if `$signup->quantity` doesn't exists.
  - Combines `getTotalQuantity()` into `getQuantity()`
- Adds a test to make sure quantities are returned correctly. 

- Does the logic in the test and code make sense? 
- All tests should pass! 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/153032066) Pivotal card. 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -34,14 +34,11 @@ class SignupTransformer extends TransformerAbstract
      */
     public function transform(Signup $signup)
     {
-        // $firstPost = $signup->posts()->first();
-
         return [
             'signup_id' => $signup->id,
             'northstar_id' => $signup->northstar_id,
             'campaign_id' => $signup->campaign_id,
             'campaign_run_id' => $signup->campaign_run_id,
-            // 'quantity' => $firstPost ? $firstPost->getTotalQuantity() : $signup->getQuantity(),
             'quantity' => $signup->getQuantity(),
             'why_participated' => $signup->why_participated,
             'signup_source' => $signup->source,
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -34,12 +34,14 @@ class SignupTransformer extends TransformerAbstract
      */
     public function transform(Signup $signup)
     {
+        $firstPost = $signup->posts()->first();
+
         return [
             'signup_id' => $signup->id,
             'northstar_id' => $signup->northstar_id,
             'campaign_id' => $signup->campaign_id,
             'campaign_run_id' => $signup->campaign_run_id,
-            'quantity' => $signup->getQuantity(),
+            'quantity' => $firstPost ? $firstPost->getTotalQuantity() : $signup->getQuantity(),
             'why_participated' => $signup->why_participated,
             'signup_source' => $signup->source,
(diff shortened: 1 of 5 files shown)


app/Http/Transformers/SignupTransformer.php
```diff
@@ -34,12 +34,14 @@ class SignupTransformer extends TransformerAbstract
...
     {
+        $firstPost = $signup->posts()->first();
+
         return [
             'signup_id' => $signup->id,
             'northstar_id' => $signup->northstar_id,
             'campaign_id' => $signup->campaign_id,
             'campaign_run_id' => $signup->campaign_run_id,
-            'quantity' => $signup->getQuantity(),
+            'quantity' => $firstPost ? $firstPost->getTotalQuantity() : $signup->getQuantity(),
```
katiecrane: hmm I wonder if `getQuantity()` on the signup should be updated to include the logic that you do here (see if there are posts, grab the quantity from them if it can, otherwise return what it can find in the signups table)

chloealee: @katiecrane yes I like that idea! I'll update now!

chloealee: the only thing is I'm worried about is the conditionals looping back and forth between these two functions (`getQuantity` and `getTotalQuantity`). Once we move the quantity to the posts, will there still be a `quantity` or `quantity_pending` on the signup?

katiecrane: ah if you add all the logic into `getQuantity()` then you can just call that here without ever calling `getTotalQuantity()`. Now I'm thinking it should have been that initially instead of the function on the post model.

katiecrane: Initially there will be some quantities that exist in both places, but we should only need to worry about the column on the post table at that point I think

chloealee: ahhh I see that makes sense! Thank you @katiecrane ! Combined `getTotalQuantity` into `getQuantity` - does this still work for the reportback transformer?

app/Models/Signup.php
```diff
@@ -131,10 +133,26 @@ public function getQuantity()
...
         // column at all. We only want to return what is in the quanity_pending column
         // if is the only place quanity is stored.
-        if (! is_null($this->quantity_pending) && is_null($this->quantity)) {
-            return $this->quantity_pending;
+
+        // If the there is at least one post and it has NULL quantity,
+        // it means that the quantity is still on the signup.
+        $posts = $this->posts;
+
+        if ($posts->count() <= 0 || is_null($posts->first()->quantity)) {
```
katiecrane: Nitty gritty but this could probably be `== 0` or could use `isEmpty()` to determine if there are posts or not

chloealee: ah yes - I like `isEmpty()` a lot better!

app/Http/Transformers/SignupTransformer.php
katiecrane: @chloealee yes I think the reportback transformer is great with this!

PR #516, opened 2017-11-29
WIP: POST api/v2/posts - store quantity on post

sbsmith86:
#### What's this PR do?
This updates the `POST api/v2/posts` endpoint so that

* It stores quantity on the post.  551c826
* Returns quantity from the post in `PostTransformer` 926fe6a
* Updates `PostApiTests.php` so that we check that quantity was stored on the `post` in our tests 9478a2a

👀  

Tests should pass!  

* For now we are storing quantity on _both_ the `post` and the `signup` so that we can continue to support both data schemas while we make this transition. 

* I made a quick little update to make the `quantity` column on `posts` be an index column because I added it the filterable index values in the model and that seemed to make sense. 

#### Relevant tickets
Addresses https://www.pivotaltracker.com/story/show/153072890

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Traits/PostRequests.php
+++ b/app/Http/Controllers/Traits/PostRequests.php
@@ -24,7 +24,6 @@ trait PostRequests
 
         $updating = ! is_null($signup);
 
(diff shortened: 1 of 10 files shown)


sbsmith86:
Closing in favor of a new PR

PR #517, opened 2017-11-29
Update POST /api/v3/signups and PATCH /api/v3/signups/:signup_id - remove storing quantity on signup & tests

chloealee:
#### What's this PR do?
- Creates new `Three\SignupService` to point to `Three\SignupRepository`. 
- Creates new `Three\SignupRepository` to not store quantity on the signup. 
- Updates so `POST /api/v3/signups` does not store quantity on signup.
- Updates so `PATCH /api/v3/signups/:signup_id` doesn't update `quantity` and only updates `why_participated`
- Updates assertions in test around quantity.
- Adds test to make sure quantity is returned correctly (if quantity is on the posts, return a sum of all the quantities across the signup's posts. Otherwise, return quantity on the signup). 
- Updates documentation. 

👀 
- Make sure tests pass! 

#### Note / Question
- **Important** No other apps should use `v3` endpoints right now until we're 100% ready for all quantity to be stored on the post and not the signup (or both signup and post).
- Question: Do we need to notify Blink of these updates? I don't _think_ we should since a signup now doesn't always have a quantity sent so I think it'll be fine but just want to double check! 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/153073363) Pivotal card.

#### Checklist
- [ ] Tested on staging.


chloealee:
The Wercker build is failing because of the Post Tests. Is this OK? I can fix these tests temporarily just to get it to pass but think that these should actually be updated when the `POST /v3/posts` endpoint gets updated too to save quantity?

chloealee:
Note: after team chat re: solution on keeping quantity on post AND signup, pausing on this for now and moving on to updating `POST v3/posts` endpoint first since we'll want to keep the logic of storing on signup.

app/Repositories/Three/SignupRepository.php
```diff
@@ -0,0 +1,63 @@
...
+        $signup = new Signup;
+
+        $signup->northstar_id = $data['northstar_id'];
+        $signup->campaign_id = $data['campaign_id'];
+        $signup->campaign_run_id = isset($data['campaign_run_id']) ? $data['campaign_run_id'] : null;
+        $signup->why_participated = isset($data['why_participated']) ? $data['why_participated'] : null;
+        $signup->source = isset($data['source']) ? $data['source'] : null;
+        $signup->details = isset($data['details']) ? $data['details'] : null;
+
+        if (isset($data['created_at'])) {
```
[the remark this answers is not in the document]
katiecrane: I agree I think we can remove this moving forward! 🔥

tests/Http/Three/SignupTest.php
```diff
@@ -262,7 +262,7 @@ public function testUpdatingASignup()
      * PATCH /api/v3/signups/186
      * @return void
      */
-    public function testValidationgForUpdatingASignup()
+    public function testValidationForUpdatingASignup()
```
sbsmith86: 👁 !!!

tests/Http/Three/SignupTest.php
```diff
@@ -281,10 +281,60 @@ public function testUnauthenticatedUserUpdatingASignup()
...
+     * @return void
+     */
+    public function testQuantityOnSignupIndex()
+    {
+        // Create one signup with a quantity_pending.
+        $firstSignup = factory(Signup::class)->create();
+
+        // Create another signup with a quantity.
+        $secondSignup = factory(Signup::class)->create();
+        $secondSignup->quantity_pending = null;
```
sbsmith86: I don't think `v3` needs to worry about `quantity_pending` because we will probably kill that after we migrate `quantity` to `posts`.

chloealee: ah true! I'll get rid of this.

tests/Http/Three/SignupTest.php
```diff
@@ -281,10 +281,60 @@ public function testUnauthenticatedUserUpdatingASignup()
...
+        $secondSignup->quantity = 8;
+        $secondSignup->save();
+
+        // Create another signup with three posts with quantities.
+        $thirdSignup = factory(Signup::class)->create();
+        $thirdSignup->quantity_pending = null;
+        $thirdSignup->save();
+
+        for ($i = 0; $i < 3; $i++) {
```
sbsmith86: If you update the `PostFactory` in `ModelFactory.php` so that it creates a `Post` with a random `quantity`, you can then create three posts with quantities with ` $posts = factory(Post::class, 3)->create();` 

Later in your test you can sum up the quantities over the set of posts you created, and confirm that the quantity you received is equal to the quantity you summed up. 

I think that would be cool, because then we can pretty easily create `posts` with quantities in our tests whenever we need them.

chloealee: Great idea! 💡

chloealee: updated this and also `testQuantityOnActivityIndex` to use the same logic with the updated `PostFactory`!

tests/Http/Api/ActivityApiTest.php
```diff
@@ -291,10 +291,14 @@ public function testQuantityOnActivityIndex()
...
-            $post = $thirdSignup->posts()->save(factory(Post::class)->make());
-            $post->quantity = 3;
+        $posts = factory(Post::class, 3)->create();
+        $postsTotalQuantity = [];
+
+        foreach ($posts as $post) {
+            $post->signup()->associate($thirdSignup);
             $post->save();
+
+            array_push($postsTotalQuantity, $post->quantity);
```
sbsmith86: And then, _I think_, to sum up all the quantities, you can grab the posts as a collection and use `sum` 

`$signup->posts->sum('quantity');`

tests/Http/Api/ActivityApiTest.php
```diff
@@ -291,10 +291,14 @@ public function testQuantityOnActivityIndex()
...
         $thirdSignup->save();
 
-        for ($i = 0; $i < 3; $i++) {
-            $post = $thirdSignup->posts()->save(factory(Post::class)->make());
-            $post->quantity = 3;
+        $posts = factory(Post::class, 3)->create();
+        $postsTotalQuantity = [];
+
+        foreach ($posts as $post) {
+            $post->signup()->associate($thirdSignup);
```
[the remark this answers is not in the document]
sbsmith86: This reads a little odd to me because it looks like we are associating a signup with a different signup.

tests/Http/Api/ActivityApiTest.php
chloealee: omg yes this is WAY better - thanks @sbsmith86 ! Updated :)

PR #518, opened 2017-11-29
Require OAuth tokens for v3 endpoints.

DFurnes:
#### What's this PR do?
This pull request updates Rogue's v3 endpoints to use OAuth tokens! This is important because it's how we'll allow requests made from other apps (like Phoenix Next or [a GraphQL gateway](https://github.com/DoSomething/graphql-prototype)) to know _who_ is making the request.

🍗 I renamed the `auth.api` middleware to `legacy-api` so that it isn't confusing next to `auth:api` (the default Laravel middleware, with the `api` guard specified).

💂 Adds a `guard` middleware to set the default guard for a route group – so for example, any call to `Auth::id()` will use a JWT on `v3/...` routes. Otherwise you'd have to remember to do `Auth::guard('api')->id()`.

🚁 Adds [test helpers](https://github.com/DoSomething/gateway/pull/97) to make it easy to test API requests that are signed with Northstar tokens!

Tests should still pass! Sorry the diff is kinda large, but happy to talk over any details! 😬

Diego's starting to use these endpoints and I realized we should really finish up this work so he's unblocked! This sets us up so we can use `auth()->id()` to grab the user, or `role:admin` middleware to lock down the `DELETE` route, and so on.

Two changes here – `v1` and `v2` endpoints will only work with `X-DS-Rogue-API-Key` now, and `v3` endpoints will only work with `Authorization` header with a [Northstar token](https://github.com/DoSomething/northstar/blob/dev/documentation/authentication.md).

#### Relevant tickets
[#152602162](https://www.pivotaltracker.com/story/show/152602162)

#### Checklist
- [ ] Tested on staging.


DFurnes:
Another note: this also means that you can't just put `X-DS-REST-API-Key` header in Postman or Paw for these v3 endpoints! I'll sit down with folks to help set up everyone's Paw/Postman to be able to fetch Northstar tokens for testing as well. I promise it's easy! 😺

composer.json
```diff
@@ -10,7 +10,7 @@
     "aws/aws-sdk-php-laravel": "^3.1",
     "barryvdh/laravel-debugbar": "^2.3",
     "doctrine/dbal": "~2.5.13",
-    "dosomething/gateway": "^1.7.0",
+    "dosomething/gateway": "dev-test-helpers",
```
sbsmith86: is this a stable version?

DFurnes: I'll run a `composer update` with the stable release as soon as DoSomething/gateway#97 is approved!

DFurnes: Updated to use stable release in 045d5ff!

PR #519, opened 2017-11-30
Some example stuff we can do with OAuth on v3 routes!

DFurnes:
#### What's this PR do?
This is just a quick demo of what we can do with OAuth now on our v3 routes! This pull request adds some rudimentary access control to the `v3/posts` index:

🤺 Users will now automatically see their own reaction count based on their OAuth token, rather than having to send an `as_user` parameter with a user ID.

✋ Users will not see other people's posts until they've been approved (unless they're an admin)!

I'd check commit-by-commit since e29b6a5 just extracts the `index` method out of the trait.

This is just to demonstrate the cool things we can do with OAuth tokens (and because I want to use this as part of my GraphQL demo 😁), I'll work with @ngjo to put together some cards with other places we can do things like this in Rogue!

#### Relevant tickets
[#152602162](https://www.pivotaltracker.com/story/show/152602162)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -2,17 +2,18 @@
 
 namespace Rogue\Http\Controllers\Three;
 
+use Illuminate\Http\Request;
 use Rogue\Models\Post;
 use Rogue\Services\PostService;
 use Rogue\Repositories\SignupRepository;
 use Rogue\Http\Requests\Three\PostRequest;
 use Rogue\Http\Transformers\PostTransformer;
 use Rogue\Http\Controllers\Api\ApiController;
-use Rogue\Http\Controllers\Traits\PostRequests;
+use Rogue\Http\Controllers\Traits\FiltersRequests;
 
 class PostsController extends ApiController
 {
-    use PostRequests;
+    use FiltersRequests;
 
     /**
      * The post service instance.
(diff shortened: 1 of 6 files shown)


DFurnes:
Rebased 5c048b1 to fix some StyleCI issues. 💨

app/Http/Controllers/Three/PostsController.php
```diff
@@ -67,14 +67,22 @@ public function index(Request $request)
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Post::$indexes);
 
-        // If user param is passed, return whether or not the user has liked the particular post.
-        if ($request->query('as_user')) {
-            $userId = $request->query('as_user');
-            $query = $query->with(['reactions' => function ($query) use ($userId) {
-                $query->where('northstar_id', '=', $userId);
+        // If a user made the request, return whether or not they liked each post.
+        if (auth()->check()) {
```
chloealee: should we document any of this stuff anywhere now without needing the `as_user` param?

DFurnes: Yes! It looks like we use the same docs for v2 and v3, so I can split those out. 👍

chloealee: Awesome, thanks Dave!

DFurnes: Done in cf960be!

PR #520, opened 2017-11-30
Run migrations & fetch public key on new releases.

DFurnes:
#### What's this PR do?
This pull request adds a [`release` task](https://devcenter.heroku.com/articles/release-phase) to the Procfile to run migrations & a [PHP build step](https://devcenter.heroku.com/articles/php-support#build-behavior) to download the public key from Northstar. (This seems to be the only way to do this, since things from the release step don't seem to persist on the filesystem).

👀

The `gateway:key` command is necessary for Rogue Thor to accept OAuth tokens! Otherwise it 500s when it can't find the `public.key` file to validate a token with.

#### Relevant tickets
[#152602162](https://www.pivotaltracker.com/story/show/152602162) / DoSomething/devops#352

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,4 @@
 web: vendor/bin/heroku-php-apache2 /public/
 
+release: php artisan migrate --force
+
--- a/composer.json
+++ b/composer.json
@@ -54,6 +54,9 @@
     "post-autoload-dump": [
       "Illuminate\\Foundation\\ComposerScripts::postAutoloadDump",
       "@php artisan package:discover"
+    ],
+    "compile": [
+      "@php artisan gateway:key"
     ]
   },
   "config": {


Procfile
```diff
@@ -1,2 +1,4 @@
 web: vendor/bin/heroku-php-apache2 /public/
 
+release: php artisan migrate --force
```
katiecrane:
```
Some migration operations are destructive, which means they may cause you to lose data. In order to protect you from running these commands against your production database, you will be prompted for confirmation before the commands are executed. To force the commands to run without a prompt, use the --force flag:
```
Does this mean that it always prompts you for any migrations in production or that it knows which ones in particular are destructive?

DFurnes: Yeah, if `APP_ENV` is `production` it'll always ask for user confirmation.

DFurnes: (Which is considerate if you're SSHing in and don't realize you're not on your local dev, but not particularly helpful when trying to do something in a script!)

PR #521, opened 2017-12-04
Quantity updates Feature Flag

sbsmith86:
#### What's this PR do?

* Hides the logic around figuring out how to return signup quantity under a feature flag. 
* Fixes quantity tests to account for when the flag is on or off. Mainly just tests if the `quantity` returned is equal to the quantity we stored, which uses the feature flag to determine what to store. 

👀 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/153371277

#### Checklist
- [ ] Tested on staging.


app/Models/Signup.php
```diff
@@ -137,6 +137,11 @@ public function toBlinkPayload()
      */
     public function getQuantity()
     {
+        if (! config('features.v3QuantitySupport')) {
+            return $this->quantity;
```
chloealee: I _think_ we still want to do the logic around checking if there is a `quantity_pending` first and return that if `quantity` is `null` (below in lines 155-161)?

sbsmith86: ah yeeepppp!!! i always forget about `quantity_pending`

app/Models/Signup.php
```diff
@@ -137,6 +137,16 @@ public function toBlinkPayload()
      */
     public function getQuantity()
     {
+        // @TODO - remove once we finish DB updates.
+        if (! config('features.v3QuantitySupport')) {
+            if (! is_null($this->quantity_pending) && is_null($this->quantity)) {
```
sbsmith86: @chloealee updated!

PR #523, opened 2017-12-05
Oauth v3 signup create update

sbsmith86:
#### What's this PR do?
Updates the `Three/SignupRepository` so that it uses `auth()->id()` as the `northstar_id` and `token()->client()` as the `source` 

👀 

This starts to address some updates needed to work nicely with the new Oauth V3 endpoints. 

This doesn't resolve the issue, but because @chloealee is working on `v3 posts` endpoints, I don't want to conflict with her work. 

#### Relevant tickets
Addresses https://www.pivotaltracker.com/story/show/153365724

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/Three/SignupRepository.php
+++ b/app/Repositories/Three/SignupRepository.php
@@ -17,11 +17,11 @@ class SignupRepository
         // Create the signup
         $signup = new Signup;
 
-        $signup->northstar_id = $data['northstar_id'];
+        $signup->northstar_id = auth()->id();
         $signup->campaign_id = $data['campaign_id'];
         $signup->campaign_run_id = isset($data['campaign_run_id']) ? $data['campaign_run_id'] : null;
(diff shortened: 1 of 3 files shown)


app/Repositories/Three/SignupRepository.php
```diff
@@ -17,19 +17,19 @@ public function create($data)
         // Create the signup
         $signup = new Signup;
 
-        $signup->northstar_id = $data['northstar_id'];
+        $signup->northstar_id = auth()->id();
```
DFurnes: It might make sense to send these along as parameters from the controller (so that we could use the repository for things that aren't HTTP requests, like for creating signups from a CSV or something).

DFurnes:
It'd also be good to swap out the test case to use `withAccessToken($id, 'user')` and then assert that the correct user ID is stored on the signup.

PR #524, opened 2017-12-05
POST api/v2/posts quantity update

sbsmith86:
#### What's this PR do?
This updates the `POST api/v2/posts` endpoint so that

* It stores quantity on the post based on [this discussion](https://github.com/orgs/DoSomething/teams/team-bleed), if the feature flag to support that is turned on. Otherwise, if the feature flag is not turned on, we just store the submitted quantity on the signup. 
* Returns quantity from the post in `PostTransformer`
* Updates `PostApiTests.php` so that we check that quantity was stored on the `post` in our tests. 
* Adds two new tests to `PostApiTests.php` so we can test our quantity storing logic with the flag turned on or off. 

Apologies in advanced if this is complicated to read! the TLDR; is that if we have the flag on, we will split quantity amongst posts, if not, we will just store quantity on the signup as we have always done. 

👀  

Tests should pass!  

See this write up for more detailed explanation of what is going on: https://github.com/orgs/DoSomething/teams/team-bleed

* For now we are storing quantity on _both_ the `post` and the `signup` so that we can continue to support both data schemas while we make this transition. 

* I made a quick little update to make the `quantity` column on `posts` be an index column.


app/Repositories/PostRepository.php
```diff
@@ -73,7 +73,7 @@ public function create(array $data, $signupId)
             $fileUrl = 'default';
         }
 
-        $signup = Signup::find($signupId);
+        $signup = Signup::withCount('posts')->find($signupId);
```
chloealee: @sbsmith86 just curious why we need to query for this `withCount`?

sbsmith86: ahh yep. This allows me to then check if this is not the first post under a signup because it adds a `post_count` property to the signup (see line 101). Depending on whether or not this is the first post or not, we handle edge cases differently. For example, if it is not the first post then when the user submits a quantity that is equal to what we already have, we store it as `0`

app/Repositories/PostRepository.php
```diff
@@ -83,10 +83,35 @@ public function create(array $data, $signupId)
             'url' => $fileUrl,
             'caption' => $data['caption'],
             'status' => isset($data['status']) ? $data['status'] : 'pending',
-            'source' => $data['source'],
-            'remote_addr' => $data['remote_addr'],
+            'source' => isset($data['source']) ? $data['source'] : null,
+            'remote_addr' => isset($data['remote_addr']) ? $data['remote_addr'] : null,
```
DFurnes: Not relevant for this PR, but this is another thing we should set internally on v3 endpoints! (With `$request->ip()` reading from `X-Forwarded-For` headers provided by whatever service is making the Rogue call).

app/Repositories/PostRepository.php
```diff
@@ -83,10 +83,35 @@ public function create(array $data, $signupId)
...
+            if ($quantityDiff < 0) {
+                $quantityDiff = $data['quantity'];
+            } elseif ($quantityDiff === 0 && $signup->posts_count > 0) {
+                // If the quantity difference equals zero, and this is not the first post,
+                // then we can assume there is no difference in quantity and store it as 0 on the post.
+                $quantityDiff = 0;
+            }
+
+            $post->quantity = $quantityDiff;
+            $post->save();
```
DFurnes: I don't think we need this since we call `$post->save()` below (so no need to hit the DB twice).

app/Repositories/PostRepository.php
sbsmith86: good call!

app/Repositories/PostRepository.php
chloealee: ahhh got it - thanks for clarifying @sbsmith86 !

PR #525, opened 2017-12-06
Syslog default

sbsmith86:
#### What's this PR do?

Updates our logging configuration to write to the `syslog` by default, unless otherwise specified by the `APP_LOG` environment variable.

👀 

Also- any suggestions about how to confirm this is working on other environments?

I am setting `APP_LOG=daily` on my local to continue getting daily log files for my development work, but using `syslog` on our other environments should allow us to see logs in heroku, logstash, etc.

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/153261132

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,7 @@
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
-APP_LOG=
+APP_LOG=daily
 APP_LOG_LEVEL=debug
 APP_URL=http://rogue.dev
 ROGUE_API_KEY=abc123
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,7 @@
 APP_ENV=local
 APP_KEY=
 APP_DEBUG=true
+APP_LOG=
 APP_LOG_LEVEL=debug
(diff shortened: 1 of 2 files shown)


katiecrane:
@sbsmith86 how does this change how we find the logs in production? I know we talked about making this change, I just don't think I completely understand all the implications

sbsmith86:
@katiecrane so I think the idea is that right now we send logs to a log file in `storage/logs` so it's hard to find in heroku/logentries/etc but this changes that setting to write to the `syslog` instead which is what heroku reads from so we will be able to see the logs we usually see.

DFurnes:
IMO, it makes sense to leave the default in `.env.example` as daily!

sbsmith86:
hooookay updated and builds seem to be passing now. For documentation sake: here are the current settings we are using on each environment: 

Rogue-QA (AWS): `daily`
Rogue-Thor (HEROKU) : `errorlog`
Rogue-Prod (AWS): `daily`

PR #526, opened 2017-12-06
v3/posts update and Start of OAuth updates to v3 endpoints

chloealee:
#### What's this PR do?
- Updates `POST v3/posts` endpoint to store quantity on the post.
- Updates `PATCH v3/posts` endpoint to not try to update quantity and allows you to update quantity.
- Updates Post transformer to return quantity for the post.
- [Updates SignupController@store to incldue authorization logic.](https://www.pivotaltracker.com/n/projects/2019429/stories/153365833)
- [Updates PostController@store to allow user tokens.](https://www.pivotaltracker.com/n/projects/2019429/stories/153365724)
- [Updates PostController@update to include authorization logic.](https://www.pivotaltracker.com/n/projects/2019429/stories/153365782)
  - Only allows admins to be able to do this. 
- Updates Signup and Post tests. 
- Adds test to make sure only admins can update a post. 
- Updates documentation.

- Check to make sure OAuth logic makes sense for storing a new signup (that is gets NS id from auth instead of `northstar_id`). 
- Go through flow of creating a new signup and new post at the same time and make sure that quantity is stored correctly.
- Go through flow of creating a new post with an existing signup and make sure that quantity is stored correctly. 
- When creating a post, make sure we're grabbing NS id from auth instead of `northstar_id`. 
- When updating a post, make sure that only an admin can do this. 
- Makes sure all tests pass (and test logic is accurate!). 

#### Relevant tickets
Pivotal Tickets: 
- [Update v3 posts endpoints](https://www.pivotaltracker.com/n/projects/2019429/stories/153073386)
- OAuth tickets:
  - https://www.pivotaltracker.com/n/projects/2019429/stories/153365724
  - https://www.pivotaltracker.com/n/projects/2019429/stories/153365782
  - https://www.pivotaltracker.com/n/projects/2019429/stories/153365833

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -148,7 +148,7 @@ class PostsController extends ApiController
      */
     public function update(PostRequest $request, Post $post)
     {
-        $post->update($request->only('status', 'caption'));
+        $post->update($request->only('status', 'caption', 'quantity'));
 
         return $this->item($post);
     }
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -5,7 +5,7 @@ namespace Rogue\Http\Controllers\Three;
 use Rogue\Models\Post;
 use Illuminate\Http\Request;
 use Rogue\Services\PostService;
-use Rogue\Repositories\SignupRepository;
+use Rogue\Repositories\Three\SignupRepository;
 use Rogue\Http\Requests\Three\PostRequest;
 use Rogue\Http\Transformers\PostTransformer;
 use Rogue\Http\Controllers\Api\ApiController;
@@ -25,7 +25,7 @@ class PostsController extends ApiController
     /**
      * The signup repository instance.
      *
-     * @var \Rogue\Repositories\SignupRepository
+     * @var \Rogue\Repositories\Three\SignupRepository
      */
     protected $signups;
 
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -149,7 +149,8 @@ class PostsController extends ApiController
      */
     public function update(PostRequest $request, Post $post)
     {
-        if (token()->role() === 'admin') {
+        // Only allow an admin or the user who owns the post to update.
+        if (token()->role() === 'admin' || auth()->id() === $post->northstar_id) {
             $post->update($request->only('status', 'caption', 'quantity'));
 
             return $this->item($post);
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -108,7 +108,7 @@ class PostsController extends ApiController
 
         // @TODO - should we eventually throw an error if a signup doesn't exist before a post is created? I create one here because we haven't implemented sending signups to rogue yet, so it will have to create a signup record for all posts.
         if (! $updating) {
-            $signup = $this->signups->create($request->all());
+            $signup = $this->signups->create($request->all(), auth()->id());
             $post = $this->posts->create($request->all(), $signup->id, $transactionId);
 
             $code = 200;
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -100,7 +100,7 @@ class PostsController extends ApiController
      */
     public function store(PostRequest $request)
     {
-        $northstarId = $request['northstar_id'] ? $request['northstar_id'] : auth()->id();
+        $northstarId = getNorthstarId($request);
 
         $transactionId = incrementTransactionId($request);
 
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -18,7 +18,7 @@ class PostsController extends ApiController
     /**
      * The post service instance.
      *
-     * @var \Rogue\Services\PostService
+     * @var \Rogue\Services\Three\PostService
      */
     protected $posts;
 
@@ -101,14 +101,13 @@ class PostsController extends ApiController
     {
         $transactionId = incrementTransactionId($request);
 
-        $signup = $this->signups->get($request['northstar_id'], $request['campaign_id'], $request['campaign_run_id']);
+        $signup = $this->signups->get(auth()->id(), $request['campaign_id'], $request['campaign_run_id']);
 
         $updating = ! is_null($signup);
 
         // @TODO - should we eventually throw an error if a signup doesn't exist before a post is created? I create one here because we haven't implemented sending signups to rogue yet, so it will have to create a signup record for all posts.
         if (! $updating) {
             $signup = $this->signups->create($request->all());
-
             $post = $this->posts->create($request->all(), $signup->id, $transactionId);
 
             $code = 200;
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -7,7 +7,7 @@ use Illuminate\Http\Request;
 use Rogue\Services\Three\PostService;
 use Rogue\Repositories\Three\SignupRepository;
 use Rogue\Http\Requests\Three\PostRequest;
-use Rogue\Http\Transformers\PostTransformer;
+use Rogue\Http\Transformers\Three\PostTransformer;
 use Rogue\Http\Controllers\Api\ApiController;
 use Rogue\Http\Controllers\Traits\FiltersRequests;
 
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -4,7 +4,7 @@ namespace Rogue\Http\Controllers\Three;
 
 use Rogue\Models\Post;
 use Illuminate\Http\Request;
-use Rogue\Services\PostService;
+use Rogue\Services\Three\PostService;
 use Rogue\Repositories\Three\SignupRepository;
 use Rogue\Http\Requests\Three\PostRequest;
 use Rogue\Http\Transformers\PostTransformer;
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -10,6 +10,7 @@ use Rogue\Http\Requests\Three\PostRequest;
 use Rogue\Http\Transformers\Three\PostTransformer;
 use Rogue\Http\Controllers\Api\ApiController;
 use Rogue\Http\Controllers\Traits\FiltersRequests;
+use Illuminate\Auth\Access\AuthorizationException;
 
 class PostsController extends ApiController
 {
@@ -148,9 +149,13 @@ class PostsController extends ApiController
      */
     public function update(PostRequest $request, Post $post)
     {
-        $post->update($request->only('status', 'caption', 'quantity'));
+        if (token()->role() === 'admin') {
+            $post->update($request->only('status', 'caption', 'quantity'));
 
-        return $this->item($post);
+            return $this->item($post);
+        }
+
+        throw new AuthorizationException('You don\'t have the correct role to do that!');
     }
 
     /**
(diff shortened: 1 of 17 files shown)


app/Http/Requests/Three/PostRequest.php
```diff
@@ -26,6 +26,7 @@ public function rules()
         return [
             'status' => 'in:pending,accepted,rejected',
             'caption' => 'string|max:140',
```
chloealee: should we make caption required here?

app/Repositories/Three/PostRepository.php
```diff
@@ -0,0 +1,254 @@
...
+            $arrayKeysToUpdate = array_keys($nonNullArrayKeys);
+
+            $signup->fill(array_only($data, $arrayKeysToUpdate));
+            $signup->save(['timestamps' => false]);
+
+            $event = $signup->events->last();
+            $event->created_at = $data['updated_at'];
+            $event->updated_at = $data['updated_at'];
+            $event->save(['timestamps' => false]);
+        } else {
```
chloealee: @katiecrane I'm pretty sure we can, but just want to double check - can I delete all of the above (lines 111-130)?

katiecrane: @chloealee for V3 YES 🔥

app/Http/Requests/Three/PostRequest.php
DFurnes: I think that makes sense (at least for how things work now). 👍

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -48,7 +48,6 @@ public function __construct(SignupService $signups)
     public function store(Request $request)
     {
         $this->validate($request, [
-            'northstar_id' => 'required|string',
             'campaign_id' => 'required',
             'campaign_run_id' => 'int',
             'why_participated' => 'string',
```
chloealee: can we delete `source` (below) since we're now storing this using `token()->client()` in the Signup Repository?

app/Repositories/Three/PostRepository.php
```diff
@@ -0,0 +1,254 @@
...
+        $post = new Post([
+            'signup_id' => $signup->id,
+            'northstar_id' => auth()->id(),
+            'campaign_id' => $signup->campaign_id,
+            'quantity' => $data['quantity'],
+            'url' => $fileUrl,
+            'caption' => $data['caption'],
+            'status' => isset($data['status']) ? $data['status'] : 'pending',
+            'source' => token()->client(),
+            'remote_addr' => isset($data['remote_addr']) ? $data['remote_addr'] : null,
```
chloealee: @DFurnes I saw in your comment in Shae's PR that we should use `$data->ip()` instead here. But when I do that and run tests, the test suite stops in the middle and I get this error `Segmentation fault (core dumped)`. Not sure why? 🤔

chloealee:
Before this is merged, do I need to update/notify Blink that a quantity will now be sent along in the payload for posts?

app/Http/Requests/Three/PostRequest.php
chloealee: oh and also make quantity required?

app/Http/Controllers/Three/SignupsController.php
DFurnes: Yep!

app/Http/Controllers/Three/PostsController.php
```diff
@@ -149,9 +149,13 @@ public function show(Post $post)
      */
     public function update(PostRequest $request, Post $post)
     {
-        $post->update($request->only('status', 'caption'));
+        if (token()->role() === 'admin') {
```
DFurnes: Do we want to allow a user to update their own post? (If `auth()->id() == $post->northstar_id`?)

app/Http/Requests/Three/PostRequest.php
sbsmith86: quantity should not be required since PN won't be asking for it often.

app/Http/Requests/PostRequest.php
```diff
@@ -24,8 +24,8 @@ public function rules()
         return [
             'northstar_id' => 'required|string',
             'campaign_id' => 'required',
+            'quantity' => 'int|required',
```
sbsmith86: This shouldn't be required. It's a nullable field and campaigns going forward might not require it so we need to be flexible.

app/Models/Post.php
```diff
@@ -157,7 +157,7 @@ public function toBlinkPayload()
         return [
             'id' => $this->id,
             'signup_id' => $this->signup_id,
-            'quantity' => $this->signup->quantity,
+            'quantity' => is_null($this->quantity) ? $this->signup->quantity : $this->quantity,
```
sbsmith86: I think we want to return `null` if we have `null` stored on the post. So we should really just return what is on the post

app/Repositories/Three/PostRepository.php
sbsmith86: I think he said it would be `$request->ip() ` not sure if that helps!

DFurnes: Oh yup, sorry I missed this the first time around. The IP is stored on the request object, so you can't access it from the `$data` array here. You can always grab it from the helper with `request()->ip()` since I don't think we pass in a request here.

app/Repositories/Three/SignupRepository.php
```diff
@@ -17,19 +17,23 @@ public function create($data)
         // Create the signup
         $signup = new Signup;
 
-        $signup->northstar_id = $data['northstar_id'];
+        // @TODO change auth()->id(); based on Dave's note:
```
sbsmith86: Can we just do this in this PR?

app/Http/Requests/Three/PostRequest.php
chloealee: hmm @DFurnes when I made the caption required and tested to not sent, it seems like the error handling was off again and broke with a `500` response but didn't have anything in the body. Should we hold off on that until we fix the error handling in Rogue (also made a new card for that [here](https://www.pivotaltracker.com/n/projects/2019429/stories/153458427)!).

app/Repositories/Three/PostRepository.php
chloealee: ah great to know, thanks!

app/Http/Controllers/Three/PostsController.php
chloealee: oh I didn't think about that! I don't think we have UI for that now but definitely could see that in the future - I'll add this!

app/Http/Requests/PostRequest.php
chloealee: ah yes!

chloealee: actually this is for v2 endpoints. But it should still be added and be `nullable|int`, right?

sbsmith86: yea we can leave it as not required for `v2`, since we are supporting both the new way of collecting quantity (if we have it) and the old way at the same time.

app/Repositories/Three/SignupRepository.php
chloealee: ah I forgot about this note to myself! I wanted to do this in this PR but wanted to do after I got everything working. Thanks for flagging, this @sbsmith86 !

app/Repositories/Three/SignupRepository.php
```diff
@@ -10,26 +10,27 @@ class SignupRepository
...
      * @return \Rogue\Models\Signup|null
      */
-    public function create($data)
+    public function create($data, $northstarIdFromAuth = null)
     {
         // Create the signup
         $signup = new Signup;
 
-        $signup->northstar_id = $data['northstar_id'];
+        $signup->northstar_id = isset($northstarIdFromAuth) ? $northstarIdFromAuth : $data['northstar_id'];
```
chloealee: Updated this based on @DFurnes 's suggestion: 

>  It might make sense to send these along as parameters from the controller (so that we could use the repository for things that aren't HTTP requests, like for creating signups from a CSV or something).

The only thing I'm worried about here is if somehow, the `auth()->id()` isn't sent and `$data['northstar_id']` also isn't set, this will break? Because we don't have a validation that requires to send if the `northstar_id` in the request anymore and all signups need to be associated to a user.

DFurnes: Yeah I think we'd always want to provide the ID as a parameter (so even if we were creating from a CSV or an admin flow or something, we'd still call `$repository->create($data, $id)` with the proper ID).

app/Services/Three/SignupService.php
```diff
@@ -42,7 +42,9 @@ public function __construct(SignupRepository $signup, Blink $blink)
      */
     public function create($data, $transactionId)
     {
-        $signup = $this->signup->create($data);
+        $northstarId = isset($data['northstar_id']) ? $data['northstar_id'] : auth()->id();
```
chloealee: @DFurnes changed this here based on your last comment to decide which to send over!

app/Http/Controllers/Three/PostsController.php
```diff
@@ -101,14 +102,13 @@ public function store(PostRequest $request)
     {
         $transactionId = incrementTransactionId($request);
 
-        $signup = $this->signups->get($request['northstar_id'], $request['campaign_id'], $request['campaign_run_id']);
+        $signup = $this->signups->get(auth()->id(), $request['campaign_id'], $request['campaign_run_id']);
```
chloealee: @DFurnes should this also be updated to check if a `northstar_id` is sent in the request, and if so, send that instead of getting it from `auth()->id()`?

app/Http/Controllers/Three/PostsController.php
```diff
@@ -101,14 +102,13 @@ public function store(PostRequest $request)
...
 
         $updating = ! is_null($signup);
 
         // @TODO - should we eventually throw an error if a signup doesn't exist before a post is created? I create one here because we haven't implemented sending signups to rogue yet, so it will have to create a signup record for all posts.
         if (! $updating) {
-            $signup = $this->signups->create($request->all());
-
+            $signup = $this->signups->create($request->all(), auth()->id());
```
chloealee: @DFurnes and here too (checking if `northstar_id` is sent with the request)?

app/Http/Controllers/Three/PostsController.php
```diff
@@ -99,15 +100,17 @@ public function index(Request $request)
      */
     public function store(PostRequest $request)
     {
+        $northstarId = $request['northstar_id'] ? $request['northstar_id'] : auth()->id();
```
chloealee: @DFurnes updated here too! I _think_ I got everywhere that needs `northstar_id`

app/Services/Three/SignupService.php
DFurnes: One catch - we want to make sure that only admins can set a custom user ID (since we wouldn't want a normal user to be able to hack someone else's reportbacks). I think it makes sense to handle this all in the controller, and have a `token()->role()` check there to see if we should use the provided `northstar_id` parameter.

app/helpers.php
```diff
@@ -95,3 +95,19 @@ function getAgeFromBirthdate($birthdate)
...
+ * @param $request
+ */
+function getNorthstarId($request)
+{
+    if (token()->role() === 'admin') {
+        return isset($request['northstar_id']) ? $request['northstar_id'] : auth()->id();
+    }
+
+    return auth()->id();
+}
```
chloealee: @DFurnes thanks for walking me through this and great catch re: only allowing admins to send through custom IDs! Added this and updated the rest of the code!

app/Models/Post.php
```diff
@@ -157,7 +157,7 @@ public function toBlinkPayload()
         return [
             'id' => $this->id,
             'signup_id' => $this->signup_id,
-            'quantity' => $this->signup->quantity,
+            'quantity' => $this->quantity,
```
chloealee: will this break anything in v2 @sbsmith86 ?

sbsmith86: nope nope! it will just return whatever is in that column (`null` for most things until the script runs)

chloealee: ah right! Thanks for confirming!

PR #528, opened 2017-12-07
Send null instead of 'default' to Blink if post has default image 

chloealee:
#### What's this PR do?
If a post has a default image, send `null` instead of `default` to Blink. 

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/152517900) Pivotal Ticket.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -102,8 +102,7 @@ class Post extends Model
     public function getMediaUrl()
     {
         if ($this->url === 'default') {
-            // @TODO: We should either return a placeholder, or `null`.
-            return 'default';
+            return null;
         }
 
         // If Glide is enabled, provide the default image URL.
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -101,7 +101,7 @@ class Post extends Model
      */
     public function getMediaUrl()
     {
-        if ($this->url === 'default') {
+        if ($this->url === null) {
             return null;
         }
 
--- a/resources/assets/helpers.js
+++ b/resources/assets/helpers.js
@@ -267,7 +267,7 @@ export function getImageUrlFromProp(photoProp) {
   }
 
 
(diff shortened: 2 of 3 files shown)


chloealee:
cc @rapala61 @sergii-tkachenko

sbsmith86:
Dope, I think we might also want to update our logic to [store this as `null` ](https://github.com/DoSomething/rogue/blob/master/app/Repositories/PostRepository.php#L73)  as well!

chloealee:
@sbsmith86 create catch! 💡 

I updated this function in the `Three/PostRepository` file as well in [this PR](https://github.com/DoSomething/rogue/pull/526/files).

app/Models/Post.php
```diff
@@ -102,8 +102,7 @@ public function tagSlugs()
     public function getMediaUrl()
     {
         if ($this->url === 'default') {
```
sbsmith86: Sorry wait! you will have to check for `null` here now as well!

chloealee:
@sbsmith86 thanks for catching the other one! I did a search through the repo to make sure I didn't miss anymore and found one more in the front end! I think because of this, we'll need to change all the `default` values to `null` in the database too?

sbsmith86:
@chloealee which component?

sbsmith86:
I think for now we can just check for `null` and `default` for this column until we clean it up

resources/assets/helpers.js
```diff
@@ -267,7 +267,7 @@ export function getImageUrlFromProp(photoProp) {
   }
 
 
-  if (photo_url == 'default') {
+  if (photo_url == 'default' || photo_url == null) {
```
chloealee: ah smart @sbsmith86 ! Updated that here in the helpers :)

PR #529, opened 2017-12-08
Removes middleware so OAuth works correctly with a non-admin user

chloealee:
#### What's this PR do?
- Removes middleware so OAuth works correctly with a non-admin user (and any user can use v3 endpoints).
- Removes tests that only allow admins to make these requests (except for `DELETE /v3/posts`).
👀 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -159,7 +159,7 @@ class PostsController extends ApiController
             return $this->item($post);
         }
 
-        throw new AuthorizationException('You don\'t have the correct role to do update this post!');
+        throw new AuthorizationException('You don\'t have the correct role to update this post!');
     }
 
     /**
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -49,7 +49,6 @@ class PostsController extends ApiController
         $this->transformer = $transformer;
 
         $this->middleware('auth:api', ['only' => ['store', 'update', 'destroy']]);
-        $this->middleware('role:admin', ['only' => ['store', 'update', 'destroy']]); // @TODO: Allow anyone to use this.
     }
 
     /**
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/Three/PostsController.php
```diff
@@ -49,7 +49,6 @@ public function __construct(PostService $posts, SignupRepository $signups, PostT
         $this->transformer = $transformer;
 
         $this->middleware('auth:api', ['only' => ['store', 'update', 'destroy']]);
-        $this->middleware('role:admin', ['only' => ['store', 'update', 'destroy']]); // @TODO: Allow anyone to use this.
```
chloealee: Right now, I don't think we allow users to destroy their own posts. Should we keep this with only destroy in it? 

`$this->middleware('role:admin', ['only' => [destroy']]);`

DFurnes: IMO, let's keep it for now?

chloealee: I agree! Updated :)

tests/Http/Three/PostTest.php
```diff
@@ -372,44 +372,6 @@ public function testValidationUpdatingAPost()
         $this->assertEquals('The caption may not be greater than 140 characters.', $json['errors']['caption'][0]);
     }
 
-    /**
-     * Test that non-authenticated user's/apps can't update posts.
-     *
-     * @return void
-     */
-    public function testUnauthenticatedUserUpdatingAPost()
```
DFurnes: Can we keep this test?

chloealee: ah yes, thanks for pointing this out! I misread. I had to update with the new error message we're getting now below. But I'm not exactly sure why this is happening now with a different error message?

DFurnes: That's because this is using the new [custom authorization logic](https://github.com/DoSomething/rogue/blob/a3452e43130634c1076a637911c209361257d301/app/Http/Controllers/Three/PostsController.php#L162) in the controller, rather than [Gateway's `role` middleware](https://github.com/DoSomething/gateway/blob/d1479758bb276e8ddb950d49cc3827322684a819/src/Server/Middleware/RequireRole.php#L31). It might make sense to update the messaging to make it more clear that it's an issue with the user's permissions on _that_ particular post, but otherwise 👍 from me!

chloealee: oh duh! Thanks so much for explaining, @DFurnes ! Updated the message to be more explicit :)

PR #531, opened 2017-12-08
More performance improvements.

DFurnes:
#### What's this PR do?
A few more performance improvements:

🏷 Eager loads the `tags` on posts, so that we don't have the ol' N+1 query problem. ([Before](https://user-images.githubusercontent.com/583202/33785726-58fd4bf0-dc34-11e7-8c8a-07e71d2334e5.png), [After](https://user-images.githubusercontent.com/583202/33788227-0d6b8a98-dc3f-11e7-8330-c906cb581967.png))

💵 Caches routes & config when compiling the app. I also switched these commands to run in a "warmup" task so they also run on environment variable changes (as suggested [here](https://stackoverflow.com/a/35862183/811624)).

⬆️ Updated dependencies, including the latest version of Laravel Debugbar. Since it now registers with package discovery, I've removed the line we'd use to register it in AppServiceProvider.

📄 Forces cursor pagination on the v3 routes – this gets rid of an otherwise gnarly `SELECT COUNT(*)` query that I don't believe we can optimize. IMO, it makes the most sense to create separate "stats" routes that cache these queries if we need stuff like this later on.

Things should be faster! It may make sense to review commit-by-commit.

References #530.

#### Relevant tickets
N/A

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -2,7 +2,6 @@
 
 namespace Rogue\Providers;
 
-use DoSomething\Gateway\Blink;
 use Illuminate\Support\ServiceProvider;
 
 class AppServiceProvider extends ServiceProvider
@@ -24,14 +23,6 @@ class AppServiceProvider extends ServiceProvider
      */
     public function register()
     {
-        // @TODO: This should be registered in Gateway's service provider!
-        $this->app->singleton(Blink::class, function () {
-            return new Blink(config('services.blink'));
-        });
-
-        // Ensure that debugbar doesn't register on production.
-        if ($this->app->environment() !== 'production') {
-            $this->app->register(\Barryvdh\Debugbar\ServiceProvider::class);
-        }
+        //
     }
 }
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
-web: vendor/bin/heroku-php-apache2 /public/
+web: composer warmup && vendor/bin/heroku-php-apache2 /public/
 
 release: php artisan migrate --force
 
--- a/app/Http/Controllers/PagesController.php
+++ b/app/Http/Controllers/PagesController.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Rogue\Http\Controllers;
+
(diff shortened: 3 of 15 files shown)


DFurnes:
Alright, this is ready for review! 👀

chloealee:
this is so cool, @DFurnes ! How did you see the queries in the Before and After screenshots?

DFurnes:
> How did you see the queries in the Before and After screenshots?

I'm using the [Clockwork extension](https://chrome.google.com/webstore/detail/clockwork/dmggabnehkmmfmdffgajcflpdjlnoemp)!

app/Providers/AppServiceProvider.php
```diff
@@ -24,14 +23,6 @@ public function boot()
      */
     public function register()
     {
-        // @TODO: This should be registered in Gateway's service provider!
-        $this->app->singleton(Blink::class, function () {
-            return new Blink(config('services.blink'));
-        });
-
-        // Ensure that debugbar doesn't register on production.
```
katiecrane: Will this register on production now?

app/Providers/AppServiceProvider.php
```diff
@@ -24,14 +23,6 @@ public function boot()
      */
     public function register()
     {
-        // @TODO: This should be registered in Gateway's service provider!
```
katiecrane: Is this registered in Gateway's service provider now?

app/Providers/AppServiceProvider.php
DFurnes: It's registered using Laravel's [package discovery](https://laravel.com/docs/5.5/packages#package-discovery) and marked as a `devDependency`, so it shouldn't. Let's double-check after next deploy just in case!

app/Providers/AppServiceProvider.php
DFurnes: [Yep](https://github.com/DoSomething/gateway/blob/d1479758bb276e8ddb950d49cc3827322684a819/src/Laravel/GatewayServiceProvider.php#L46-L48)!

PR #534, opened 2017-12-11
Adding queueing functionality for Heroku

blisteringherb:
#### What's this PR do?
We need queueing support for Rogue. In Heroku, you have to add a `queue` line to the Procfile to support this.

Submit a reportback to see if you receive an email.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
 web: vendor/bin/heroku-php-apache2 /public/
 
 release: php artisan migrate --force
-
+queue: php artisan queue:work --daemon --sleep=5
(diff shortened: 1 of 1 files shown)


Procfile
```diff
@@ -1,4 +1,4 @@
 web: vendor/bin/heroku-php-apache2 /public/
 
 release: php artisan migrate --force
-
+queue: php artisan queue:work --daemon --sleep=5
```
DFurnes: The `--daemon` option is no longer necessary.

PR #537, opened 2017-12-12
Add script that will clean up extraneous posts we received

sbsmith86:
#### What's this PR do?

This PR adds a command `php artisan rogue:postcleanup {date}` that will go through all the posts we received on the given date, find the duplicate submissions (by checking for duplicate `captions`) and then delete all but the most recent submission from the DB. 

👀 

Definitely double check the deletion logic: 

I used `$post->forceDelete()` to truly delete the record from the DB because we don't want a record of these bad posts. 

I used `Storage::delete({path_to_file})` to delete the image from storage. 

This script is meant to clean up the duplicate posts we receive because of this bug: 
https://github.com/orgs/DoSomething/teams/team-bleed/discussions/5

#### Relevant tickets
Fixes 🐛 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Console/Commands/PostCleanup.php
+++ b/app/Console/Commands/PostCleanup.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Rogue\Console\Commands;
+
+use DB;
+use Rogue\Models\Post;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Storage;
+
+class PostCleanup extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'rogue:postcleanup {date=2017-12-11}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Clean up duplicate posts caused by blink erroring out and resending post requests';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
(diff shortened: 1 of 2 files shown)


app/Console/Commands/PostCleanup.php
```diff
@@ -0,0 +1,84 @@
...
+            ->having('caption_count', '>', 1)
+            ->get();
+
+        $this->info('There are ' . $postsByCaption->count() . ' posts with more than 1 of the same caption');
+
+        // Loop through all the posts that have the
+        // same caption per signup and delete all but the most recent one.
+        foreach ($postsByCaption as $postCaption) {
+            $this->info('Getting posts under signup ' . $postCaption->signup_id);
+            $userPosts = Post::where('signup_id', $postCaption->signup_id)
```
katiecrane: I'm trying to work through this and this part seems to me like it is deleting all posts under that signup on that day except for the most recent one. What if a user had submitted multiple posts with different captions?

sbsmith86: And then I go through each of those `signup_id` and get all those posts and delete all but the most recent one.

sbsmith86: Yea, I guess that is a possibility but I wasn't really seeing that. I can add that extra filter!

katiecrane: Woo! Then we can 🔥 🔥 🔥 !!

sbsmith86:
@katiecrane yea i figure dump some fake posts in there and then run this to make sure they get cleaned up.

PR #538, opened 2017-12-13
Slack Integration 

chloealee:
#### What's this PR do?
Work done with @sbsmith86 , @katiecrane , @DFurnes , @ngjo , and @lkpttn during HEXUPLE coding! 

When a post is tagged as "Good For Storytelling", Rogue sends a Slack nofitication with the Post's info.! 

:eyes:

Before we merge, we'll need to find out what channel people want this to be in (#badass-members?) and then re-do the webhook portion of this work. 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/152084069) Pivotal ticket

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/TagsController.php
```diff
@@ -5,6 +5,7 @@
 use Rogue\Repositories\PostRepository;
 use Rogue\Http\Transformers\PostTransformer;
 use Rogue\Http\Controllers\Traits\TagsRequests;
+use Rogue\Events\PostTagged;
```
chloealee: do we need this here?

ngjo:
@chloealee can we add a line that we'll be continuing to add other post information to this slack notification (like member info!)

chloealee:
@ngjo sure thing! Do you mean in the message that gets posted to Slack? So maybe in the link "Click here for the full post on Rogue" we can change to "Click here for the full post on Rogue (we'll be continuing to add other post info. to this Slack notification (like member info!)" That might be too long though? 
screen shot 2017-12-13 at 11 20 27 am

app/Notifications/SlackTagNotification.php
```diff
@@ -0,0 +1,71 @@
...
+        return (new SlackMessage)
+                    ->from('Tongue Cat', ':tonguecat:')
+                    ->content('Another badass member submitted a post!')
+                    ->attachment(function ($attachment) {
+                        $attachment->title('Click here for the full post on Rogue', route('signups.show', [$this->post->signup_id], true))
+                                ->fields([
+                                    'Caption' => str_limit($this->post->caption, 140),
+                                    'Why Participated' => str_limit($this->post->signup->why_participated),
+                                ])
+                                ->color('#4e2b63')
```
lkpttn: We should change this to 
`->color(array_random(['#FCD116', '#23b7fb', '#4e2b63']))`
so we get a different brand color each time!

chloealee: love it! updated! :)

app/Notifications/SlackTagNotification.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toSlack($notifiable)
+    {
+        return (new SlackMessage)
```
katiecrane: Should we only do this if the Slack URL env var is set?

chloealee: great catch - updated!

app/Notifications/SlackTagNotification.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toSlack($notifiable)
+    {
+        if (isset(SLACK_WEBHOOK_INTEGRATION_URL)) {
```
katiecrane: I thiiiiink you can use `if (config('services.slack.url'))`?

ngjo:
@chloealee maybe: Click here for more details in Rogue! We'll be adding member info here soon.

??

app/Notifications/SlackTagNotification.php
```diff
@@ -0,0 +1,73 @@
...
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toSlack($notifiable)
+    {
+        if (config('services.slack.url')) {
+            return (new SlackMessage)
+                ->from('Tongue Cat', ':tonguecat:')
```
DFurnes: This is a lame suggestion, but would it make sense to have the post by "Rogue" instead of "Tongue Cat"? (Thinking that could help more people on staff associate what Rogue is used for.)

chloealee:
@ngjo you're a wordsmith! Changed it and here is what a looks like now (@DFurnes also updated to Rogue instead of from Tongue Cat!):
screen shot 2017-12-13 at 1 13 11 pm

chloealee:
yay! Ok so I think everything is all set to go besides re-running the webhook to route to the correct room in Slack. @ngjo should we have this go to #badass-members?

katiecrane:
This is kind of nitty gritty but to me "we'll be adding member info here soon" sounds like it is being added in Rogue, so I wonder if it would be clearer if that said "We'll be adding member info in Slack soon" 🤔

.env.example
```diff
@@ -54,3 +54,5 @@ PHOENIX_API_VERSION=v1
 BLINK_URL=
 BLINK_USERNAME=
 BLINK_PASSWORD=
+
+SLACK_WEBHOOK_INTEGRATION_URL=
```
chloealee: I actually could just update the room in the Slack user interface but use the same URL! Updated this here since this is a new thing we'll need. Should I add this to Heroku now? 

After that's added, this seems like it's good to go! Unless I should test? But that might ruin the surprise/could be the surprise reveal?!

app/Events/PostTagged.php
```diff
@@ -0,0 +1,40 @@
...
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Broadcasting\PrivateChannel;
+use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Broadcasting\InteractsWithSockets;
+
+class PostTagged
+{
+    use Dispatchable, InteractsWithSockets, SerializesModels;
+
+    public $post;
```
DFurnes: Can we add docblocks for these two properties?

app/Events/PostTagged.php
```diff
@@ -0,0 +1,40 @@
...
+use Rogue\Models\Post;
+use Illuminate\Broadcasting\Channel;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Broadcasting\PrivateChannel;
+use Illuminate\Foundation\Events\Dispatchable;
+use Illuminate\Broadcasting\InteractsWithSockets;
+
+class PostTagged
+{
+    use Dispatchable, InteractsWithSockets, SerializesModels;
```
DFurnes: Since this event is internal to the application (it's just triggering a listener class), I think we can kill `InteractsWithSockets` and the `broadcastOn` methods here. 📻

app/Notifications/SlackTagNotification.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Get the mail representation of the notification.
+     *
+     * @param  mixed  $notifiable
+     * @return \Illuminate\Notifications\Messages\MailMessage
+     */
+    public function toSlack($notifiable)
+    {
+        if (config('services.slack.url')) {
```
DFurnes: Nitpicky, but it could be cleaner to reformat this as a [guard clause](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html):

```php
if (! config('services.slack.url')) {
    return;
}

return (new SlackMessage)
    ->from('Rogue', ':tonguecat:')
    // ...
```

.env.example
DFurnes: It couldn't hurt to test one more time in `#team-bleed`, but then let's do it!!!

katiecrane: @chloealee I think we only want the url to be set for production so tags from Thor and Staging don't hit `#badass_members` ? Unless we can use a different url that sends to #team-bleed for us to test stuff?

chloealee: @katiecrane ah so smart!! I'll remove those all those except for on Prod after I test on Thor!

PR #541, opened 2017-12-13
Script to make default urls null

katiecrane:
#### What's this PR do?
This is a shiny new artisan command to go through all of our posts with a `url` of `default` and set that to `null` instead. We probably do not want to run this until we are storing incoming borked images as `null` or we will have to run it twice (running it twice is no problem technically though).

Will this do what I said it would?

#### Relevant tickets
[card](https://www.pivotaltracker.com/story/show/153491711)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Console/Commands/MakeDefaultURLsNull.php
+++ b/app/Console/Commands/MakeDefaultURLsNull.php
@@ -40,11 +40,14 @@ class MakeDefaultURLsNull extends Command
     {
         info('rogue:nullurl: Starting!');
 
-        // Create the progress bar
-        $bar = $this->output->createProgressBar(Post::where('url', 'default')->count());
+        // All posts that have "default" set as their url
+        $query = Post::where('url', 'default');
 
-        // Get all posts with "default" for the url
-        $posts = Post::where('url', 'default')->chunkById(100, function ($posts) use ($bar) {
+        // Create the progress bar
+        $bar = $this->output->createProgressBar($query->count());
+
+        // Update all the posts
+        $query->chunkById(100, function ($posts) use ($bar) {
             // Set all of the "default"s to null
             foreach ($posts as $post) {
                 info('rogue:nullurl: Nulling url for post ' . $post->id);
@@ -54,6 +57,7 @@ class MakeDefaultURLsNull extends Command
             }
         });
 
+        // Finish the progress bar
         $bar->finish();
         info('rogue:nullurl: All done!');
     }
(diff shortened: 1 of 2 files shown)


app/Console/Commands/MakeDefaultURLsNull.php
```diff
@@ -0,0 +1,60 @@
...
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        info('rogue:nullurl: Starting!');
+
+        // Get all posts with "default" for the url
+        $posts = Post::where('url', 'default')->get();
```
DFurnes: This might be something to handle [chunked query or cursor](https://laravel.com/docs/5.5/queries#chunking-results) (depending on how many `default` posts there are).

katiecrane: I thought about that too, there are ~60k. Worth chunking?

DFurnes: Mmm yeah, I think so. That's a lot of stuff to cram into memory otherwise.

app/Console/Commands/MakeDefaultURLsNull.php
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public function handle()
+    {
+        info('rogue:nullurl: Starting!');
+
+        // Create the progress bar
+        $bar = $this->output->createProgressBar(Post::where('url', 'default')->count());
+
+        // Get all posts with "default" for the url
+        $posts = Post::where('url', 'default')->chunkById(100, function ($posts) use ($bar) {
```
DFurnes: Lil' nitpick: this `$posts` variable isn't storing anything.

app/Console/Commands/MakeDefaultURLsNull.php
```diff
@@ -0,0 +1,60 @@
...
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        info('rogue:nullurl: Starting!');
+
+        // Create the progress bar
+        $bar = $this->output->createProgressBar(Post::where('url', 'default')->count());
```
DFurnes: If you want, you could extract the query to a variable and use it for both the count & iterator:

```php
$query = Post::where('url', 'default');
$bar = $this->output->createProgressBar($query->count());

// Get all posts with "default" for the url
$query->chunkById(100, function ($posts) use ($bar) {
   // ...
```

app/Console/Commands/MakeDefaultURLsNull.php
katiecrane: whoops that's from pre-chunk

app/Console/Commands/MakeDefaultURLsNull.php
katiecrane: YES I didn't like doing it twice felt un-DRY. did not realize my dreams could come true 🌠

PR #552, opened 2017-12-14
Blink queue job

chloealee:
#### What's this PR do?
- Extracts the Blink API request into a queued job so that we don't fail the main thread if it gives an error or times out.

👀 

#### Questions
- Do we want to delay this job? If so, by how long? 
- How do I test this on my local? 😬 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/153605502) Pivotal Card

#### Checklist
- [ ] Tested on staging.


DFurnes:
Hmm, not sure what's up with the build – let's check what queue driver we use in Wercker.

chloealee:
ahh ok - thanks @DFurnes ! So should I just add `QUEUE_DRIVER=sync` to [these](https://app.wercker.com/dosomething/rogue/environment) env variables?

DFurnes:
I think the `.env.example` will override those if we don't comment out the `QUEUE_DRIVER` line.

chloealee:
OH 🤦‍♀️ sorry! When I clicked that link the first time I _swear_ that it was already set to `sync`. I'll change that now!

chloealee:
@DFurnes it looks like it isn't actually getting to the `SendToBlink` file and when `            SendToBlink::dispatch($payload);` is called, it returns a `500 Internal Server Error` with no message besides this. Nothing in the logs either 🤔  I made sure to use the correct pattern for a `campaign_run` and `campaign_run_id` as well to avoid any of the validation errors we were seeing with Blink that were similar in the past. Is there something I'm missing I don't see?

DFurnes:
Hmm, yeah... weird! I'll pull down on my local and see if I can help find out what's going on, but it looks like you did everything right from what I can tell! 🤔

chloealee:
Thanks @DFurnes !!

app/Jobs/SendToBlink.php
```diff
@@ -0,0 +1,50 @@
...
+     * @var \DoSomething\Gateway\Blink
+     */
+    protected $blink;
+
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct(Blink $blink)
```
DFurnes: Oh! This is super unintuitive, but dependencies are injected as parameters of the `handle` method for jobs ([example](https://laravel.com/docs/5.5/queues#class-structure)). So the `Blink $blink` parameter would move down, and this line would take your job payload.

app/Jobs/SendToBlink.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle($payload)
+    {
+        // If there is a signup_id, this is a userSignupPost.
+        // Otherwise, it is a userSignup.
+        if ($payload['signup_id']) {
```
DFurnes: Can we make two separate queue jobs for this? (One `SendSignupToBlink` and one `SendPostToBlink`, with the `Signup $signup` or `Post $post` as the job's argument?)

app/Jobs/SendToBlink.php
chloealee: omg I totally missed that!! Thank you!

app/Jobs/SendToBlink.php
chloealee: yes that makes more sense!

app/Jobs/SendPostToBlink.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle(Blink $blink)
+    {
+        $this->blink = $blink;
```
DFurnes: You can just use `$blink` directly here, no need to save to a property!

```php
public function handle(Blink $blink)
{
    $payload = $this->post->toBlinkPayload();
    $blink->userSignupPost($payload);
    // ...

File: app/Jobs/SendPostToBlink.php [hunk 4]
```diff
@@ -0,0 +1,46 @@
...
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+class SendPostToBlink implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
+     *
+     * @var Post to send to Blink.
```
DFurnes: Can you reformat this to be a standard docblock?

```php
/**
 * The post to send to Blink.
 *
 * @var Post
 */
```

app/Jobs/SendSignupToBlink.php
```diff
@@ -0,0 +1,47 @@
...
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+class SendSignupToBlink implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
```
DFurnes: Same formatting comment as above!

app/Services/PostService.php
```diff
@@ -43,8 +43,7 @@ public function create($data, $signupId, $transactionId)
 
         // Save the new post in Customer.io, via Blink.
         if (config('features.blink') && $should_send_to_blink) {
```
DFurnes: Not needed for this PR, but it'd be nice to extract this check inside the job as well (so we always dispatch `SendPostToBlink`, but only make the API call inside the job if this feature flag is toggled on). That way the job always runs and logs the expected payload when we're running on local machines.

PR #553, opened 2017-12-14
Install Clockwork for easier API debugging.

DFurnes:
#### What's this PR do?
This pull request installs [Clockwork](https://underground.works/blog/introducing-clockwork-2.0) when running Rogue in development. I've been using it with the [built-in integration](https://github.com/DoSomething/rogue/blob/a7008ec5f06c1b6acfb0a9b4a68d2fb1964dc41c/config/debugbar.php#L65-L74) in `laravel-debugbar`, but this adds some extra features that are really handy - most importantly, you can see details of JSON requests made in Postman/Paw!!

screen shot 2017-12-13 at 12 46 46 pm

👀

#### Relevant tickets
N/A

#### Checklist
- [ ] Tested on staging.


composer.json
```diff
@@ -8,7 +8,6 @@
     "laravel/framework": "5.5.*",
     "laravel/tinker": "^1.0",
     "aws/aws-sdk-php-laravel": "^3.1",
-    "barryvdh/laravel-debugbar": "^3.1.0",
```
katiecrane: I thought that this was to add clockwork with debugbar. Are we removing debugbar?

DFurnes: Clockwork works with debugbar already (with a limited feature set)!

The downside is that doesn't support a lot of Clockworks most handy features – the ability to work with JSON API responses (which I feel like are most often what we want to inspect) & the `clock()->startEvent()` and `clock()->stopEvent()` methods for measuring performance.

I'd love to try this out for a little bit and see what people think, and we can always switch back to Debugbar if it doesn't seem like an improvement.

katiecrane: Ooooo you mean Clockwork contains debugbar? When you said that before I thought you meant they work in conjunction! Cool that sounds good to me!

DFurnes: Sorry, that was awkwardly phrased! The [Clockwork chrome extension](https://chrome.google.com/webstore/detail/clockwork/dmggabnehkmmfmdffgajcflpdjlnoemp?hl=en) can function in a "limited" mode while the `barryvdh/laravel-debugbar` extension is installed w/ Composer, since it reads some of the same data. That's what I've been doing for a while now.

Removing the `barryvdh/laravel-debugbar` _will_ remove the debugbar interface, so you'll have to visit the `rogue.dev/__clockwork` page on your local app (or install the Chrome extension) in order to see the same kind of debugging information.

From my time playing with it, I think Clockwork should handle all the exact same things we were using debugbar for, in addition to having a better workflow for inspecting API requests & better tools for measuring performance.

DFurnes: Let's just consider this a test drive, and we can always switch back in a few weeks! 🚗

katiecrane: Ahhh I get it now! Makes sense and sounds good

PR #554, opened 2017-12-14
Add support for using ImageMagick on Heroku.

DFurnes:
#### What's this PR do?
This pull request adds a `IMAGE_DRIVER` environment variable, and instructs Heroku to install the [ImageMagick PHP extension](http://php.net/manual/en/book.imagick.php) so we can test out `IMAGE_DRIVER = imagick` there.

👀

Switching to use ImageMagick on Heroku should provide a good performance boost when resizing/cropping images, which will help us more efficiently handle larger image uploads.

#### Relevant tickets
N/A

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
     "league/csv": "^9.0",
     "dfurnes/environmentalist": "0.0.2",
     "ext-gd" : "*",
-    "ext-exif":"*"
+    "ext-exif":"*",
+    "ext-imagick": "*"
   },
   "require-dev": {
     "fzaninotto/faker": "^1.6",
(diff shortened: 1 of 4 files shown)


sheyd:
🥇 📈 🥃

DFurnes:
One downside is that this would require `ext-imagick` to be installed on Homestead boxes as well for Compose commands not to complain (since that has to be listed as an app dependency for it to get installed on the Heroku dyno). It's something we can always handle in `after.sh`, but still a lil' bummer:

```sh
# after.sh

sudo apt-get update
sudo apt-get install -y php-imagick
sudo service php7.0-fpm restart
```

sheyd:
That...kind of sucks. 👿

DFurnes:
We had tested deploying this directly to Thor before break and it didn't seem to have a noticeable performance improvement. I'm going to close this for now since it doesn't seem worth complicating our dev environments without a real payoff.

PR #558, opened 2018-01-02
Excludes soft deleted posts in count for pending posts per campaign 

chloealee:
#### What's this PR do?
- This PR fixes the bug we were seeing with inconsistent Pending Post counts on the Campaign Overview page (the Pending number would be 2 but then when clicking review, there would be no posts to review in the inbox). 
- This was because it was counting posts that had been soft deleted when making the count query. However, when pulling the actual posts that needed to be reviewed in the campaign inbox page, it did not include posts that have been soft deleted.
- This PR adds a line in the query to only count those posts whose `deleted_at = NULL`

👀 
- To test on local, find a campaign that has 0 pending posts to review. 
- Change a post's status to pending (that has been soft deleted). 
- You should still see the Pending count on the Campaign Overview page for the campaign to say 0. 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/152892786) Pivotal Ticket. 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Services/CampaignService.php
+++ b/app/Services/CampaignService.php
@@ -220,7 +220,7 @@ class CampaignService
                 ->leftJoin('posts', 'signups.id', '=', 'posts.signup_id')
                 ->selectRaw('signups.campaign_id, count(posts.id) as pending_count')
                 ->where('status', '=', 'pending')
-                ->where('posts.deleted_at', '=', NULL)
+                ->where('posts.deleted_at', '=', null)
                 ->wherein('signups.campaign_id', $ids)
                 ->groupBy('signups.campaign_id')
                 ->get();
--- a/app/Services/CampaignService.php
+++ b/app/Services/CampaignService.php
@@ -220,6 +220,7 @@ class CampaignService
                 ->leftJoin('posts', 'signups.id', '=', 'posts.signup_id')
                 ->selectRaw('signups.campaign_id, count(posts.id) as pending_count')
                 ->where('status', '=', 'pending')
+                ->where('posts.deleted_at', '=', NULL)
                 ->wherein('signups.campaign_id', $ids)
                 ->groupBy('signups.campaign_id')
                 ->get();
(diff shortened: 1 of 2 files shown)


resources/assets/components/Table/CampaignRow.js
```diff
@@ -10,6 +10,7 @@ class CampaignRow extends React.Component {
   }
 
   createCampaignRow(campaign) {
+    console.log('hi');
```
DFurnes: 👀

app/Services/CampaignService.php
```diff
@@ -220,6 +220,7 @@ public function getPendingPostTotals($campaigns)
                 ->leftJoin('posts', 'signups.id', '=', 'posts.signup_id')
                 ->selectRaw('signups.campaign_id, count(posts.id) as pending_count')
                 ->where('status', '=', 'pending')
+                ->where('posts.deleted_at', '=', null)
```
DFurnes: Ahh, good catch! I wonder if we could re-write this query to use the Eloquent builder (which would automatically take this into account), but I could see that getting tricky perhaps.

sbsmith86:
I feel like another issue we were seeing is that even if you `approved` all the posts it would still include them in the count. But I could have imagined that. would check with @ngjo to confirm

chloealee:
thanks for flagging, @sbsmith86 ! Do you mean that the count was the total number of approved posts?

resources/assets/components/Table/CampaignRow.js
chloealee: oops! thanks, @DFurnes !

sbsmith86:
@chloealee I think it was if it said there were 2 pending posts, and then you approved them, when you went back to the homepage it still said there were 2 pending posts.

chloealee:
@sbsmith86 ohh I see - tricky! @ngjo do you have an example of this?

ngjo:
@chloealee I don't have one off the top of my head - I do remember that happening in the past, but I wonder if it was happening before the post status had actually saved (before we added the loading symbol to the status button). I just tried saving some on Rogue Thor and those all checked out, but I'll keep an eye out for it!

PR #560, opened 2018-01-03
Log post event anytime we send to blink

sbsmith86:
#### What's this PR do?
Just a small update to follow up on my previous PR (#559 )

We need to send the `post_created` event when we move through the update functionality even though we are creating a post. This follows the logic of when we send posts to blink.

👀 

We have a card in our backlog to work through this logic. When we create a new post we shouldn't consider it an update.

### DIFF
--- a/app/Services/PostService.php
+++ b/app/Services/PostService.php
@@ -73,6 +73,9 @@ class PostService
         // Save the new post in Customer.io, via Blink.
         if (config('features.blink') && $postOrSignup instanceof Post && $should_send_to_blink) {
             SendPostToBlink::dispatch($postOrSignup);
+
+            // Log that a post was created.
+            info('post_created', ['id' => $postOrSignup->id, 'signup_id' => $postOrSignup->signup_id]);
         }
 
         // Add new transaction id to header.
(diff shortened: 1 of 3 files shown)


app/Services/PostService.php
```diff
@@ -73,6 +73,9 @@ public function update($signup, $data, $transactionId)
         // Save the new post in Customer.io, via Blink.
         if (config('features.blink') && $postOrSignup instanceof Post && $should_send_to_blink) {
             SendPostToBlink::dispatch($postOrSignup);
+
+            // Log that a post was created.
+            info('post_created', ['id' => $postOrSignup->id, 'signup_id' => $postOrSignup->signup_id]);
```
katiecrane: Could this actually be a Post or a Signup? Asking because if it's a Signup I don't think it will have a `signup_id`

DFurnes: I was gonna ask the same thing!! There's a very sneaky `instanceof` check above. 🕵️

sbsmith86: it's in a conditional that makes sure it's a `Post` instance. But ideally we wouldn't need to handle the Post or Signup situation.

katiecrane: oooo I see the sneaky logic now! solid this looks all good then. YAY LOGGING!!!!!

PR #561, opened 2018-01-03
Adds ability to delete a signup in Rogue admin UI

chloealee:
#### What's this PR do?
- Exposes token in front end.
- Adds ability for admin to delete a signup.
- Updates latest gateway version.

When an admin deletes a signup, the page will look like this: 
screen shot 2018-01-03 at 3 33 17 pm
Then, when you refresh the page, since the signup doesn't exist anymore, the standard Laravel "Page Not Found" page will pop up:
screen shot 2018-01-03 at 3 33 24 pm
@lkpttn let me know if you want me to change anything about this! 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/152487740) Pivotal Ticket.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "laravel/tinker": "^1.0",
     "aws/aws-sdk-php-laravel": "^3.1",
     "doctrine/dbal": "~2.5.13",
-    "dosomething/gateway": "^1.12.0",
+    "dosomething/gateway": "^1.13.0",
     "fideloper/proxy": "^3.3",
     "guzzlehttp/guzzle": "^6.2",
     "intervention/image": "^2.3",
--- a/resources/assets/components/Signup/index.js
+++ b/resources/assets/components/Signup/index.js
@@ -80,6 +80,7 @@ class Signup extends React.Component {
     this.showHistory = this.showHistory.bind(this);
     this.hideHistory = this.hideHistory.bind(this);
     this.deletePost = this.deletePost.bind(this);
+    this.deleteSignup = this.deleteSignup.bind(this);
     this.showUploader = this.showUploader.bind(this);
     this.hideUploader = this.hideUploader.bind(this);
     this.submitReportback = this.submitReportback.bind(this);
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -9,7 +9,7 @@
     "laravel/tinker": "^1.0",
     "aws/aws-sdk-php-laravel": "^3.1",
     "doctrine/dbal": "~2.5.13",
-    "dosomething/gateway": "^1.12.0",
+    "dosomething/gateway": "^1.13.0",
```
DFurnes: Make sure to run `composer update` to update the lockfile as well!

resources/assets/components/Signup/index.js
```diff
@@ -262,6 +263,34 @@ class Signup extends React.Component {
...
   }
 
+  // Delete a signup.
+  deleteSignup(signupId, event) {
+    event.preventDefault();
+    const confirmed = confirm('🚨🔥🚨 Are you sure you want to delete this? 🚨🔥🚨');
+
+    if (confirmed) {
+      // Make API request to Rogue to delete the signup on the backend.
+      const authorizedRequest = new RestApiClient(window.location.origin, {
```
DFurnes: This probably makes sense on the `RestApiClient` defined at `this.api`.

resources/assets/components/Signup/index.js
```diff
@@ -262,6 +263,34 @@ class Signup extends React.Component {
...
+        headers: { 'Authorization' : `Bearer ${window.AUTH}`}
+      });
+
+      const response = authorizedRequest.delete('api/v3/signups/'.concat(signupId));
```
DFurnes: This is subjective, but this might read a little better with a template tag:

```js
const response = authorizedRequest.delete(`api/v3/signups/${signupId}`);
```

resources/assets/components/Signup/index.js
```diff
@@ -321,67 +350,74 @@ class Signup extends React.Component {
...
-                    id={this.state.historyModalId}
-                    onUpdate={this.updateQuantity}
-                    onClose={e => this.hideHistory(e)}
-                    signup={signup}
-                    campaign={campaign}
-                    signupEvents={this.state.signupEvents}
-                  />
-                  : null}
-              </ModalContainer>
+      this.state.signup != undefined ?
```
DFurnes: It might make sense to have this as a guard clause instead (since there's so much in this ternary):

```js
if (! this.state.signup) {
  return <Empty header={`This signup has been deleted`} />;
}

return (
  <div className="signup">
    <div className="container__row">
        // ...

```

resources/assets/components/Signup/index.js
chloealee: @DFurnes will this break other things though since we are only using the tokens for the v3 endpoints?

DFurnes:
It's definitely outside the scope of the PR, but it'd be nice to rethink routing in Rogue somewhere down the line (and probably use [React Router](https://reacttraining.com/react-router/)) since it's a little awkward that it renders differently from the server after you refresh the page...

resources/assets/components/Signup/index.js
DFurnes: Oh, that's a good thought. I thiiiiiiink other endpoint should just ignore the `Authorization` header since they're relying on `Session`, but it's worth double-checking just to make sure!

chloealee: oh you're right ! I just tried with other functions (updating quantity, history modal, deleting post, tagging) and all still works with the change - thanks @DFurnes !

chloealee:
@DFurnes I agree! I've created the card [here](https://www.pivotaltracker.com/n/projects/2019429/stories/154037337). Thanks for flagging! Besides that, everything else is updated - is it ok to merge?

PR #562, opened 2018-01-04
Some updates to fix issues with latest Gateway release.

DFurnes:
#### What's this PR do?
This pull request cleans up some issues I'd introduced with the latest [Gateway release](https://github.com/DoSomething/gateway/releases/tag/v1.13.0).

🆙 Updates to [v1.13.1](https://github.com/DoSomething/gateway/releases/tag/v1.13.1) with fixes to key command & test suite. 805f38b

🔃 Switches config location for authorization server URL & public key path to `config/auth.php`, following the new preferred config from that Gateway release. 19b9a3d

💂 Use the included `SetGuard` middleware in Gateway, rather than custom one. 6b72b5a

🔕 Removes some unnecessary overrides we were making in the Handler, which were either unnecessary or prevented Laravel's [renderable exceptions](https://laravel.com/docs/5.5/errors#renderable-exceptions) (used by Gateway 1.13) from working. Also removes our custom JSON exception renderer since this is [in the framework now](https://github.com/laravel/framework/blob/6d1f0695a95a507772d205987895720341fd4d2d/src/Illuminate/Foundation/Exceptions/Handler.php#L450-L469). cd75372

#### How should this be reviewed?
I linked each commit in the description above!

#### Any background context you want to provide?
I hadn't fully tested the changes I had made in this Gateway release since I wanted to get Chloe's fix in DoSomething/gateway#102 out quickly. Sorry for the trouble!

#### Relevant tickets
N/A. Will make an "unplanned work" Pivotal card!

([Slack conversation](https://dosomething.slack.com/archives/C0YGXUE01/p1515017166000346))

#### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.
- [ ] Pinged a PM if this is a larger PR that would benefit from some additional testing love.

PR #563, opened 2018-01-04
Add named 'login' route.

DFurnes:
#### What's this PR do?
The [default "unauthenticated" handler](https://github.com/laravel/framework/blob/6d1f0695a95a507772d205987895720341fd4d2d/src/Illuminate/Foundation/Exceptions/Handler.php#L216-L221) in Laravel looks for a named `login` route. This creates one so that these continue to work as they did [before](https://github.com/DoSomething/rogue/blob/1a5db305e47934714583b0fc6d6b63f51d04383d/app/Exceptions/Handler.php#L88-L95).

👀

References my cleanup in #562.

#### Relevant tickets
N/A

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -10,7 +10,7 @@
  */
 
 // Homepage
-$router->get('/', 'PagesController@home');
+$router->get('/', 'PagesController@home')->name('login');
 
 // Authentication
 $router->get('login', 'Auth\AuthController@getLogin');


DFurnes:
Oh, one more note before merging – I added this name to the "home" route instead of the actual "login" route since that shows the user the "Press login to continue" prompt (rather than automatically starting the Northstar OAuth flow, which could be kinda jarring).

PR #564, opened 2018-01-04
Adds New Tags and Ability to Tag on Accepted and Rejected Posts 

chloealee:
#### What's this PR do?
- Adds the following new tags: 
  - Irrelevant
  - Inappropriate
  - Unrealistic Quantity
  - Test
  - Incomplete Action
- Adds ability to tag both accepted and rejected posts
- Tag button will be disabled until a post's status is accepted or rejected 
- Once a post's status is accepted or rejected, tag buttons become enabled and when you hover over the tags, they will be highlighted in blue 

👀 
**Does all the above make sense? Those scenarios should work!** 
Approved and tagged post:
screen shot 2018-01-05 at 10 49 00 am

**Rejected and tagged post:**
screen shot 2018-01-05 at 10 49 37 am

**Reviewed post with enabled tags vs unreviewed post with disabled tags:**
screen shot 2018-01-05 at 10 45 51 am

#### Relevant tickets
Fixes these Pivotal Cards:
- https://www.pivotaltracker.com/n/projects/2019429/stories/150443391 
- https://www.pivotaltracker.com/n/projects/2019429/stories/152892878

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/ReviewBlock/index.js
+++ b/resources/assets/components/ReviewBlock/index.js
@@ -22,7 +22,7 @@ class ReviewBlock extends React.Component {
         <ul className="form-actions -inline">
           <li><button className="button delete -tertiary" onClick={e => this.props.deletePost(post.id, e)}>Delete</button></li>
         </ul>
-        <Tags id={post.id} tagged={post.tags} onTag={this.props.onTag} disableTags={disableTags}/>
+        <Tags id={post.id} tagged={post.tags} onTag={this.props.onTag} disabled={disableTags}/>
       </div>
     );
   }
--- a/resources/assets/components/ReviewBlock/index.js
+++ b/resources/assets/components/ReviewBlock/index.js
@@ -11,7 +11,7 @@ class ReviewBlock extends React.Component {
 
   render() {
     const post = this.props.post;
-    const disableTags = post.status === 'accepted' || post.status === 'rejected' ? false : true;
+    const disableTags = ! (post.status === 'accepted' || post.status === 'rejected');
 
     return (
       <div>
--- a/resources/assets/components/ReviewBlock/index.js
+++ b/resources/assets/components/ReviewBlock/index.js
@@ -11,6 +11,7 @@ class ReviewBlock extends React.Component {
 
   render() {
     const post = this.props.post;
+    const disableTags = post.status === 'accepted' || post.status === 'rejected' ? false : true;
 
     return (
       <div>
@@ -21,9 +22,7 @@ class ReviewBlock extends React.Component {
         <ul className="form-actions -inline">
           <li><button className="button delete -tertiary" onClick={e => this.props.deletePost(post.id, e)}>Delete</button></li>
         </ul>
-        {post.status === 'accepted' ?
-          <Tags id={post.id} tagged={post.tags} onTag={this.props.onTag} />
-          : null}
+        <Tags id={post.id} tagged={post.tags} onTag={this.props.onTag} disableTags={disableTags}/>
       </div>
     );
   }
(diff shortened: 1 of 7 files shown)


lkpttn:
I think there needs to be more of a visual difference between active and inactive tag styles. If you look s the Master document in Abstract you can see there’s an opacity difference and the tags are very subdued until they’re actionable.

chloealee:
@lkpttn thanks for looking at this so quickly! I was thinking that since the hover highlight only works when the tag buttons are active, that would be the indicator but I can definitely add the opacity difference as well!

resources/assets/components/ReviewBlock/index.js
```diff
@@ -11,6 +11,7 @@ class ReviewBlock extends React.Component {
 
   render() {
     const post = this.props.post;
+    const disableTags = post.status === 'accepted' || post.status === 'rejected' ? false : true;
```
DFurnes: This reads a little awkwardly to me with the ternary... could we reformat like this?

```js
const disableTags = ! (post.status === 'accepted' || post.status === 'rejected');
```

resources/assets/components/Tags/index.js
```diff
@@ -60,7 +65,7 @@ class Tags extends React.Component {
         <ul className="aligned-actions">
           {map(tags, (label, key) => (
             <li key={key}>
-              <Tag isActive={this.props.tagged.includes(key)} isClicked={this.props.onTag} label={label} post={this.props.id} />
+              <Tag isActive={this.props.tagged.includes(key)} isClicked={this.props.onTag} label={label} post={this.props.id} disable={this.props.disableTags} />
```
DFurnes: Could we renamed this prop to `disabled` so it's clearer it's a boolean, and not an event handler?

resources/assets/components/ReviewBlock/index.js
```diff
@@ -21,9 +22,7 @@ class ReviewBlock extends React.Component {
         <ul className="form-actions -inline">
           <li><button className="button delete -tertiary" onClick={e => this.props.deletePost(post.id, e)}>Delete</button></li>
         </ul>
-        {post.status === 'accepted' ?
-          <Tags id={post.id} tagged={post.tags} onTag={this.props.onTag} />
-          : null}
+        <Tags id={post.id} tagged={post.tags} onTag={this.props.onTag} disableTags={disableTags}/>
```
DFurnes: Would it make sense to rename this `disableTags` to `disabled` as well?

chloealee:
@lkpttn updated the screenshots for your review!

lkpttn:
This is good to merge, we might come back and clean up some padding/space issues in some upcoming changes to the reviewer item.

chloealee:
@DoSomething/team-bleed this is ready for review with all tests passing now! Should I add a Dusk test [here](https://github.com/DoSomething/rogue/blob/master/tests/Browser/ReviewPostTest.php) to check that you can now tag a rejected post?

resources/assets/components/Tags/index.js
```diff
@@ -52,15 +52,22 @@ class Tags extends React.Component {
...
+      'test': 'Test',
+      'incomplete-action': 'Incomplete Action',
     };
 
+    const h2ClassName = this.props.disabled ? 'disabled' : 'enabled';
+
     return (
       <div>
-        <h4>Tags</h4>
+        <h4 className={h2ClassName}>Tags</h4>
```
sbsmith86: It is a little confusing to name this `h2ClassName` when it is styling an `h4` tag. Maybe we could call it like `showTags` which would describe the `enabled` or `disabled` status.

chloealee: love this idea!

PR #574, opened 2018-01-09
Script to change source from sms-mobilecommons to sms in signups and posts tables

chloealee:
#### What's this PR do?
- Creates a script to change source from `sms-mobilecommons` to `sms` in the signups and posts tables. 

Run `php artisan rogue:smssource` and all sources that are `sms-mobilecommons` should be changed to `sms`! 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/153820600) Pivotal Ticket.

#### Checklist
- [ ] Tested on staging.


app/Console/Commands/MakeSourceSms.php
```diff
@@ -0,0 +1,80 @@
...
+    public function handle()
+    {
+        // Start updating signups
+        info('rogue:smssource: Starting to update signups!');
+
+        // Get all signups that have "sms-mobilecommons" set as their source
+        $query = Signup::where('source', 'sms-mobilecommons');
+
+        // Update all the signups
+        $query->chunkById(100, function ($signups) {
```
chloealee: There are about 90-100k records that need to be changed. Should I increase this?

katiecrane: Great question! On the quantity script I just changed it from 100 to 1000 but we haven't tried it yet. There are A LOT of signups so 1000 makes sense to me

katiecrane: Also @chloealee 100% up to you, but this line and `$query = Signup::where('source', 'sms-mobilecommons');` can be condensed into one line like 
```
Signup::where('source', 'sms-mobilecommons')->chunkById(100, function ($signups) {
``` 
but totally go with whatever you find more readable.

chloealee: oooo I like that a lot more - thanks @katiecrane !

PR #577, opened 2018-01-12
Image cache key

sbsmith86:
#### What's this PR do?
Adds a `Surrogate-Key` header to `image/{post}` responses which will allow fastly to map objects stored in cache to a key that we can target and purge when needed. 

Here is an example of the headers that will be returned when getting an image with Glide turned on 

image

Using the `Surrogate-Key` as `image-30` we can clear this particular image from the fastly cache if it gets updated or deleted. 

👀 

This is one part of the work to start using Glide in rogue. Coming in another PR will be the work to target these images in Fastly using these keys. 

[Read up on Fastly Surrogate Keys here](https://docs.fastly.com/guides/purging/getting-started-with-surrogate-keys)

#### Relevant tickets
Addresses https://www.pivotaltracker.com/story/show/153072096

#### Checklist
- [ ] Tested on staging.


app/Http/Middleware/AddCacheKey.php
```diff
@@ -0,0 +1,26 @@
...
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @return mixed
+     */
+    public function handle($request, Closure $next, $type)
+    {
+        $params = $request->route()->parameters();
+        $post = $params['post'];
```
DFurnes: Since this is so specific to this particular route, I feel like it might make more sense to inline [in the controller](https://github.com/DoSomething/rogue/blob/0297d021e46aa06f615b8de1a5f13ef55f2e9a38/app/Http/Controllers/ImagesController.php#L63):

```php
return $server->getImageResponse($post->getMediaPath(), $request->all())
  ->withHeader('Surrogate-Key', 'post-'.$post->id);
```

sbsmith86: It was getting a little messy in the controller because `$server->getImageResponse($post->getMediaPath(), $request->all())` returns a `StreamedResponse` instance, so I thought it would be nicer to work with it in the middleware. Also if we ever decided to cache/purge other things you can just use this middleware to do it.

DFurnes: Oh, I made a bad assumption on what that class would be. If we wanted to keep it in the controller, it looks like this will work:

```php
$response = $server->getImageResponse($post->getMediaPath(), $request->all());
$response->headers->set('Surrogate-Key', 'post-'.$post->id);

return $response;
```

I'm not opposed to making a more generalized middleware here, but it feels a bit suspect to be pulling the `$post` out of the route parameters since that's tightly coupled to this use-case.

sbsmith86: otay! will update

sbsmith86:
@dave updated!

PR #579, opened 2018-01-12
Fix an issue with grabbing client ID.

DFurnes:
#### What's this PR do?
I'm a real dummy and didn't commit a fix I'd made to #578 on my local! Sorry! 🙃 This updates the ternary to check whether or not the token actually exists (`token()` itself is always a thing), and uses the correct [getter](https://github.com/DoSomething/gateway/blob/3126df34d278548b72cb924409419fe60f72be4f/src/Server/Token.php#L66-L74)).

😩

🤦‍♂️

#### Relevant tickets
🙅‍♂️

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -112,5 +112,5 @@ function client_id()
 {
     // If this is an API request, use embedded client ID from the JWT. For
     // web requests, check what this instance of Rogue is configured to use!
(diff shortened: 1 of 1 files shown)


DFurnes:
🙇

PR #581, opened 2018-01-16
Fastly purge

sbsmith86:
#### What's this PR do?
Purges images from the Fastly cache based on a cache key. 

Images are purged from cache when a post gets deleted or when it gets rotated. 

:eyes: 

Added some log messages so we can see if we are successfully making calls to fastly. 

I wonder if we should clean up some of the logic around the "edited image"?

Right now we automatically crop images and store them with the `edited_` prefix in S3 but if we can turn on glide then we should be able to just request the appropriate sized image without having to maintain "edited" versions. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/153072096

#### Checklist
- [ ] Tested on staging.


app/Http/Controllers/ImagesController.php
```diff
@@ -102,6 +104,17 @@ public function update($id, Request $request)
         $originalImage = $this->aws->storeImageData($originalImage->__toString(), $originalFilename);
         $editedImage = $this->aws->storeImageData((string) $editedImage, 'edited_' . $post->id);
 
+        if (config('features.glide')) {
```
DFurnes: Should we have a separate feature flag here, so we can disable this on local environments?

DFurnes: Or perhaps just check against `APP_ENV`, or whether or not Fastly config is empty?

sbsmith86: yea I can add that!

app/Http/Controllers/ImagesController.php
```diff
@@ -102,6 +104,17 @@ public function update($id, Request $request)
         $originalImage = $this->aws->storeImageData($originalImage->__toString(), $originalFilename);
         $editedImage = $this->aws->storeImageData((string) $editedImage, 'edited_' . $post->id);
 
+        if (config('features.glide') && ! empty(config('services.fastly'))) {
```
sbsmith86: @DFurnes I added the extra check - wanted to make sure this is what you meant before merging.

DFurnes: Will this array itself ever be empty (since it has the nulled keys `['url' => null, 'key' => null]`)? It might make more sense to specifically check that none of those keys are empty.

I imagine we'd want also this in other places where we interact with Fastly too, so is this maybe something that makes sense to move into the `purgeKey` method itself? (So we always call `purgeKey` but short-circuit it if the client isn't fully configured?)

sbsmith86: ahh yea. That is a great idea!

sbsmith86:
@DFurnes okie updated! Can you take another 👀

PR #584, opened 2018-01-18
Gambit unit tests

chloealee:
#### What's this PR do?
- Adds three new unit tests based on [Gambit Flow](https://github.com/DoSomething/gambit-campaigns/wiki/Rogue) to ensure that when we deploy, we're not breaking Gambit! 
  - `/activity?filter[northstar_id]=123&filter[campaign_run_id]=456` test 
  - test for submitting multiple posts to `/v2/posts`
  - test for submitting multiple posts to `/v3/posts`
- Allows you to create a post to the `POST /v3/posts` endpoint without a `quantity`

Make sure all tests pass! 

Did I cover all the tests that we need? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/152980920) Pivotal Card.

cc @aaronschachter 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Requests/Three/PostRequest.php
+++ b/app/Http/Requests/Three/PostRequest.php
@@ -29,7 +29,7 @@ class PostRequest extends Request
             'northstar_id' => 'nullable|objectid',
             'why_participated' => 'nullable|string',
             'caption' => 'required|nullable|string|max:140',
-            'quantity' => 'required|nullable|integer',
+            'quantity' => 'nullable|integer',
             'file' => 'image|dimensions:min_width=400,min_height=400',
             'status' => 'in:pending,accepted,rejected',
         ];
--- a/documentation/endpoints/v3/posts.md
+++ b/documentation/endpoints/v3/posts.md
@@ -140,7 +140,7 @@ POST /api/v3/posts
     The Drupal/Contentful ID of the campaign that the user's post is associated with.
   - **campaign_run_id**: (int) optional.
     The drupal campaign run node id of the campaign that the user's post is associated with.
-  - **quantity**: (int|nullable).
+  - **quantity**: (int|nullable) optional.
     The number of reportback nouns verbed. Can be `null`.
   - **why_participated**: (string).
     The reason why the user participated.
--- a/documentation/endpoints/v3/posts.md
+++ b/documentation/endpoints/v3/posts.md
@@ -140,8 +140,8 @@ POST /api/v3/posts
     The Drupal/Contentful ID of the campaign that the user's post is associated with.
   - **campaign_run_id**: (int) optional.
     The drupal campaign run node id of the campaign that the user's post is associated with.
-  - **quantity**: (int).
-    The number of reportback nouns verbed. 
+  - **quantity**: (int|nullable).
+    The number of reportback nouns verbed. Can be `null`.
   - **why_participated**: (string).
     The reason why the user participated.
   - **caption**: (string).
--- a/tests/Http/Api/ActivityApiTest.php
+++ b/tests/Http/Api/ActivityApiTest.php
@@ -308,7 +308,6 @@ class ActivityApiTest extends TestCase
 
     /**
      * Test for retrieving a user's activity with northstar_id and campaign_run_id query params.
-     * This test is to ensure we are not breaking the Gambit flow when it checks to see if a signup exists.
      *
      * GET /activity?filter[]=
      * @return void
--- a/tests/Http/Api/ActivityApiTest.php
+++ b/tests/Http/Api/ActivityApiTest.php
@@ -325,8 +325,8 @@ class ActivityApiTest extends TestCase
             'data' => [
                 [
                     'signup_id' => $signup->id,
-                    'northstar_id' => "17",
-                    'campaign_run_id' => "143",
+                    'northstar_id' => '17',
+                    'campaign_run_id' => '143',
                 ],
             ],
             'meta' => [
--- a/tests/Http/Api/ActivityApiTest.php
+++ b/tests/Http/Api/ActivityApiTest.php
@@ -305,4 +305,36 @@ class ActivityApiTest extends TestCase
             ],
         ]);
     }
+
+    /**
+     * Test for retrieving a user's activity with northstar_id and campaign_run_id query params.
+     * This test is to ensure we are not breaking the Gambit flow when it checks to see if a signup exists.
+     *
+     * GET /activity?filter[]=
+     * @return void
+     */
+    public function testActivityIndexWithNorthstarIdAndCampaignRunIdFilters()
+    {
+        $signup = factory(Signup::class)->create(['northstar_id' => 17, 'campaign_run_id' => 143]);
+        factory(Signup::class, 5)->create();
+
+        $response = $this->getJson('api/v2/activity?filter[northstar_id]=17&filter[campaign_run_id]=143');
+
+        $response->assertStatus(200);
+        $response->assertJson([
+            'data' => [
+                [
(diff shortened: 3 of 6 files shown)


tests/Http/Api/ActivityApiTest.php
```diff
@@ -305,4 +305,36 @@ public function testQuantityOnActivityIndex()
...
     }
+
+    /**
+     * Test for retrieving a user's activity with northstar_id and campaign_run_id query params.
+     * This test is to ensure we are not breaking the Gambit flow when it checks to see if a signup exists.
+     *
+     * GET /activity?filter[]=
+     * @return void
+     */
+    public function testActivityIndexWithNorthstarIdAndCampaignRunIdFilters()
```
sbsmith86: hmm I think we are deprecating this, we can keep this test so this is not a comment blocking merging. But lets not invest anymore in support for this endpoint. We can tell devs they can start using the more RESTful `/v3` endpoints

tests/Http/Api/PostApiTest.php
```diff
@@ -125,6 +125,101 @@ public function testCreatingAPost()
         ]);
     }
 
+    /**
+     * Test that a POST request to /posts with an existing post creates an additional new photo post.
+     * This test ensures we are not breaking Gambit when it creates a post for a signup that already has posts.
```
sbsmith86: A very nitpicky thing- I don't think we need to contextualize these tests to only Gambit flows. This is a good test to have for any other service that will be sending over multiple posts

tests/Http/Api/PostApiTest.php
```diff
@@ -125,6 +125,101 @@ public function testCreatingAPost()
...
+
+        $this->assertDatabaseHas('posts', [
+            'signup_id' => $signup->id,
+            'northstar_id' => $signup->northstar_id,
+            'campaign_id' => $signup->campaign_id,
+            'status' => 'pending',
+        ]);
+
+        // Create a second post.
+        // Note: when a post already exists, Gambit doesn't send why_participated again.
```
sbsmith86: Again, we should just be testing that you can create a second post without a `why_participated` . I could see us adding more tests around creating secondary posts without any number of fields.

sbsmith86:
This looks good to me! Are there any other tests you can think of us running. For example, submitting a post without certain fields, or signups without certain fields?

chloealee:
@sbsmith86 oooo I didn't think about those! 

Just asked in our [discussion](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/12?from_comment=16#discussion-12-comment-16), but have we made a decision around if each post has to have a `quantity`?

sbsmith86:
@chloealee yea let's definitely add a test around that!

chloealee:
Cool, just updated @sbsmith86 ! I had a few questions though that are in the comments!

app/Http/Requests/Three/PostRequest.php
```diff
@@ -29,7 +29,7 @@ public function rules()
             'northstar_id' => 'nullable|objectid',
             'why_participated' => 'nullable|string',
             'caption' => 'required|nullable|string|max:140',
-            'quantity' => 'required|nullable|integer',
```
chloealee: should this still be required but if there is no quantity, `null` should be sent? I tried to do this in the test (send through a `quantity` of `null` but it failed...

tests/Http/Api/PostApiTest.php
```diff
@@ -125,6 +125,99 @@ public function testCreatingAPost()
...
+            'remote_addr'      => $this->faker->ipv4,
+            'file'             => UploadedFile::fake()->image('photo.jpg'),
+            'crop_x'           => 0,
+            'crop_y'           => 0,
+            'crop_width'       => 100,
+            'crop_height'      => 100,
+            'crop_rotate'      => 90,
+        ]);
+
+        $response->assertStatus(200);
```
chloealee: Now that I think about it, shouldn't all of these codes be `201`? When I `dd` the response in the code, it is `201` since these are all created but for some reason the tests return a `200` status code...

chloealee: oh wait now I see - it checks to see if the [signup is created or not](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Three/PostsController.php#L114) and [determines the status code based on that](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Three/PostsController.php#L123) - we might want to change this (I know we talked about this before) since this is a little misleading...

app/Http/Requests/Three/PostRequest.php
sbsmith86: I think setting `nullable` should allow you to send `null` as the quantity value. We should also allow quantity parameter to just be not sent at all and we don't receive a `quantity` param we should store it as `null` in the DB. 

From what I can tell ` 'nullable|integer` should be all we need. Was that not working in the tests?

tests/Http/Api/PostApiTest.php
sbsmith86: yea +1 it really does need to be refactored.

app/Http/Requests/Three/PostRequest.php
chloealee: ah ok! Right now, the test sends `null` as the quantity and it works! 

Should I also write a test for not sending the quantity parameter at all? It's throwing a `500` for that but I can write a separate test to test both scenarios and fidget with it!

sbsmith86: Yea!

sbsmith86: if the quantity work becomes more than just creating the test- like application logic needs to be tweaked- we can do that in a separate ticket/PR to reduce the scope of this.

chloealee: ah i figured it out - I needed to use `isset` 🤦‍♂️

chloealee:
@sbsmith86 added two tests (one to test that a post is created if `null` is sent as `quantity` and one to test that a post can be made with no `quantity` param at all) and all tests are passing now - ready for a final look!

documentation/endpoints/v3/posts.md
```diff
@@ -140,8 +140,8 @@ POST /api/v3/posts
     The Drupal/Contentful ID of the campaign that the user's post is associated with.
   - **campaign_run_id**: (int) optional.
     The drupal campaign run node id of the campaign that the user's post is associated with.
-  - **quantity**: (int).
-    The number of reportback nouns verbed. 
+  - **quantity**: (int|nullable).
```
sbsmith86: should we note it is optional like we did for `campaign_run_id`?

chloealee: love this idea! added :)

tests/Http/Three/PostTest.php
```diff
@@ -138,6 +138,168 @@ public function testCreatingAPost()
...
+            'quantity' => $quantity,
+        ]);
+    }
+
+    /**
+     * Test that a POST request to /posts without a quantity creates a new post.
+     *
+     * @return void
+     */
+    public function testCreatingAPostWithoutQuantity()
```
sbsmith86: could we call this `testCreatingAPostWithANullQuantity`? - sorry that's a bit nitpicky, but I think it describes the test better?

chloealee: not nitpicky - that is a lot more clear and agree! Actually updated that when adding the new post but not sure it updated in time :) Does `testCreatingAPostWithNullAsQuantity` work?

sbsmith86: wonderful

PR #585, opened 2018-01-18
Add action columns

sbsmith86:
#### What's this PR do?
Adds columns to the `posts` table to store `type`, `action_bucket`, and `details`

the `details` column should be `JSON` but our current DB version doesn't support that column type so it is `TEXT` for now.

👀 

A full thread on how we decided to add these columns: https://github.com/orgs/DoSomething/teams/team-bleed/discussions/12

**Question** 
* Should we have a default value for `action_bucket`? Old posts won't have a value in it, but it is not nullable because we want it to be required. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/154212087

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/database/migrations/2018_01_18_191700_add_columns_to_categorize_posts.php
+++ b/database/migrations/2018_01_18_191700_add_columns_to_categorize_posts.php
@@ -14,8 +14,8 @@ class AddColumnsToCategorizePosts extends Migration
     public function up()
     {
         Schema::table('posts', function (Blueprint $table) {
-            $table->string('type')->after('northstar_id')->default('photo')->comment('Describes the type of post submitted e.g. photo, call, voter-reg');
-            $table->string('action_bucket')->after('type')->comment('Describes the bucket the action is tied to. A campaign could ask for multiple types of actions throught the life of the campaign.');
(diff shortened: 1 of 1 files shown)


database/migrations/2018_01_18_191700_add_columns_to_categorize_posts.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('posts', function (Blueprint $table) {
+            $table->string('type')->after('northstar_id')->default('photo')->comment('Describes the type of post submitted e.g. photo, call, voter-reg');
+            $table->string('action_bucket')->after('type')->comment('Describes the bucket the action is tied to. A campaign could ask for multiple types of actions throught the life of the campaign.');
```
DFurnes: I wonder if we'll want an index on either of these fields (for example, so we could query only posts for the `v1/reportbacks` or `v2/activity` endpoint, or so we could have separate galleries on PN for each action.)

sbsmith86: yes! indexes!

PR #586, opened 2018-01-22
Log only every 1000 signups

katiecrane:
#### What's this PR do?
1. Now we log when we've updated every 1000th signup, not every single atomic thing that the script does to save log space.
2. Remove progress bar since that doesn't really jive with Heroku

👀 

Saving log space!

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Console/Commands/PostQuantity.php
+++ b/app/Console/Commands/PostQuantity.php
@@ -52,13 +52,11 @@ class PostQuantity extends Command
 
                 // If no posts, move on
                 if ($posts->isEmpty()) {
-                    info('rogue:postquantity: No posts for signup ' . $signup->id);
                     $bar->advance();
                     continue;
                 }
 
                 // Sum quant of posts
-                info('rogue:postquantity: Updating posts for signup ' . $signup->id);
                 $postQuantityTotal = $posts->sum('quantity');
 
                 // If >0, see if any posts have null quant
@@ -88,6 +86,7 @@ class PostQuantity extends Command
                 // Advance the progress bar
                 $bar->advance();
             }
+            info('rogue:postquantity: Updated posts for signup ' . $signup->id);
         });
         // We did it!
         $bar->finish();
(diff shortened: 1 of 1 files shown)


app/Console/Commands/PostQuantity.php
```diff
@@ -74,23 +68,17 @@ public function handle()
...
                 }
 
                 // If no posts have a quantity yet
                 $quantity = $signup->getQuantity();
                 $this->putQuantityOnPosts($posts, $quantity);
-
-                // Advance the progress bar
-                $bar->advance();
             }
+            info('rogue:postquantity: Updated posts for signup ' . $signup->id);
```
sbsmith86: are you going to have a `$signup` at this point? it's not in the foreach loop.

katiecrane: @sbsmith86 I wasn't sure if it would still be set there either, but it worked locally

sbsmith86: hmm is it for sure the 1000th signup or just the first one in that chunk?

katiecrane: @sbsmith86 it's the 1000th signup and the last one in the chunk. here is the last bit of example logs from my local:
```
[2018-01-22 16:14:23] local.INFO: rogue:postquantity: Updated posts for signup 27000  {"user_id":null,"client_id":null,"request_id":null}
[2018-01-22 16:14:23] local.INFO: rogue:postquantity: Updated posts for signup 28000  {"user_id":null,"client_id":null,"request_id":null}
[2018-01-22 16:14:23] local.INFO: rogue:postquantity: Updated posts for signup 29000  {"user_id":null,"client_id":null,"request_id":null}
[2018-01-22 16:14:23] local.INFO: rogue:postquantity: Updated posts for signup 29556  {"user_id":null,"client_id":null,"request_id":null}
```
So the last chunk had less than 1000 signups so the signup id is `29556`

PR #592, opened 2018-01-23
Update SignupController@index and @show to include authorization logic. Adds Query Scope.

chloealee:
#### What's this PR do?
- Adds the `?include=posts` query param to `GET /v3/signups`
- Updates SignupController@index and @show to include authorization logic (using query scopes!). 
  - If the `?include=posts` query param is used, only show posts that are accepted for anonymous users. 
  - If the user is an admin, they can see all posts! 
  - If the signup belongs to the user, they can also see all posts (accepted, rejecting, pending). 
- Update the PostsController@index to user query scopes. 
- Adds tests to check that only admins/staff and signup owners can see posts when using `?include=posts` for `GET /v3/signups` and `GET /v3/signups/:signup_id`
- Updates documentation

- All tests should pass! 
- Does the logic make sense? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/153365879) Pivotal Card.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -89,6 +89,7 @@ class SignupsController extends ApiController
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Signup::$indexes);
 
+        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
         if ($request->query('include') === 'posts') {
             $query = $query->whereVisible();
         }
@@ -106,6 +107,7 @@ class SignupsController extends ApiController
      */
     public function show(Request $request, Signup $signup)
     {
+        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
         if ($request->query('include') === 'posts') {
             $signup = Signup::whereVisible()->first();
         }
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -89,6 +89,10 @@ class SignupsController extends ApiController
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Signup::$indexes);
 
+        if ($request->query('include') === 'posts') {
+            $query = $query->whereVisible();
+        }
+
         return $this->paginatedCollection($query, $request);
     }
 
@@ -103,9 +107,7 @@ class SignupsController extends ApiController
     public function show(Request $request, Signup $signup)
     {
         if ($request->query('include') === 'posts') {
-            if (! is_staff_user() && auth()->id() != $signup->northstar_id) {
-                throw new AuthorizationException('You don\'t have the correct role to view this signup\'s post!');
-            }
+            $signup = Signup::whereVisible()->first();
         }
 
         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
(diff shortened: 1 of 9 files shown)


app/Http/Controllers/Three/SignupsController.php
```diff
@@ -102,6 +102,12 @@ public function index(Request $request)
      */
```
chloealee: @DFurnes I'm a little confused how to use query scope on this. I tried to do something like 
```
Signup::scopeWhereVisible($query, $northstarId)->get();
```
But not sure how we can do that here because it sends the whole `$query` to `$this->paginatedCollection` and does more work later to find the exact results we're querying for? Am I thinking of this incorrectly?

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -102,6 +102,12 @@ public function index(Request $request)
...
     public function show(Request $request, Signup $signup)
     {
+        if ($request->query('include') === 'posts') {
```
chloealee: @DFurnes also for this one, since the query happens later when we send to the transformer, and just includes all posts if it sees the `includes=posts` flag, not sure how to weed out the ones with the status not approved?

app/Http/Controllers/Three/SignupsController.php
DFurnes: You should be able to access the scopes here since it's a `Signup` query:

```php
$query = $this->newQuery(Signup::class);

// ...

$query = $query->whereVisible($northstarId);
```

chloealee: ohhhhhh that makes a lot more sense! Thank you!!

app/Http/Controllers/Three/SignupsController.php
chloealee: figured it out - disregard!

app/Models/Signup.php
```diff
@@ -156,4 +156,19 @@ public function getQuantity()
         // If we are supporting quantity on posts then we can just return the summed quantity across all posts under the signup.
         return $this->posts->sum('quantity');
     }
+
+    /**
+     * Scope a query to only return signups if a user is an admin, staff, or is owner of signup.
+     *
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function scopeWhereVisible($query)
```
DFurnes: Should we rename this one to `withVisiblePosts`?

tests/Http/Three/SignupTest.php
```diff
@@ -199,6 +299,65 @@ public function testSignupWithIncludedPosts()
...
+     *
+     * GET /api/v3/signups/186?include=posts
+     * @return void
+     */
+    public function testSignupWithIncludedPostsWithMultipleCredentials()
+    {
+        $signup = factory(Signup::class)->create();
+        $post = factory(Post::class)->create();
+        $post->signup()->associate($signup);
+        $post->status = 'rejected';
```
DFurnes: If you want, there's a [factory]() for this!

```php
$post = factory(Post::class, 'rejected')->create();
$signup = $post->signup;

```

chloealee: oh awesome, thank you!!

PR #593, opened 2018-01-24
Log less in the PostQuantity script by chunking by 10,000

katiecrane:
#### What's this PR do?
Turns out I don't know how to estimate how much storage stuff takes up in PaperTrail! We need to log less! Going to try doing this by chunking by 10,000.

Can anything bad happen in production if we chunk 10x as many things? 5000 would be another option to chunk by.

#### Checklist
- [ ] Tested on staging.


katiecrane:
@chloealee I just updated it so that we can set the "chunk" number when we run the command and don't have to update the code!

chloealee:
@katiecrane brilliant!

PR #596, opened 2018-01-26
Creates /v3/reviews endpoint

chloealee:
#### What's this PR do?
- Added `/v3/reviews` endpoint using OAuth logic 
- Adds tests
- Updates documentation to have Reactions to have it's own v3 file for consistency) and adds documentation for Reviews under v3 endpoints. 

tell us, how can we review, to test that this works

Do we want to also great `create` and `delete` routes? Not sure we need this right now but in order to follow CRUD? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/154640860) Pivotal Story

#### Checklist
- [ ] Tested on staging.


tests/Http/Three/ReviewsTest.php
```diff
@@ -0,0 +1,58 @@
...
+        $post = factory(Post::class)->create();
+
+        $response = $this->postJson('api/v3/reviews', [
+            'post_id' => $post->id,
+            'status' => 'accepted',
+        ]);
+
+        $response->assertStatus(401);
+    }
+}
```
chloealee: The other reactions endpoint tests test that an event record is made when a reaction is created. I think we hold off on this until we refactor the events logic - is anyone opposed to this?

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,99 @@
...
+use Illuminate\Auth\Access\AuthorizationException;
+use Rogue\Http\Transformers\Three\PostTransformer;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+
+class ReviewsController extends ApiController
+{
+    /**
+     * The post service instance.
+     *
+     * @var Rogue\Repositories\ThreePostRepository
```
DFurnes:
```diff
- * The post service instance.
+ * The post repository instance.
  *
- * @var Rogue\Repositories\ThreePostRepository
+ * @var \Rogue\Repositories\Three\PostRepository
```

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,99 @@
...
+            $review = $request->all();
+            $post = Post::where('id', $review['post_id'])->first();
+            $review['signup_id'] = $post->signup_id;
+            $review['northstar_id'] = $post->northstar_id;
+            $review['old_status'] = $post->status;
+
+            // Append admin's ID to the request for the "reviews" service.
+            $review['admin_northstar_id'] = auth()->id();
+            $reviewedPost = $this->post->reviews($review);
+            $reviewedPostCode = $this->code($reviewedPost);
```
DFurnes: I'm not sure these conditionals will ever be used.

Unless it hits an exception, [`$this->post->reviews`](https://github.com/DoSomething/rogue/blob/94a743e4c9a14243e4ca58b8c4d05953016dc594/app/Repositories/Three/PostRepository.php#L154-L181) will always return a post.

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,99 @@
...
+     * Update a post(s)'s status when reviewed.
+     *
+     * @param Request $request
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function reviews(Request $request)
+    {
+        $request->validate([
+            'post_id' => 'required',
```
DFurnes: Let's check here whether a post exists: `required|exists:posts`.

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,99 @@
...
+        if (token()->role() === 'admin') {
+            $review = $request->all();
+            $post = Post::where('id', $review['post_id'])->first();
+            $review['signup_id'] = $post->signup_id;
+            $review['northstar_id'] = $post->northstar_id;
+            $review['old_status'] = $post->status;
+
+            // Append admin's ID to the request for the "reviews" service.
+            $review['admin_northstar_id'] = auth()->id();
+            $reviewedPost = $this->post->reviews($review);
```
DFurnes: I think we should consider moving away from passing arrays like this in our `Three` repositories/services. 

Instead, we can send named parameters for the things we care to know:

```php
$post = Post::find($request['post_id']);
$reviewedPost = $this->post->reviews($post, $request['status'], $request['comment']);
```


```php
/**
 * Updates a post's status after being reviewed.
 *
 * @param Post $post
 * @param string $status
 *
 * @return Post
 */
public function reviews(Post $post, $status, $comment)
{
    // Create the Review.
    $review = Review::create([
        'signup_id' => $post->signup_id,
        'admin_northstar_id' => auth()->id(),
        'status' => $status,
        'old_status' => $post->status,
        'comment' => $comment,
        'post_id' => $post->id,
    ]);

    // Update the status on the Post.
    $post->status = $data['status'];
    $post->save();

    return $post;
}
```

This makes it clearer what the actual dependencies and inputs are at each step of the process.

app/Http/Controllers/Three/ReviewsController.php
DFurnes: @chloealee Let's add this in before merging!

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public function reviews(Request $request)
+    {
+        $request->validate([
+            'post_id' => 'required',
+            'status' => 'in:pending,accepted,rejected',
+        ]);
+
+        // Only allow an admin to review the post.
+        if (token()->role() === 'admin') {
```
DFurnes: We can handle this with the `role:admin` middleware, like we do elsewhere!

```diff
    /**
     * Create a controller instance.
     *
     * @param  PostContract $posts
     * @return void
     */
    public function __construct(PostRepository $post)
    {
        $this->post = $post;
        $this->transformer = new PostTransformer;

        $this->middleware('auth:api');
+       $this->middleware('role:admin');
    }
```

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,91 @@
...
+            $review = $request->all();
+            $post = Post::where('id', $request['post_id'])->first();
+
+            // Append admin's ID to the request for the "reviews" service.
```
DFurnes: I don't think this comment is relevant anymore.

app/Http/Controllers/Three/ReviewsController.php
chloealee: @DFurnes shoot that got deleted by accident - re-added! :)

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $request->validate([
+            'post_id' => 'required',
+            'status' => 'in:pending,accepted,rejected',
+        ]);
+
+        // Only allow an admin to review the post.
+        if (token()->role() === 'admin') {
+            $review = $request->all();
+            $post = Post::where('id', $request['post_id'])->first();
```
DFurnes: Let's use `findOrFail` so we get auto-404s if it can't load the post:

```php
$post = Post::findOrFail($request['post_id']);
```

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,91 @@
...
+
+            // Append admin's ID to the request for the "reviews" service.
+            $reviewedPost = $this->post->reviews($post, $request['status'], isset($request['comment']) ? $request['comment'] : null);
```
DFurnes: You shouldn't need to do an `isset` here since this is a Request object.

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,91 @@
...
+
+            $reviewedPostCode = $this->code($reviewedPost);
```
DFurnes: I'm still not sure when this would get evaluated. What's a scenario where `$reviewedPost` would be null?

app/Http/Controllers/Three/ReviewsController.php
chloealee: oh right!

app/Http/Controllers/Three/ReviewsController.php
chloealee: you're totally right!

chloealee:
@DFurnes fixed what you suggested and also added one more test to make sure that you can't review a post that doesn't exist!

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,66 @@
...
+     * Update a post(s)'s status when reviewed.
+     *
+     * @param Request $request
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function reviews(Request $request)
+    {
+        $request->validate([
+            'post_id' => 'required:exists:posts',
```
DFurnes: This looks like a typo: `required:exists:posts` → `required|exists:posts`.

...but it seems like we can safely rely on the `findOrFail` below now, so I think you're safe to remove it!

app/Http/Controllers/Three/ReviewsController.php
```diff
@@ -0,0 +1,66 @@
...
+            'status' => 'in:pending,accepted,rejected',
+        ]);
+
+        $review = $request->all();
```
DFurnes: This variable is never used!

app/Http/Controllers/Three/ReviewsController.php
chloealee: oops you're right! removed!

PR #597, opened 2018-01-26
Signup Import script updates: Remove progress bar, log at given frequency

katiecrane:
#### What's this PR do?
1. Removed progress bar because that is not useful with Heroku
2. Add an argument to only log every so many signups imported. Defaults to 1000. 

Is this a reasonable way to log every X signups imported?

We most likely can't log every single signup imported when running this script because it will fill our logs right up!

#### Relevant tickets
[Card](https://www.pivotaltracker.com/n/projects/2019429/stories/152608407)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Console/Commands/ImportSignupsCommand.php
+++ b/app/Console/Commands/ImportSignupsCommand.php
@@ -14,7 +14,7 @@ class ImportSignupsCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'rogue:signupimport {path}';
+    protected $signature = 'rogue:signupimport {path} {logfreq=1000}';
 
     /**
      * The console command description.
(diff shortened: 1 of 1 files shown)


app/Console/Commands/ImportSignupsCommand.php
```diff
@@ -14,7 +14,7 @@ class ImportSignupsCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'rogue:signupimport {path}';
+    protected $signature = 'rogue:signupimport {path} {logfreq=1000}';
```
DFurnes: This is so nitpicky, but I think this feels like it might make more sense as an [option](https://laravel.com/docs/5.5/artisan#options):

```sh
php artisan rogue:signupimport example.csv --logfreq=1000
# versus:
php artisan rogue:signupimport example.csv 1000
```

katiecrane: @DFurnes I like that better too!

katiecrane:
@DFurnes updated!

PR #598, opened 2018-01-26
Add migrations to make new post columns nullable

sbsmith86:
#### What's this PR do?
When I added these columns, I didn't make them `nullable` because I figured we would need them. But we won't always need them and now that I am digging into storing things in these columns, tests are mad about these columns not being nullable. 

So this PR makes them nullable. 

👀 

#### Relevant tickets
Fixes 🐛

### DIFF
--- a/database/migrations/2018_01_26_211937_make_bucket_and_details_columns_nullable.php
+++ b/database/migrations/2018_01_26_211937_make_bucket_and_details_columns_nullable.php
@@ -14,8 +14,9 @@ class MakeBucketAndDetailsColumnsNullable extends Migration
     public function up()
     {
         Schema::table('posts', function (Blueprint $table) {
-            $table->string('action_bucket')->nullable()->change();
             $table->text('details')->nullable()->change();
+            $table->dropColumn('action_bucket');
+            $table->string('action')->index()->after('type')->default('default')->comment('Describes the bucket the action is tied to. A campaign could ask for multiple types of actions throught the life of the campaign.');
         });
     }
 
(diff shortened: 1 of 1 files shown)


database/migrations/2018_01_26_211937_make_bucket_and_details_columns_nullable.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('posts', function (Blueprint $table) {
+            $table->string('action_bucket')->nullable()->change();
```
katiecrane: I see why `details` would be `null`, but why would `action_bucket` be `null` for anything new?

DFurnes: Yeah, I'm curious if we keep this non-nullable but have it default to `default` (or w/e)? Otherwise I could see bugs popping up since we'd have to always keep in mind both possible "states" of this column.

```
$table->string('action_bucket')->default('default')->change();
```

sbsmith86: To support old tests, old posts, or old requests that might not be sending it. I mean we can probably deal with support for `action_bucket` in more detail. But it is odd that old posts now have an empty string in that column since it was before "buckets" were a thing. 

If we don't want to allow this to be `nullable` then we probably need a sensible default value.

sbsmith86: Ok, i just remember running into issues storing `default` everywhere before (for images). But if we are good with that then I can change it.

sbsmith86:
@DoSomething/team-bleed Can I get another set of eyes on this. This includes the changes we discusses IRL (renaming `action_bucket` to `action` and setting it's default value)

PR #600, opened 2018-01-30
CSV upload

sbsmith86:
#### What's this PR do?

* Adds a `/import` route that shows the user a form they can use to import a CSV
* In `ImportController` we grab the field and store it in S3 and then kick off a queue job to import the turbovote csv. 
* Redirects the user to import page with a message that says the import is happening 
* `ImportTurboVotePosts.php` is the queue job that processes the CSV and imports it into the DB.

👀 

tests should still pass.

Will do more testing on thor.

* This will send new signups to Blink and new posts to Blink. Team O'doyle is aware of the changes we are making to the blink payload and I have suggested they validate the request appropriately. [See the full discussion here](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/13)

* I think we should do some work around showing the import progess to the admin with clear messages if something fails. 

#### Relevant tickets
Fixes  https://www.pivotaltracker.com/story/show/154154143

#### Checklist
- [ ] Tested on staging.


app/Jobs/ImportTurboVotePosts.php
```diff
@@ -0,0 +1,193 @@
...
+        // Get the file from storage based on path.
+        $file = Storage::get($this->filepath);
+
+        // Read in the file and get the records.
+        $csv = Reader::createFromString($file);
+        $csv->setHeaderOffset(0);
+        $records = $csv->getRecords();
+
+        foreach ($records as $record) {
+            info('Importing record ' . $record['id'], ['record' => $record]);
```
katiecrane: @sbsmith86 Do we have any sense of how large we expect these CSVs to be? I want to be mindful of not overloading the logs, but I also recognize that we may want detailed records so we can verify that each signup got created.

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -0,0 +1,193 @@
...
+                        ];
+
+                        $signup = $signupService->create($signupData, $referralCodeValues['northstar_id']);
+                    }
+                    // Check if a post already exists.
+                    $post = Post::where([
+                        'signup_id' => $signup->id,
+                        'northstar_id' => $referralCodeValues['northstar_id'],
+                        'campaign_id' => $referralCodeValues['campaign_id'],
+                        'type' => 'voter-reg',
```
katiecrane: So exciting to see this!!!

app/Jobs/ImportTurboVotePosts.php
sbsmith86: I will double check with @ngjo but I don't think we will have a good idea about this until it is out in the wild. I prefer to have more logs and cut back if we need to since I want to make sure we can track how these run.

app/Jobs/ImportTurboVotePosts.php
```diff
@@ -0,0 +1,193 @@
...
+                    if (! $post) {
+                        $tvCreatedAtMonth = strtolower(Carbon::parse($record['created-at'])->format('F-Y'));
+                        $postDetails = $this->extractDetails($record);
+
+                        $postData = [
+                            'campaign_id' => $referralCodeValues['campaign_id'],
+                            'northstar_id' => $referralCodeValues['northstar_id'],
+                            'type' => 'voter-reg',
+                            'action' => $tvCreatedAtMonth . '-turbovote',
+                            'status' => $record['voter-registration-status'],
```
katiecrane: @sbsmith86 I might have missed something somewhere but I didn't think that we could directly store status as the TurboVote status because it might be something like `Registered` instead of `Accepted` or `Rejected`.

app/Jobs/ImportTurboVotePosts.php
katiecrane: Sounds like a good game plan to me!

app/Jobs/ImportTurboVotePosts.php
sbsmith86: We don't do any validation that would not allow different strings to be stored (as long as they are strings). There is some validation around API requests only being able to accept `accepted`, `rejected`, and `pending` but that will not come into play here since we are adding the post ourselves.

katiecrane: @sbsmith86 I thought we were going to decide if it should be `accepted`, `pending`, or `rejected` based on the TurboVote status, and then store the raw status in the `details` column, but since we bypass the validation creating the posts this way we're going to go ahead and store the TurboVote status directly in the `status` column. Am I understanding this right?

sbsmith86: @katiecrane ahh right right. I thiink we did discuss this but I did not implement any translation of turbovote statuses to our statuses. Let me bring this to the team and try to remember what we should do here.

sbsmith86: @katiecrane thanks for catching this! I forgot about that discussion (I need to write things down ugh) but I spoke with Jen and I think the best/easiest approach for now is just to store the TV status as is. https://dosomething.slack.com/archives/C0YGXUE01/p1517342932000844

katiecrane: Haha I have to write everything down too or it disappears so I feel that! That sounds good!

PR #601, opened 2018-01-30
Adds `/v3/posts/:post_id/tags` RESTful endpoints

chloealee:
#### What's this PR do?
- Adds `/v3/posts/:post_id/tags` RESTful endpoints:
  - `POST /v3/posts/:post_id/tags` 
  - `DELETE /v3/posts/:post_id/tags` 
- Adds testing
- Adds documentation

- All tests should pass! 
- 👀 

Right now, we should only need the POST and DELETE endpoints. An index, show, and update methods are not useful to us right now. 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/154640911) Pivotal Card.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/tests/Http/Three/TagsTest.php
+++ b/tests/Http/Three/TagsTest.php
@@ -4,7 +4,6 @@ namespace Tests\Http\Three;
 
 use Tests\TestCase;
 use Rogue\Models\Post;
-use Rogue\Models\User;
 
 class TagsTest extends TestCase
 {
--- a/tests/Http/Three/TagsTest.php
+++ b/tests/Http/Three/TagsTest.php
@@ -62,115 +62,108 @@ class TagsTest extends TestCase
     public function testDeleteTagOnAPost()
     {
         // Create a post with a tag.
-        // $post = factory(Post::class)->create();
-        // $post->tag('Good Photo');
+        $post = factory(Post::class)->create();
+        $post->tag('Good Photo');
 
-        // $this->assertContains('Good Photo', $post->tagNames());
+        $this->assertContains('Good Photo', $post->tagNames());
 
-        // $response = $this->withAdminAccessToken()->deleteJson('api/v3/posts/' . $post->id . '/tag', [
-        //     'tag_name' => 'Good Photo',
-        // ]);
+        $response = $this->withAdminAccessToken()->deleteJson('api/v3/posts/' . $post->id . '/tag', [
+            'tag_name' => 'Good Photo',
+        ]);
 
-        // // Make sure that the tag is deleted.
-        // $response->assertStatus(200);
-        // $this->assertEmpty($post->tagNames());
+        // Make sure that the tag is deleted.
+        $response->assertStatus(200);
+        $this->assertEmpty($post->tagNames());
 
         // @TODO: Make sure we created a event for the tag once events are refactored.
     }
 
-    // /**
-    //  * Test deleting one tag on a post only deletes that tag
-    //  *
-    //  * POST /tags
-    //  * @return void
-    //  */
-    // public function testAddMultipleTagsAndDeleteOne()
-    // {
-    //     $this->actingAsAdmin();
+    /**
+     * Test that a non-admin cannot untag a post.
+     *
+     * DELETE /v3/posts/:post_id/tag
+     * @return void
+     */
+    public function testUnauthenticatedUserCannotUnTagAPost()
+    {
+        // Create the models that we will be using
+        $post = factory(Post::class)->create();
 
-    //     // Create a post with a tag.
-    //     $post = factory(Post::class)->create();
-    //     $post->tag('Good Photo');
-    //     $post->tag('Tag To Delete');
+        // Apply the tag to the post
+        $response = $this->deleteJson('api/v3/posts/' . $post->id . '/tag', [
+            'tag_name' => 'Good Photo',
+        ]);
 
-    //     // Make sure both tags actually exist
-    //     $this->assertContains('Good Photo', $post->tagNames());
-    //     $this->assertContains('Tag To Delete', $post->tagNames());
+        $response->assertStatus(401);
+    }
 
-    //     // Send request to remove "Tag To Delete" tag
-    //     $response = $this->postJson('tags', [
-    //         'post_id' => $post->id,
-    //         'tag_name' => 'Tag To Delete',
-    //     ]);
+    /**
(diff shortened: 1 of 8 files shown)


tests/Http/Three/TagsTest.php
```diff
@@ -0,0 +1,168 @@
...
+     */
+    public function testDeleteTagOnAPost()
+    {
+        // Create a post with a tag.
+        $post = factory(Post::class)->create();
+        $post->tag('Good Photo');
+
+        $this->assertContains('Good Photo', $post->tagNames());
+
+        $response = $this->withAdminAccessToken()->deleteJson('api/v3/posts/' . $post->id . '/tag', [
```
chloealee: for some reason, this is throwing an unauthorized 401 error and not sure why?

tests/Http/Three/TagsTest.php
```diff
@@ -0,0 +1,168 @@
...
+        $post = factory(Post::class)->create();
+        $post->tag('Good Photo');
+        $post->tag('Tag To Delete');
+
+        // Make sure both tags actually exist
+        $this->assertContains('Good Photo', $post->tagNames());
+        $this->assertContains('Tag To Delete', $post->tagNames());
+
+        // Send request to remove "Tag To Delete" tag
+        $response = $this->withAdminAccessToken()->deleteJson('api/v3/posts/' . $post->id . '/tag', [
```
chloealee: same with this - throwing an unauthorized 401 error?

tests/Http/Three/TagsTest.php
chloealee: @sbsmith86 I get a very looong response but seems ok? Towards the end I am seeing:
```
  #serverVariables: array:1 [
    "HTTP_AUTHORIZATION" => "Bearer {token}"
  ]
```
I also tried something like this but didn't work:
```    
$northstarId = $this->faker->northstar_id;
$response = $this->withAccessToken($northstarId, 'admin')->json('DELET', 'api/v3/posts/' . $post->id . '/tag', [
            'tag_name' => 'Good Photo',
        ]);
```

sbsmith86: hmmm what is the actual error message you are getting?

chloealee:
```
1) Tests\Http\Three\TagsTest::testDeleteTagOnAPost
Expected status code 200 but received 401.
Failed asserting that false is true.

/home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Foundation/Testing/TestResponse.php:78
/home/vagrant/DoSomething.org/rogue/tests/Http/Three/TagsTest.php:74

2) Tests\Http\Three\TagsTest::testAddMultipleTagsAndDeleteOne
Expected status code 200 but received 401.
Failed asserting that false is true.

/home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Foundation/Testing/TestResponse.php:78
/home/vagrant/DoSomething.org/rogue/tests/Http/Three/TagsTest.php:122
```

I think it may have something to do with the `$this->middleware('auth:api');` because when I comment that out, I'm getting different errors. Digging more into this!

sbsmith86: yea it is something in the middleware, i think also! but it seems like this is set up similar to other routes so not sure why these tests are failing in particular. I'll punt this one to @DFurnes for authorization expertise!

DFurnes: This is so strange!! Going to pull this branch down locally and dig in.

chloealee: thanks, @DFurnes !!

chloealee:
@DoSomething/team-bleed another q - when we untag something, do we just want to delete it from the `post_tag` table or do we want to add soft deletes and a `deleted_at` column?

sbsmith86:
@chloealee My vote is to just detach the relationship https://laravel.com/docs/5.5/eloquent-relationships#updating-many-to-many-relationships

chloealee:
perfect, we're already doing that! Thanks, @sbsmith86 !

https://github.com/DoSomething/rogue/blob/master/app/Models/Post.php#L187

chloealee:
hmmm something interesting @DFurnes - when I comment out
```
        // $post->tag('Good Photo');
        // $this->assertContains('Good Photo', $post->tagNames());
```
in `testDeleteTagOnAPost()`, it passes 🤔

chloealee:
if I tag it via the api first and then untag, the test passes. I don't think this is ideal though...
```
    public function testDeleteTagOnAPostv3()
    {

        // // Apply the tag to the post
         $this->withAdminAccessToken()->postJson('api/v3/posts/' . $post->id . '/tag', [
             'tag_name' => 'Good Photo',
         ]);

        $response = $this->withAdminAccessToken()->deleteJson('api/v3/posts/' . $post->id . '/tag', [
            'tag_name' => 'Good Photo',
        ]);

        // Make sure that the tag is deleted.
        $response->assertStatus(200);
        $this->assertEmpty($post->tagNames());

        // @TODO: Make sure we created a event for the tag once events are refactored.
    }
```

chloealee:
ah wait - if I tag via the api first, and then use the `DELETE /api/v3/posts/:post_id/tag` endpoint to untag, it goes through, but doesn't _actually_ remove the tag? BUT when testing this manually everything works!

DFurnes:
That's so weird!! I didn't get a chance to dig into this as much as I wanted to the other day 😅 but I did notice that it works fine when I make the same requests via Paw so there must be _something_ getting borked in the test config... let me swing by and we can try to figure it out together!

chloealee:
yes that is what I'm thinking too (which I guess is good that the endpoints are working as expected!) Thanks, Dave! :)

chloealee:
@DFurnes figured out the issue! 

```
Gateway keeps the "Token" from this PHPUnit call for later, and so we always think requests are anonymous. 
```

sbsmith86:
sorry only one more question to me: 

The endpoint is `POST /v3/posts/:post_id/tags`? where in the endpoint do you specify the actual tag you are posting?

chloealee:
@sbsmith86 great q! It comes in part of the request body as `tag_name => 'Good Photo'`

sbsmith86:
@chloealee ohhh yes yes i see that in the documentation.

PR #603, opened 2018-01-31
Push data to Quasar

katiecrane:
#### What's this PR do?
Wrote 2 new jobs to handle making the actual request to Blink which then goes to Quasar:
- `SendPostToQuasar.php`
- `SendSignupToQuasar.php`

Each of these jobs calls `toQuasarPayload()` (new method on `Signup` and `Post` to format the data for Quasar, similar to what we do for the regular Blink pushes) and uses the new `Blink` service (similar to the `Phoenix` service we use) to send the request to the `events/quasar-relay` endpoint in Blink.

The jobs are triggered in the corresponding `Post/SignupService`, both the originals and the one in the `Three` folder. 

A new `.env` variable has been added called `DS_ENABLE_PUSH_TO_QUASAR` so that we can have all of this behind a feature flag.

QUESTION:
In the original `PostService` the model we are working with is called `$postOrSignup` but I can't figure out any scenarios where it is a signup. I'm checking on this because if it IS a signup, I don't call the job and we don't call the existing Blink job either, and I want to make sure that we aren't missing any updates! Could this ever be a signup?

REQUEST FORMAT:
Data is sent pretty much the same way data is returned from the `/activity` endpoint, but with the recently added fields on the `post` table included.

Post
```
'id' => 543,
'signup_id' => 453,
'northstar_id' => 543jk534,
'type' => 'photo',
'action' => 'default',
'quantity' => 53,
'media' => [
        'url' => https://s3.amazonaws.com/ds-rogue-prod/uploads/reportback-items/689.jpeg,
        'caption' => 'The stuff I did',
],
'tags' => 'Good Photo',
'status' => 'pending',
'source' => 'phoenix-next',
'remote_addr' => '35.456.532',
'details' => '{stuff: thing, blah: blah}',
'created_at' => as Iso8601String(),
'updated_at' => as Iso8601String(),
'meta' => [
        'message_source' => 'rogue',
],
```
Signup
```
'signup_id' => 345,
'northstar_id' => 453iuh435hiu54,
'campaign_id' => 897,
'campaign_run_id' => 534,
'quantity' => 4,
'why_participated' => 'I wanted to help',
'signup_source' => 'a-place',
'details' => 'affiliate-opt-out',
'created_at' => as Iso8601String(),
'updated_at' => as Iso8601String(),
'meta' => [
    'message_source' => 'rogue',
],
```

Is this how we want this setup? How I tested this locally was to see if the payload was printing out right, but I can't actually send it to Blink. Do we think we need to do more testing around the actual sending?

No more scraping! 🤕 

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/153611842)

#### Checklist
- [ ] Tested on staging.


chloealee:
@katiecrane love the branch name :)

app/Models/Post.php
```diff
@@ -177,6 +177,44 @@ public function toBlinkPayload()
...
+            'id' => $this->id,
+            'signup_id' => $this->signup_id,
+            'northstar_id' => $this->northstar_id,
+            'type' => $this->type,
+            'action' => $this->action,
+            'quantity' => $this->quantity,
+            // Add cache-busting query string to urls to make sure we get the
+            // most recent version of the image.
+            // @NOTE - Remove if we get rid of rotation.
+            'media' => [
```
sbsmith86: have we always been giving them these two urls? I feel like we should only need to sent them the `original_url` since they don't do any work displaying the images and shouldn't care about edited vs not edited versions.

app/Models/Post.php
```diff
@@ -177,6 +177,44 @@ public function toBlinkPayload()
...
+            'tags' => $this->tagSlugs(),
+            'reactions' => [
+                'total' => isset($this->reactions_count) ? $this->reactions_count : null,
+            ],
+            'status' => $this->status,
+            'source' => $this->source,
+            'remote_addr' => $this->remote_addr,
+            'details' => $this->details,
+            'created_at' => $this->created_at->toIso8601String(),
+            'updated_at' => $this->updated_at->toIso8601String(),
```
chloealee: would there ever be a need for us to send `deleted_at`? Not sure how they count signups/posts that have been deleted?

app/Services/Blink.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Rogue\Services;
+
+use DoSomething\Gateway\Common\RestApiClient;
+
+class Blink extends RestApiClient
```
sbsmith86: Could we extend the `DoSomething\Gateway\Blink` class instead of creating a new one?

app/Models/Signup.php
```diff
@@ -131,6 +131,32 @@ public function toBlinkPayload()
...
+            'signup_id' => $this->id,
+            'northstar_id' => $this->northstar_id,
+            'campaign_id' => $this->campaign_id,
+            'campaign_run_id' => $this->campaign_run_id,
+            'quantity' => $this->getQuantity(),
+            'why_participated' => $this->why_participated,
+            'signup_source' => $this->source,
+            'details' => $this->details,
+            'created_at' => $this->created_at->toIso8601String(),
+            'updated_at' => $this->updated_at->toIso8601String(),
```
chloealee: same thing here with `deleted_at`?

app/Services/PostService.php
```diff
@@ -45,6 +46,13 @@ public function create($data, $signupId)
             SendPostToBlink::dispatch($post);
         }
 
+        info('post created');
```
sbsmith86: We already log this below on line 56.

chloealee:
I think we also want to send push updates to Quasar when a signup is updated [here](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Three/SignupsController.php#L126)!

app/Services/PostService.php
```diff
@@ -73,6 +81,12 @@ public function update($signup, $data)
             info('post_created', ['id' => $postOrSignup->id, 'signup_id' => $postOrSignup->signup_id]);
         }
 
+        // Dispatch job to send signup to Quasar
+        // is this ever a signup?
+        if (config('features.pushToQuasar') && $postOrSignup instanceof Post) {
+            SendPostToQuasar::dispatch($postOrSignup);
+        }
+
```
chloealee: @katiecrane I think we would want to do this when the signup's `why_participated` (and `quantity` before we switch) is updated?

sbsmith86:
> In the original PostService the model we are working with is called $postOrSignup but I can't figure out any scenarios where it is a signup. I'm checking on this because if it IS a signup, I don't call the job and we don't call the existing Blink job either, and I want to make sure that we aren't missing any updates! Could this ever be a signup?

The issue here is from this [`update` method](https://github.com/DoSomething/rogue/blob/master/app/Repositories/PostRepository.php#L146-L195) in the `PostRepostitory` where we were deciding whether the update should just be on the signup or if it should be a new post. I believe this logic is funky because in the PA world, we might get a new RB but it was just an update to `why_participated` (which in rogue is a signup update). 

We have a ticket to refactor this, but for now I think we need to be safe and account for it being a post or a signup.

sbsmith86:
> Do we think we need to do more testing around the actual sending?

Let's follow up with Rafa on this. I have some questions about it as well, but we should be using the `blink-staging` url for sending test requests. I will follow up with him re: making sure stuff got into the queue successfully

app/Models/Post.php
katiecrane: @sbsmith86 yeah, this is what the activity endpoint returns for a post too.

@sheyd do you use both of these values? `url` and `original_image_url`?

app/Models/Post.php
katiecrane: @chloealee ooooo great question! currently I think deleting a post would update the timestamp on the signup and then Quasar re-ingests that and sees that the deleted post is no longer there. I guess I need to make sure that deleting a post goes through the PostService?

app/Services/Blink.php
katiecrane: That is a great idea!

app/Services/PostService.php
katiecrane: whoops this was for testing!

app/Services/PostService.php
katiecrane: I'll add a check to see if it's a Signup and if so, trigger the Signup job 👍

app/Models/Post.php
sheyd: @katiecrane I currently only ingest the ['media']['url'] part, so I'm guessing that's the `url`. The reason we wanted the URI on Quasar side was for the Campaigns team to be able to directly click on a reportback from Looker and see the media asset. Whatever URI enables this would be preferred.

katiecrane: @sheyd either could do that, but the one you currently use is the re-sized square crop one that we display elsewhere so let's stick with that one unless the admins would need the fullsize original image?

sheyd: Re-sized square crop should be 💯

app/Models/Post.php
chloealee: @katiecrane oh that's right!

app/Models/Post.php
```diff
@@ -177,6 +177,43 @@ public function toBlinkPayload()
...
+            // Add cache-busting query string to urls to make sure we get the
+            // most recent version of the image.
+            // @NOTE - Remove if we get rid of rotation.
+            'media' => [
+                'url' => $this->getMediaUrl(),
+                'caption' => $this->caption,
+            ],
+            'tags' => $this->tagSlugs(),
+            'reactions' => [
+                'total' => isset($this->reactions_count) ? $this->reactions_count : null,
```
DFurnes: Let's remove this `reactions` property, since this `reactions_count` won't be set here. (And can be better inferred with a join or whatnot in the warehouse itself)>

app/Services/Blink.php
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Send a POST request to be sent to Quasar.
+     *
+     * @param  array $params
+     * @return array|null
+     */
+    public function sendToQuasar($request)
+    {
+        return $this->post('events/quasar-relay', $request);
```
DFurnes: Rather than creating a new `Blink` class, can we just call this directly in `SendPostToQuasar` and co.?

```php
    /**
     * Execute the job.
     *
     * @return void
     */
    public function handle(Blink $blink)
    {
        // Format the payload & send to Quasar.
        $payload = $this->post->toQuasarPayload();
        gateway('blink')->post('events/quasar-relay', $payload);
```

We can then always refactor this later to add the real method in Gateway.

app/Services/Three/PostService.php
```diff
@@ -46,6 +47,10 @@ public function create($data, $signupId)
             SendPostToBlink::dispatch($post);
         }
 
+        if (config('features.pushToQuasar')) {
+            SendPostToQuasar::dispatch($postOrSignup);
```
DFurnes: Is there a `$postOrSignup` here?

app/Services/Blink.php
katiecrane: Oh sure! @sbsmith86 what do you think about this instead of extending `Blink.php`? I'm into this because it's one line and we'll have less stuff to change should this method end up in Gateway.

app/Services/Three/PostService.php
katiecrane: good catch, it is just `$post` here!

app/Services/Blink.php
sbsmith86: sounds good

katiecrane: I forgot that we could even do this but `Blink.php` sure does extend `RestApiClient`

app/Services/PostService.php
```diff
@@ -105,6 +104,13 @@ public function destroy($postId)
             'id' => $postId,
         ]);
 
-        return $this->repository->destroy($postId);
+        $trashed = $this->repository->destroy($postId);
+
+        // Dispatch job to send post to Quasar
+        if (config('features.pushToQuasar')) {
+            SendPostToQuasar::dispatch(Post::withTrashed()->find($postId));
```
DFurnes: Is this how we want to handle situations like this, and then Quasar differentiates based on `deleted_at`? I'm tempted to think it's a bit cleaner to send them a separate delete event saying "we deleted a X with this ID".

cc: @sheyd

app/Models/Post.php
DFurnes: Oh seeing this now! 🆙

I'm a little concerned we're "leaking" implementation details by sending along the `deleted_at` column rather than a distinct "deleted" event. Let's double-check with Team Storm to make sure we're all on the same page.

app/Services/PostService.php
DFurnes: JK JK, see the existing discussion on this [above](https://github.com/DoSomething/rogue/pull/603/files/c4611d85313825e08f8ba6964de99f4692233c3c..d42b1d2a7db33f9f2b9ada9e01d46e8c36760288#r165102243).

katiecrane: @DFurnes This is what I talked to Sena about but I also didn't think of that! @sheyd if Dave's suggestion sounds better then we can certainly do that!

app/Models/Post.php
katiecrane: Whatever works best for Storm is cool with me!

app/Services/PostService.php
sheyd: @DFurnes I've been coding the consumer on the assumption that if a `deleted_at` date is present, we'll handle the record updates on the Quasar side. I'm happy to accommodate whatever changes you think make sense here, but at the moment, I'm fine with sticking with what we have here. 

On the Quasar side, what I'm checking on per event is if `deleted_at` is NULL. If not, then it triggers a record delete function in the Quasar data warehouse (we don't actually delete the record, just mark status as `deleted`). This entails a lookup per Rogue event for an actual value, which I'm doing for the Rogue queue routing code. Key lookups generally are more efficient, so if something like `[data][delete_post]` exists vs checking for value, I'm happy to go with that, but I'm happy to live with the current performance hit of per event value check and throw more hardware at this, since we're using so little CPU in general, if it saves Rogue code increasing in complexity.

DFurnes: Just talked this over with Sena IRL – let's keep things in Quasar & Rogue working exactly as we've got 'em, but just send this payload for a "deleted" event (rather than re-loading the full soft-deleted record from the database):

```php
'data' => [
  'id' => $postId,
  'deleted_at' => Carbon::now(),
],
'meta' => [
  'message_source' => 'rogue',
]
```

We can always re-visit if there's a better general "contract" for this for a v1.1.

katiecrane: @DFurnes Okay! It's okay that the `deleted_at` timestamp we send here may not match the `deleted_at` on the post in Rogue?

sheyd:
@katiecrane Just talked with @DFurnes about the difference between the pull from API vs push to Quasar logic I've coded so far to parse out signups and posts. Going to the push model allows for a simpler logic parse on my end. Currently, with pull model, I iterate through each sign-up per page of activity results, and parse through each post, if they exist. Since with the push model, each event will be separately sent to Blink, i.e. one message will have a sign-up, one message will have a post, etc.  can we add a `type` field to the `meta` section of the message, that would contain fields like `type: signup` and `type:post`, so I can extract data based on that, rather than trying to see if certain keys exist per message?

app/Services/PostService.php
sheyd: I've updated Rogue consumer code on Quasar side to account for these changes.

DFurnes: > Okay! It's okay that the `deleted_at` timestamp we send here may not match the `deleted_at` on the post in Rogue?

@katiecrane Yeah, I think so. It should be pretty darn close.

app/Models/Post.php
```diff
@@ -180,6 +180,42 @@ public function toBlinkPayload()
...
+                'caption' => $this->caption,
+            ],
+            'tags' => $this->tagSlugs(),
+            'status' => $this->status,
+            'source' => $this->source,
+            'remote_addr' => $this->remote_addr,
+            'details' => $this->details,
+            'created_at' => $this->created_at->toIso8601String(),
+            'updated_at' => $this->updated_at->toIso8601String(),
+            'deleted_at' => $this->updated_at->toIso8601String(),
```
chloealee: @katiecrane sorry - still a little confused by this! Can you explain why you did `updated_at` value here?

katiecrane: @chloealee oops that's a typo!

katiecrane: and we now send a separate job, `SendDeletedPostToQuasar` for deletions so it doesn't need that field at all!

app/Jobs/SendPostToQuasar.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        // Format the payload
+        $payload = $this->post->toQuasarPayload($deleted);
```
chloealee: @katiecrane what is the `$deleted` value here?

app/Models/Post.php
chloealee: got it - thanks for explaining that!

app/Jobs/SendPostToQuasar.php
katiecrane: another oops!

app/Services/PostService.php
```diff
@@ -88,6 +104,13 @@ public function destroy($postId)
...
```
DFurnes: This should be `SendDeletedPostToBlink`, right?

katiecrane: @DFurnes it sure should be!

PR #604, opened 2018-02-01
`/v3/signups` updated viewing permissions 

chloealee:
#### What's this PR do?
- Updates `GET /v3/signups` and `GET /v3/signups/{signup}` to only return a user's own signup(s) unless they're an admin so that we're keeping more data protected
- Updates tests and deletes tests that are now irrelevant 
- Updates documentation

👀 

In particular, we want to gate this information as we can see the `why_participated` values containing sensitive user information

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/154611572) Pivotal Card.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -111,13 +111,13 @@ class SignupsController extends ApiController
     public function show(Request $request, Signup $signup)
     {
         // Only allow an admin, staff, or the user who owns the signup to see the signup.
-        $signup = Signup::whereVisible()->first();
-
         if ($signup) {
-            return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
-        }
+            if (is_staff_user() || auth()->id() === $signup->northstar_id) {
+                return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
+            }
 
-        throw new AuthorizationException('You don\'t have the correct role to see this signup!');
+            throw new AuthorizationException('You don\'t have the correct role to see this signup!');
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


chloealee:
@DoSomething/team-bleed I believe the `GET /v3/posts` and `GET /v3/posts/{post}` endpoints return all posts for admins and owners of the post, and accepted posts for non admins and non owners. 

Similar to these changes, I think we should probably only return the posts to admins and the owners. Thoughts?

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -107,12 +110,14 @@ public function index(Request $request)
      */
     public function show(Request $request, Signup $signup)
     {
-        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
-        if ($request->query('include') === 'posts') {
-            $signup = Signup::withVisiblePosts()->first();
+        // Only allow an admin, staff, or the user who owns the signup to see the signup.
+        $signup = Signup::whereVisible()->first();
```
DFurnes: This is building a new database query from the `Signup` model & overwriting the instance we got from the router, so it'll give us the first post by the currently authenticated user. Instead, let's check against the instance we have:

```php
if (is_staff_user() || auth()->id() === $signup->northstar_id) {
  return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
}
```

DFurnes:
I was just thinking about this a little more... I wonder if it's good enough to just wrap `$signup->why_participated` in a conditional in the transformer and continue to return all signups to everyone?

I could see it being useful for non-admin sessions to load other user's signups (for example to get "total signups" on a campaign page, or to show a gallery of posts grouped by signup) as long as we're not giving away anything too personal.

chloealee:
agree with the above! More convo [here](https://dosomething.slack.com/archives/C0YGXUE01/p1517588216000325). I'm going to close this PR and start fresh, including work on hiding `tags`, `source`, and `remote_addr` in the Post Transformer for non-staff.

PR #607, opened 2018-02-02
Add a message that Jen is the only one to use this

sbsmith86:
#### What's this PR do?

Adds a little message to the upload form that Jen is the only one to use this feature. 

image

👁 

Check my grammar...I do not have good grammar and I also am a typo queen 👑 

We are still testing this feature and do not want folks using it in the meantime. 

\

### DIFF
--- a/resources/views/pages/import.blade.php
+++ b/resources/views/pages/import.blade.php
@@ -7,7 +7,16 @@
     <div class="container -padded">
         <div class="wrapper">
             <div class="container__block -narrow">
-                <p>For the time being, This feature should only be used by Jen Ng. If you have any questions please reach out to Jen directly or post a message in the #rogue channel in slack</p>
+                <article class="figure margin-bottom-none -left -small">
+                    <div class="figure__media rounded-figure">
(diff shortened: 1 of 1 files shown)


resources/views/pages/import.blade.php
```diff
@@ -6,6 +6,9 @@
 
     <div class="container -padded">
         <div class="wrapper">
+            <div class="container__block -narrow">
+                <p>For the time being, This feature should only be used by Jen Ng. If you have any questions please reach out to Jen directly or post a message in the #rogue channel in slack</p>
+            </div>
```
katiecrane: Should there be a picture of Jen here?

DFurnes: "Is this a picture of you? IF NOT, KEEP OUT."

ngjo: Hahahaha all of these are funny things and I fully support. 

But for the text that's there right now, lowercase `this` and a period after slack!

sbsmith86:
you are all correct. Updated the page: 

image

PR #609, opened 2018-02-02
Hit PostService when updating or deleting posts through v3 endpoints

katiecrane:
#### What's this PR do?
This touches `v3/posts` `delete` and `patch`. Documentation is updated to reflect changes.

Delete:
- Call the existing `destroy` method in `PostService`
- No changes to `PostService` or `PostRepository`

Patch:
More changes here!
- Do not accept `status` because those changes will come in through `v3/reviews`
- Hit the `update` method of `PostService`
- In `PostService`, assume that we are always working with a Post and never a Signup
- In `PostRepository`, same deal, only update the Post and don't mess with Signups

Does this contain any breaking changes? See a better way to do anything?

We need to always do alterations through the service classes to be sure that any business logic happens.

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/154877666)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/Three/PostRepository.php
+++ b/app/Repositories/Three/PostRepository.php
@@ -139,7 +139,6 @@ class PostRepository
      */
     public function destroy($postId)
     {
-        info('destroy - PostRepository');
         $post = Post::findOrFail($postId);
 
         // Delete the image file from AWS.
--- a/app/Repositories/Three/PostRepository.php
+++ b/app/Repositories/Three/PostRepository.php
@@ -150,6 +150,7 @@ class PostRepository
      */
     public function destroy($postId)
     {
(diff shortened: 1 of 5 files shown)


app/Repositories/Three/PostRepository.php
```diff
@@ -117,29 +117,18 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
...
+     * @param \Rogue\Models\Post $post
      * @param array $data
      *
-     * @return Signup|Post
+     * @return Post
      */
-    public function update($signup, $data)
+    public function update($post, $data)
     {
-        // Should only update why_participated on the signup.
```
chloealee: 🔪

PR #611, opened 2018-02-05
POST /api/v3/posts - Update to store new post columns

sbsmith86:
#### What's this PR do?
This updates the `POST /api/v3/posts` API endpoint to validate, store, and return the new `type`, `action`, and `details` columns that we added to support multiple actions. 

Updates the `Three\PostTransformer` to return these new columns, but only returns them for admins and post-owners. *Thoughts on gating all of these columns?*

Updates V3 PostAPITest. 

👀 

Going to go through each endpoint individually to see how it could be updated to support these new columns. This is just the first one but I think updating `PostTransformer` gets us most of the way there. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/154154145

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -106,20 +106,13 @@ class PostsController extends ApiController
 
         $signup = $this->signups->get($northstarId, $request['campaign_id'], $request['campaign_run_id']);
 
-        $signupExists = ! is_null($signup);
-
-        if (! $signupExists) {
+        if (! $signup) {
             $signup = $this->signups->create($request->all(), $northstarId);
-            $post = $this->posts->create($request->all(), $signup->id);
-        } else {
-            $post = $this->posts->create($request->all(), $signup->id);
         }
 
-        // @TODO - We should probably only ever return a 201 here since all
-        // this does is create a resource.
-        $code = $signupExists ? 200 : 201;
+        $post = $this->posts->create($request->all(), $signup->id);
 
-        return $this->item($post, $code, [], null, 'signup');
+        return $this->item($post, 201, [], null, 'signup');
     }
 
     /**
--- a/app/Http/Controllers/Three/PostsController.php
+++ b/app/Http/Controllers/Three/PostsController.php
@@ -115,8 +115,8 @@ class PostsController extends ApiController
             $post = $this->posts->create($request->all(), $signup->id);
         }
 
-        // @TODO - We should probably only ever return a 201 here since all this does is
-        // create a resource.
+        // @TODO - We should probably only ever return a 201 here since all
+        // this does is create a resource.
         $code = $signupExists ? 200 : 201;
 
         return $this->item($post, $code, [], null, 'signup');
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Three/PostsController.php
```diff
@@ -106,16 +106,18 @@ public function store(PostRequest $request)
 
         $signup = $this->signups->get($northstarId, $request['campaign_id'], $request['campaign_run_id']);
 
-        $updating = ! is_null($signup);
+        $signupExists = ! is_null($signup);
```
DFurnes: This is much clearer! 👍

app/Http/Controllers/Three/PostsController.php
```diff
@@ -106,16 +106,18 @@ public function store(PostRequest $request)
...
 
-        if (! $updating) {
+        if (! $signupExists) {
             $signup = $this->signups->create($request->all(), $northstarId);
             $post = $this->posts->create($request->all(), $signup->id);
```
DFurnes: Can we simplify this conditional since the `$this->posts->create` is shared between both cases?

```php
$signup = $this->signups->get($northstarId, $request['campaign_id'], $request['campaign_run_id']);

if (! $signup) {
    $signup = $this->signups->create($request->all(), $northstarId);
}

$post = $this->posts->create($request->all(), $signup->id);
```

app/Http/Controllers/Three/PostsController.php
```diff
@@ -106,16 +106,18 @@ public function store(PostRequest $request)
...
         } else {
             $post = $this->posts->create($request->all(), $signup->id);
         }
 
-        $code = $updating ? 200 : 201;
+        // @TODO - We should probably only ever return a 201 here since all
+        // this does is create a resource.
+        $code = $signupExists ? 200 : 201;
```
DFurnes: Agreed, this is super weird! Can we just start doing this now?

app/Http/Requests/Three/PostRequest.php
```diff
@@ -27,11 +27,14 @@ public function rules()
             'campaign_id' => 'required',
             'campaign_run_id' => 'required_with:campaign_id|integer',
             'northstar_id' => 'nullable|objectid',
+            'type' => 'required|string',
```
DFurnes: Do we want to validate what types are set here? `in:photo,voterreg`

app/Http/Requests/Three/PostRequest.php
```diff
@@ -27,11 +27,14 @@ public function rules()
...
+            'action' => 'required|string',
```
DFurnes: Do we want to also validate format a bit here with [`alpha_dash`](https://laravel.com/docs/5.5/validation#rule-alpha-dash)?

app/Http/Controllers/Three/PostsController.php
sbsmith86: ah yes. nice!

app/Http/Requests/Three/PostRequest.php
sbsmith86: @DFurnes I was wondering that as well... I can add it! We will just have to keep an eye on new types coming in and update this validation accordingly.

app/Http/Requests/Three/PostRequest.php
sbsmith86: Sure! I did not know about that rule. Very nice.

PR #613, opened 2018-02-06
Add additional fields to post Quasar payload

katiecrane:
#### What's this PR do?
1. Adds `campaign_id` (which was supposed to be there all along!)
2. Added some signup fields that were needed to ingest posts which were previously retrieved from the signup portion of the `/activity` response.

1. Is the signup reloaded each time we call `$this->signup`? I want to make sure this isn't too bad performance wise.
2. The signup fields were already sent with the signup, but this is how data currently expects it and it is a large lift on their side to alter that, while on our side we just need to add 5 easy lines.
3. From @sheyd : Next sprint, Team Storm will discuss changing Quasar data model to not necessitate sending signup data on posts anymore (as part of MySQL -> Postgres).

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/154725135)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -190,10 +190,13 @@ class Post extends Model
         return [
             'id' => $this->id,
             'signup_id' => $this->signup_id,
+            'campaign_id' => $this->campaign_id,
+            'campaign_run_id' => $this->signup->campaign_run_id,
             'northstar_id' => $this->northstar_id,
             'type' => $this->type,
             'action' => $this->action,
             'quantity' => $this->quantity,
+            'why_participated' => $this->signup->why_participated,
             // Add cache-busting query string to urls to make sure we get the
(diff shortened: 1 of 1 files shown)


DFurnes:
> Is the signup reloaded each time we call $this->signup? I want to make sure this isn't too bad performance wise.

I believe Eloquent caches relationships in-memory after the first load, but worth double-checking!

> From @sheyd : Next sprint, Team Storm will discuss changing Quasar data model to not necessitate sending signup data on posts anymore (as part of MySQL -> Postgres).

👍 👍

katiecrane:
> I believe Eloquent caches relationships in-memory after the first load, but worth double-checking!

@DFurnes I have found other folks saying this as well, but I can't find it in the Laravel docs anywhere.

PR #614, opened 2018-02-06
Adds orderBy filter for v3 Signup Index and tests & updates logic to save summed quantity on signup when post is created or updated

chloealee:
#### What's this PR do?
- Adds the `orderByQuantity` filter for `GET /v3/signups` 
  - Choose to return results with quantity in ascending or descending order
- Adds logic to update and save summed quantity on the signup when a post is created or updated via the v3 endpoints
- Adds tests for:
  - `?filter[northstar_id]`, `?filter[campaign_id]`, and `?filter[campaign_run_id]`
  - `?orderByQuantity=desc` and `?orderbyQuantity=asc`
  - updates PostTests to check for summing of quantity on signup
- Updates documentation 

- All tests should pass! 
- All filters should work in Postman/Paw! 
- Did I miss any logic for updating the total quantity on the signup? Does this need to happen anywhere else outside of the `POST /v3/posts` and `PATCH `/v3/posts/{post}` endpoints?

#### Relevant tickets
Fixes the first task in [this](https://www.pivotaltracker.com/n/projects/2019429/stories/154579464) Pivotal Card

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -94,7 +94,7 @@ class SignupsController extends ApiController
             $query = $query->withVisiblePosts();
         }
 
-        $orderBy = $request->query('order_by');
+        $orderBy = $request->query('orderBy');
 
         if ($orderBy) {
             list($column, $direction) = explode(',', $orderBy, 2);
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -94,10 +94,14 @@ class SignupsController extends ApiController
             $query = $query->withVisiblePosts();
         }
 
-        if ($request->query('orderByQuantity') === 'desc') {
-            $query = $query->orderBy('quantity', 'desc');
-        } elseif ($request->query('orderByQuantity') === 'asc') {
-            $query = $query->orderBy('quantity', 'asc');
+        $orderBy = $request->query('order_by');
+
+        if ($orderBy) {
+            list($column, $direction) = explode(',', $orderBy, 2);
+
+            if (in_array($column, Signup::$indexes)) {
+                $query = $query->orderBy($column, $direction);
+            }
         }
 
         return $this->paginatedCollection($query, $request);
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -96,6 +96,8 @@ class SignupsController extends ApiController
 
         if ($request->query('orderByQuantity') === 'desc') {
             $query = $query->orderBy('quantity', 'desc');
+        } elseif ($request->query('orderByQuantity') === 'asc') {
+             $query = $query->orderBy('quantity', 'asc');
         }
 
         return $this->paginatedCollection($query, $request);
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/Three/SignupsController.php
```diff
@@ -94,6 +94,12 @@ public function index(Request $request)
             $query = $query->withVisiblePosts();
         }
 
+        if ($request->query('orderByQuantity') === 'desc') {
```
DFurnes: Would this make more sense as an `order_by[column]` query parameter (instead of a special case for quantity)?

chloealee: ooo I like that idea @DFurnes ! I know we did this before but not sure how - would we use regex or something to get the column out of the filter?

DFurnes: If separating with a comma, you should be able to use `explode`. Something like this:

```php
$orderBy = $request->query('order_by');
if ($orderBy && in_array($orderBy, Signup::$indexes)) {
   list($column, $direction) = explode(',' $orderBy, 2)
}
```

chloealee: brilliant! Thank you!

app/Models/Signup.php
```diff
@@ -41,7 +41,7 @@ class Signup extends Model
      * @var array
      */
     public static $indexes = [
-        'campaign_id', 'campaign_run_id', 'updated_at', 'northstar_id', 'id',
+        'campaign_id', 'campaign_run_id', 'updated_at', 'northstar_id', 'id', 'quantity',
```
DFurnes: Oh, hmm! Two questions here ––

1. Does this column exist on the signup anymore? I think we [decided](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/10) it wouldn't.
2. If it does, we need it to be indexed!

chloealee: oh @DFurnes you're so right! I totally blanked on that! 

This column still exists in the database but when we switch to saving quantity on the post, we will no longer be storing the quantity in this column. 

This is what you were talking about re: the heavy query to order these by. Should we discuss this in the discussions you linked if this is the time we want to store this?

DFurnes: I think this definitely makes sense to discuss as a team! 👍

chloealee: for documentation: we decided to store the total quantity on the signup! https://github.com/orgs/DoSomething/teams/team-bleed/discussions/10?from_comment=13#discussion-10-comment-13

documentation/endpoints/v3/signups.md
```diff
@@ -49,10 +49,25 @@ GET /api/v3/signups
 Only admins and signup owners will have `why_participated`, `source`, and `details` returned in the response.
 
 When using `?include=posts`, anonymous requests will only return accepted posts. Logged-in users can see accepted posts & any of their own pending or rejected posts. Staff can see anything!
+
 ### Optional Query Parameters
 - **include** _(string)_
   - Include additional related records in the response: `posts`
   - e.g. `/v3/signups?include=posts`
+- **filter[column]** _(string)_
```
sbsmith86: 🥇

PR #615, opened 2018-02-08
Use Review & Signup Service & Repository for Patch and Delete and send that to Quasar

katiecrane:
#### What's this PR do?
1. `PATCH v3/signups` now calls new `update` method in `SignupService`, which in turn calls new `update` method in `SignupRepository`. This mirrors `PATCH v3/posts`.
    - `SignupService` also triggers the `SendSignupToQuasar` job.
2. `DELETE v3/signups` now calls new `destroy` method in `SignupService`, which in turn calls new `destroy` method in `SignupRepository`. This mirrors `PATCH v3/posts`.
    - `SignupService` also triggers the new `SendDeletedSignupToQuasar` job.

Do both of these flows make sense?

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/154968896)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -149,8 +149,12 @@ class SignupsController extends ApiController
      */
     public function destroy(Signup $signup)
     {
-        $this->signups->destroy($signup->id);
+        $trashed = $this->signups->destroy($signup->id);
 
-        return $this->respond('Signup deleted.', 200);
+        if ($trashed) {
+            return $this->respond('Signup deleted.', 200);
+        }
+
+        return response()->json(['code' => 500, 'message' => 'There was an error deleting the post']);
     }
 }
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -125,15 +125,14 @@ class SignupsController extends ApiController
      */
     public function update(Request $request, Signup $signup)
     {
-        $this->validate($request, [
+        $validatedRequest = $this->validate($request, [
             'why_participated' => 'required',
         ]);
 
         // Only allow an admin or the user who owns the signup to update.
         if (token()->role() === 'admin' || auth()->id() === $signup->northstar_id) {
-            $signup->update(
-                $request->only('why_participated')
-            );
+            // why_participated is the only thing that can be changed
+            $this->signups->update($signup, $validatedRequest);
 
             return $this->item($signup);
         }
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -149,7 +149,7 @@ class SignupsController extends ApiController
      */
     public function destroy(Signup $signup)
     {
-        $signup->delete();
+        $this->signups->destroy($signup->id);
 
         return $this->respond('Signup deleted.', 200);
     }
--- a/app/Repositories/Three/SignupRepository.php
+++ b/app/Repositories/Three/SignupRepository.php
@@ -30,6 +30,20 @@ class SignupRepository
     }
 
     /**
+     * Update a signup.
+     *
+     * @param array $data
+     * @param array $data
+     * @return \Rogue\Models\Signup
+     */
+    public function update($signup, $data)
+    {
+        $signup->update($data);
+
+        return $signup;
+    }
+
+    /**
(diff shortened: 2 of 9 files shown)


app/Http/Controllers/Three/SignupsController.php
```diff
@@ -150,7 +149,7 @@ public function update(Request $request, Signup $signup)
      */
     public function destroy(Signup $signup)
     {
-        $signup->delete();
+        $this->signups->destroy($signup->id);
 
         return $this->respond('Signup deleted.', 200);
```
katiecrane: Should this only be returned if `$this->signups->destroy($signup->id)` is `true`?

app/Services/Three/SignupService.php
```diff
@@ -56,6 +57,50 @@ public function create($data, $northstarId)
...
+
+    /**
+     * Handle all business logic around deleting a signup.
+     *
+     * @param int $signupId
+     * @return bool
+     */
+    public function destroy($signupId)
+    {
+        info('signup_deleted', [
```
sbsmith86: Should probably log this after the delete is successful.

app/Http/Controllers/Three/SignupsController.php
sbsmith86: I think so, just for safety, else return an error.

katiecrane: @sbsmith86 I'm not sure what kind of exception to throw in this case, what do you think?

sbsmith86: we could just throw a `500` since if this method fails something weird is going on

app/Repositories/Three/SignupRepository.php
```diff
@@ -47,4 +61,20 @@ public function get($northstarId, $campaignId, $campaignRunId)
...
+     * Delete a signup.
+     *
+     * @param int $signupId
+     * @return $post;
+     */
+    public function destroy($signupId)
+    {
+        $signup = Signup::findOrFail($signupId);
+
+        // Soft delete the post.
```
sbsmith86: > // Soft delete the post.

should be 

> //// Soft delete the signup. 

😬

app/Services/Three/PostService.php
```diff
@@ -87,6 +87,31 @@ public function update($post, $data)
         return $post;
     }
 
+    /**
+     * Handles all business logic around reviewing posts.
+     *
+     * @param \Rogue\Models\Signup $signup
```
sbsmith86: This should be switched to accepting a post instance

app/Services/Three/PostService.php
```diff
@@ -87,6 +87,31 @@ public function update($post, $data)
...
+     * @param array $data
+     * @return \Rogue\Models\Post|\Rogue\Models\Signup
```
sbsmith86: and this as well.

app/Services/Three/PostService.php
```diff
@@ -87,6 +87,31 @@ public function update($post, $data)
...
+     */
+    public function review($post, $data)
```
sbsmith86: what is the difference between `$post` and `$data` here? in the other `PostService` we only pass `$data`

app/Services/Three/SignupService.php
```diff
@@ -56,6 +57,50 @@ public function create($data, $northstarId)
...
     }
 
+    /**
+     * Handles all business logic around updating signups.
+     *
+     * @param array $data
+     * @param string $northstarId
+     * @return Illuminate\Database\Eloquent\Model $model
+     */
+    public function update($signup, $data)
```
sbsmith86: Need to update the comment block to match the actual params

katiecrane: @sbsmith86 thank you for catching all of these!

app/Services/Three/PostService.php
katiecrane: @sbsmith86 hmm this is a good question. This is what the PostRepository asks for (it actually needs `$comment` too which I've added) so I went with that

DFurnes: When possible, I think it's a lot clearer to pass real models and named parameters around, rather than big arrays (where the dependency is not as explicit).

A side benefit is that we can check whether the post exists & use that same instance elsewhere in the request, whereas the `post_id` parameter in the old repository always needed to make a fresh query to load it anew from the DB.

PR #616, opened 2018-02-09
updates migration for quantity to not be unique 

chloealee:
#### What's this PR do?
- Fixes the mistake made in [this PR](https://github.com/DoSomething/rogue/pull/614) where I accidentally made the `quantity` column a unique index instead of just an index. 

👀 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/database/migrations/2018_02_09_150930_add_quantity_index_to_signups_table.php
+++ b/database/migrations/2018_02_09_150930_add_quantity_index_to_signups_table.php
@@ -26,7 +26,7 @@ class AddQuantityIndexToSignupsTable extends Migration
     public function down()
     {
         Schema::table('signups', function (Blueprint $table) {
-            $table->dropIndex('quantity');
+            $table->dropIndex(['quantity']);
         });
     }
 }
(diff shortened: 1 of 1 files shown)


database/migrations/2018_02_09_150930_add_quantity_index_to_signups_table.php
```diff
@@ -26,7 +26,7 @@ public function up()
     public function down()
     {
         Schema::table('signups', function (Blueprint $table) {
-            $table->dropUnique('quantity');
+            $table->dropIndex('quantity');
```
sbsmith86: To drop an index. I believe you either have to do `$table->dropIndex(['quantity']);` (by column name) or `$table->dropIndex('signups_quantity_index');` (by index name). I would make sure the rollback works along with the migration. 

https://laravel.com/docs/5.5/migrations#dropping-indexes

chloealee: ah yes thank you @sbsmith86 !

PR #617, opened 2018-02-09
Send updates through PostService

sbsmith86:
#### What's this PR do?
Updated the import CSV logic to run status updates through the `PostService@update` method so that all business logic is executed with the post update. That means that c.io should get the updated post status on their end. 

👀 

Found this bug with @rapala61 and @ngjo . We did a test CSV import to rogue-thor with registration records from Rafa. We saw that while the initial signups and posts got into c.io successfully, one of the posts had an updated status that was not being sent to c.io. 

This fixes it so that post updates during an import will be sent to blink successfully.

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/155090650

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/Three/PostRepository.php
+++ b/app/Repositories/Three/PostRepository.php
@@ -133,7 +133,7 @@ class PostRepository
         $post->update($data);
 
         // If the quantity was updated, update the total quantity on the signup.
-        if ($data['quantity']) {
+        if (isset($data['quantity'])) {
             $signup = Signup::find($post->signup_id);
             $signup->quantity = $signup->posts->sum('quantity');
             $signup->save();
--- a/app/Services/Three/PostService.php
+++ b/app/Services/Three/PostService.php
@@ -62,9 +62,9 @@ class PostService
     /**
      * Handles all business logic around updating posts.
      *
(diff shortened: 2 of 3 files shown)


app/Repositories/Three/PostRepository.php
```diff
@@ -133,7 +133,7 @@ public function update($post, $data)
         $post->update($data);
 
         // If the quantity was updated, update the total quantity on the signup.
-        if ($data['quantity']) {
+        if (isset($data['quantity'])) {
```
chloealee: ah great catch!

app/Services/Three/PostService.php
```diff
@@ -81,8 +81,8 @@ public function update($post, $data)
             SendPostToQuasar::dispatch($post);
         }
 
-        // Log that a post was created.
-        info('post_created', ['id' => $post->id, 'signup_id' => $post->signup_id]);
+        // Log that a post was updated.
+        info('post_updated', ['id' => $post->id, 'signup_id' => $post->signup_id]);
```
chloealee: 🕵️‍♀️

PR #618, opened 2018-02-12
Adds laravel-cors package

chloealee:
#### What's this PR do?
- Adds laravel-cors package 

👀 

We weren't able to successfully make requests from Domino to Rogue. When making a request, the following was returned:
```
Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'null' is therefore not allowed access. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```
This package allows us to make successful calls.

#### Relevant tickets
Needed for [this](https://www.pivotaltracker.com/n/projects/2019429/stories/154579464) Pivotal Card. 

#### Checklist
- [ ] Tested on staging.


app/Http/Kernel.php
```diff
@@ -41,6 +41,7 @@ class Kernel extends HttpKernel
             'guard:api',
             'bindings',
             // 'throttle:60,1',
+            'cors' => \Rogue\Http\Middleware\Cors::class,
```
chloealee: also tried just `\Rogue\Http\Middleware\Cors::class,` here too and still throwing this error: 
```
Failed to load http://rogue.test/api/v3/signups?filter[campaign_id]=1283&orderBy=quantity%2Cdesc: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'null' is therefore not allowed access. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

sbsmith86:
@DFurnes We tried adding the header in the middleware and that didn't seem to work. This [package](https://github.com/barryvdh/laravel-cors) was suggested for handling laravel CORS requests. It looks like it add some more headers and handles preflight requests. When @chloealee added the package, it seems to work. I can't tell if this package is doing more than we need or if we should just go with using this. I'm inclined to just use this package and see if we have more CORS like functionality that we need as we move forward.

DFurnes:
Good by me! I'm honestly pretty surprised this isn't built in to the framework yet. 👍

PR #619, opened 2018-02-12
/v3/signups?include=user param 

chloealee:
#### What's this PR do?
- Adds ability to access user info. with `?include=user param ` on v3 signups endpoint.
- Updates documentation.
- Adds tests.

👀 

We need more user info. for the leaderboard! 

#### Relevant tickets
Is needed for [this](https://www.pivotaltracker.com/n/projects/2019429/stories/154579464) Pivotal card so we can list user name in the leaderboard.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Transformers/Three/SignupTransformer.php
+++ b/app/Http/Transformers/Three/SignupTransformer.php
@@ -5,7 +5,6 @@ namespace Rogue\Http\Transformers\Three;
 use Rogue\Models\Signup;
 use Rogue\Services\Registrar;
 use League\Fractal\TransformerAbstract;
-use Rogue\Http\Transformers\UserTransformer;
 
 class SignupTransformer extends TransformerAbstract
 {
@@ -66,11 +65,9 @@ class SignupTransformer extends TransformerAbstract
      */
     public function includeUser(Signup $signup)
     {
-        if (is_staff_user() || auth()->id() === $signup->northstar_id) {
-            $registrar = app(Registrar::class);
-            $northstar_id = $signup->northstar_id;
+        $registrar = app(Registrar::class);
+        $northstar_id = $signup->northstar_id;
 
-            return $this->item($registrar->find($northstar_id), new UserTransformer);
-        }
+        return $this->item($registrar->find($northstar_id), new UserTransformer);
     }
 }
--- a/app/Http/Transformers/Three/SignupTransformer.php
+++ b/app/Http/Transformers/Three/SignupTransformer.php
@@ -5,7 +5,6 @@ namespace Rogue\Http\Transformers\Three;
 use Rogue\Models\Signup;
 use Rogue\Services\Registrar;
 use League\Fractal\TransformerAbstract;
-use Rogue\Http\Transformers\Three\PostTransformer;
 use Rogue\Http\Transformers\UserTransformer;
 
 class SignupTransformer extends TransformerAbstract
--- a/app/Http/Transformers/Three/SignupTransformer.php
+++ b/app/Http/Transformers/Three/SignupTransformer.php
@@ -3,8 +3,10 @@
 namespace Rogue\Http\Transformers\Three;
 
 use Rogue\Models\Signup;
+use Rogue\Services\Registrar;
 use League\Fractal\TransformerAbstract;
-use Rogue\Http\Transformers\PostTransformer;
+use Rogue\Http\Transformers\Three\PostTransformer;
+use Rogue\Http\Transformers\UserTransformer;
 
 class SignupTransformer extends TransformerAbstract
 {
@@ -14,7 +16,7 @@ class SignupTransformer extends TransformerAbstract
      * @var array
      */
     protected $availableIncludes = [
-        'posts',
+        'posts', 'user',
     ];
 
     /**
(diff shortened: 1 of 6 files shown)


app/Http/Transformers/Three/SignupTransformer.php
```diff
@@ -3,8 +3,9 @@
 namespace Rogue\Http\Transformers\Three;
 
 use Rogue\Models\Signup;
+use Rogue\Services\Registrar;
 use League\Fractal\TransformerAbstract;
-use Rogue\Http\Transformers\PostTransformer;
+use Rogue\Http\Transformers\UserTransformer;
```
katiecrane: Looks like `PostTransformer` is still used in `includePosts`

chloealee: @katiecrane I don't think we need that though since it is in the same directory! (Using `\Transformers\Three\PostTransformer` now :))

tests/Http/Three/SignupTest.php
```diff
@@ -282,6 +282,66 @@ public function testSignupIndexWithIncludedPostsAsOwner()
...
+     * GET /api/v3/signups?include=user
+     * @return void
+     */
+    public function testSignupIndexWithIncludedUserAsNonAdminNonOwner()
+    {
+        $post = factory(Post::class)->create();
+        $signup = $post->signup;
+
+        // Test with annoymous user that no posts are returned.
+        $response = $this->getJson('api/v3/signups' . '?include=posts');
```
katiecrane: Should this be `include=user`?

app/Http/Transformers/Three/SignupTransformer.php
sbsmith86: shouldn't we have a `Three/UserTransformer`?

tests/Http/Three/SignupTest.php
chloealee: ah yes! Thanks for catching this! :)

app/Http/Transformers/Three/SignupTransformer.php
katiecrane: oh cool! nice @chloealee !

chloealee: @sbsmith86 we totally should! I initially didn't do a new `UserTransformer` because I thought it'd be exactly the same file, but actually now that I think about this more, I think we should do some logic with who can see what about the user in the new `UserTransformer` instead of in the `includeUser` block in the `SignupTransformer`. 

Is it safe to return a user's first name and hide everything else for non-admins/non-owners of the signup? And then for admins/owners of the signup, they can see everything (`first_name`, `last_name`, `birthdate`, `email`, `mobile`)?

sbsmith86: @chloealee That makes sense to me. Do we display any other user information  (in the gallery, for example) ? Maybe age?

In general we should keep all of the V3 files unique and separate (even if it is a dupe), it will make it easier to archive and re-arrange files later.

DFurnes: I'm not sure how things are hooked up now, but ideally we can forward along the `Authorization` header when asking Northstar for user details & that'll handle showing/hiding fields.

chloealee: ooo interesting @DFurnes ! So when grabbing the user, will we need to change the [`asClient()`](https://github.com/DoSomething/rogue/blob/master/app/Services/Registrar.php#L33) part to grab user info based on `Authorization` header?

DFurnes: Yeah, that potentially interferes with our ability to cache users (since what we cache depends on what a particular user can see) so it might not make sense to run through the Registrar.

DFurnes: Longer-term, this is a great example of where we'd want to use an API gateway (like [the GraphQL prototype](https://graphql.dosomething.org/explore) I worked on last quarter), rather than having Rogue piece together entire responses.

chloealee: ah yes! This totally makes sense. Matt and I were chatting about GraphQL at my last checkin too and how it could be a great use case when we need to pull in a user's college (and college logo) for more Domino work. cc @katiecrane 

@DFurnes for now, is it ok to just do this logic in the new `UserTransformer`?

chloealee:
Added new v3 `UserTransformer` and logic but LMK if this doesn't work!

app/Http/Transformers/Three/SignupTransformer.php
katiecrane: @chloealee I'm confused about how that would help us because the colleges and college logos will live in Domino, unless the suggestion is that those live somewhere else?

chloealee: @katiecrane sorry for the confusion! I might have missed this but where does the college data live (where the user attends college)? We were just briefly chatting that if we'd have to ping NS for this info. for a user's college, GraphQL might be a nice way to just make one call instead of multiple API calls to get this info.

katiecrane: @chloealee ahh yeah that would make sense, but I believe the college info will all just live in Domino

chloealee: @katiecrane ah got it! Makes sense! :)

documentation/endpoints/activity.md
```diff
@@ -17,7 +17,7 @@ GET /api/v2/activity
   - For pagination, specify page of activity to return in the response.
   - e.g. `/activity?page=2`
 - **include** _(string)_
-  - Include additional related records in the response: `user`
+  - Include additional related records in the response: `user`, `posts`
```
sbsmith86: Nitpicky - But can multiple values be pass in this param? If so, can we add an example and specify that to add multiple include params they need to be comma-separated?

chloealee: yes great idea!

PR #621, opened 2018-02-13
If quantity is not stored on post, send Quasar quantity from signup

katiecrane:
#### What's this PR do?
When a user submits a post, we send the updated post to Quasar. However, while the quantity lives on the signup, they don't get that updated quantity! For now, let's send the quantity on the post only if it is not null, otherwise grab the quantity from the signup.

👀 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -195,7 +195,7 @@ class Post extends Model
             'northstar_id' => $this->northstar_id,
             'type' => $this->type,
             'action' => $this->action,
-            'quantity' => $this->quantity ? $this->quantity : $this->signup->quantity,
+            'quantity' => $this->getQuantity(),
             'why_participated' => $this->signup->why_participated,
             // Add cache-busting query string to urls to make sure we get the
             // most recent version of the image.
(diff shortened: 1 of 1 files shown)


app/Models/Post.php
```diff
@@ -195,7 +195,7 @@ public function toQuasarPayload()
             'northstar_id' => $this->northstar_id,
             'type' => $this->type,
             'action' => $this->action,
-            'quantity' => $this->quantity,
+            'quantity' => $this->quantity ? $this->quantity : $this->signup->quantity,
```
sbsmith86: I don't think I understand this issue. Why would we ever want to send the signup quantity (which will be the total quantity across all posts) here?

katiecrane: @sbsmith86 because as of right now we are not updating the quantity column on the post when a post is submitted

katiecrane:
@DFurnes ah we have that on the signup! I figured that since we are trying to turn on quantity on the post really soon that we won't need this check for long

DFurnes:
Yeah, that's totally fair too. 👍

sbsmith86:
@katiecrane should hide this logic under the feature flag like we do in `signup->getQuantity` ?

katiecrane:
@sbsmith86 I was wondering that too! I wasn't sure if it was totally necessary since if that flag is on, then the quantity on the post will not be `null` and we'll still get what we want. Probably safer to do it that way though, I'll bust it out into a method that does that so we can rest easy

PR #623, opened 2018-02-13
Fixing edge cases where the signup is updated in Post Repository and not sent to Quasar

katiecrane:
#### What's this PR do?
1. Update the send signup and post to Quasar jobs to check for the sending to quasar feature flag in the job itself so locally you can still get (altered to say that they WOULD HAVE BEEN sent to Quasar) log messages and know that you are hitting the job. This has been a pain point for me when working on these jobs.

2. There are two times in the `PostService` where we go on to alter a signup, so we should be calling the signup job alongside the post job.

Passing tests? Feature flags still working?

When testing it became evident that we were missing some signup updates!

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/155199023)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/Three/PostRepository.php
+++ b/app/Repositories/Three/PostRepository.php
@@ -134,7 +134,7 @@ class PostRepository
 
         // If the quantity was updated, update the total quantity on the signup.
         if (isset($data['quantity'])) {
-            $signup = Signup::find($post->signup_id);
+            $signup = $post->signup;
             $signup->quantity = $signup->posts->sum('quantity');
             $signup->save();
         }
--- a/app/Services/Three/PostService.php
+++ b/app/Services/Three/PostService.php
@@ -49,9 +49,7 @@ class PostService
             SendPostToBlink::dispatch($post);
         }
 
-        if (config('features.pushToQuasar')) {
-            SendPostToQuasar::dispatch($post);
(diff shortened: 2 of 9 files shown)


app/Services/Three/PostService.php
```diff
@@ -49,9 +49,7 @@ public function create($data, $signupId, $authenticatedUserRole = null)
             SendPostToBlink::dispatch($post);
         }
 
-        if (config('features.pushToQuasar')) {
-            SendPostToQuasar::dispatch($post);
-        }
+        SendPostToQuasar::dispatch($post);
```
sbsmith86: it's much cleaner!

app/Repositories/Three/PostRepository.php
```diff
@@ -134,7 +134,7 @@ public function update($post, $data)
 
         // If the quantity was updated, update the total quantity on the signup.
         if (isset($data['quantity'])) {
-            $signup = Signup::find($post->signup_id);
+            $signup = $post->signup;
```
sbsmith86: 👀

katiecrane: credit to @DFurnes , always leaving things cleaner than he found them!

PR #627, opened 2018-02-21
History modal bug

chloealee:
#### What's this PR do?
- Fixes history modal so you can update the quantity on the post when `DS_ENABLE_V3_QUANTITY_SUPPORT` is set to `true`. **This only works for when this is set to `true` and will break when it is set to `false`.** 
- Disables history events log table for now (until we refactor events system)
- Disables showing the quantity at the top of the signup page until we switch to support for quantity on the post and revisit what this looks like (quantity is now visible next to each post)
- Only allows admins to update the quantity if they enter a number (button only becomes disabled if a number is entered):
feb-21-2018 15-33-54

1. Set the `DS_ENABLE_V3_QUANTITY_SUPPORT` to be `true`. 
2. Visit `/campaign/:nid` 
3. Update a post's quantity. Make sure the change is reflected in the database. 
4. Visit `/campaign/:nid/inbox`
5. Update a post's quantity. Make sure the change is reflected in the database. 
6. Visit `/signups/:sid`. Make sure there is no quantity at the top.
7. Try to update quantity with non-integer. This should not let you submit the form.
#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/154347334) pivotal card.

#### Checklist
- [ ] Tested on staging.


resources/assets/components/WithReviewing/index.js
```diff
@@ -149,29 +151,51 @@ const reviewComponent = (Component, data) => {
...
+      if (window.ENV['DS_ENABLE_V3_QUANTITY_SUPPORT']) {
+        // Field to send to /api/v3/posts/:post_id
+        const field = {
+          quantity: parseInt(newQuantity),
+        };
 
-      // Make API request to Rogue to update the quantity on the backend
-      let request = this.api.post('posts', fields);
+        // Make API request to Rogue to update the quantity on the backend
+        let request = this.api.patch('api/v3/posts/'.concat(postOrSignup['id']), field);
```
sbsmith86: hmm I am a little confused on all of this. If we are using `V3` endpoints for this then the feature flag doesn't need to be checked. V3 endpoints should all be considering quantity on the post. 

For example, we should never be making a request to `api/v3/post/:signup_id`

chloealee: @sbsmith86 right (and sorry this is all a little jumbled!)! I didn't want to break anything before the feature flag is turned on so that's why did the check to see if it is on or not so it works when it is both on and off.

sbsmith86: The feature flag should be relatively temporary. Let's add some `@TODO`s to remind us that we need to remove support for both ways since the world we want to move towards is simply quantity on the post and rogue admin using v3 endpoints.

chloealee: sounds good! I'll add those in all places we check for the feature flag now!

chloealee: is this something I should add to the list of things to refactor? I think you started a doc but unsure where it is?

sbsmith86: @chloealee I started a list [here](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/14). We certainly need refactor all of this. One of the points in that list is to switch the rogue frontend to use V3 endpoints and we will need to figure out a way to break up that work reasonably. Maybe add this as a sub-bullet to that point so we can track it as a discrete piece of work that needs to happen.

chloealee: awesome, thanks @sbsmith86 ! Updated that :)

resources/assets/components/WithReviewing/index.js
```diff
@@ -149,25 +151,21 @@ const reviewComponent = (Component, data) => {
...
+    // Update a post's quantity.
+    updateQuantity(post, newQuantity) {
+      // Field to send to /api/v3/posts/:post_id
+      const field = {
+        quantity: parseInt(newQuantity),
       };
 
       // Make API request to Rogue to update the quantity on the backend
-      let request = this.api.post('posts', fields);
+      let request = this.api.patch('api/v3/posts/'.concat(post['id']), field);
```
DFurnes: This might read a little better with a template string:

```js
let request = this.api.patch(`api/v3/posts/${post['id']}`, field);
```

DFurnes: I'm not sure why this would have been passing before & not now, but we may want to add a named method to the `RogueClient` so we can [mock this call](https://github.com/DoSomething/rogue/blob/e64bfcedab841a5a1cef9a5733541e0c08f1d222/resources/assets/utilities/__mocks__/RogueClient.js) if it's run [in the test](https://github.com/DoSomething/rogue/blob/e64bfcedab841a5a1cef9a5733541e0c08f1d222/resources/assets/components/WithReviewing/test.js)?

chloealee: @DFurnes ah cool! I looked at that but wasn't sure if there was a way to add the OAuth authorization to that without affecting the old calls?

DFurnes: Ahhh right, I forgot that's how we did this. It might make sense to make a `RogueOAuthClient` class that we could use for v3 endpoints & then we could configure and mock it separately?

PR #629, opened 2018-02-23
Box shows 0 if there is no quantity on the post 

chloealee:
#### What's this PR do?
- Box shows `0` if there is no quantity on the post (instead of the box not showing up all together)
Before:
screen shot 2018-02-23 at 2 16 52 pm

After:
screen shot 2018-02-23 at 2 46 07 pm

👀 

This shouldn't matter for most of the posts, only those posts that were created before we turned on the flag to support quantity on post. 

The only posts that should show this `0` are the posts that the script turns to `0` and puts the signup quantity on the latest post. 

#### Relevant tickets
Relevant to [this](https://www.pivotaltracker.com/n/projects/2019429/stories/154347334) pivotal card.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/resources/assets/components/Post/index.js
+++ b/resources/assets/components/Post/index.js
@@ -59,7 +59,7 @@ class Post extends React.Component {
     const user = this.props.user ? this.props.user : null;
     const signup = this.props.signup;
     const campaign = this.props.campaign;
-    let quantity = post.quantity != null ? post.quantity : 0;
+    const quantity = post.quantity != null ? post.quantity : 0;
 
     return (
       <div className="post container__row">
--- a/resources/assets/components/Post/index.js
+++ b/resources/assets/components/Post/index.js
@@ -95,7 +95,7 @@ class Post extends React.Component {
         {/* User and Post information */}
         <div className="container__block -third">
           <UserInformation user={user} linkSignup={signup.signup_id}>
-            {post.quantity && this.props.showQuantity ?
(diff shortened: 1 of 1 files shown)


resources/assets/components/Post/index.js
```diff
@@ -59,6 +59,7 @@ class Post extends React.Component {
     const user = this.props.user ? this.props.user : null;
     const signup = this.props.signup;
     const campaign = this.props.campaign;
+    let quantity = post.quantity != null ? post.quantity : 0;
```
katiecrane: Could this be `const`?

chloealee: yes I think so! I just wasn't sure since it could change?

sbsmith86:
@chloealee in the screenshot, i don't see a `0` just an empty box?

resources/assets/components/Post/index.js
katiecrane: doesn't look like it ever gets changed so I think it'd be good as `const`!

sbsmith86: hrrrmmmm i think it could be `const` ! it doesn't change throughout the lifecycle of the component.

chloealee: makes sense! updated :)

chloealee:
@sbsmith86 shoot sorry wrong screenshot! updated it!

PR #631, opened 2018-02-27
Send reviews to Customer.io

katiecrane:
#### What's this PR do?
- Added `SendReviewedPostToCIO.php` which works like our other jobs to send data along. It hits `v1/events/user-signup-post-review` in Blink and sends the same payload that we sent to Blink when a post is created.
- The job is dispatched in the `review` methods of both the original and the v3 `PostService`
- In the reviews tests, call `Bus::fake()` and `        Bus::assertDispatched(SendReviewedPostToCIO::class)` to make sure the job is getting dispatched (but don't actually call it)

All the tests are passing, but are there other tests where the job should be getting dispatched?

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/149862801)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Jobs/SendReviewedPostToCIO.php
+++ b/app/Jobs/SendReviewedPostToCIO.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Rogue\Jobs;
+
+use Rogue\Models\Post;
+use Illuminate\Bus\Queueable;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+class SendReviewedPostToCIO implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
+     * The post to send to Customer.io.
+     *
+     * @var Post
+     */
+    protected $post;
+
+    /**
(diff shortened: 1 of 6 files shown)


app/Jobs/SendReviewedPostToCIO.php
```diff
@@ -0,0 +1,53 @@
...
+namespace Rogue\Jobs;
+
+use Rogue\Models\Post;
+use Illuminate\Bus\Queueable;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+class SendReviewedPostToCIO implements ShouldQueue
```
DFurnes: This is kinda nitpicky, but I think this might read better as `SendReviewedPostToCustomerIo` (so you don't need to figure out what `CIO` stands for). If we do keep the abbreviation, Laravel's style rules also write them title-case (like `Http`). :neckbeard:

katiecrane: I'm down with spelling it out!

katiecrane: @DFurnes updated! cool to merge?

DFurnes: Good by me! ⛵️

PR #632, opened 2018-03-03
Tag posts command

sbsmith86:
#### What's this PR do?
Adds a command that lets you programatically tag a group of posts. 

To run the command: 
`php artisan rogue:tagposts --tag=1 --posts=1,3,6`

which would tag posts 1, 3, and 6 as `hide-in-gallery`

👀 

So we can quickly tag some posts for a campaign when we need to. 

#### Relevant tickets

Needed to fix some posts on prod. 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Console/Commands/TagPosts.php
+++ b/app/Console/Commands/TagPosts.php
@@ -47,6 +47,7 @@ class TagPosts extends Command
         if ($posts) {
             foreach ($posts as $post) {
                 $this->info('Tagging post_id: '.$post->id.' as '.$tag->tag_slug);
+                // @TODO - Send to quasar when tagged.
                 $post->tag($tag->tag_name);
             }
         }
--- a/app/Console/Commands/TagPosts.php
+++ b/app/Console/Commands/TagPosts.php
@@ -2,8 +2,8 @@
 
 namespace Rogue\Console\Commands;
 
-use Rogue\Models\Post;
 use Rogue\Models\Tag;
+use Rogue\Models\Post;
 use Illuminate\Console\Command;
 
 class TagPosts extends Command
@@ -44,8 +44,7 @@ class TagPosts extends Command
         $posts = Post::find($postIds);
         $tag = Tag::findOrFail($this->option('tag'));
 
-        foreach ($posts as $post)
-        {
+        foreach ($posts as $post) {
             $post->tag($tag->tag_name);
         }
     }
--- a/app/Console/Commands/TagPosts.php
+++ b/app/Console/Commands/TagPosts.php
@@ -13,7 +13,7 @@ class TagPosts extends Command
      *
      * @var string
      */
-    protected $signature = 'rogue:tagposts {post* : The ids of the posts to update, separate by spaces} {--tag=}';
+    protected $signature = 'rogue:tagposts {--tag=: The id of the tag to use.} {--posts=: Comma-separated list of the post_ids to update.}';
 
     /**
      * The console command description.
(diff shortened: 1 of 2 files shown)


app/Console/Commands/TagPosts.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $postIds = explode(',', $this->option('posts'));
+        $posts = Post::find($postIds);
+
+        $tag = Tag::findOrFail($this->option('tag'));
+
+        if ($posts) {
+            foreach ($posts as $post) {
+                $this->info('Tagging post_id: '.$post->id.' as '.$tag->tag_slug);
+                $post->tag($tag->tag_name);
```
katiecrane: If the post has already been tagged with that particular tag, will this untag it?

app/Console/Commands/TagPosts.php
```diff
@@ -0,0 +1,54 @@
...
+use Illuminate\Console\Command;
+
+class TagPosts extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'rogue:tagposts {--tag=: The id of the tag to use.} {--posts=: Comma-separated list of the post_ids to update.}';
```
katiecrane: Don't want to overcomplicate this, but in other commands where do things like that we usually read in things like this from a CSV (but that in itself is kind of a pain because it has to get uploaded to AWS). Wondering how many post IDs will be passed here and if you think passing them in through a CSV is more trouble than it's worth?

sbsmith86: Yea I think that could make sense for larrrrge batches of updates but that doesn't mean that it is not useful to have an easy command to reach to when we need to update a set of post. Also, this is really something that can just be done directly in the DB but having an easy command to reach to seems like a win. There has only been one use case for this really so far, but I can see this being useful in the future. Doesn't really bother me if is a large set of posts but yea I can see that being cumbersome. Something to keep in mind if this becomes a frequent request!

app/Console/Commands/TagPosts.php
sbsmith86: Nope! The logic just attaches the relationship and duplicates are not allowed in the table: https://github.com/DoSomething/rogue/blob/master/app/Models/Post.php#L242-L254

app/Console/Commands/TagPosts.php
katiecrane: @sbsmith86 awesome! makes total sense to me to have a quick way to add these tags, I'm on board 🚢

app/Console/Commands/TagPosts.php
katiecrane: @sbsmith86 cool! So since this command updates the Posts directly without passing through the PostService, I think we will want to fire off a `SendPostToQuasar` job here?

katiecrane: @sbsmith86 given that we are already not sending the tags over to Quasar (🐛 ) maybe we don't need to worry about this here?

app/Console/Commands/TagPosts.php
sbsmith86: yea hmmm. let me make a `@TODO` to send these quasar. I rather us continue using the service-repository pattern so we can really see what all the business logic is around events in rogue, even though it requires the mental lift to remember to do it.

katiecrane: @sbsmith86 sounds good to me!

sbsmith86: @katiecrane updated!

PR #633, opened 2018-03-05
Update signup field command

sbsmith86:
#### What's this PR do?

Adds a command to update a target field/value on the signup record given a set of query contraints.

Here is an example for updating `campaign_run_id` to `8044` for all signups where `campagin_run_id = 7669` (currently) and that were created after 3/1/18 

`php artisan rogue:updatesignup --target=campaign_run_id --campaign_run=7669 -targetValue=8044 --date=2018-03-01`

:eyes:

This is a command to help with a specific use case outlined in this [slack thread](https://dosomething.slack.com/archives/C2BLZGBPH/p1520264003000003)

I tried to make it flexible if we need it for other things, but this should at least take care of this use case and we can flush it out if more requests like this come in. 

Also this moves through the `SignupService` so it will send the signup update to quasar. 

#### Relevant tickets
Fixes 🐛 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Console/Commands/UpdateSignup.php
+++ b/app/Console/Commands/UpdateSignup.php
@@ -23,13 +23,22 @@ class UpdateSignup extends Command
     protected $description = 'Update a target field on a set of signups contrained by the provided parameters';
 
     /**
+     * The signup service instance.
+     *
+     * @var Rogue\Services\Three\SignupService
+     */
+    protected $signups;
+
+    /**
      * Create a new command instance.
      *
      * @return void
      */
-    public function __construct()
+    public function __construct(SignupService $signups)
     {
         parent::__construct();
+
+        $this->signups = $signups;
     }
 
     /**
@@ -40,5 +49,6 @@ class UpdateSignup extends Command
     public function handle()
     {
         // dd($this->option('target'), $this->option('campaign'), $this->option('campaign_run'), $this->option('date'));
+        dd(get_class($this->signups));
     }
 }
--- a/app/Console/Commands/UpdateSignup.php
+++ b/app/Console/Commands/UpdateSignup.php
@@ -83,13 +83,11 @@ class UpdateSignup extends Command
 
         if ($signups->isNotEmpty()) {
             foreach ($signups as $signup) {
-                $this->info('Updating '.$targetField.' for signup '.$signup->id);
-
                 $this->signups->update($signup, [$targetField => $targetValue]);
             }
         } else {
             $this->error('No signups found with that criteria.');
-            $return;
+            return;
         }
     }
 }
--- a/app/Console/Commands/UpdateSignup.php
+++ b/app/Console/Commands/UpdateSignup.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Rogue\Console\Commands;
+
+use Rogue\Models\Signup;
+use Illuminate\Console\Command;
+use Rogue\Services\Three\SignupService;
+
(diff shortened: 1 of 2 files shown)


app/Console/Commands/UpdateSignup.php
```diff
@@ -0,0 +1,95 @@
...
+            $this->error('No target field specified.');
+
+            return;
+        }
+
+        if (! $targetValue) {
+            $this->error('No target value specified.');
+
+            return;
+        }
```
katiecrane: I like that you made these options even though they are required because it feels easier to keep track of what is what when you are writing out the command to run like this.

app/Console/Commands/UpdateSignup.php
```diff
@@ -0,0 +1,95 @@
...
+
+        if ($this->option('campaign_run')) {
+            $query = $query->where('campaign_run_id', $this->option('campaign_run'));
+        }
+
+        // Only take in one date and we assume to be looking for things on or after that date.
+        // @TODO - Allow this to be more flexible (i.e accept two bounding dates for date ranges)
+        if ($this->option('date')) {
+            $query = $query->where('created_at', '>=', $this->option('date'));
+        }
```
katiecrane: These 3 `if` statements, if all true, do they continue to add more `where` statements to the query, and not reset the query to just have one `where` statement?

katiecrane: Oh wait. I see it now. Nevermind!

app/Console/Commands/UpdateSignup.php
```diff
@@ -0,0 +1,95 @@
...
+
+        $signups = $query->get();
+
+        if ($signups->isNotEmpty()) {
+            foreach ($signups as $signup) {
+                $this->info('Updating '.$targetField.' for signup '.$signup->id);
```
katiecrane: In this case (<1000 updates) logging this is probably fine, but I think we should make a note that if we ever run this command on a larger batch we will need to update this part so we don't overflow PaperTrail

app/Console/Commands/UpdateSignup.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+namespace Rogue\Console\Commands;
+
+use Rogue\Models\Signup;
+use Illuminate\Console\Command;
+use Rogue\Services\Three\SignupService;
```
katiecrane: THREEE!!! Yay!

app/Console/Commands/UpdateSignup.php
```diff
@@ -0,0 +1,95 @@
...
+
+        if ($signups->isNotEmpty()) {
+            foreach ($signups as $signup) {
+                $this->info('Updating '.$targetField.' for signup '.$signup->id);
+
+                $this->signups->update($signup, [$targetField => $targetValue]);
+            }
+        } else {
+            $this->error('No signups found with that criteria.');
+            $return;
```
katiecrane: Should this be plain `return`?

app/Console/Commands/UpdateSignup.php
sbsmith86: Ah right, I don't even know if we need this line I was just debugging with it. I can remove it for now.

app/Console/Commands/UpdateSignup.php
sbsmith86: eek yes, good 👁

PR #634, opened 2018-03-06
154979852 Turn off image editing

sbsmith86:
#### What's this PR do?

Moves the cropping we do under the glide feature flag so that, if glide is on, we don't do it.

The hope is that this won't cause any weird issues with images being oriented or cropped incorrectly. 

👀 

Detailed background on this issue: https://github.com/orgs/DoSomething/teams/team-bleed/discussions/16

#### Relevant tickets
addresses https://www.pivotaltracker.com/story/show/154979852

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -279,18 +279,20 @@ class PostRepository
     {
         $editedImage = Image::make($data['file']);
 
-        // If we have crop values, then use 'em.
+        // If glide is off and we have crop values, use those for editing.
         $cropValues = array_only($data, $this->cropProperties);
-        if (count($cropValues) > 0) {
+
+        if (!config('features.glide') && count($cropValues) > 0) {
             $editedImage = $editedImage
                 // Intervention Image rotates images counter-clockwise, but we get values assuming clockwise rotation, so we negate it to rotate clockwise.
                 ->rotate(-$cropValues['crop_rotate'])
-                ->crop($cropValues['crop_width'], $cropValues['crop_height'], $cropValues['crop_x'], $cropValues['crop_y']);
+                ->crop($cropValues['crop_width'], $cropValues['crop_height'], $cropValues['crop_x'], $cropValues['crop_y'])
+                ->encode('jpg', 75);
+        // Otherwise, use default crop (400x400).
+        } else {
+            $editedImage = $editedImage->fit(400, 400)->encode('jpg', 75);
         }
 
-        $editedImage = $editedImage->fit(400)
-            ->encode('jpg', 75);
-
-        return $this->aws->storeImageData((string) $editedImage, 'edited_' . $postId);
+        return $this->aws->storeImageData((string)$editedImage, 'edited_' . $postId);
     }
 }
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -65,9 +65,6 @@ class PostRepository
     public function create(array $data, $signupId)
     {
         if (isset($data['file'])) {
-            // Auto-orient the photo by default based on exif data.
-            $image = Image::make($data['file'])->orientate();
-
             $fileUrl = $this->aws->storeImage((string) $image->encode('data-url'), $signupId);
         } else {
             $fileUrl = null;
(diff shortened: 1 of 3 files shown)


app/Repositories/PostRepository.php
```diff
@@ -127,9 +127,11 @@ public function create(array $data, $signupId)
             $signup->save();
         }
 
-        // Edit the image if there is one
-        if (isset($data['file'])) {
-            $this->crop($data, $post->id);
+        if (! config('features.glide')) {
```
DFurnes: Do we want to keep creating these, in case we decide to turn Glide off again?

DFurnes: I know this is not part of the PR, but we might want to remove the `->orientate()` from above too:

```diff
- $image = Image::make($data['file'])->orientate();
+ $image = Image::make($data['file']);
 $fileUrl = $this->aws->storeImage((string) $image->encode('data-url'), $signupId);
```

sbsmith86: I think the flag should allow us to continue creating these.  If it is off, we crop, else we don't?

sbsmith86: ah yes!

sbsmith86:
@DFurnes updated!

app/Repositories/PostRepository.php
DFurnes: I think this would mean we'd stop creating `edited_12313.jpg` crops once we turn on Glide, which would leave us with some 404ing images if we decided we had to turn it back off again after flipping the switch.

DFurnes: Maybe a nice interim solution is to remove the `crop_rotate`, `crop_width`, etc. code paths from that function so we don't support custom crops anymore but still make the "default" edited crops?

sbsmith86: hmm I see, ok

sbsmith86: @DFurnes updated!

sbsmith86:
@DFurnes This is ready for another 👀

app/Repositories/PostRepository.php
katiecrane: @sbsmith86 It looks like you removed the pieces we don't need and just make the cropped image, but I thought the idea was to make that cropped image regardless of if glide is on or not, in case we have to turn it off?

sbsmith86: @katiecrane hmmm i don't think I understand the question, sorry! So this will still crop the image regardless of glide being on or off and store it in S3 as `edited_{post-id}.jpeg` but instead of using the crop values passed in the request, we just default to a 400x400 crop.

katiecrane: @sbsmith86 hmm I'm seeing this inside `if (! config('features.glide'))` so it looks to me like this will only happen if glide is off. There are a lot of comments between that and the next line of code though, so I might be reading it wrong

katiecrane: @sbsmith86 oh weird, I think I was somehow on an outdated version of the PR, nevermind!

sbsmith86:
@katiecrane I see what you are saying! i updated `Three/PostRepository`  but not the `PostRepostitory` class. This has been updated. It should be more clear now.

app/Repositories/PostRepository.php
katiecrane: @sbsmith86 I'm not seeing this change (remove call to `orientate` on original image)

katiecrane: @sbsmith86 hmm yeah I was expecting a change to remove `orientate` here: https://github.com/DoSomething/rogue/pull/634/files#diff-158da9223a5d32ce24f63cb94477a423R69

sbsmith86: OH! looks like I missed that one. Good 👀 !!!

sbsmith86:
@katiecrane updated!

app/Repositories/PostRepository.php
```diff
@@ -65,9 +65,6 @@ public function find($id)
     public function create(array $data, $signupId)
     {
         if (isset($data['file'])) {
-            // Auto-orient the photo by default based on exif data.
-            $image = Image::make($data['file'])->orientate();
```
katiecrane: @sbsmith86 I think you still gotta make it because you use it below, but remove the call to `orientate()`

sbsmith86: @katiecrane yep! Refresh

PR #637, opened 2018-03-14
Throttle /images route

katiecrane:
#### What's this PR do?
- Added throttling (at 15 requests/minute) to the `/images` route
- Added a test to make sure that a user gets a `Too Many Attempts` response after 15 images

👀  Is there anything else we should be testing for? Do we need to test with more than one image?

#### Relevant tickets
[Card](https://www.pivotaltracker.com/n/projects/2019429/stories/155853452)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/tests/Http/ImagesTest.php
+++ b/tests/Http/ImagesTest.php
@@ -15,24 +15,23 @@ class ImagesTest extends TestCase
     public function testImagesThrottle()
     {
         // Make a post to view
-        $post = factory(Post::class)->create();
+        $posts = factory(Post::class, 10)->create();
 
         // View the post 15 times (using 3 different versions)
         for ($i = 0; $i < 5; $i++) {
-            $response = $this->getJson('images/' . $post->id);
-            $response->assertStatus(200);
+            $response = $this->getJson('images/' . $posts->random()->id)->assertSuccessful(200);
         }
         for ($i = 5; $i < 10; $i++) {
-            $response = $this->getJson('images/' . $post->id . '?w=500&h=500&fit=crop');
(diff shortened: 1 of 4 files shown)


tests/Http/ImagesTest.php
```diff
@@ -0,0 +1,38 @@
...
+     *
+     * @return void
+     */
+    public function testImagesThrottle()
+    {
+        // Make a post to view
+        $post = factory(Post::class)->create();
+
+        // View the post 15 times (using 3 different versions)
+        for ($i = 0; $i < 5; $i++) {
```
sbsmith86: Just making sure -  this makes the requests fast enough to model an aggressive user loading too many images?

DFurnes:
> Do we need to test with more than one image?

I think that could be a good idea! Maybe something like this:

```php
$posts = factory(Post::class, 10)->create();

for ($i = 0; $i < 10; $i++) {
    $this->get('images/' . $posts->random()->id)->assertSuccessful();
}
```

tests/Http/ImagesTest.php
katiecrane: @sbsmith86 I believe so since the test takes way less than a minute to run. Here is the test in Northstar that it's modeled after: https://github.com/DoSomething/northstar/blob/5a98445784fa94c5fc0a80a92f66f9621293b200/tests/Http/Web/WebAuthenticationTest.php#L231-L245

DFurnes: Yeah, this for-loop should run incredibly quickly.

PR #638, opened 2018-03-15
Allows text value for type when creating a post 

chloealee:
#### What's this PR do?
- Allows text value to be passed for `type` when creating a post
- Doesn't require `campaign_run_id` when creating a post for v3 endpoints

👀 

We don't want `campaign_run_id` to be required with `campaign_id` anymore since Phoenix-Next doesn't have `campaign_run_id`s correct?

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/155793181) Pivotal Card

#### Checklist
- [ ] Tested on staging.


app/Http/Requests/Three/PostRequest.php
```diff
@@ -25,9 +25,9 @@ public function rules()
     {
         return [
             'campaign_id' => 'required',
-            'campaign_run_id' => 'required_with:campaign_id|integer',
+            'campaign_run_id' => 'integer',
```
sbsmith86: nice catch!

chloealee:
oh wait @katiecrane that is less than so that is why it is failing :-/ Is it supposed to stop after that decimal?

katiecrane:
@chloealee hmmm yeah that definitely is greater and I can't remember having issues with this in the past. could you try re-running the wercker build so we know for sure that it's not working?

chloealee:
@katiecrane hmmm just re-ran and still breaking :-/ this seems to also be failing when I switch to my local master - are there any updates that I need to run perhaps? I ran `composer install` and `composer update`

chloealee:
Got 👍 from team to mark this test an incomplete and address in buggy unit tests [card](https://www.pivotaltracker.com/n/projects/2019429/stories/155479565). Slack convo [here](https://dosomething.slack.com/archives/C0YGXUE01/p1521147300000122)!

PR #640, opened 2018-03-16
Update root readme

sbsmith86:
#### What's this PR do?
The root readme was holding content from the API specific readme. This updates that. 

👀

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,96 +1,30 @@
-# Rogue API
-This is the Rogue API, it is used to capture activity from members.
-All `POST` and `PUT` endpoints (except v3 endpoints) require an api key (`X-DS-Rogue-API-Key`) in the header to be submitted with the request.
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1,96 +1,30 @@
...
-### v1
-#### Reportbacks
-Endpoint                                       | Functionality
----------------------------------------------- | --------------------------------------------------------
-`GET /api/v1/reportbacks`                      | [Get reportbacks](endpoints/reportbacks.md#reportbacks)
+To get started with development, follow the [installation](./docs/development/installation.md) and [contributing](./docs/development/contributing.md) documentation.
 
+To get started using the API, see our [API Documentation](./docs/endpoints)
 
-### v2
```
katiecrane: I'm not seeing v2 indexed in table form like this anywhere else, just making sure we don't want to keep this!

sbsmith86: Also it should still be there in some format in `/endpoints/*`

katiecrane: sounds good!!

PR #642, opened 2018-03-16
Updates POST /api/v2/posts endpoint to support type and action

chloealee:
#### What's this PR do?
- Updates `POST /api/v2/posts` endpoint to support `type` and `action` so that it unblocks Gambit from using `v2` to send these over.
- Adds test to make sure creating a post with and without `type` and `action` works. 
- Updates documentation (including for #641 's update that `type` and `action` are now returned in the transformer!)

- All tests should pass!

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/155993693) Pivotal Card.

#### Checklist
- [ ] Tested on staging.


app/Repositories/PostRepository.php
```diff
@@ -85,7 +85,8 @@ public function create(array $data, $signupId)
             'status' => isset($data['status']) ? $data['status'] : 'pending',
             'source' => isset($data['source']) ? $data['source'] : null,
             'remote_addr' => isset($data['remote_addr']) ? $data['remote_addr'] : null,
-
+            'type' => isset($data['type']) ? $data['type'] : 'photo',
+            'action' => isset($data['action']) ? $data['action'] : 'default',
         ]);
```
chloealee: are these the defaults we want to include if `type` and `action` are not sent in the request?

app/Repositories/Three/PostRepository.php
```diff
@@ -83,8 +83,8 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
             'northstar_id' => $signup->northstar_id,
             'campaign_id' => $signup->campaign_id,
             'quantity' => isset($data['quantity']) ? $data['quantity'] : null,
-            'type' => isset($data['type']) ? $data['type'] : 'photo',
-            'action' => isset($data['action']) ? $data['action'] : null,
+            'type' => $data['type'],
+            'action' => $data['action'],
             'url' => $fileUrl,
```
chloealee: I removed the `isset` check because in `v3`, these are required.

docs/endpoints/posts.md
```diff
@@ -139,6 +139,9 @@ POST /api/v2/posts
     `Y-m-d H:i:s` format. When the post was created.
   - **updated_at**: (string) optional.
     `Y-m-d H:i:s` format. When the post was last updated.
+  - **type**: (string).
+    The type of post submitted e.g. photo, voter-reg, text
+  - **action**: (string).
```
sbsmith86: We define this as: `Describes the bucket the action is tied to. A campaign could ask for multiple types of actions throughout the life of the campaign.` in the DB

app/Repositories/PostRepository.php
sbsmith86: yea

docs/endpoints/posts.md
chloealee: ah ok that is great! I'll add to the documentation :)

app/Repositories/Three/PostRepository.php
sbsmith86: I think we should keep these because this repository could be used in other places other than `v3` routes and we want to allow them to be flexible.

chloealee: ah ok that makes sense! Changed back :)

PR #643, opened 2018-03-16
Review command

sbsmith86:
#### What's this PR do?

Adds a command that will review all posts under a given signup id with a given status. 

It's run like so: `php artisan rogue:reviewsignup 1 rejected --admin=559442cca59dbfc9578b4bf4`

^^ which would update all posts under signup 1 with a rejected status and using the admin id passed

The main gotcha here was that we store the `admin_northstar_id` when we review a post. We need to be able to set this manually when running the command instead of using just the authenticated user. 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/156055464


app/Console/Commands/ReviewSignup.php
```diff
@@ -0,0 +1,69 @@
...
+        $status = $this->argument('status');
+        $posts = $signup->posts;
+        $admin = $this->option('admin');
+
+        $this->info('Updating status of all posts under signup ID '.$signup->id);
+        $this->info('There are ' .$posts->count(). ' posts to update.');
+
+        if ($posts) {
+            foreach ($posts as $post) {
+                $this->posts->review($post, $status, $admin);
```
chloealee: i think this should be `reviews`?

chloealee: oh wait no I'm wrong! Sorry!

app/Console/Commands/ReviewSignup.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $signup = Signup::findOrFail($this->argument('signup'));
```
katiecrane: Would it make sense to grab it with posts here?

katiecrane: I guess it's only one signup so probably easier this way since it's already done!

PR #644, opened 2018-03-19
Adds activity and write scopes

chloealee:
#### What's this PR do?
- Requires `activity` scope to read/write to Rogue for `v3` endpoints (based on [this comment](https://dosomething.slack.com/archives/C02BBRN5A/p1521229020000254?thread_ts=1521228167.000555&cid=C02BBRN5A), I _think_ we only want this on `v3` endpoints?)
- Requires `write` scope for any create/update/delete endpoints, so we can easily make read-only clients.
- Updates all tests to add `activity` and `write` scopes where needed.
- Adds tests to make sure `activity` and `write` scopes work as expected.

- All `v3` endpoints must have `activity` scope.
- All create/update/delete `v3` endpoints must have `write` scope.
- All tests should pass!

#### Before this gets merged...
We need to update/notify all places that are using `v3` endpoints to add the proper scopes. 
  - Frontend Rogue relies on some `v3` endpoints. 
    - [ `DELETE /api/v3/signups/${signupId}`](https://github.com/DoSomething/rogue/blob/406c5bae333962c2bfd75e5fde56b097b6e00783/resources/assets/components/Signup/index.js#L281)
    - [`PATCH /api/v3/posts/${postId}`](https://github.com/DoSomething/rogue/blob/406c5bae333962c2bfd75e5fde56b097b6e00783/resources/assets/components/Signup/index.js#L281)
  - Any other apps using `v3` endpoints right now? Let's also give other teams a heads up for when they start using `v3` endpoints. 
  - Runscope test client
  - Anything else I'm missing? 

#### Relevant tickets
Fixes second task of [this](https://www.pivotaltracker.com/n/projects/2019429/stories/155367548) Pivotal card.

#### Checklist
- [ ] Tested on staging.


chloealee:
@DFurnes will other apps using `v3` endpoints need to add new scopes in their `services.php` file like we did in this PR?

chloealee:
@sbsmith86 I think these are the same tests that were failing for @katiecrane you mentioned today in planning? Should I mark them as incomplete for now?

chloealee:
Not sure why the Dusk tests aren't passing! They are all passing on my local so I'm wondering if it is the same issue we're seeing on prod for Grab the Mic (Slack convo [here](https://dosomething.slack.com/archives/C0YGXUE01/p1522092837000010)). Wondering if I should also mark these incomplete/comment them out and re-visit with the card mentioned above?

tests/Http/Three/ReviewsTest.php
```diff
@@ -74,7 +122,7 @@ public function testUpdatedPostAndSignupWithReview()
 
         // Review the post.
         $northstarId = $this->faker->northstar_id;
-        $this->withAccessToken($northstarId, 'admin')->postJson('api/v3/reviews', [
+        $this->withAccessToken($northstarId, 'admin', ['activity', 'write'])->postJson('api/v3/reviews', [
```
DFurnes: I'd love to make an update to Gateway's [WithOAuthTokens](https://github.com/DoSomething/gateway/blob/master/src/Testing/WithOAuthTokens.php) trait that allows it to read default scopes from a property on the class, so that we can keep using `withAdminAccessToken` here. (Or even just add `activity` to the list of scopes hard-coded into that trait.)

routes/api.php
```diff
@@ -48,7 +48,7 @@
 });
 
 // v3 routes
-$router->group(['prefix' => 'api/v3', 'middleware' => ['guard:api']], function () {
+$router->group(['prefix' => 'api/v3', 'middleware' => ['guard:api', 'scopes:activity']], function () {
```
DFurnes: I think I slightly prefer attaching this `scopes:activity` in each of the controllers, so it's more easily discoverable (alongside where we require other scopes, authorization, or roles).

DFurnes:
Let's make sure to communicate this change to Team Rocket before deploying it!

DFurnes:
> @DFurnes will other apps using `v3` endpoints need to add new scopes in their `services.php` file like we did in this PR?

Yep! We'll need to make sure they're on the appropriate clients in Aurora, and updated in each app.

chloealee:
@DFurnes great! I will put this in the #product-eng-data channel and specifically tag Rocket !

tests/Http/Three/ReviewsTest.php
chloealee: @DFurnes Ah ok that makes more sense! So should I just add it [here](https://github.com/DoSomething/gateway/blob/master/src/Testing/WithOAuthTokens.php#L54) in the `scopes` array?

DFurnes:
@chloealee For the Dusk tests, I wonder if it's because we haven't pulled the public key for that server (whereas we do as part of our setup steps on local)? Although not sure why it wouldn't have failed earlier then... 

Could you try adding `php artisan gateway:key` above [this line](https://github.com/DoSomething/rogue/blob/406c5bae333962c2bfd75e5fde56b097b6e00783/wercker.yml#L34)?

tests/Http/Three/ReviewsTest.php
DFurnes: Yeah, I think that'd be the easiest fix!

routes/api.php
chloealee: makes sense! Updated :)

chloealee:
@DFurnes I added that and it still didn't work :( should I still include?

tests/Http/Three/ReviewsTest.php
chloealee: great! I'll add the `activity` scope there and then keep the `write` scopes in these tests in Rogue!

chloealee: [here's](https://github.com/DoSomething/gateway/compare/scopes?expand=1) the PR :)

DFurnes:
I still think it makes sense to include, but bummer that it didn't fix the errors! 😭

chloealee:
@DFurnes hmm so a bunch of the phpunit tests are failing with the new gateway update but because the `write` scope is overriding all the others in `withAccessToken` - I think I need to include `write` as a default after all. PR for that [here](https://github.com/DoSomething/gateway/pull/113)!

chloealee:
yay  @DFurnes everything is passing now (with commented out tests)! Can you give one more look for final 👍 before I merge?

tests/Http/Three/PostTest.php
```diff
@@ -460,7 +491,7 @@ public function testPostsIndexAsNonAdminNonOwner()
         factory(Post::class, 'accepted', 10)->create();
         factory(Post::class, 'rejected', 5)->create();
 
-        $response = $this->getJson('api/v3/posts');
+        $response = $this->withAccessToken($this->randomUserId(), 'user')->getJson('api/v3/posts');
```
DFurnes: Could this be `->withStandardAccessToken()`?

tests/Http/Three/PostTest.php
```diff
@@ -510,7 +541,7 @@ public function testPostsIndexAsAdmin()
         factory(Post::class, 'accepted', 10)->create();
         factory(Post::class, 'rejected', 5)->create();
 
-        $response = $this->withAdminAccessToken()->getJson('api/v3/posts');
+        $response = $this->withAccessToken($this->randomUserId(), 'admin')->getJson('api/v3/posts');
```
DFurnes: Same here, could this be `->withAdminAccessToken()`?

tests/Http/Three/PostTest.php
```diff
@@ -553,6 +584,24 @@ public function testPostsIndexAsAdmin()
...
+     * GET /api/v3/posts
+     * @return void
+     */
+    public function testPostsIndexWithoutRequiredScope()
+    {
+        // Admins should see all posts.
+        factory(Post::class, 'accepted', 10)->create();
+        factory(Post::class, 'rejected', 5)->create();
+
+        $response = $this->getJson('api/v3/posts');
```
DFurnes: Oh hmm, I think there's a catch here! 🎣 We want to make sure this route is publicly accessible (for example, so that anonymous users visiting a campaign page can see a gallery before signing up).

I'm thinking that we only want Gateway's `scope` middleware to check scopes at all if there isn't a token... it feels like they're irrelevant in cases where we allow anonymous traffic. What do you think about adding a conditional like this to the [`scope` middleware](https://github.com/DoSomething/gateway/blob/00e3e567f0ab8fdbf85c75c5d4a1b894427120ab/src/Server/Middleware/RequireScope.php#L38)?

```php
    public function handle($request, Closure $next, ...$requestedScopes)
    {
        if (! $this->token->exists()) {
            return $next($request);
        }

        $providedScopes = $this->token->scopes;
        // ...
```

chloealee: @DFurnes that makes sense to me! And then edit all the tests so that anonymous users _can_ see all the `GET` v3 endpoints?

chloealee: @DFurnes just bumping this in case you missed it! I _think_ that all `GET` v3 endpoints should be ok for anonymous users to see but wanted to double check!

DFurnes: Yeah, I think that sounds good! 👍

chloealee:
shoooot I was afraid this was going to happen! @DFurnes I started working on another card to restructure the directories and it somehow to into this PR by accident when I switched branches! Is there anyway to omit those files from this PR? I always have trouble with this!

DFurnes:
Oh no! I can swing by and we can try to sort it out. 🐙

chloealee:
🙌

tests/Http/Three/SignupTest.php
```diff
@@ -616,7 +675,7 @@ public function testValidationForUpdatingASignup()
     {
         $signup = factory(Signup::class)->create();
 
-        $response = $this->withAdminAccessToken()->patchJson('api/v3/signups/' . $signup->id);
```
DFurnes: Not required to merge, but you should be able to change all these back to `->withAdminAccessToken()` now.

DFurnes:
I think this should only affect Phoenix Next (which has [been updated](https://github.com/DoSomething/phoenix-next/pull/791)), but let's be sure to give a heads up in `#prod-eng-data` to anyone else who might be thinking about using the `v3` endpoints!

tests/Http/Three/SignupTest.php
chloealee: oh right ! I will add this in the card when I dig into the `AsOwner` tests to also change these back!

PR #646, opened 2018-03-20
Optionally including "accepted_quantity" when getting signups

katiecrane:
#### What's this PR do?
- Added a new transformer for `accepted_quantity`
- Added a new optional `includes` to the Signups Transformer to access `accepted_quantity` (and updated docs to reflect this)
- `getAcceptedQuantity()` method added to Signup to calculate the total quantity of accepted posts

Does this let Domino get the information that it needs?

This has been a real whirlwind! 🌪 

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/156024795)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -102,10 +102,6 @@ class SignupsController extends ApiController
             if (in_array($column, Signup::$indexes)) {
                 $query = $query->orderBy($column, $direction);
             }
-
-            if ($column === 'accepted_quantity') {
-                // maybe do some sorting stuff here
-            }
         }
 
         return $this->paginatedCollection($query, $request);
--- a/app/Http/Controllers/Three/SignupsController.php
+++ b/app/Http/Controllers/Three/SignupsController.php
@@ -100,13 +100,21 @@ class SignupsController extends ApiController
             list($column, $direction) = explode(',', $orderBy, 2);
 
             if (in_array($column, Signup::$indexes)) {
+                dd('got a column');
                 $query = $query->orderBy($column, $direction);
             }
 
-            // if ($column === 'accepted_quantity') {
+            if ($column === 'accepted_quantity') {
+                // $query = $query->withAcceptedQuantity();
+                $query->with(['posts' => function ($query) {
+                    $query->where('status', 'accepted');
+                }]);
+                // $query->select('*', \DB::raw('SUM(posts.quantity) as accepted_quantity'));
 
-            //     $query = $query->orderBy(DB::raw());
-            // }
+                // $query = $query->orderBy(\DB::raw('accepted_quantity'), $direction);
+
+                // $query->select('*', \DB::raw('SUM(quantity) as accepted_quantity'));
+            }
         }
 
         return $this->paginatedCollection($query, $request);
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Three/SignupsController.php
```diff
@@ -102,6 +102,10 @@ public function index(Request $request)
             if (in_array($column, Signup::$indexes)) {
                 $query = $query->orderBy($column, $direction);
             }
+
+            if ($column === 'accepted_quantity') {
```
sbsmith86: where is `$column` coming from?

app/Http/Controllers/Three/SignupsController.php
```diff
@@ -102,6 +102,10 @@ public function index(Request $request)
...
+                // maybe do some sorting stuff here
+            }
         }
 
         return $this->paginatedCollection($query, $request);
```
katiecrane: Here we pass off the query we have built to `paginatedCollection` and it takes care of the rest. Currently, other orderBy parameters are baked into the query, but since we are trying to order by the sum of the quantity of the accepted posts associated with each signup, it is less straightforward to add to the query (and I haven't figured out a way to get it to work).

app/Http/Controllers/Three/SignupsController.php
katiecrane: line 100 above

sbsmith86:
hmm I think what I am not understanding is that you are already able to include posts in the request and order them? [here](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Three/SignupsController.php#L78-L108)

So could you maybe create a new query scope like `withVisiblePosts()` that only grabs accepted posts (regardless of user role) and then allow the ordering to happen as it already does?

katiecrane:
@sbsmith86 Ahh so we don't actually want to include the posts. What Domino wants is just the signups, and it would like those ordered based on the accepted quantity (being the sum of the quantities of the accepted posts under that signup). Right now we are only able to order signups by the quantity that lives on the signup (which includes quantities from all posts, regardless of status).

In that code chunk that you linked, we are able to include the posts, but can only order the results based on a column in the `signups` table.

app/Models/Signup.php
```diff
@@ -182,6 +182,18 @@ public function getQuantity()
...
     }
 
+    /**
+     * Get the quantity total associated with approved posts under this signup
+     *
+     * @return int
+     */
+    public function getAcceptedQuantity()
+    {
+        $accepted_posts = $this->posts->where('status', 'accepted');
```
DFurnes: How's this look in terms of performance? (Specifically curious what kind of SQL queries this ends up running, and if it still performs well with production-sized data sets).

katiecrane: good question! I'll check on it

katiecrane: @DFurnes I used clockwork and my local Domino for this (since currently this optional include should only be used by Domino) and I'm clocking times in the ~700ms range but there are a bunch of queries. But are there too many queries? Of that, I am not sure! Want to swing by?

DFurnes: Wowza, 91 queries?! That's a lot, but yeah doesn't seem related to this change.

katiecrane: @DFurnes yup! there are more in the case when stuff is no longer in cache, but seems consistent whether or not `include=accepted_quantity` is used

PR #649, opened 2018-03-27
Text Reportback Updates 

chloealee:
#### What's this PR do?
This updates the front end so Rogue admins understand that they're reviewing a text post versus other post.

Specifically, this PR: 
- Changes "Photo Caption" to "Text" in the Post component if the post is a `text` post (if it is a photo post, it remains as "Photo Caption"!)
- Extends the width of the Post component if this the Post type is text 
- Changes the validation to show up to 256 characters

**Signup page text Post:** 
screen shot 2018-03-28 at 10 18 35 am

**Signup page photo Post:** 
screen shot 2018-03-28 at 10 17 39 am

**Inbox page text Post:**

**Inbox page photo Post:**
screen shot 2018-03-28 at 10 42 22 am

👀 

Is this clear that this is a text post and not that the photo is missing? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156173991) Pivotal card.

#### Checklist
- [ ] Documentation added for new feature
screen shot 2018-03-28 at 10 43 40 am
s/changed endpoints.
- [ ] Tested on staging.

### DIFF
--- a/app/Http/Requests/Three/PostRequest.php
+++ b/app/Http/Requests/Three/PostRequest.php
@@ -30,7 +30,7 @@ class PostRequest extends Request
             'type' => 'required|string|in:photo,voter-reg,text',
             'action' => 'required|string',
             'why_participated' => 'nullable|string',
-            'text' => 'required|nullable|string|max:255',
+            'text' => 'required|nullable|string|max:256',
             'quantity' => 'nullable|integer',
             'file' => 'image|dimensions:min_width=400,min_height=400',
             'status' => 'in:pending,accepted,rejected',
--- a/app/Http/Requests/Three/PostRequest.php
+++ b/app/Http/Requests/Three/PostRequest.php
@@ -30,7 +30,7 @@ class PostRequest extends Request
             'type' => 'required|string|in:photo,voter-reg,text',
             'action' => 'required|string',
             'why_participated' => 'nullable|string',
-            'text' => 'required|nullable|string|max:140',
+            'text' => 'required|nullable|string|max:255',
             'quantity' => 'nullable|integer',
             'file' => 'image|dimensions:min_width=400,min_height=400',
             'status' => 'in:pending,accepted,rejected',
--- a/resources/assets/components/Post/__snapshots__/test.js.snap
+++ b/resources/assets/components/Post/__snapshots__/test.js.snap
@@ -5,7 +5,7 @@ exports[`it renders correctly 1`] = `
   className="post container__row"
 >
   <div
-    className="container__block -third images"
+    className="container__block -third"
   >
     <div
       className="post__image"
@@ -44,9 +44,7 @@ exports[`it renders correctly 1`] = `
           >
             Photo Caption
           </h4>
-          <p>
-            Here is my awesome caption!
-          </p>
+          <p />
         </div>
       </div>
       <div
--- a/resources/assets/components/Post/index.js
+++ b/resources/assets/components/Post/index.js
@@ -61,6 +61,7 @@ class Post extends React.Component {
     const campaign = this.props.campaign;
     const quantity = post.quantity != null ? post.quantity : 0;
     const containerSize = post.type === 'photo' ? '-third' : '-half';
+    const textOrCaption = post.type === 'photo' ? 'Photo Caption' : 'Text';
 
     return (
       <div className="post container__row">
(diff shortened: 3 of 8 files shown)


sbsmith86:
> Is this clear that this is a text post and not that the photo is missing?

My two cents is that it does seem to just look like the photo is missing. I think there was some work slated for adding `type` to the post card so that we can use that for context.

app/Http/Requests/Three/PostRequest.php
```diff
@@ -30,7 +30,7 @@ public function rules()
             'type' => 'required|string|in:photo,voter-reg,text',
             'action' => 'required|string',
             'why_participated' => 'nullable|string',
-            'text' => 'required|nullable|string|max:140',
+            'text' => 'required|nullable|string|max:255',
```
sbsmith86: did this not work at 256?

chloealee: oh oops - I thought that `255` would let you do 256 characters - I can change that! But the database doesn't need to be updated, was that it?

chloealee:
@sbsmith86 ok that makes sense to me!

resources/assets/components/Post/index.js
```diff
@@ -60,39 +60,42 @@ class Post extends React.Component {
...
-                map(this.getOtherPosts(post), (post, key) => <PostTile key={key} details={post} />)
-              }
-            </ul>
-            : null}
-        </div>
+          : null}
 
         {/* User and Post information */}
-        <div className="container__block -third">
+        <div className={"container__block " + containerSize}>
```
sbsmith86: Can a template string work here?

`container__block  -${containerSize}`

resources/assets/components/Post/index.js
```diff
@@ -105,7 +108,7 @@ class Post extends React.Component {
               : null}
 
             <div className="container -padded">
-              <TextBlock title="Photo Caption" content={displayCaption(post)} />
+              <TextBlock title="Text" content={displayCaption(post)} />
```
sbsmith86: Could we call this `Text` if it is of `type=text` and "photo caption" if it is of `type=photo`?

app/Http/Requests/Three/PostRequest.php
sbsmith86: In the DB we set the column type to be VARCHAR 255. And that means the column should be able to hold 256 character strings. 

So then in the validation we just want to limit it to 256. But it might be worth double checking that a 256 character string is able to be stored in that column which will determine if any DB updates need to be made.

sbsmith86:
Could we possible ignore those failing tests in this PR so that it unblocks other PRs while the scope one is waiting to be merged?

resources/assets/components/Post/index.js
chloealee: hmm I tried to do that but it didn't like it :-/

sbsmith86: Oh sorry you need the llittle ticks 
```
`container__block -${containerSize}`
```

resources/assets/components/Post/index.js
chloealee: ooo that is a good idea!

resources/assets/components/Post/index.js
chloealee: oh yes - this reads so much better!

chloealee:
Yes on that now!

chloealee:
@sbsmith86 everything is passing now!

resources/assets/components/Post/__snapshots__/test.js.snap
```diff
@@ -44,9 +44,7 @@ exports[`it renders correctly 1`] = `
           >
             Photo Caption
           </h4>
-          <p>
-            Here is my awesome caption!
-          </p>
+          <p />
```
sbsmith86: whoa is this a self closing `<p>` tag?

chloealee: oh weird! I just updated the snapshot and it added that in!

resources/assets/components/Post/index.js
```diff
@@ -60,39 +60,43 @@ class Post extends React.Component {
     const signup = this.props.signup;
     const campaign = this.props.campaign;
     const quantity = post.quantity != null ? post.quantity : 0;
+    const containerSize = post.type === 'photo' ? '-third' : '-half';
+    const textOrCaption = post.type === 'photo' ? 'Photo Caption' : 'Text';
```
sbsmith86: This is a `TODO` for later, but it would be cool if we can return a standard property for this value in the JSON response for a post so the post can decide how the `text` field should be described instead of us having to hard code it. Related to this ticket: https://www.pivotaltracker.com/story/show/154744799

chloealee: oh yes I love that!

PR #652, opened 2018-03-30
Update import signups command to use for RRR imports

katiecrane:
#### What's this PR do?
- Added business logic into `ImportSignupsCommand` which happens after we create a signup (`SendSignupToQuasar`, `SendSignupToBlink`)
- updated `SendSignupToBlink` to be local dev friendly (you are able to hit it regardless of if you are sending to Blink and it will log accordingly, so you can be sure it is getting hit at the right times)
- `ImportSignupsCommand` no longer requires a `created_at` and will use `now()` if one is not given
- updated our test `.env` to not send to Blink
- lock file is updated because I had to run `composer update`
- updated script to use source `phoenix-next` instead of `sms`

Is the update to our test `.env` okay? 

#### Relevant tickets
[Card](https://www.pivotaltracker.com/story/show/156354146)

#### How to deploy

### DIFF
--- a/app/Console/Commands/ImportSignupsCommand.php
+++ b/app/Console/Commands/ImportSignupsCommand.php
@@ -74,7 +74,7 @@ class ImportSignupsCommand extends Command
                     'campaign_id' => $missing_signup['campaign_node_id'],
                     'campaign_run_id' => $missing_signup['campaign_run_id'],
                     'source' => 'sms',
-                    'created_at' => $missing_signup['signup_created_at_timestamp'] ? $missing_signup['signup_created_at_timestamp']: Carbon::now(),
+                    'created_at' => $missing_signup['signup_created_at_timestamp'] ? $missing_signup['signup_created_at_timestamp'] : Carbon::now(),
                 ]);
 
                 if ($signup->id % $logfreq == 0) {
--- a/app/Console/Commands/ImportSignupsCommand.php
+++ b/app/Console/Commands/ImportSignupsCommand.php
@@ -7,6 +7,8 @@ use League\Csv\Reader;
 use Rogue\Models\Signup;
 use Rogue\Services\Registrar;
 use Illuminate\Console\Command;
+use Rogue\Jobs\SendSignupToBlink;
+use Rogue\Jobs\SendSignupToQuasar;
 
 class ImportSignupsCommand extends Command
 {
--- a/app/Console/Commands/ImportSignupsCommand.php
+++ b/app/Console/Commands/ImportSignupsCommand.php
@@ -78,6 +78,10 @@ class ImportSignupsCommand extends Command
                 if ($signup->id % $logfreq == 0) {
                     info('rogue:signupimport: Created signup ' . $signup->id);
                 }
+
+                // Business Logic
+                SendSignupToQuasar::dispatch($signup);
+                SendSignupToBlink::dispatch($signup);
             } else {
                 if ($existing_signup->id % $logfreq == 0) {
                     info('rogue:signupimport: Signup ' . $existing_signup->id . ' already exists! Moving on.');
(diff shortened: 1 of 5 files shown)


app/Console/Commands/ImportSignupsCommand.php
```diff
@@ -70,13 +73,17 @@ public function handle()
                     'northstar_id' => $missing_signup['northstar_id'],
                     'campaign_id' => $missing_signup['campaign_node_id'],
                     'campaign_run_id' => $missing_signup['campaign_run_id'],
-                    'source' => 'sms',
-                    'created_at' => $missing_signup['signup_created_at_timestamp'],
+                    'source' => 'phoenix-next',
+                    'created_at' => $missing_signup['signup_created_at_timestamp'] ? $missing_signup['signup_created_at_timestamp'] : Carbon::now(),
```
sbsmith86: If this is just doing an insert into the `signups` table, it should set a `created_at` automatically, right? I don't think we will need to set it manually as `now()`

app/Console/Commands/ImportSignupsCommand.php
```diff
@@ -70,13 +73,17 @@ public function handle()
...
                 ]);
 
                 if ($signup->id % $logfreq == 0) {
                     info('rogue:signupimport: Created signup ' . $signup->id);
                 }
+
+                // Business Logic
```
sbsmith86: shouldn't se use the `SignupService` so that we can just handle this automatically?

app/Console/Commands/ImportSignupsCommand.php
```diff
@@ -70,13 +73,17 @@ public function handle()
                     'northstar_id' => $missing_signup['northstar_id'],
                     'campaign_id' => $missing_signup['campaign_node_id'],
                     'campaign_run_id' => $missing_signup['campaign_run_id'],
-                    'source' => 'sms',
-                    'created_at' => $missing_signup['signup_created_at_timestamp'],
+                    'source' => 'phoenix-next',
```
sbsmith86: can the `source` be provided in the csv so that we don't have to hard code this?

app/Console/Commands/ImportSignupsCommand.php
DFurnes: 💼

DFurnes: That makes good sense to me.

app/Console/Commands/ImportSignupsCommand.php
katiecrane: @sbsmith86 that is correct but we still allow the option to backdate so we need to be able to set the `created_at` if provided

sbsmith86: @katiecrane if we do something like @DFurnes outlined, I think that will allow us to either add the column for backdating or not.

app/Console/Commands/ImportSignupsCommand.php
katiecrane: @sbsmith86 @DFurnes we'd have to use `v2` because this needs to support backdating, so I wasn't sure if that was a path we wanted to go down

app/Console/Commands/ImportSignupsCommand.php
katiecrane: @sbsmith86 yeah it could be! I wasn't sure how future-proof I should make this

app/Console/Commands/ImportSignupsCommand.php
katiecrane: @sbsmith86 @DFurnes I went the quick and less elegant route since I think we need to run this today. Would you both be okay holding off on these updates until next time we need to import signups?

app/Console/Commands/ImportSignupsCommand.php
katiecrane: Oh and some history! I believe that this originally did not use the signup service because we were backfilling millions of mobile commons signups and we did not want those to be sent to Customer.io at the time

PR #657, opened 2018-04-03
Fix reaction counts & boolean on `v3/posts/:id` route.

DFurnes:
#### What's this PR do?
We were previously only loading the `reacted` and `reaction_count` properties on the [post index](https://github.com/DoSomething/rogue/blob/b0af355ddb2e29c126a71aa5d55a13bceda7453c/app/Http/Controllers/PostsController.php#L64-L97), which means the same posts ran through `PostTransformer` on `posts/{postId}` (or anywhere else) would appear to have no reactions. 🙅‍♂️ 

This pull request updates the `Post` model to always eager-load the `reaction_count` property on posts, and adds an (also eager-loaded) `reaction` has-one relationship to get the current user's reaction if it exists. This simplifies logic in that index method a bit & makes everything consistent!

I added a failing test in f207c79, which [shows how the response is incorrect](https://app.wercker.com/dosomething/rogue/runs/build/5ac3b6783ada9c00016a3a9f?step=5ac3b6b696f3f30001371e43).

I then fixed `reactions.total` in 4566058 and `reactions.reacted` in 7f418b7, and a little bonus fix to make sure we always load `tags` in `4c6c04f`.

This was blocking my work on using the v3 APIs to power the new Phoenix Next galleries.

#### Relevant tickets
[#155944555](https://www.pivotaltracker.com/story/show/155944555)

### DIFF
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -71,7 +71,6 @@ class PostsController extends ApiController
     public function index(Request $request)
     {
         $query = $this->newQuery(Post::class)
-            ->withCount('reactions')
             ->with('tags')
             ->orderBy('created_at', 'desc');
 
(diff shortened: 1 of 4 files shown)


DFurnes:
Database queries look nice and clean as well. For __PostController@index__:

postcontroller index


And for __PostController@show__:

postcontroller show

PR #659, opened 2018-04-03
Cleans up tests and makes sure AsOwner tests are always passing 

chloealee:
#### What's this PR do?
- Cleans up Signup, Post, Reaction, and Tags Tests to use `withAdminAccessToken()` where possible and deletes unneccessary params in some `withAccessToken()` cases. 
- Comments back in incomplete Signup Tests that are now passing (this was a wercker config issue)!
- Explicitly sets rejected posts' to a new `northstar_id` variable to get `testPostsIndexAsOwner` tests to pass every time
- Fixes the `testImportingSignupsWithNoDuplicates` test

- All tests should pass! 

- Wercker was using [`rogue-qa-client`](https://aurora-qa.dosomething.org/clients/rogue-qa-client) in [Aurora](https://aurora-qa.dosomething.org/clients). This for some reason isn't allowing us to update this client and it is stuck on Authorization code grant instead of Machine (Client Credentials grant). Other clients aren't updating as expected either and just redirect to https://aurora-qa.dosomething.org/users when pressing "Save" Card to explore [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156516089) here!

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/155479565) Pivotal Card.

#### How to deploy

### DIFF
--- a/tests/Console/ImportSignupsCommandTest.php
+++ b/tests/Console/ImportSignupsCommandTest.php
@@ -11,7 +11,7 @@ class ImportSignupsCommandTest extends TestCase
     public function testImportingSignupsWithNoDuplicates()
     {
         // Timestamp right before running the script
-        $start_time = Carbon::now();
+        $start_time = $this->mockTime(Carbon::now());
 
         // Run the signup import command
         $this->artisan('rogue:signupimport', ['path' => 'tests/Console/example-signups.csv']);
@@ -37,14 +37,9 @@ class ImportSignupsCommandTest extends TestCase
         $all_signups = Signup::all();
         $this->assertTrue($all_signups->count() == 2);
 
-        // The below is failing and we're not sure why!
-        // @TODO: come back and investigate!
-
-        $this->markTestIncomplete();
-
         // Make sure the 'updated_at' timestamps are not backdated
-        // foreach ($all_signups as $signup) {
-        //     $this->assertTrue($signup->updated_at->gte($start_time));
-        // }
+        foreach ($all_signups as $signup) {
+            $this->assertTrue($signup->updated_at->gte($start_time));
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


chloealee:
haha of course the wercker build can't get those tests I thought were fixed to pass! I'm looking into that more now! :)

chloealee:
I'd like to also fix up the [`ImportSignupsCommandTest`](https://github.com/DoSomething/rogue/blob/d1df55dccc4d6a1b9716772be73dc4402ea5be7c/tests/Console/ImportSignupsCommandTest.php) and it looks like it is failing on [this](https://github.com/DoSomething/rogue/blob/d1df55dccc4d6a1b9716772be73dc4402ea5be7c/tests/Console/ImportSignupsCommandTest.php#L47) line.

When `dd($start_time, $signup->updated_at);` I get the below: 

```
Carbon\Carbon @1522867681 {#1334
  date: 2018-04-04 18:48:01.616521 UTC (+00:00)
}
Illuminate\Support\Carbon @1522867681 {#1473
  date: 2018-04-04 18:48:01.0 UTC (+00:00)
}
```

Wondering if there is something wrong with this because it looks as if the `$start_time` is after the `$signup->updated_at` which doesn't make sense...  cc @katiecrane

chloealee:
actually figured out how to get rid of the milliseconds that it is comparing by sending through the [`mockTime()`](https://github.com/DoSomething/rogue/blob/9d68e5fd550d3cd2f0686c17663c4fa6d9cb0f1a/tests/WithMocks.php#L41) function! This also makes sense why this would be failing because we don't save milliseconds in the database so but the `$start_time` did include this so failed because it thought the `$start_time` was after the `$signup->updated_at` time.

PR #660, opened 2018-04-05
Update Signup Component to use v3 endpoints

chloealee:
#### What's this PR do?
- Updates the Signup Component to use v3 endpoints 
- Updates Post Repository to set `status` or `source` to what is sent in the request, if it is set. This is because we want the `status` to not always be pending (when uploading a file on behalf of a user it should automatically be accepted) and the `source` to be `rogue-admin` if it is being uploaded instead of something like `dev-oauth`. 

👀 

#### Relevant tickets
Fixes tasks 1 and 3 in [this](https://www.pivotaltracker.com/n/projects/2019429/stories/155907208) Pivotal Card.

#### How to deploy


app/Http/Requests/PostRequest.php
```diff
@@ -30,7 +30,7 @@ public function rules()
             'why_participated' => 'nullable|string',
             'text' => 'required|nullable|string|max:256',
             'quantity' => 'nullable|integer',
-            'file' => 'image|dimensions:min_width=400,min_height=400',
+            // 'file' => 'image|dimensions:min_width=400,min_height=400',
             'status' => 'in:pending,accepted,rejected',
```
chloealee: Do we need this validation? With this validation, I wasn't able to upload a photo in Rogue Admin because it threw a `422 (Unprocessable Entity)` error.

sbsmith86: i'm fine with removing this. but can we add a `TODO` to figure out the file validation?

DFurnes: Wait, we definitely want this! (Unless it's triggering for `text` posts? What's the error?)

chloealee: @DFurnes when I try to upload a photo in Rogue admin via the uploader, with this validation, it throws a `422 (Unprocessable Entity)` :-/

DFurnes: What are the details of the validation error?

app/Repositories/PostRepository.php
```diff
@@ -87,8 +87,8 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
             'action' => isset($data['action']) ? $data['action'] : null,
             'url' => $fileUrl,
             'text' => isset($data['text']) ? $data['text'] : null,
-            'status' => 'pending',
-            'source' => token()->client(),
+            'status' => isset($data['status']) ? $data['status'] : 'pending',
```
sbsmith86: I'm not sure we want to do this here. 

Down below on [lines 100 -> 104](https://github.com/DoSomething/rogue/blob/ee23e4dbd699ed45ae9526b78c37f4b7a8b1a46f/app/Repositories/PostRepository.php#L100-L104) I added some logic that checks the token scope or admin role and decides if it is allow to update those fields. 

If  request is coming from Rogue Admin that is using a `client_id=rogue-qa` (or whatever) then We check the scopes on that client token that says, yes you can update status and we will use your client id as the source. 

the client `dev-outh` should only be used on dev machines, so it's more likely that we want to use the `client_id` for source than not. For example, Chompy uses a `importer-client` id so then when we create posts (because that client has the right scopes) we set the status that is passed in the request and set the source as the `token()->client()`

resources/assets/components/ReportbackUploader/index.js
```diff
@@ -49,11 +49,11 @@ class ReportbackUploader extends React.Component {
 
     const reportback = {
       media: this.state.media,
-      caption: this.caption.value,
+      text: this.caption.value,
```
sbsmith86: A total aside but can we update the name of this to `PostUploader`

app/Http/Requests/PostRequest.php
sbsmith86: also can we remove the requirement for `text` since not all posts will have this.

resources/assets/components/ReportbackUploader/index.js
```diff
@@ -49,11 +49,11 @@ class ReportbackUploader extends React.Component {
...
       campaignId: this.props.campaignId,
       campaignRunId: this.props.campaignRunId,
       northstarId: this.props.northstarId,
-      source: this.props.source,
+      source: 'rogue-admin',
```
sbsmith86: Instead of passing this, we should let the `PostRepository` set the source as the incoming client id.

resources/assets/components/ReportbackUploader/index.js
```diff
@@ -120,8 +120,8 @@ class ReportbackUploader extends React.Component {
 }
 
 ReportbackUploader.propTypes = {
-  campaignId: PropTypes.number.isRequired,
-  campaignRunId: PropTypes.number.isRequired,
+  campaignId: PropTypes.string.isRequired,
```
sbsmith86: It is odd that `campaignID` and `campaign_run_id` are different types, any reason why?

resources/assets/components/Signup/index.js
```diff
@@ -296,27 +296,29 @@ class Signup extends React.Component {
     }
   }
 
-  // Submit a new reportback on behalf of the user.
+  // @TODO: make this work for any type of post and rename to submitPost?
+  // Submit a new photo reportback on behalf of the user.
   submitReportback(reportback) {
```
sbsmith86: can we rename this to `submitPost`

app/Repositories/PostRepository.php
DFurnes: Yeah, we for sure only want this to be overridden by admins/machines. ➕

resources/assets/components/Signup/index.js
```diff
@@ -296,27 +296,29 @@ class Signup extends React.Component {
...
       quantity: reportback.impact,
       why_participated: reportback.whyParticipated,
-      caption: reportback.caption,
+      text: reportback.text,
       source: reportback.source,
       status: reportback.status,
       file: reportback.media.dataURL,
-      remote_addr: '',
+      type: 'photo',
+      // @TODO: add action to the form as an optional field?
```
sbsmith86: hmmm interesting... let's chat with Jen about this. But I think an optional field will make sense at somepoint

resources/assets/components/Signup/index.js
```diff
@@ -296,27 +296,29 @@ class Signup extends React.Component {
...
+      action: 'default',
     };
-
     // Make API request to Rogue to upload post
-    const request = this.api.post('posts', fields);
+    const request = this.api.post('api/v3/posts', fields);
 
     request.then((result) => {
-      // Update the state
+       // Update the state
```
sbsmith86: 👾

app/Http/Requests/PostRequest.php
chloealee: oh right @sbsmith86 ! 

@DFurnes it doesn't give much more info. than that :-/ 
screen shot 2018-04-05 at 2 08 32 pm

app/Repositories/PostRepository.php
chloealee: ah yes this totally makes sense and didn't see the code below! It wasn't working before so that is why I had to update the `$isAdmin` line - I'll write more about it in the change!

resources/assets/components/ReportbackUploader/index.js
DFurnes: (Idle thought, not something to fix in this PR:) We should really standardize that all IDs are requested and returned as strings, regardless of actual values. We'd made that call a long while back in Ashes, but didn't really do a great job of sticking to it. Otherwise it adds some awkward cognitive overhead when switching between querying different apps.

app/Repositories/PostRepository.php
```diff
@@ -94,7 +94,7 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
             'remote_addr' => request()->ip(),
         ]);
 
-        $isAdmin = isset($data['status']) && isset(auth()->user()->role) && auth()->user()->role === 'admin';
+        $isAdmin = auth()->user()->role && auth()->user()->role === 'admin';
         $hasAdminScope = in_array('admin', token()->scopes());
```
chloealee: I don't think we need the `isset($data['status']` check here since we do this below! Also, for some reason `isset(auth()->user()->role)` was returning an error so that is why when I uploaded a photo via the uploader, the post in was marked as `pending`. I updated this and is working now!

sbsmith86: hmm i think we still need to check if exists. If is a machine client_id, auth role won't be available to us, just `token()->scope()`

resources/assets/components/ReportbackUploader/index.js
sbsmith86: @DFurnes yea I agree, that should just be a normal convention.

app/Repositories/PostRepository.php
DFurnes: Yeah, I'd write this as `auth()->user() && auth()->user()->role === 'admin';`

resources/assets/components/ReportbackUploader/index.js
chloealee: cool! I added a card [here](https://www.pivotaltracker.com/n/projects/2019429/stories/156552592)! cc @DFurnes @sbsmith86

resources/assets/components/Signup/index.js
chloealee: cool! I created a card [here](https://www.pivotaltracker.com/n/projects/2019429/stories/156552721) :)

resources/assets/components/Signup/index.js
```diff
@@ -296,24 +299,26 @@ class Signup extends React.Component {
...
-      remote_addr: '',
+      northstar_id: post.northstarId,
+      campaign_id: post.campaignId,
+      campaign_run_id: post.campaignRunId,
+      quantity: post.impact,
+      why_participated: post.whyParticipated,
+      text: post.text,
+      source: post.source,
+      status: post.status,
+      file: post.media.dataURL,
```
DFurnes: We want to upload the actual file here, not a Data URL string.

I believe that's `post.media.file` (set by the [MediaUploader](https://github.com/DoSomething/rogue/blob/master/resources/assets/components/MediaUploader/index.js)).

DFurnes: Let's also make sure we're on the latest Gateway by running an `npm update`!

chloealee: ah! I changed that but still getting this: 
screen shot 2018-04-06 at 10 48 57 am

DFurnes: Can you see what's in the details of the 422 response (click the "arrow" on the left of the `GATEWAY: Uprocessable Entity` error? Is it still complaining about the `file` field?

I'm also curious what the request payload looks like - you should be able to find the post request in Chrome's web inspector's "Network" tab, then click on it and scroll to the bottom of the "headers" tab to see what's in "Request Payload".

chloealee: ah so interesting I didn't know you could do that! 

Here's what the 422 response says when I click the "arrow":
screen shot 2018-04-06 at 11 03 01 am

Digging around I'm seeing some inconsistencies though. Here's the response that says I didn't pass these fields but they are definitely being passed...
screen shot 2018-04-06 at 11 05 04 am

Then it says that I'm not sending `multipart/form-data` when we just updated so it should be...
screen shot 2018-04-06 at 11 05 50 am

And lastly, it looks like the file is empty? 
screen shot 2018-04-06 at 11 06 01 am

chloealee: looks like I am sending the blob in the request (`console.log(fields)` before sending the request): 
screen shot 2018-04-06 at 11 10 41 am

DFurnes:
Ah, found it. Gateway JS is a little tricky and requires a `FormData` to upload as `multipart/form-data`:

```diff
+/* global FormData */
+
 // Utilities
 import React from 'react';
 import PropTypes from 'prop-types';
-import { map, startCase, keyBy, filter, isEmpty } from 'lodash';
+import { map, forEach, startCase, keyBy, filter, isEmpty } from 'lodash';
 import { RestApiClient } from '@dosomething/gateway';

 // Components
@@ -317,8 +319,12 @@ class Signup extends React.Component {
       // @TODO: add action to the form as an optional field?
       action: 'default',
     };
+
+    const payload = new FormData();
+    forEach(fields, (value, key) => payload.append(key, value));
+
     // Make API request to Rogue to upload post
-    const request = this.api.post('api/v3/posts', fields);
+    const request = this.api.post('api/v3/posts', payload);

     request.then((result) => {
       // Update the state
```

chloealee:
ah tricky @DFurnes ! Thanks so much for all your help on this! I updated to what you did above but when I load the page now get this error: 
screen shot 2018-04-06 at 12 01 57 pm

Not sure why because this is pointing to an error in the [`Tags/index.js`](https://github.com/DoSomething/rogue/blob/master/resources/assets/components/Tags/index.js#L41)?

DFurnes:
@chloealee Can you try re-starting `npm start`? Seems like that same undefined import bug.

resources/assets/components/PostUploader/index.js
```diff
@@ -47,33 +47,32 @@ class ReportbackUploader extends React.Component {
...
-      source: this.props.source,
       status: 'accepted',
     };
 
     const signup = this.props.signup;
 
     if (this.quantity.value) {
-      reportback.impact = this.quantity.value;
-      signup.quantity = reportback.impact;
+      post.impact = this.quantity.value;
```
DFurnes: It might make sense to rename this to `quantity` since that's what it's called in the v3 API.

resources/assets/components/Signup/index.js
```diff
@@ -296,24 +301,29 @@ class Signup extends React.Component {
     }
   }
 
-  // Submit a new reportback on behalf of the user.
-  submitReportback(reportback) {
-    // Fields to send to /posts
+  // @TODO: make this work for any type of post?
+  // Submit a new photo post on behalf of the user.
+  submitPost(post) {
+    console.log(post);
```
DFurnes: Can remove this `console.log`! 👾

PR #663, opened 2018-04-06
Command re-send posts and signups to Quasar

katiecrane:
#### What's this PR do?
- New command to send posts and signups within a given date range (or all, if none given) to Quasar
    - Dates are based on the `updated_at` timestamps on the posts and signups
- Adds 3 tests for each scenario that the command can be run with (no arguments, start date given, and start and end date both given)

Does using `updated_at` make sense? Is it okay to assume we always want to send both posts and signups?

#### Relevant tickets
[Card](https://www.pivotaltracker.com/n/projects/2019429/stories/155196205)

#### How to deploy

### DIFF
--- a/app/Console/Commands/SendToQuasar.php
+++ b/app/Console/Commands/SendToQuasar.php
@@ -56,7 +56,7 @@ class SendToQuasar extends Command
         }
 
         $signups->chunkById(1000, function ($signups) {
-            foreach($signups as $signup) {
+            foreach ($signups as $signup) {
                 SendSignupToQuasar::dispatch($signup);
             }
         });
@@ -69,7 +69,7 @@ class SendToQuasar extends Command
         }
 
         $posts->chunkById(1000, function ($posts) {
-            foreach($posts as $post) {
+            foreach ($posts as $post) {
                 SendPostToQuasar::dispatch($post);
             }
         });
--- a/app/Console/Commands/SendToQuasar.php
+++ b/app/Console/Commands/SendToQuasar.php
@@ -2,7 +2,11 @@
 
 namespace Rogue\Console\Commands;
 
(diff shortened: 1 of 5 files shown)


app/Console/Commands/SendToQuasar.php
```diff
@@ -0,0 +1,79 @@
...
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        info('rogue:quasar - Starting to send to Quasar');
+
+        $start = $this->argument('start');
+        $end = $this->argument('end');
```
chloealee: will this be defined is nothing is passed in as the `start` param?

app/Console/Commands/SendToQuasar.php
```diff
@@ -0,0 +1,79 @@
...
+
+class SendToQuasar extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'rogue:quasar
+                            {start=0000-00-00 : Include all signups and posts updated on or after this day. Format: YYYY-MM-DD}
```
sbsmith86: lol is the default the beginning of time?

katiecrane: haha yeah to grab all of them! I started out by checking if a `start` was provided or not, but it seemed like this way was going to be less code and easier to follow.

chloealee: ah I see now!

app/Console/Commands/SendToQuasar.php
katiecrane: If only `start` (or no arguments) are passed, then `$end` gets set to `null`

PR #664, opened 2018-04-09
Updates CampaignSingle component to use v3 endpoints

chloealee:
#### What's this PR do?
- Updates CampaignSingle component to use v3 endpoints. 
- `PostTransformer` now includes the `user` as well if the signup is included (user is returned under the signup). 

- 👀 
- `CamapaignSingle` page `/campaigns/:campaign_id` should work exactly the same as before with no errors!

#### Relevant tickets
Fixes second task in [this](https://www.pivotaltracker.com/n/projects/2019429/stories/155907208) Pivotal Card.

#### How to deploy


resources/assets/utilities/RogueClient.js
```diff
@@ -2,7 +2,7 @@ import { RestApiClient } from '@dosomething/gateway';
 
 class RogueClient extends RestApiClient {
   getPosts(options) {
-    return this.get('posts', options);
+    return this.get('api/v3/posts', options);
```
chloealee: This is the initial call on page load where the campaign single page first grabs posts!

app/Http/Transformers/PostTransformer.php
```diff
@@ -69,7 +69,10 @@ public function transform(Post $post)
      */
     public function includeSignup(Post $post)
     {
-        return $this->item($post->signup, new SignupTransformer);
+        // Don't include posts but include the user.
```
sbsmith86: I'm not sure I understand this comment. You might not need it anyway because I think the code is pretty clear!

resources/assets/components/CampaignSingle/index.js
```diff
@@ -93,18 +95,18 @@ class CampaignSingle extends React.Component {
...
       .then((json) => {
         this.setState({ loadingNewPosts: false });
         this.props.setNewPosts(json);
       });
   }
 
-  // Make API call to GET /posts to get posts by filtered status and/or tag(s).
+  // Make API call to GET api/v3/posts to get posts by filtered status and/or tag(s).
   getPostsByFilter(filters) {
     this.setState({ loadingNewPosts: true });
```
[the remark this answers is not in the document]
chloealee: @sbsmith86 yup! That is done [here](https://github.com/DoSomething/rogue/blob/master/resources/assets/components/CampaignSingle/index.js#L112)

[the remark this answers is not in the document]
sbsmith86: hmmm I see that in an old version of the code, but in this PR it doesn't seem like it is there? I might be losing my mind.

resources/assets/utilities/RogueClient.js
chloealee: Actually it's a little tricky! Here's how (I _think_) it works: 
- [`campaign_single.blade.php` passes `data-reviewing="true"` in the `wrapper`](https://github.com/DoSomething/rogue/blob/5cf34bb0adf4d88ff231a16a5b279bd20ba58930/resources/views/pages/campaign_single.blade.php#L8)
- The `data-reviewing="true"` then called the `WithReviewing` component to [`getPostsByStatus`](https://github.com/DoSomething/rogue/blob/5cf34bb0adf4d88ff231a16a5b279bd20ba58930/resources/assets/components/WithReviewing/index.js#L43) to populate the initial posts. 
- The line that you referenced up there gets new posts when you filter on the `CampaignSingle` page. 

Does that seem right?

resources/assets/components/CampaignSingle/index.js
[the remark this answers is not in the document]
chloealee: [Here](https://github.com/DoSomething/rogue/pull/664/files/c6daf9550a023b0d187e3ad4dcd9d6cedcb6efe9#diff-0335bdd7424f67314da72e781e5e027cR114) it is, @sbsmith86 ! You're right tho - it's hidden!

resources/assets/utilities/RogueClient.js
sbsmith86: yea, i guess i'm just clarifying that you are just defining `getPosts` as a method on the `RogueClient` here. You aren't making an actually call. So somewhere you must be calling `RogueClient->getPosts()` which will really make the call. (sorry about that php syntax)

resources/assets/components/CampaignSingle/index.js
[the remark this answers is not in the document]
sbsmith86: gotcha! thanks

resources/assets/utilities/RogueClient.js
chloealee: oh yes!

PR #666, opened 2018-04-10
Allow sending posts/signups to Quasar without logging

katiecrane:
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥  
#### What's this PR do?
- `SendPostToQuasar` and `SendSignupToQuasar` now accept a parameter to suppress logging (if nothing is passed, they still log)
- `SendToQuasar` command now takes in an option (defaults to 1) for how often we should log that a Post or Signup has been sent to Quasar (aka how often we should NOT suppress logging)

Will this keeps the logging to a reasonable volume when running this command over all Posts and Signups?

Without a change like this, we would way overflow our logs.

#### Relevant tickets
[Card](https://www.pivotaltracker.com/n/projects/2019429/stories/155196205)

#### How to deploy

### DIFF
--- a/app/Console/Commands/SendToQuasar.php
+++ b/app/Console/Commands/SendToQuasar.php
@@ -16,6 +16,7 @@ class SendToQuasar extends Command
      * @var string
      */
     protected $signature = 'rogue:quasar
+                            {logFreq=1}
                             {start=0000-00-00 : Include all signups and posts updated on or after this day. Format: YYYY-MM-DD}
                             {end? : Include signups and posts updated up to but not including this day. Format: YYYY-MM-DD}';
 
@@ -45,6 +46,7 @@ class SendToQuasar extends Command
     {
         info('rogue:quasar - Starting to send to Quasar');
 
+        $logFreq = $this->argument('logFreq');
         $start = $this->argument('start');
         $end = $this->argument('end');
 
@@ -55,9 +57,14 @@ class SendToQuasar extends Command
             $signups = $signups->where('updated_at', '<=', $end);
         }
 
-        $signups->chunkById(1000, function ($signups) {
+        $signups->chunkById(1000, function ($signups) use ($logFreq) {
             foreach ($signups as $signup) {
-                SendSignupToQuasar::dispatch($signup);
+                if ($signup->id % $logFreq === 0) {
+                    $log = true;
+                } else {
+                    $log = false;
+                }
+                SendSignupToQuasar::dispatch($signup, $log);
             }
         });
 
@@ -68,9 +75,14 @@ class SendToQuasar extends Command
             $posts = $posts->where('updated_at', '<=', $end);
         }
 
-        $posts->chunkById(1000, function ($posts) {
+        $posts->chunkById(1000, function ($posts) use ($logFreq) {
             foreach ($posts as $post) {
-                SendPostToQuasar::dispatch($post);
+                if ($post->id % $logFreq === 0) {
+                    $log = true;
+                } else {
+                    $log = false;
+                }
+                SendPostToQuasar::dispatch($post, $log);
             }
         });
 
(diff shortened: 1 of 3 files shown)


app/Console/Commands/SendToQuasar.php
```diff
@@ -55,9 +57,14 @@ public function handle()
             $signups = $signups->where('updated_at', '<=', $end);
         }
 
-        $signups->chunkById(1000, function ($signups) {
+        $signups->chunkById(1000, function ($signups) use ($logFreq) {
             foreach ($signups as $signup) {
-                SendSignupToQuasar::dispatch($signup);
+                if ($signup->id % $logFreq === 0) {
```
chloealee: ah smart way to do this!!

app/Jobs/SendPostToQuasar.php
```diff
@@ -47,8 +55,10 @@ public function handle()
         }
 
         // Log
-        $verb = $shouldSendToQuasar ? 'sent' : 'would have been sent';
-        info('Post ' . $this->post->id . ' ' . $verb . ' to Quasar');
+        if ($this->log) {
+            $verb = $shouldSendToQuasar ? 'sent' : 'would have been sent';
+            info('Post ' . $this->post->id . ' ' . $verb . ' to Quasar');
```
chloealee: I'm confused - is this still logging every time but just changes the verb?

katiecrane: @chloealee great question! The if statement determines if we want to log or not, and the stuff inside is what does the actual logging. When we log, we figure out if we are actually sending it to Quasar or not, and log whether we actually sent it, or that we hit the job and would have sent it if Blink was on. So if Blink is on we log:
Post 4 sent to Quasar
If Blink is off we log:
Post 4 would have been sent to Quasar
Let me know if that's still confusing or doesn't make sense!

app/Console/Commands/SendToQuasar.php
```diff
@@ -55,9 +57,14 @@ public function handle()
...
+                    $log = true;
+                } else {
+                    $log = false;
```
DFurnes: I think you could condense this into a one-liner... your call if that's readable or not! 😜

```php
$shouldWriteToLog = (bool) ($signup->id % $logFreq === 0);
SendSignupToQuasar::dispatch($signup, $shouldWriteToLog);
```

app/Jobs/SendPostToQuasar.php
chloealee: that makes sense! But wasn't the problem before that we were running into problems because we were logging everything here and we're still logging the same amount? I think I might be misunderstanding something!

katiecrane: @chloealee Oh! So in the script where we dispatch the job (https://github.com/DoSomething/rogue/pull/666/files#diff-a29f7802c195289adb5fc1ba1033f452R67) we usually send in an argument of `false` so that `$this->log` will be false and it won't log

app/Console/Commands/SendToQuasar.php
sbsmith86: i like it

katiecrane: @DFurnes I like this especially because it works without the casting (like this:                 `$shouldWriteToLog = ($signup->id % $logFreq === 0);`) too! 🍲

DFurnes: Oh right, even better!! 🎬

app/Jobs/SendPostToQuasar.php
chloealee: @katiecrane ohohoh I'm a dummy I was totally reading that wrong - that makes total sense thanks for explaining!

katiecrane: haha no it's confusing! thanks @chloealee !

PR #667, opened 2018-04-10
Allows staff to update posts

chloealee:
#### What's this PR do?
- Adds staff (in addition to admin or user who owns the post) to update a post. 

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156633114) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -144,7 +144,7 @@ class PostsController extends ApiController
             'quantity' => 'nullable|integer',
         ]);
 
-        // Only allow an admin or the user who owns the post to update.
+        // Only allow an admin/staff or the user who owns the post to update.
         if (in_array(token()->role, ['admin', 'staff']) || auth()->id() === $post->northstar_id) {
             $this->posts->update($post, $validatedRequest);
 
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/PostsController.php
```diff
@@ -144,8 +144,8 @@ public function update(Request $request, Post $post)
             'quantity' => 'nullable|integer',
         ]);
 
-        // Only allow an admin or the user who owns the post to update.
-        if (token()->role() === 'admin' || auth()->id() === $post->northstar_id) {
+        // Only allow an admin/staff or the user who owns the post to update.
+        if (in_array(token()->role, ['admin', 'staff']) || auth()->id() === $post->northstar_id) {
```
sbsmith86: Maybe out of the scope of this...but I wonder if there should be a little like `isAllowed()` helper. Cause I see us doing these checks a lot of places. And that helper would also check client scopes.

sbsmith86: Or if something else make more sense, then we could do that...just trying to centralize this logic a little more.

PR #669, opened 2018-04-11
Allow Staff to Review and Tag a Post 

chloealee:
#### What's this PR do?
- Allows staff to review and tag a post.
- Also updates `PostsController` to use `is_staff_user()` helper (suggestion from @sbsmith86 [here](https://github.com/DoSomething/rogue/pull/667#discussion_r180554588)!). 

👀 

#### Relevant tickets
Fixes 🐛 

#### How to deploy

### DIFF
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -145,7 +145,7 @@ class PostsController extends ApiController
         ]);
 
         // Only allow an admin/staff or the user who owns the post to update.
-        if (in_array(token()->role, ['admin', 'staff']) || auth()->id() === $post->northstar_id) {
+        if (is_staff_user() || auth()->id() === $post->northstar_id) {
             $this->posts->update($post, $validatedRequest);
 
             return $this->item($post);
--- a/app/Http/Controllers/ReviewsController.php
+++ b/app/Http/Controllers/ReviewsController.php
@@ -35,6 +35,7 @@ class ReviewsController extends ApiController
 
         $this->middleware('auth:api');
         $this->middleware('role:admin');
+        $this->middleware('role:staff');
         $this->middleware('scopes:write');
         $this->middleware('scopes:activity');
     }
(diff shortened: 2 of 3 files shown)


app/Http/Controllers/ReviewsController.php
```diff
@@ -35,6 +35,7 @@ public function __construct(PostManager $post)
 
         $this->middleware('auth:api');
         $this->middleware('role:admin');
+        $this->middleware('role:staff');
```
DFurnes: I think these should be combined - this will check first "is this user an admin", and then "is this user staff"...

```php
$this->middleware('role:admin,staff');
```

app/Http/Controllers/PostsController.php
```diff
@@ -145,7 +145,7 @@ public function update(Request $request, Post $post)
         ]);
 
         // Only allow an admin/staff or the user who owns the post to update.
-        if (in_array(token()->role, ['admin', 'staff']) || auth()->id() === $post->northstar_id) {
+        if (is_staff_user() || auth()->id() === $post->northstar_id) {
```
DFurnes: At some point, I'd love to extract these into [policy classes](https://laravel.com/docs/5.6/authorization#creating-policies)!

app/Http/Controllers/ReviewsController.php
chloealee: ah yes that is helpful bc the tests were not passing bc it thought it needed _both_ roles, `admin` AND `staff`!

app/Http/Controllers/PostsController.php
chloealee: oohh yes @sbsmith86 what you were talking about [here](https://github.com/DoSomething/rogue/pull/667#discussion_r180554588)!

chloealee: [Here](https://www.pivotaltracker.com/n/projects/2019429/stories/156700437) is the card for this :)

PR #670, opened 2018-04-11
Restructuring Rogue's Directory Structure - Documentation 

chloealee:
#### What's this PR do?
- We did the majority of directory restructuring in #654 but I forgot to restructure the docs! This PR does that. 

#### How should this be reviewed?
👀 

#### Relevant tickets
Finishes work in #654 

#### How to deploy
https://github.com/DoSomething/rogue/blob/master/docs/development/deployments.md

PR #674, opened 2018-04-13
Updates v3 endpoints for consistency 

chloealee:
#### What's this PR do?
- Updates v3 endpoints for consistency 
  - updates `/post/{post}/reactions` to `/posts/{post}/reactions`
  - updates `/reviews` to `/posts/{post}/reviews`
  - updates front end to reflect updates, tests, and documentation 
- Fixes documentation mistakes in #670

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156379108) Pivotal Card.

#### How to deploy


tests/Http/SignupTest.php
```diff
@@ -393,7 +393,7 @@ public function testSignupsIndexAsAdminWithFilters()
         ]);
 
         // Create three more signups with different northstar_id, campaign_id, and campaign_run_id
-        $secondNorthstarId = $this->faker->northstar_id;
+        $secondNorthstarId = '5554eac1a59dbf117e8b456z';
```
katiecrane: Why do we have to set this one explicitly but not the other one?

chloealee: @katiecrane good q! I have seen on my local that since it is randomized, sometimes they get set to the same string. So the results get borked since `$northstarId` and `$secondNorthstarId` think they're the same user.

DFurnes: Ooooh, _tricky!!_ You should also be able to avoid this using Faker's `unique()` provider:

```php
$northstarId = $this->faker->unique()->northstar_id;
//...
$secondNorthstarId = $this->faker->unique()->northstar_id;
```

chloealee: omg that is way better and good to know - thanks @DFurnes ! I'll change now!

docs/endpoints/Legacy/Two/posts.md
```diff
@@ -100,73 +96,53 @@ Example Response:
...
-    Describes the bucket the action is tied to. A campaign could ask for multiple types of actions throughout the life of the campaign.
-  - **quantity**: (int|nullable) optional.
-    The number of reportback nouns verbed. Can be `null`.
+  - **quantity**: (int).
+    The number of reportback nouns verbed. 
   - **why_participated**: (string).
     The reason why the user participated.
-  - **text**: (string).
-    Corresponding text for the post (could be photo caption or other words). 256 max characters.
+  - **caption**: (string).
```
sbsmith86: Did we change this back to `caption`?

chloealee: oh shoot thanks for catching this @sbsmith86 ! I messed up and moved the v3 docs to legacy by accident - will fix now!

chloealee: @sbsmith86 actually this is correct - this is for the `v2` endpoint - I think it just got a little mixed up when I was moving files around but this should be good!

docs/endpoints/posts.md
```diff
@@ -1,42 +1,44 @@
 ## Posts
+All `v3 /posts` endpoints require the `activity` scope. `Create`/`update`/`delete` endpoints also require the `write` scope.
```
chloealee: @sbsmith86 here is it switched around!

PR #675, opened 2018-04-16
v3/posts quantity bug

chloealee:
#### What's this PR do?
- We were seeing a bug that if a post and signup were created at the same time, the quantities were saved as `0` both on the post and the signup. 
- This PR fixes that bug - it adds the missing check if the difference between the new quantity (coming in as the request) and the signup quantity is `0` and this is the first post, set the post and the signup's quantity to `0`. Before it was not check this and thus set the quantities as `0` (the difference). 
  - When the signup is first created, the quantity is correctly set as `2`. However, that gets overriden and set to `0` [here](https://github.com/DoSomething/rogue/blob/master/app/Repositories/Legacy/Two/PostRepository.php#L127) when it re-sets the signup's quantity to a sum of the posts' quantities. In this case, since it thought the post's quantity was `0`, it summed that one quantity and set the signup's quantity to `0`.  

👀 
- Does this logic make sense? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156817767) Pivotal Card.

#### How to deploy

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,7 +1,5 @@
 preset: laravel
 
-linting: true
-
 disabled:
  - concat_without_spaces
  - no_useless_return
--- a/app/Repositories/Legacy/Two/PostRepository.php
+++ b/app/Repositories/Legacy/Two/PostRepository.php
@@ -97,7 +97,10 @@ class PostRepository
 
             // If the quantity difference is negative than we recieved an incremental submission
             // and should just add that to the post.
-            if ($quantityDiff < 0) {
+            // If the quantity difference equals zero and this is the first post,
+            // the post and signup were created at the same time -
+            // store this quantity on both the post and signup.
+            if ($quantityDiff < 0 || ($quantityDiff === 0 && $signup->posts_count === 0)) {
                 $quantityDiff = $data['quantity'];
             } elseif ($quantityDiff === 0 && $signup->posts_count > 0) {
                 // If the quantity difference equals zero, and this is not the first post,


.styleci.yml
```diff
@@ -1,7 +1,5 @@
 preset: laravel
 
-linting: true
```
sbsmith86: You might need to rebase this with master, I added this in a [previous PR.](#676 )

app/Repositories/Legacy/Two/PostRepository.php
```diff
@@ -97,7 +97,10 @@ public function create(array $data, $signupId)
 
             // If the quantity difference is negative than we recieved an incremental submission
             // and should just add that to the post.
-            if ($quantityDiff < 0) {
+            // If the quantity difference equals zero and this is the first post,
+            // the post and signup were created at the same time -
+            // store this quantity on both the post and signup.
+            if ($quantityDiff < 0 || ($quantityDiff === 0 && $signup->posts_count === 0)) {
```
katiecrane: Does the post not exist because it hasn't been saved yet?

app/Repositories/Legacy/Two/PostRepository.php
```diff
@@ -97,7 +97,10 @@ public function create(array $data, $signupId)
 
             // If the quantity difference is negative than we recieved an incremental submission
             // and should just add that to the post.
-            if ($quantityDiff < 0) {
+            // If the quantity difference equals zero and this is the first post,
+            // the post and signup were created at the same time -
```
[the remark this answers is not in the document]
chloealee: Ah that is true that isn't the best check - I can add checking timestamps!

app/Repositories/Legacy/Two/PostRepository.php
sbsmith86: I think we just need to be clear about what we are checking for. This logic doesn't have anything to do with the the time it was created. What happens when someone signsup for a campaign, and comes back a day later to reportback. If this logic is sound then great! But it doesn't mean they were created at the same time.

katiecrane: I guess by "same time" we really mean "via this particular api call to create a new post"

sbsmith86: > Via this particular api call to create a new post

I'm not sure what this means? I think if we needed to solve for the case of when the first post was created, then dope, this works. If we need to solve for a post being created at the same time as a signup then I'm not sure this does that. If we need to solve for only posts being created via the runscope API call, the we should be checking for `source` or something.

sbsmith86: > when a post was created and a signup didn't exist yet. 

So the post was created _before_ the signup? I don't think that should ever happen since we have a `signup_id` constraint in the DB.

sbsmith86: or at least I think we should.....

katiecrane: @sbsmith86 ah sorry. I should have said "when a request was made to create a post, but that user did not yet have a signup for that campaign"

chloealee: @katiecrane yes that is correct! This is updating for the edge case when a post is being created but doesn't yet have a signup. 

I don't think I explained well before but hopefully this is a little clear: 
- A post is being created via the `v2/posts` endpoint.
- A signup for this campaign doesn't exist yet, so it creates a signup [here](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Traits/PostRequests.php#L27) before creating the post.
- The [quantity is saved](https://github.com/DoSomething/rogue/blob/master/app/Repositories/Legacy/Two/SignupRepository.php#L23) on the signup from the request.
- Then when the post is created, it says the `$quantityDiff` is `0` [here](https://github.com/DoSomething/rogue/blob/master/app/Repositories/Legacy/Two/PostRepository.php#L96).
- Since it doesn't go through any of [this](https://github.com/DoSomething/rogue/blob/master/app/Repositories/Legacy/Two/PostRepository.php#L100-L106), it sets the `$post->quantity` to `0` [here](https://github.com/DoSomething/rogue/blob/master/app/Repositories/Legacy/Two/PostRepository.php#L108). 
- The `$signup->quantity` then gets set to `0` [here](https://github.com/DoSomething/rogue/blob/master/app/Repositories/Legacy/Two/PostRepository.php#L127) when it [sums all its posts quantities](https://github.com/DoSomething/rogue/blob/master/app/Models/Signup.php#L182) (which is `0`).

sbsmith86: Ok! thanks for explaining that @chloealee ! This works as a quick fix :)

I would be also interested in your thoughts about how we initially store quantity on the signup when creating a signup for a post. I feel like, the real fix for this bug is in that piece. If we go to create a post that does not have a signup, and the feature flag is on, we should create a signup but we should not store quantity in the signup. Then we should create the post, store the quantity there, and proceed as normal. That seems to be how it works when the signup already exists. 

However, given this is v2 and we want to move away from it, this will work.

chloealee: @sbsmith86 I was thinking that exactly as I was writing out the above! That seems more clear and so saving the quantity on the signup happens all in one place versus saving it here and there etc.

PR #677, opened 2018-04-17
Force delete test records

chloealee:
#### What's this PR do?
- Force deletes test records created by Ghost Inspector and Runscope
  - Signups and Posts from runscope are found by `source` (either `runscope` or `runscope-oauth`). 
  - Signups from Ghost Inspector are found by `why_participated` (`Why! I love to test! Team Bleed for the win! Tongue Cat forever!`)
  - Posts from Ghost Inspector are found by `text` (`Caption! I love to test! Team Bleed for the win! Tongue Cat forever!`)
- Updates some documentation irrelevant to this but wanted to just sneak in here!

👀 

@sheyd re: [sending deleted events to the queue](https://dosomething.slack.com/archives/C6J24ADQW/p1523391687000195), since this is just working directly with the database and force deleting, I don't think it will actually send any notifications when this is deleted. Can you explain a little more to me what you need here so I can make sure I write this in?  

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156512296) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Console/Commands/ForceDeleteTestRecords.php
+++ b/app/Console/Commands/ForceDeleteTestRecords.php
@@ -51,7 +51,7 @@ class ForceDeleteTestRecords extends Command
             $runscopePost->forceDelete();
         }
 
-        $this->info('All Runscope and Ghost Inspector Signups deleted.');
+        $this->info('All Runscope signups and posts deleted.');
 
         // Force delete all signups created by Ghost Inspector tests, identified by 'why_participated.'
         $ghostInspectorSignups = Signup::where('why_participated', 'Why! I love to test! Team Bleed for the win! Tongue Cat forever!')->get();
@@ -59,12 +59,12 @@ class ForceDeleteTestRecords extends Command
             $ghostInspectorSignup->forceDelete();
         }
 
-        // Force delete all psots created by Ghost Inspector tests, identified by
+        // Force delete all posts created by Ghost Inspector tests, identified by caption.
         $ghostInspectorPosts = Post::where('text', 'Caption! I love to test! Team Bleed for the win! Tongue Cat forever!')->get();
         foreach ($ghostInspectorPosts as $ghostInspectorPost) {
             $ghostInspectorPost->forceDelete();
         }
 
-        $this->info('All Runscope and Ghost Inspector Posts deleted.');
+        $this->info('All Ghost Inspector signups and posts deleted.');
     }
 }
--- a/app/Console/Commands/ForceDeleteTestRecords.php
+++ b/app/Console/Commands/ForceDeleteTestRecords.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace Rogue\Console\Commands;
+
+use Rogue\Models\Post;
+use Rogue\Models\Signup;
+use Illuminate\Console\Command;
+
+class ForceDeleteTestRecords extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'rogue:forceDeleteTestRecords';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Force deletes signups and posts made in Runscope and Ghost Inspector tests';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        // Force delete all signups created by Runscope, identified by source.
+        $runscopeSignups = Signup::where('source', 'runscope')->orWhere('source', 'runscope-oauth')->get();
+        foreach ($runscopeSignups as $runscopeSignup) {
+            $runscopeSignup->forceDelete();
+        }
+
+        // Force delete all posts created by Runscope, identified by source.
+        $runscopePosts = Post::where('source', 'runscope')->orWhere('source', 'runscope-oauth')->get();
+        foreach ($runscopePosts as $runscopePost) {
+            $runscopePost->forceDelete();
+        }
+
+        $this->info('All Runscope and Ghost Inspector Signups deleted.');
+
+        // Force delete all signups created by Ghost Inspector tests, identified by 'why_participated.'
+        $ghostInspectorSignups = Signup::where('why_participated', 'Why! I love to test! Team Bleed for the win! Tongue Cat forever!')->get();
+        foreach ($ghostInspectorSignups as $ghostInspectorSignup) {
+            $ghostInspectorSignup->forceDelete();
+        }
+
+        // Force delete all psots created by Ghost Inspector tests, identified by
+        $ghostInspectorPosts = Post::where('text', 'Caption! I love to test! Team Bleed for the win! Tongue Cat forever!')->get();
+        foreach ($ghostInspectorPosts as $ghostInspectorPost) {
+            $ghostInspectorPost->forceDelete();
+        }
+
+        $this->info('All Runscope and Ghost Inspector Posts deleted.');
+    }
+}
--- a/app/Console/Commands/ForceDeleteTestRecords.php
+++ b/app/Console/Commands/ForceDeleteTestRecords.php
@@ -48,10 +48,10 @@ class ForceDeleteTestRecords extends Command
         $this->info('All Runscope signups and posts deleted.');
 
         // Force delete all signups created by Ghost Inspector tests, identified by 'why_participated.'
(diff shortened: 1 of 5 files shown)


sheyd:
@chloealee If we don't receive a delete event to the queue, this will inflate our numbers. However... I think it's worth considering either filtering out runscope tests on our consumer side, or at the very least excluding them from the analytics we report. I'll talk to ⚡️ about it further.

chloealee:
@sheyd totally makes sense! Thanks for following up and let me know what ⚡️ decides to do and I can adjust this based on that! :)

sheyd:
@chloealee Talked to Sohaib, and you don't need to make any changes on your end, we'll just filter out by our queries at reporting layer. Consider ⚡️ signed-off on this PR.

chloealee:
@sheyd great! Thank you for letting me know so quickly! :)

app/Console/Commands/ForceDeleteTestRecords.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        // Force delete all signups created by Runscope, identified by source.
+        $runscopeSignups = Signup::where('source', 'runscope')->orWhere('source', 'runscope-oauth')->get();
```
katiecrane: I don't think I've ever used `orWhere` with the query builder - cool!!

app/Console/Commands/ForceDeleteTestRecords.php
```diff
@@ -0,0 +1,70 @@
...
+        foreach ($runscopeSignups as $runscopeSignup) {
+            $runscopeSignup->forceDelete();
+        }
+
+        // Force delete all posts created by Runscope, identified by source.
+        $runscopePosts = Post::where('source', 'runscope')->orWhere('source', 'runscope-oauth')->get();
+        foreach ($runscopePosts as $runscopePost) {
+            $runscopePost->forceDelete();
+        }
```
katiecrane: This chunk is a duplicate of the chunk above!

app/Console/Commands/ForceDeleteTestRecords.php
```diff
@@ -0,0 +1,70 @@
...
+
+        $this->info('All Runscope and Ghost Inspector Signups deleted.');
```
katiecrane: I think this log should be moved to line 61 after you delete the ghost inspector signups. I think you can also use plain `info('All Runscope and Ghost Inspector Signups deleted.');`

app/Console/Commands/ForceDeleteTestRecords.php
```diff
@@ -0,0 +1,70 @@
...
+
+        // Force delete all signups created by Ghost Inspector tests, identified by 'why_participated.'
+        $ghostInspectorSignups = Signup::where('why_participated', 'Why! I love to test! Team Bleed for the win! Tongue Cat forever!')->get();
+        foreach ($ghostInspectorSignups as $ghostInspectorSignup) {
+            $ghostInspectorSignup->forceDelete();
+        }
+
+        // Force delete all psots created by Ghost Inspector tests, identified by
```
katiecrane: psots -> posts 😉 
and I think you want to add 'caption' to the end of this line

app/Console/Commands/ForceDeleteTestRecords.php
chloealee: me either! :)

app/Console/Commands/ForceDeleteTestRecords.php
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        // Force delete all signups created by Runscope, identified by source.
+        $runscopeSignups = Signup::where('source', 'runscope')->orWhere('source', 'runscope-oauth')->get();
+        foreach ($runscopeSignups as $runscopeSignup) {
```
sbsmith86: Do you need to loop through these or can you just forceDelete the whole collection?

app/Console/Commands/ForceDeleteTestRecords.php
chloealee: @katiecrane this is for posts made by Runscope and the above is for signups made by Runscope!

app/Console/Commands/ForceDeleteTestRecords.php
chloealee: oh shoot you're right! I meant to say `All Runscope Signups and Posts deleted` :) Thanks for catching!

app/Console/Commands/ForceDeleteTestRecords.php
katiecrane: AH I see now! thanks!

app/Console/Commands/ForceDeleteTestRecords.php
```diff
@@ -0,0 +1,70 @@
...
+        // Force delete all posts created by Runscope, identified by source.
+        $runscopePosts = Post::where('source', 'runscope')->orWhere('source', 'runscope-oauth')->get();
+        foreach ($runscopePosts as $runscopePost) {
+            $runscopePost->forceDelete();
+        }
+
+        $this->info('All Runscope and Ghost Inspector Signups deleted.');
+
+        // Force delete all signups created by Ghost Inspector tests, identified by 'why_participated.'
+        $ghostInspectorSignups = Signup::where('why_participated', 'Why! I love to test! Team Bleed for the win! Tongue Cat forever!')->get();
```
sbsmith86: I would love to look at an easier string to search for. I know this would also mean updating the ghost inspector test but I think there is more room for error (especially between teams like bleed and storm) using a string like this. 

Could we update why and caption test statements to be like `why_participated_ghost_test`, `caption_ghost_test` or at least just include these strings so it is easier to isolate?

app/Console/Kernel.php
```diff
@@ -34,6 +35,6 @@ class Kernel extends ConsoleKernel
      */
     protected function schedule(Schedule $schedule)
     {
-        //
+        $schedule->command(ForceDeleteTestRecords::class, ['--force'])->hourly();
```
sbsmith86: do you know if this is going to work on heroku? I know we have to update a cron configuration on the server but I haven't really looked into it.

docs/endpoints/signups.md
```diff
@@ -10,6 +10,8 @@ POST /api/v3/signups
     The drupal node id of the campaign the user is signing up for.
   - **campaign_run_id**: (int) optional.
     The drupal campaign run node id of the campaign run the user is signing up for.
+  - **northstar_id**: (int) optional.
```
sbsmith86: Do we need to be clear about the requirements making this request `asAUser` vs `asClient` because `northstar_id` is only optional when acting as a client right?

app/Console/Commands/ForceDeleteTestRecords.php
chloealee: @katiecrane I tried doing just `info('All Runscope and Ghost Inspector Signups deleted.');` but that didn't output any message!

app/Console/Kernel.php
sbsmith86: I _think_ we need to add this addon https://devcenter.heroku.com/articles/scheduler

And just doing a quick search, I found some other people who have blogged about the set up:

http://chandrewz.github.io/blog/laravel-and-scheduled-jobs-on-heroku
https://www.neontsunami.com/posts/laravel-scheduler-on-heroku

Happy to look into this with you once it is on thor!

app/Console/Commands/ForceDeleteTestRecords.php
katiecrane: hmmm that's weird, that seems to be how our other commands do it

app/Console/Commands/ForceDeleteTestRecords.php
chloealee: oh @sbsmith86 brilliant! Much better - updated!

app/Console/Commands/ForceDeleteTestRecords.php
chloealee: 🤔

app/Console/Commands/ForceDeleteTestRecords.php
chloealee: @sbsmith86 that is a great idea! I'll ping Jen about that now!

docs/endpoints/signups.md
chloealee: @sbsmith86 just updated this! Does this make more sense?

sbsmith86: yes! amazing

app/Console/Kernel.php
chloealee: these are super helpful - thank you @sbsmith86 !

PR #679, opened 2018-04-18
Makes updates to set us up to use Phoenix endpoints instead of Ashes

chloealee:
#### What's this PR do?
- Updates `Phoenix.php` to be `Ashes.php` and adds new `Phoenix.php` service refers to Phoenix, not Ashes so we are ready to use Phoenix endpoints instead of Ashes when they're ready! 
  - Updates all dependencies so everything still works! 

👀 

- Note: this is only updates Rogue to use the Phoenix endpoint to grab one campaign instead of using the Ashes endpoint. Bulk campaigns are still retrieved via Ashes. Once this is available in Phoenix, we'll make the total switch! 
- Right now, the `GET /api/v2/campaigns/{id}` that we use here doesn't return the campaign cause, staff pick, noun, or verb. We use this endpoint to [resolve any missing campaigns](https://github.com/DoSomething/rogue/blob/master/app/Services/CampaignService.php#L79) if not in cache, so this might cause some issues for new campaigns only in Phoenix (and not in Ashes). I think we should wait to merge this until this info. is also returned? 

**UPDATE** updated per @sbsmith86 's [comment below](https://github.com/DoSomething/rogue/pull/679#issuecomment-383145993) to just set this up to use Phoenix endpoints when we're ready! Everything still uses Ashes endpoints.

#### Relevant tickets
Fixes part of [this](https://www.pivotaltracker.com/n/projects/2019429/stories/155672260) Pivotal card. 

#### How to deploy


sbsmith86:
can we use this endpoint just for grabbing stuff that has a mongo db (a string instead of an int)?  Or if there is a way to just have the endpoint ready to use with out actively using it, we can do that for now. 

I also want us to revisit all the campaigns caching logic to make it simpler.

chloealee:
@sbsmith86 I like that idea! I'm wondering if it will cause issues if we fetch from Phoenix now since it doesn't return campaign cause, staff pick, noun, or verb though in the response? 

I guess that raises another q I have - how are we getting Phoenix campaign info from the Ashes endpoints we are using? cc @weerd 

> I also want us to revisit all the campaigns caching logic to make it simpler.

@sbsmith86 by this, do you mean making the `resolveMissingItems` logic simpler?

sbsmith86:
@chloealee yep! we could also just ask @weerd to add more data to the response for campaigns. 

I think even if we don't use what you have set up, lets merge in at least the class, and connections so when we are actually ready to make the call we can. We can make cards for Rocket about the things we need from them.

chloealee:
@sbsmith86 cool that is a great suggestion, thank you! I'll work on updating this now :) 

Re: asks for Rocket, I believe @ngjo is going to connect with Ashely about this (additional things we need outlined in [this card!](https://www.pivotaltracker.com/n/projects/2019429/stories/155672260)).

chloealee:
@sbsmith86 updated and also updated the description & title of this PR so it is reflective of what this PR actually does!

.env.dusk.ci
```diff
@@ -46,8 +46,8 @@ NORTHSTAR_URL=https://northstar-qa.dosomething.org
 # NORTHSTAR_EMAIL=
 # NORTHSTAR_PASSWORD=
 
-PHOENIX_URI=https://staging.dosomething.org
-PHOENIX_API_VERSION=v1
+ASHES_URI=https://staging.dosomething.org
```
sbsmith86: ooo nice! keepin' us up to date

.env.example
```diff
@@ -49,8 +49,8 @@ NORTHSTAR_AUTH_SECRET=
 NORTHSTAR_CLIENT_ID=
 NORTHSTAR_CLIENT_SECRET=
 
-PHOENIX_URI=https://staging.dosomething.org
```
sbsmith86: We still have a `PHOENIX_URI` in here right? except now it is set to next?

chloealee: oh yes I forgot to update the `.env.example`! I'll do that now!

app.json
```diff
@@ -68,16 +68,16 @@
...
+    "ASHES_PASSWORD": {
       "required": true
     },
-    "PHOENIX_URI": {
+    "ASHES_URI": {
       "required": true
     },
-    "PHOENIX_USERNAME": {
+    "ASHES_USERNAME": {
       "required": true
```
chloealee: @sbsmith86 should I add this now or later when we actually start using Phoenix endpoints (and when we add the ENV variables to Heroku)?

sbsmith86: @chloealee ahh i think it is fine to leave it as it is! That is a good point.

PR #684, opened 2018-04-20
Clears cache when we deploy

chloealee:
#### What's this PR do?
- Adds `php artisan cache:clear` to `composer.json` to clear cache when we deploy 

👀 

- Adds this to check that everything is still working as expected after #679 

Related to [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156929053) Pivotal Card.
#### How to deploy

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,8 @@
     "warmup": [
       "@php artisan gateway:key",
       "@php artisan route:cache",
-      "@php artisan config:cache"
+      "@php artisan config:cache",
+      "@php artisan cache:clear"
     ]
   },
   "config": {


chloealee:
@DFurnes I added this to test out #679 and to make sure it didn't break anything! @sbsmith86 and I chatted about this potentially adding extra time after deployment but thought it would be ok since it would be an expected delay only on deploys. 

Another thing @sbsmith86 and I chatted about would be to add to the [Procfile](https://dosomething.slack.com/archives/C0YGXUE01/p1524253541000232?thread_ts=1524253383.000166&cid=C0YGXUE01) but not sure if that would make a difference from how it is in this PR?

DFurnes:
I'm a little wary of clearing cache on every deploy. If we're just interested in testing this out on a particular instance (rather than as an ongoing step for every deploy), would running `heroku run php artisan cache:clear --app=rogue-thor` work?

chloealee:
@DFurnes that is a great suggestion! I will remove this in another PR and test that out on Heroku! I wasn't aware that we would be able to just run that command to do a one-off :)

chloealee:
@DFurnes PR [here](https://github.com/DoSomething/rogue/pull/685) to reverse this! Also confirming that when cache is cleared on thor, #679 doesn't break :)

PR #686, opened 2018-04-23
Fixes bug to display post text properly and deletes unused helper

chloealee:
#### What's this PR do?
- Photo Caption is not being pulled correctly and is blank in Rogue - this PR fixes that!
- This PR also deletes unused `displayCaption` helper now that we've moved to using the v3 endpoint, the result is normalized when we grab posts. 
- Updates Jest test. 

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156813946) Pivotal Card.

#### How to deploy

### DIFF
--- a/resources/assets/components/Post/index.js
+++ b/resources/assets/components/Post/index.js
@@ -61,7 +61,7 @@ class Post extends React.Component {
     const quantity = post.quantity != null ? post.quantity : 0;
     const containerSize = post.type === 'photo' ? '-third' : '-half';
     const textOrCaption = post.type === 'photo' ? 'Photo Caption' : 'Text';
-
+    console.log(post);
     return (
       <div className="post container__row">
         {/* Post Images */}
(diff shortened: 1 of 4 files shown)


resources/assets/components/Post/index.js
```diff
@@ -55,14 +55,13 @@ class Post extends React.Component {
...
     const post = this.props.post;
-    const caption = displayCaption(post);
     const user = this.props.user ? this.props.user : null;
     const signup = this.props.signup;
     const campaign = this.props.campaign;
     const quantity = post.quantity != null ? post.quantity : 0;
     const containerSize = post.type === 'photo' ? '-third' : '-half';
     const textOrCaption = post.type === 'photo' ? 'Photo Caption' : 'Text';
-
+    console.log(post);
```
katiecrane: Do we still want this?

chloealee: ah I thought I got rid of that in a prior commit - thanks for flagging!

PR #687, opened 2018-04-24
Re-adds text helper back in and re-names 

chloealee:
#### What's this PR do?
- Re-adds `displayCaption` helper method back in that was removed from #686 
  - I realized we actually do need this because this is how we display on the signup page! 
- Renames the `displayCaption` helper method to `displayPostText` so it is more reflective of the data schema. 
- Updates Jest test. 

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156813946) Pivotal card (for real this time!)

#### How to deploy

### DIFF
--- a/resources/assets/helpers.js
+++ b/resources/assets/helpers.js
@@ -74,12 +74,12 @@ export function displayCityState(city, state) {
 }
 
 /**
- * Returns a readable caption string.
+ * Returns a readable Post text string.
  *
  * @param {Array} post
- * @return {String|null} caption string.
+ * @return {String|null} text string.
  */
-export function displayCaption(post) {
+export function displayPostText(post) {
   if (post.text) {
     return post.text;
   } else if (post.media) {
--- a/resources/assets/helpers.js
+++ b/resources/assets/helpers.js
@@ -74,6 +74,22 @@ export function displayCityState(city, state) {
 }
 
 /**
+ * Returns a readable caption string.
+ *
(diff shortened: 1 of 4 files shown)


resources/assets/helpers.js
```diff
@@ -73,6 +73,22 @@ export function displayCityState(city, state) {
   return `${city || ''}${city && state ? ', ' : ''}${state || ''}`;
 }
 
+/**
+ * Returns a readable Post text string.
+ *
+ * @param {Array} post
+ * @return {String|null} text string.
+ */
+export function displayPostText(post) {
```
katiecrane: Didn't you remove this helper in your other PR? Does it turn out we still need it?

chloealee: @katiecrane yes sorry that was confusing! I thought we didn't need it but turns out we do in order for the text to display on the signup page for posts! We grab posts from the database on the signup page and via the API on inbox and campaign view pages.

katiecrane: Ahhh so the format is different depending on if it comes from the API or the db?

chloealee: yes exactly! From the API, the [transformer returns it under `media`](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/PostTransformer.php#L41) while we're just grabbing the post from the database so can do `post.text`.

PR #689, opened 2018-04-24
v3/signups/:signup_id?include=posts bug

chloealee:
#### What's this PR do?
- This PR fixes the bug when you hit `/v3/signup/:signup_id?include=posts`, it returns the incorrect signup (without the `?include=posts`, everything works fine).
  - This is because `withVisiblePosts()` returns a collection and it was grabbing the first post from this collection and always returning that instead of the one specified in the endpoint! 
  - This fix explicitly grabs the correct signup from the collection. 
- Also adds tests to make sure this is returning properly!
- Also fixes bug @aaronschachter pointed out that was introduced in #688 that deleting signup no longer works (we were sending `signup.signup_id` instead of `signup.id` which was changed when we changed to using v3 endpoint).

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157064679) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Http/Controllers/SignupsController.php
+++ b/app/Http/Controllers/SignupsController.php
@@ -119,11 +119,6 @@ class SignupsController extends ApiController
      */
     public function show(Request $request, Signup $signup)
     {
-        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
-        if ($request->query('include') === 'posts') {
-            $signup = Signup::with('visiblePosts')->find($signup->id);
-        }
-
         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
     }
 
--- a/app/Http/Controllers/SignupsController.php
+++ b/app/Http/Controllers/SignupsController.php
@@ -121,8 +121,7 @@ class SignupsController extends ApiController
     {
         // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
         if ($request->query('include') === 'posts') {
-            $signups = Signup::withVisiblePosts()->get();
-            $signup = $signups->keyBy('id')[$signup->id];
+            $signup = Signup::with('posts')->find($signup->id);
         }
 
         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
--- a/app/Http/Controllers/SignupsController.php
+++ b/app/Http/Controllers/SignupsController.php
@@ -121,7 +121,8 @@ class SignupsController extends ApiController
     {
         // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
         if ($request->query('include') === 'posts') {
-            $signup = Signup::withVisiblePosts()->first();
+            $signups = Signup::withVisiblePosts()->get();
+            $signup = $signups->keyBy('id')[$signup->id];
         }
 
         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
--- a/app/Http/Controllers/SignupsController.php
+++ b/app/Http/Controllers/SignupsController.php
@@ -121,7 +121,7 @@ class SignupsController extends ApiController
     {
         // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
         if ($request->query('include') === 'posts') {
-            $signup = Signup::with('posts')->find($signup->id);
+            $signup = Signup::with('visiblePosts')->find($signup->id);
         }
 
         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -61,6 +61,16 @@ class Signup extends Model
      */
     public function posts()
     {
+        return $this->hasMany(Post::class)->with('tags');
+    }
+
+    /**
+     * Get the visible posts associated with this signup.
+     *
+     * @return \Illuminate\Database\Eloquent\Relations\HasMany
+     */
+    public function visiblePosts()
+    {
         if (! is_staff_user()) {
             return $this->hasMany(Post::class)->where('status', '=', 'accepted')
                                               ->orWhere('northstar_id', auth()->id())
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -61,6 +61,12 @@ class Signup extends Model
      */
     public function posts()
     {
+        if (! is_staff_user()) {
+            return $this->hasMany(Post::class)->where('status', '=', 'accepted')
+                                              ->orWhere('northstar_id', auth()->id())
+                                              ->with('tags');
+        }
+
         return $this->hasMany(Post::class)->with('tags');
     }
 
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -52,7 +52,8 @@ class SignupTransformer extends TransformerAbstract
      */
     public function includePosts(Signup $signup)
     {
-        $post = $signup->posts;
+        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
+        $post = $signup->visiblePosts;
 
         return $this->collection($post, new PostTransformer);
     }
--- a/resources/assets/components/Signup/index.js
+++ b/resources/assets/components/Signup/index.js
@@ -422,7 +422,7 @@ class Signup extends React.Component {
             </div>
 
             <div className="container__row">
-              <button className="button delete -tertiary" onClick={e => this.deleteSignup(signup.signup_id, e)}>Delete Signup</button>
+              <button className="button delete -tertiary" onClick={e => this.deleteSignup(signup.id, e)}>Delete Signup</button>
             </div>
 
             <MetaInformation
--- a/tests/Http/SignupTest.php
+++ b/tests/Http/SignupTest.php
@@ -309,6 +309,68 @@ class SignupTest extends TestCase
     }
 
     /**
+     * Test for signup show with included pending post info. as non-admin/non-owner.
+     * Only admins/owners should be able to see pending/rejected posts.
+     *
+     * GET /api/v3/signups/:signup_id?include=posts
+     * @return void
+     */
+    public function testSignupShowWithIncludedPostsAsNonAdminNonOwner()
+    {
+        $post = factory(Post::class)->create();
+        $signup = $post->signup;
+
+        // Test with annoymous user that no posts are returned.
+        $response = $this->getJson('api/v3/signups/' . $signup->id . '?include=posts');
+        $response->assertStatus(200);
+        $decodedResponse = $response->decodeResponseJson();
+
+        $this->assertEquals(true, empty($decodedResponse['data']['posts']['data']));
+    }
+
+    /**
+     * Test for signup show with included pending post info. as admin
+     * Only admins/owners should be able to see pending/rejected posts.
+     *
+     * GET /api/v3/signups/:signup_id?include=posts
+     * @return void
+     */
+    public function testSignupShowWithIncludedPostsAsAdmin()
+    {
+        $post = factory(Post::class)->create();
+        $signup = $post->signup;
+
+        // Test with admin that posts are returned.
+        $response = $this->withAdminAccessToken()->getJson('api/v3/signups/' . $signup->id . '?include=posts');
+        $response->assertStatus(200);
+        $decodedResponse = $response->decodeResponseJson();
+
+        $this->assertEquals(false, empty($decodedResponse['data']['posts']['data']));
+        $this->assertEquals($signup->id, $decodedResponse['data']['posts']['data'][0]['signup_id']);
+    }
+
+    /**
+     * Test for signup show with included pending post info. as owner
+     * Only admins/owners should be able to see pending/rejected posts.
+     *
+     * GET /api/v3/signups/:signup_id?include=posts
+     * @return void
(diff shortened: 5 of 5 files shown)


app/Http/Controllers/SignupsController.php
```diff
@@ -121,7 +121,8 @@ public function show(Request $request, Signup $signup)
     {
         // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
         if ($request->query('include') === 'posts') {
-            $signup = Signup::withVisiblePosts()->first();
+            $signups = Signup::withVisiblePosts()->get();
```
katiecrane: Are we grabbing all Signups and their visible posts here?

chloealee: @katiecrane yup! I wanted to still use this scope logic instead of recreating (thought it would get a little messy). Scopes return collections so we'll have to filter this more to grab the one signup we want.

katiecrane: Hmmm does it not work to grab the Signup we want and then add `->withVisiblePosts()->get()`?

chloealee: @katiecrane that is what I thought would work too! But when trying that, it still returns a collection. From the [docs](https://laravel.com/docs/5.5/eloquent#local-scopes), it sounds like scopes query all Signups with certain constraints the scope defines. That is how I understand it but I could be wrong? cc @DFurnes @sbsmith86

katiecrane: @chloealee when you do that it returns a collection of all posts or just of the post you initially grabbed?

chloealee: @katiecrane all signups with their visible posts!

katiecrane: oh weirddd I would not have expected it to work that way!

chloealee: I know it took me a long time to figure that out! Again though, seems a little strange, so maybe I am wrong?

sbsmith86: @chloealee i think you are right about how scopes work. But, I don't think I understand the issue about the signup being returned as a collection? 

If you grab the signup: `Signup::find($signup_id)->withVisiblePosts()->get()`
You are saying you get a collection back, but that "collection" should only have one signup in it, right?

chloealee: @sbsmith86 I'm getting a collection back of all the signups :-/ 
screen shot 2018-04-25 at 1 13 32 pm

DFurnes: This makes me pretty nervous - if this is loading _all_ signups and their visible posts, that's basically trying to fit all of Rogue's database into memory. Instead of using a query scope here, let's try making a [custom relationship](https://github.com/DoSomething/rogue/blob/0b5c135dc97ebbbef9f24bdc869a76e6125c67be/app/Models/Post.php#L90-L98) like we do for the `reaction` property on the Post.

sbsmith86: hmmmm also I am confused about the role check we are doing here: https://github.com/DoSomething/rogue/blob/22279d3705e67f53fca22693e360b756a6a12e3b/app/Models/Signup.php#L204-L209

when you call it as a staff user does this even do anything?

DFurnes: Yeah, that also seems suspect.

chloealee: @DFurnes I thought about doing that too but then got confused how to add that to the query? Here, we want to only show the posts that the user should be able to see depending on their role but we don't necessarily want to grab those posts here, so I'm not sure we can do that? 

 @DFurnes @sbsmith86 yeah hmm looks like it shouldn't...

chloealee: oh @DFurnes @sbsmith86 I think the posts get included if you are staff through [this line](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/SignupsController.php#L127) so we don't need to include there because staff/admins should be able to see everything?

DFurnes: @chloealee You'd be operating on the provided `$signup` instance (rather than _all_ signups with `Signup`). If configured as a relationship, it should load as soon as you ask for the related model (e.g. when the transformer reads `$signup->visiblePosts`).

DFurnes: We could also potentially do this in the `posts` relationship itself... and only apply the additional `where` queries if a user is logged in. Not sure if we'd run into any unexpected issues there, but it could be nice and clean!

chloealee: ah i like that idea, thanks for explaining @DFurnes !

chloealee: wait sorry - I am still a bit confused. If I did `$signup = $signup->visiblePosts;` (or updating the `posts` relation and grabbing that way), wouldn't it just return the signup's posts to me instead of the signup _with_ its posts?

sbsmith86: All of this is making me think it might be worth really planning to move towards policy classes to help with some of the "who can do what logic"

DFurnes: @chloealee The relationships should be loaded in the transformer (where you linked above)! We want to keep `$signup` as the signup that we're trying to look at.

chloealee: @sbsmith86 agreed! Especially since Aaron just flagged another bug that admins can only delete signups, not staff members and there is another card for staff to upload photos in this sprint.

sbsmith86: @chloealee can you add a card for that so it is on our radar?

chloealee: @sbsmith86 for what Aaron mentioned or the policy work? [Here's](https://www.pivotaltracker.com/n/projects/2019429/stories/156700437) the card for the policy work! Checking with Jen right now about if I should tackle what Aaron mentioned along with the upload card that is set for this sprint or during bigger refactoring and creating policies.

sbsmith86: Ah great! never mind me :)

chloealee: > The relationships should be loaded in the transformer (where you linked above)! We want to keep $signup as the signup that we're trying to look at.

OH I see now @DFurnes! I didn't understand that part fully and was getting tripped up. Thanks!

app/Http/Controllers/SignupsController.php
```diff
@@ -121,7 +121,7 @@ public function show(Request $request, Signup $signup)
     {
         // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
         if ($request->query('include') === 'posts') {
-            $signup = Signup::withVisiblePosts()->first();
+            $signup = Signup::with('posts')->find($signup->id);
         }
```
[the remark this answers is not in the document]
chloealee: `$signup = $signup->visiblePosts();` erroring out makes sense because that is sending the relation, not the signup, to the transformer [here](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/SignupsController.php#L127)?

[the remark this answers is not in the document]
chloealee: ah interesting! So `$signup = Signup::with('visiblePosts')->find($signup->id);` actually ends up calling `posts()` in `Signup.php` that returns all posts, regardless of status. Is this to be expected? cc @DoSomething/team-bleed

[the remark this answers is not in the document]
sbsmith86: I'm really fuzzy on this bug and I don't think I fully understand the `visiblePosts` thing. What is it that you are trying to accomplish and what is not happening?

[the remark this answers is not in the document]
chloealee: @sbsmith86 yes it's confusing! So for this endpoint, we'd like to grab a specific signup and include their posts (`/v3/signups/:signup_id?include=posts`). When using the `?include=posts` flag, the bug is that it returns the incorrect signup and was grabbing the first signup in the database.

I want to build a relation of `visiblePosts` for each signup here. The `visiblePosts` function will determine the following logic: 
- If you are an admin/staff, you can see all posts. 
- If you own the signup, you can see all posts. 
- If you aren't a admin/staff and you don't own the signup, you can only see `pending` posts. 

I'm trying to accomplish returning the `visiblePosts` based on the user but having trouble getting this to work - I'm not sure if I'm just doing something silly incorrectly based on the above? LMK if this is still unclear!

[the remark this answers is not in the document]
sbsmith86: That is super clear! Thank you for that and I'm sorry this is taking me a while to wrap my head around. 

* I don't think I see where the `visiblePosts` method is defined? is this a laravel convention?

* RE: the `posts()` method returning all posts regardless of status. It looks like we only apply a status filter to folks that are not staff members so if you are making the request as staff or an admin, I am guessing that is why the status is not being applied.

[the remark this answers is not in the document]
chloealee: Don't be - there has been a ton of conversation in this PR so it's hard to jump around - thanks for your patience! :) 

> I don't think I see where the `visiblePosts` method is defined? is this a laravel convention?
Sorry - again unclear! That is not in this PR (didn't push up the changes) - I was trying to build that out but was still returning all the posts (even pending for a non staff/admin/owner user). I realized that was because it actually then went through the `posts()` method in `Signup.php` after to return all of the posts. 

I think the best next step might be to try to get how it is working right now to work and dig into why it is breaking the v2 tests?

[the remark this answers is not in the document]
sbsmith86: yea i think move forward as you see best, and then we will take a look at the PR!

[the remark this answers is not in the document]
chloealee: @sbsmith86 the more I think about this, I think it's actually better to write this logic in it's own relation (not in `posts()` in `Signup.php`). 

The`v2` tests are breaking in the logic where it finds `posts_count` [here](https://github.com/DoSomething/rogue/blob/master/app/Repositories/Legacy/Two/PostRepository.php#L103) - since we aren't sure of the user's role in v2 endpoints, it returns `0` when we call `posts_count` with this logic. If I remove the new logic in this PR from the `posts()` method, it returns `1` when we call `posts_count`. 

For this reason, I think it's better to explicitly create a `visiblePosts` method and call that for this `v3` endpoint. 

However, this brings me back to the original problem! I can't seem to get the `visiblePosts` method working without going through the `posts()` method in `Signup.php` I'll push up what I have so it is easier to see.

app/Models/Signup.php
```diff
@@ -64,6 +64,22 @@ public function posts()
...
+    {
+        if (! is_staff_user()) {
+            return $this->hasMany(Post::class)->where('status', '=', 'accepted')
+                                              ->orWhere('northstar_id', auth()->id())
+                                              ->with('tags');
+        }
+
+        return $this->hasMany(Post::class)->with('tags');
+    }
+
```
chloealee: @sbsmith86 continuing from [this comment](https://github.com/DoSomething/rogue/pull/689#discussion_r184439169), calling `Signup::with('visiblePosts')->find($signup->id);` returns all the posts for a non staff/admin/owner user. 

It seems like it goes through the `posts()` method in `Signup.php` because when I place a `dd()` there, it reaches it and I'm not sure how to avoid this?

sbsmith86: @chloealee have you pushed up the `visiblePosts` method? would like to see what is happening so I can follow along.

chloealee: @sbsmith86 yup! it is this method! :)

chloealee: @sbsmith86 no worries it is hard to see when it cuts off like that! 

Yes! I tried doing `$signup = $signup->visiblePosts()` and it gave me this error: 

```
Type error: Argument 1 passed to Rogue\Http\Transformers\SignupTransformer::transform() must be an instance of Rogue\Models\Signup, instance of Illuminate\Database\Eloquent\Relations\HasMany given, called in /home/vagrant/DoSomething.org/rogue/vendor/league/fractal/src/Scope.php on line 338
``` 
since it seems like it is passing the relationship, not the Signup model, to the transformer [here](https://github.com/DoSomething/rogue/pull/689/files#diff-28d9d92df597ce839f7f0cfc3a8dfb94R127). 

Same thing happened when I deleted the `with('tags');` :-/ 

I'm not sure what else I might be doing wrong?

sbsmith86: where were you calling `$signup = $signup->visiblePosts()`? I think you should be doing this in the transformer itself when deciding to embed posts or not, so i'm tryna figure out why the transformer is being mad it isn't given the right thing.

chloealee: OMG @sbsmith86 that was it! I was calling that in the controller 🤦‍♀️ !! This makes TOTAL sense now thank you so much!!! I'll push up the changes now!

sbsmith86: YAY so glad we figured it out. And I am sorry I made you dig all the way through that hahaha

chloealee: no this was really good to get me to better understand the relation stuff that I thought I had a handle on! Thank you so much for your patience with me and going through this so much!! :)

chloealee: pr is updated with adding to the transformer!

PR #693, opened 2018-05-08
v3 posts index to not return posts tagged "Hide In Gallery" by default

chloealee:
#### What's this PR do?
- Updates `/v3/posts` index to not return posts tagged "Hide In Gallery" by default

👀 
- Hit this endpoint
- No posts that are tagged "Hide In Gallery" should be returned! 

The You've Got The Power gallery was showing posts that were tagged "Hide In Gallery" because it is using the `v3` endpoint and it hasn't been updated to not include these! 

Update [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157411893) Pivotal Card. 

#### How to deploy

### DIFF
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -71,15 +71,16 @@ class PostsController extends ApiController
     public function index(Request $request)
     {
         $query = $this->newQuery(Post::class)
-            ->orderBy('created_at', 'desc')
-            // by default, do not return posts hidden from the gallery
-            ->withoutTag('hide-in-gallery');
+            ->orderBy('created_at', 'desc');
 
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Post::$indexes);
 
         // Only allow admins or staff to see un-approved posts from other users.
         $query = $query->whereVisible();
+
+        // Only return posts tagged "Hide In Gallery" if staff user or if is owner of the post.
+        $query = $query->withHiddenPosts();
 
         // If tag param is passed, only return posts that have that tag.
         if (array_has($filters, 'tag')) {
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/PostsController.php
```diff
@@ -71,7 +71,9 @@ public function __construct(PostManager $posts, SignupRepository $signups, PostT
     public function index(Request $request)
     {
         $query = $this->newQuery(Post::class)
-            ->orderBy('created_at', 'desc');
+            ->orderBy('created_at', 'desc')
+            // by default, do not return posts hidden from the gallery
+            ->withoutTag('hide-in-gallery');
```
DFurnes: Would it make sense to apply the same logic as we do for `whereVisible` here (where staff members & users can see their own hidden posts)?

chloealee:
Instead of this endpoint not returning posts that are tagged by "Hide In Gallery" by default, do we want to add a flag `exclude=` instead for the gallery to use?

app/Http/Controllers/PostsController.php
chloealee: yes I like this idea better @DFurnes ! I'll update that!

app/Models/Post.php
```diff
@@ -338,4 +338,19 @@ public function scopeWhereVisible($query)
...
+     *
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function scopeWithHiddenPosts($query)
+    {
+        if ((! token()->exists()) || (! in_array(token()->role, ['admin', 'staff']))) {
+            return $query->whereDoesntHave('tags', function ($query) {
+                $query->where('tag_slug', 'hidden-in-gallery');
+            })
+                ->orWhere('northstar_id', auth()->id());
```
chloealee: actually, this returns before it reaches the `orWhere` part - trying to figure out the correct syntax for this!

chloealee: sorry I was wrong - this query is working as expected!

chloealee:
I'm going to merge this so Aki can continue reviewing and it's not a blocker but separate PR coming with tests!

PR #694, opened 2018-05-09
Adds test for scopeWithHiddenPosts and updates typo so it actually works! 

chloealee:
#### What's this PR do?
- Changes to `hide-in-gallery` which is what it actually is in the database so that #693 works!
- Adds tests. 

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157411893) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -274,7 +274,6 @@ class Post extends Model
         $this->touch();
 
         event(new PostTagged($this, $tag));
-
         return $this;
     }
 
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -348,7 +348,7 @@ class Post extends Model
     {
         if (! is_staff_user()) {
             return $query->whereDoesntHave('tags', function ($query) {
-                $query->where('tag_slug', 'hidden-in-gallery');
+                $query->where('tag_slug', 'hide-in-gallery');
             })
                 ->orWhere('northstar_id', auth()->id());
         }
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -346,7 +346,7 @@ class Post extends Model
      */
     public function scopeWithHiddenPosts($query)
     {
-        if ((! token()->exists()) || (! in_array(token()->role, ['admin', 'staff']))) {
+        if (! is_staff_user()) {
             return $query->whereDoesntHave('tags', function ($query) {
                 $query->where('tag_slug', 'hidden-in-gallery');
             })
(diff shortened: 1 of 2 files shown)


app/Models/Post.php
```diff
@@ -346,7 +346,7 @@ public function scopeWhereVisible($query)
      */
     public function scopeWithHiddenPosts($query)
     {
-        if ((! token()->exists()) || (! in_array(token()->role, ['admin', 'staff']))) {
+        if (! is_staff_user()) {
```
chloealee: @DFurnes I changed this to use the helper function here. I initially didn't use the helper function because I thought that the token was that of Phoenix, not the user, so wanted to omit the client check. I think this is more accurate to use!

DFurnes: FWIW, the `is_staff_user` helper should work for clients as well!

chloealee: yes! I initially thought we didn't want it to work for clients but now I see I misunderstood ! :)

chloealee:
@DFurnes for some reason the tests aren't passing but confirming that looks good when I test manually! Is there something I'm doing wrong?

DFurnes:
Hmm, let me pull down and take a peek!

chloealee:
thanks @DFurnes !

tests/Http/PostTest.php
```diff
@@ -585,6 +607,28 @@ public function testPostsIndexAsAdmin()
...
+    {
+        // Admins should see all posts.
+        factory(Post::class, 'accepted', 10)->create();
+
+        $hiddenPost = factory(Post::class, 'accepted')->create();
+        $hiddenPost->tag('Hide In Gallery');
+        $hiddenPost->save();
+
+        $response = $this->withAdminAccessToken()->getJson('api/v3/posts');
+
```
chloealee: @DFurnes it's weird - when I do `dd($response->decodeResponseJson());` here, the response doesn't include a `tags` array for each post like we expect - could this be part of the issue?

DFurnes: Hmm, that's interesting. It shoooould. 🤔

chloealee: hmm so when I comment out `$hiddenPost->tag('Hide In Gallery);` and saving it, then the `tags` array is returned. It seems like it is something with those lines that is making it upset!

chloealee:
@DFurnes and I tested thoroughly and have marked these tests as incomplete in order to merge this. Through our tests, everything seems to work as intended. We think the tests are failing due to something wonky with Gateway's helper test tokens. We're going to look into this but want to merge this fix so it doesn't block campaigns further!

chloealee:
Here is the card to fix the tests! https://www.pivotaltracker.com/n/projects/2019429/stories/157481061

PR #695, opened 2018-05-10
Adds text post unit test

chloealee:
#### What's this PR do?
- Adds unit test to make sure a `text` post is correctly created. 

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157202065) Pivotal Card.

#### How to deploy

### DIFF
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -91,11 +91,11 @@ class PostTest extends TestCase
     }
 
     /**
-     * Test that a POST request to /posts creates a new post.
+     * Test that a POST request to /posts creates a new photo post.
      *
      * @return void
      */
-    public function testCreatingAPost()
+    public function testCreatingAPhotoPost()
     {
         $signup = factory(Signup::class)->create();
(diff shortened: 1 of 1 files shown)


tests/Http/PostTest.php
```diff
@@ -159,6 +159,75 @@ public function testCreatingAPost()
...
+        $this->mock(Blink::class)->shouldReceive('userSignupPost');
+
+        // Create the post!
+        $response = $this->withAccessToken($signup->northstar_id)->postJson('api/v3/posts', [
+            'northstar_id'     => $signup->northstar_id,
+            'campaign_id'      => $signup->campaign_id,
+            'campaign_run_id'  => $signup->campaign_run_id,
+            'type'             => 'text',
+            'action'           => 'test-action',
+            'quantity'         => $quantity,
```
DFurnes: Do we use think we'll ever use quantities with text posts? I know You've Got The Power is a quantity-less text submission, so if that's the template we follow moving forward it might be something to omit in our test case.

PR #697, opened 2018-05-10
Hidden Posts tests

chloealee:
#### What's this PR do?
- Adds tests to make sure hidden post logic is working - admins and owners should be returned hidden posts (posts tagged as "Hide In Gallery." Anonymous users shouldn't be able to see hidden posts when hitting `GET /v3/posts`.
- Updates TagsTests to tag with `->tag();` instead of via the API.
- Updates `composer.json` to use latest Gateway release.

All tests should pass!

Writes tests for work done in #694 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157481061) Pivotal Card.

#### How to deploy

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "laravel/tinker": "^1.0",
     "aws/aws-sdk-php-laravel": "^3.1",
     "doctrine/dbal": "~2.5.13",
-    "dosomething/gateway": "^1.1.6",
+    "dosomething/gateway": "dev-logout",
     "fideloper/proxy": "^3.3",
     "guzzlehttp/guzzle": "^6.2",
     "intervention/image": "^2.3",
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     "laravel/tinker": "^1.0",
     "aws/aws-sdk-php-laravel": "^3.1",
     "doctrine/dbal": "~2.5.13",
-    "dosomething/gateway": "dev-logout",
+    "dosomething/gateway": "^1.14.8",
     "fideloper/proxy": "^3.3",
     "guzzlehttp/guzzle": "^6.2",
     "intervention/image": "^2.3",
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -2,7 +2,6 @@
 
 use Faker\Generator;
 use Rogue\Models\Post;
-use Rogue\Models\Tag;
 use Rogue\Models\User;
 use Rogue\Services\AWS;
 use Rogue\Models\Signup;
@@ -95,17 +94,4 @@ $factory->define(User::class, function (Generator $faker) {
 
 $factory->defineAs(User::class, 'admin', function () use ($factory) {
     return array_merge($factory->raw(User::class), ['role' => 'admin']);
-});
-
-// Tag Factory
-$factory->define(Tag::class, function (Generator $faker) {
-    $tagName = str_random(10);
-    return [
-        'tag_name' => $tagName,
-        'tag_slug' => str_slug($tagName, '-')
-    ];
-});
-
-$factory->defineAs(Tag::class, 'Hide In Gallery', function () use ($factory) {
-    return array_merge($factory->raw(Tag::class), ['tag_name' => 'Hide In Gallery', 'tag_slug' => 'hide-in-gallery']);
 });
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -2,6 +2,7 @@
 
 use Faker\Generator;
 use Rogue\Models\Post;
+use Rogue\Models\Tag;
 use Rogue\Models\User;
 use Rogue\Services\AWS;
 use Rogue\Models\Signup;
@@ -94,4 +95,17 @@ $factory->define(User::class, function (Generator $faker) {
 
 $factory->defineAs(User::class, 'admin', function () use ($factory) {
     return array_merge($factory->raw(User::class), ['role' => 'admin']);
+});
+
+// Tag Factory
+$factory->define(Tag::class, function (Generator $faker) {
+    $tagName = str_random(10);
+    return [
+        'tag_name' => $tagName,
+        'tag_slug' => str_slug($tagName, '-')
+    ];
+});
+
+$factory->defineAs(Tag::class, 'Hide In Gallery', function () use ($factory) {
+    return array_merge($factory->raw(Tag::class), ['tag_name' => 'Hide In Gallery', 'tag_slug' => 'hide-in-gallery']);
 });
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -4,7 +4,6 @@ namespace Tests\Http;
 
 use Tests\TestCase;
 use Rogue\Models\Post;
-use Rogue\Models\Tag;
 use Rogue\Models\User;
 use Rogue\Models\Signup;
 use Rogue\Models\Reaction;
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -4,6 +4,7 @@ namespace Tests\Http;
 
 use Tests\TestCase;
 use Rogue\Models\Post;
+use Rogue\Models\Tag;
 use Rogue\Models\User;
 use Rogue\Models\Signup;
 use Rogue\Models\Reaction;
@@ -611,8 +612,9 @@ class PostTest extends TestCase
         // Anonymous requests should only see posts that are not tagged with "Hide In Gallery."
         factory(Post::class, 'accepted', 10)->create();
 
+        $tag = factory(Tag::class, 'Hide In Gallery')->create();
         $hiddenPost = factory(Post::class, 'accepted')->create();
-        $hiddenPost->tag('Hide In Gallery');
+        $hiddenPost->tags()->attach($tag);
 
         $response = $this->getJson('api/v3/posts');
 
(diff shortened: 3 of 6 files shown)


tests/Http/PostTest.php
```diff
@@ -615,12 +615,15 @@ public function testPostsIndexAsAdmin()
...
-        $this->markTestIncomplete();
+        // $this->markTestIncomplete();
         // Admins should see all posts.
         factory(Post::class, 'accepted', 10)->create();
 
         $hiddenPost = factory(Post::class, 'accepted')->create();
         // $hiddenPost->tag('Hide In Gallery');
+        $this->withAdminAccessToken()->postJson('api/v3/posts/' . $hiddenPost->id . '/tags', [
+            'tag_name' => 'Hide In Gallery',
+        ]);
```
chloealee: @DFurnes I don't think there is anything wrong with the Gateway test tokens but rather I think there is something that the test doesn't like about `->tag()` in line 623. When I tag a post via the API, the test passes! 

Also, when I comment out this line, the `tags` array is returned in the response (we were seeing this disappeared in the response as soon as we added the `$hiddenPost->tag('Hide In Gallery');` line).  

When tagging with the API, not only does the test pass but also the `tags` array is returned in the response as expected as well.

sbsmith86: @chloealee sorry to just bop in, but is there nothing in the logs about this failure?

chloealee: @sbsmith86 don't be sorry - thanks for this suggestion! The logs say that the post is successfully tagged: 
```
[2018-05-10 19:30:52] local.INFO: post_tagged {"id":11,"tag":"hide-in-gallery"} {"user_id":null,"client_id":null,"request_id":null}
```

chloealee: This is what the log says when I tag via the API: 
```
[2018-05-10 19:33:44] local.INFO: post_tagged {"id":11,"tag":"hide-in-gallery"} {"user_id":"5589c991a59dbfa93d8b45ae","client_id":"phpunit","request_id":null}
```
I don't think it should get mad by the `null`ed `user_id` and `client_id`?

sbsmith86: when the test runs and fails, the message you get isn't very descriptive, but sometimes I look in the logs to see more details. It might be nice to see what happens in the logs on the failure instead of the successful API request?

chloealee: oh interesting idea, thanks @sbsmith86 !

tests/Http/PostTest.php
```diff
@@ -615,15 +617,17 @@ public function testPostsIndexAsAdmin()
...
         // Admins should see all posts.
         factory(Post::class, 'accepted', 10)->create();
 
+        $tag = factory(Tag::class, 'Hide In Gallery')->create();
         $hiddenPost = factory(Post::class, 'accepted')->create();
-        // $hiddenPost->tag('Hide In Gallery');
+        $hiddenPost->tags()->attach($tag);
+        // $this->withAdminAccessToken()->postJson('api/v3/posts/' . $hiddenPost->id . '/tags', [
+        //     'tag_name' => 'Hide In Gallery',
+        // ]);
```
chloealee: will delete this if this route (using Model Factory) is an ok way to go!

chloealee:
@DoSomething/team-bleed still unsure why the tests didn't like when I used the regular old $hiddenPost->tag('Hide In Gallery');` but was able to get it working by creating a Tag Model Factory and then attaching it to the post. 

I was also able to get it working by tagging via the API in the post (shown in the commented out section). I think I like tagging via the API a bit better since it explicitly goes through the exact logic of what happens when something is tagged instead of mimicking, but I don't have strong opinions on which route to go!

Is going through the API a safe way to go about these tests?

database/factories/ModelFactory.php
```diff
@@ -95,3 +96,17 @@
...
+$factory->define(Tag::class, function (Generator $faker) {
+    $tagName = str_random(10);
+
+    return [
+        'tag_name' => $tagName,
+        'tag_slug' => str_slug($tagName, '-'),
+    ];
+});
+
+$factory->defineAs(Tag::class, 'Hide In Gallery', function () use ($factory) {
```
chloealee: I tried to pass a variable `$tagName` instead of 'Hide In Gallery` here but it didn't like that!

sbsmith86: `$tagName` is scoped to the callback function in the statement before, which is why you probably can not use it.

sbsmith86:
@chloealee I agree about making the call to the API 👍 , this is similar to how we do it in the Post tests!

I think I am a little hazy about the tag factory stuff. It seems like in the [`TagsController`](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/TagsController.php#L57)  we make a call to `Post::tag()` and that method [will create the tag on the fly](https://github.com/DoSomething/rogue/blob/master/app/Models/Post.php#L269) if we don't have it yet. To your point about making sure this test is correctly mimicking what is happening, I am concerned that creating the tag first is not really aligned with the functionality. 

Is the issue with the tests in that line where we try to create the tag when we don't have it? I am wondering if there is a configuration issue where the tests are using the right test database

database/factories/ModelFactory.php
chloealee: ah you're totally right, thanks @sbsmith86 !

chloealee:
@sbsmith86 thanks for your input here! 

> Is the issue with the tests in that line where we try to create the tag when we don't have it? I am wondering if there is a configuration issue where the tests are using the right test database

Re: your q above, it seems to correctly make the tag - when I `dd($hiddenPost)`  after creating and tagging it (doing `$hiddenPost->tag('Hide In Gallery');`), it returns the a post that is tagged but seems like it isn't saving in the database. The response from the API includes all except for that `$hiddenPost`. In addition, I was seeing some odd responses with the API response that didn't include the `tags` array in the response when it should have (when accessing as an admin). 

Is it ok to proceed with this just tagging via the API?

sbsmith86:
interesting! So, if you just tag it making the call to the api endpoint it works? What is the tag factory doing then?

chloealee:
Yes! When tagging the post via the API, everything works as expected! 

Using the tag factory way also works - just wanted to show two ways we could write this test (Tag Factory and via the API) but commenting out the API way and leaving the tag factory way was an easier way of showing both in this PR! 

Like you mentioned above though, I'm leaning more towards just tagging via the API route for this test since it goes exactly through the real logic!

tests/Http/PostTest.php
```diff
@@ -718,17 +718,24 @@ public function testPostsIndexAsOwnerHiddenPosts()
...
         $secondHiddenPost = factory(Post::class, 'accepted')->create();
-        $secondHiddenPost->tag('Hide In Gallery');
         $secondHiddenPost->northstar_id = $this->faker->unique()->northstar_id;
         $secondHiddenPost->save();
 
+        $this->withAdminAccessToken()->postJson('api/v3/posts/' . $secondHiddenPost->id . '/tags', [
+            'tag_name' => 'Hide In Gallery',
+        ]);
+
         $response = $this->withAccessToken($ownerId)->getJson('api/v3/posts');
```
chloealee: @DFurnes this seems to fail bc I _think_ it is still using the `AdminAccessToken` and is actually grabbing hidden posts we don't want to see. Is there a way to make sure it's using the `withAccessToken($ownerId)` instead of the `withAdminAccessToken()` first used?

DFurnes: A ha!! I think you're totally right that we'd be keeping the same cached user in the guard! 

We should maybe add an `Auth::logout()` to [the Gateway guard](https://github.com/DoSomething/gateway/blob/master/src/Server/GatewayGuard.php) so that we can unset that property in tests (since `Auth::setUser` is type-hinted to require an actual user instance).

chloealee: awesome @DFurnes ! Here's a [the PR](https://github.com/DoSomething/gateway/compare/logout?expand=1) !

chloealee:
Because of the fixes made in Gateway here, we can now use `$post->tag('Hide In Gallery);` instead of tagging via the API in our tests! Also updated the TagsTest to use this :)

composer.json
```diff
@@ -9,7 +9,7 @@
     "laravel/tinker": "^1.0",
     "aws/aws-sdk-php-laravel": "^3.1",
     "doctrine/dbal": "~2.5.13",
-    "dosomething/gateway": "^1.1.6",
+    "dosomething/gateway": "^1.1.8",
```
DFurnes: The caret should take care of this, but let's set this to `^1.14.8` to be a bit clearer!

chloealee:
@DFurnes hm so it seems like a lot of the errors are returning `500` instead of `200`! When making the request in Postman, I'm getting something like this:
```
"message": "Trying to get property 'role' of non-object",
    "exception": "ErrorException",
    "file": "/home/vagrant/DoSomething.org/rogue/vendor/dosomething/gateway/src/Server/Middleware/RequireRole.php",
```
I was seeing that yesterday and tried to fix it in [this commit](https://github.com/DoSomething/gateway/pull/116/commits/3f2a822923f8a731f416308075d5e40ceedd3d01), but then saw it broke tests in Gateway and then also saw when I removed it again, all was working. 

Do you think I should try to add that again and get things to pass in Gateway? I feel like I maybe did something silly that I am overlooking since this seems so inconsistent...

chloealee:
Doing some more digging, It seems like the token isn't getting to `RequireRole`. It _is_ getting to `RequireScope` but I'm not seeing any difference between them so not sure why?

PR #701, opened 2018-05-18
if there is no campaign id, handles error gracefully

chloealee:
#### What's this PR do?
- If there is no `campaign_id`, mock campaign array we would get from Ashes to handle error gracefully and prevent entire page from crashing. 

👀 

This is related to #700 when we saw an error in production where `/campaigns` was throwing a `500` error. This was because there was a signup record with no `campaign_id` and caused the page to crash. If this happens in the future, we want to handle this error in a more graceful manner.

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157673595) Pivotal card.

#### How to deploy

### DIFF
--- a/app/Services/CampaignService.php
+++ b/app/Services/CampaignService.php
@@ -78,7 +78,6 @@ class CampaignService
             $this->cache->store($id, $campaign['data'], 1440);
 
             $campaign = $campaign['data'];
-            // dd($campaign);
         }
 
         return $campaign;
--- a/app/Services/CampaignService.php
+++ b/app/Services/CampaignService.php
@@ -55,14 +55,30 @@ class CampaignService
             // @TODO: change this to grab from Phoenix when Phoenix returns all info. we need!
             $campaign = $this->ashes->getCampaign($id);
 
+            // Handle cases where the ID was not found.
             if (empty($campaign['data']['id'])) {
-                info('missing_campaign_data', compact('id', 'campaign'));
+                $placeholder = [
+                    'id' => $id,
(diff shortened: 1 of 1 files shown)


app/Services/CampaignService.php
```diff
@@ -55,12 +55,27 @@ public function find($id)
...
+                    'staff_pick' => false,
+                    'causes' => [
+                        'primary' => [
+                            'name' => null,
+                        ],
+                    ],
+                    'uri' => null,
+                ];
+
+                $this->cache->store($id, $placeholder, 1440);
```
katiecrane: Hmm I'm wondering about caching this. In theory, couldn't a signup id share the same id as a campaign id? Also does this make things much faster since it's not making any external calls or anything?

chloealee: @katiecrane yes I agree with you that a signup id could be the same as a campaign id! I'm a little confused why you bring that up here though? 

> Also does this make things much faster since it's not making any external calls or anything?

We're sending this as a placeholder and having it work the same way mainly so the page doesn't break if for some reason we get an empty or `''` as a campaign id.

katiecrane: OOOOO oops I thought we were in a different file, I was thinking it was `signup_id` since you said there had been a signup missing the `campaign_id` but I was very mistaken!! Don't mind me!

chloealee: no worries I would have made the same mistake! So many 🆔 s!

PR #702, opened 2018-05-18
Post Permission Policies

chloealee:
#### What's this PR do?
- Adds Post Permission Policies for `show` and `update` methods. 
- Adds `CustomGate` so we can send an anonymous `user` to the policy.
- Updates the Post Model Factory to create a `northstar_id`
- Deletes some tests that are redundant. 

👀 
All tests should pass. 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156700437) Pivotal Card. 

#### How to deploy

### DIFF
--- a/app/Auth/CustomGate.php
+++ b/app/Auth/CustomGate.php
@@ -2,7 +2,6 @@
 
 namespace Rogue\Auth;
 
-use Rogue\Models\User;
 use Illuminate\Support\Arr;
 use Illuminate\Auth\Access\Gate;
 use Illuminate\Contracts\Auth\Access\Gate as GateContract;
@@ -12,25 +11,29 @@ class CustomGate extends Gate implements GateContract
     protected function raw($ability, $arguments = [])
     {
         if (! $user = $this->resolveUser()) {
-            // return false;
-            $user = new User;
+            $user = null;
         }
+
         $arguments = Arr::wrap($arguments);
+
         // First we will call the "before" callbacks for the Gate. If any of these give
         // back a non-null response, we will immediately return that result in order
         // to let the developers override all checks for some authorization cases.
         $result = $this->callBeforeCallbacks(
             $user, $ability, $arguments
         );
+
         if (is_null($result)) {
             $result = $this->callAuthCallback($user, $ability, $arguments);
         }
+
         // After calling the authorization callback, we will call the "after" callbacks
         // that are registered with the Gate, which allows a developer to do logging
         // if that is required for this application. Then we'll return the result.
         $this->callAfterCallbacks(
             $user, $ability, $arguments, $result
         );
+
         return $result;
     }
 }
--- a/app/Auth/CustomGate.php
+++ b/app/Auth/CustomGate.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Rogue\Auth;
+
+use Rogue\Models\User;
+use Illuminate\Support\Arr;
+use Illuminate\Auth\Access\Gate;
+use Illuminate\Contracts\Auth\Access\Gate as GateContract;
+
+class CustomGate extends Gate implements GateContract
+{
+    protected function raw($ability, $arguments = [])
+    {
+        if (! $user = $this->resolveUser()) {
+            // return false;
+            $user = new User;
+        }
+        $arguments = Arr::wrap($arguments);
+        // First we will call the "before" callbacks for the Gate. If any of these give
+        // back a non-null response, we will immediately return that result in order
+        // to let the developers override all checks for some authorization cases.
+        $result = $this->callBeforeCallbacks(
+            $user, $ability, $arguments
+        );
+        if (is_null($result)) {
+            $result = $this->callAuthCallback($user, $ability, $arguments);
+        }
+        // After calling the authorization callback, we will call the "after" callbacks
+        // that are registered with the Gate, which allows a developer to do logging
+        // if that is required for this application. Then we'll return the result.
+        $this->callAfterCallbacks(
+            $user, $ability, $arguments, $result
+        );
+        return $result;
+    }
+}
--- a/app/Auth/CustomGate.php
+++ b/app/Auth/CustomGate.php
@@ -8,6 +8,14 @@ use Illuminate\Contracts\Auth\Access\Gate as GateContract;
 
 class CustomGate extends Gate implements GateContract
 {
+    /**
+     * Get the raw result from the authorization callback. We've overridden this
+     * method to allow users to be `null` so policies can apply to anonymous users.
+     *
+     * @param  string  $ability
+     * @param  array|mixed  $arguments
+     * @return mixed
+     */
     protected function raw($ability, $arguments = [])
     {
         if (! $user = $this->resolveUser()) {
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -8,7 +8,6 @@ use Rogue\Managers\PostManager;
 use Rogue\Http\Requests\PostRequest;
 use Rogue\Repositories\SignupRepository;
 use Rogue\Http\Transformers\PostTransformer;
-use Illuminate\Auth\Access\AuthorizationException;
 use Rogue\Http\Controllers\Traits\FiltersRequests;
 use Rogue\Http\Controllers\Legacy\Two\ApiController;
 
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -121,13 +121,7 @@ class PostsController extends ApiController
     public function show(Post $post)
     {
         // Only allow an admin or the user who owns the post to see their own unapproved posts.
-        if ($post->status != 'accepted') {
-            if (is_staff_user() || auth()->id() === $post->northstar_id) {
-                return $this->item($post);
-            } else {
-                throw new AuthorizationException('You don\'t have the correct role to view this post!');
-            }
-        }
+        $this->authorize('show', $post);
 
         return $this->item($post);
     }
@@ -144,13 +138,10 @@ class PostsController extends ApiController
     {
         // Only allow an admin/staff or the user who owns the post to update.
         $this->authorize('update', $post);
-        // if (is_staff_user() || auth()->id() === $post->northstar_id) {
-            $this->posts->update($post, $request->validated());
 
-            return $this->item($post);
-        // }
+        $this->posts->update($post, $request->validated());
 
-        // throw new AuthorizationException('You don\'t have the correct role to update this post!');
+        return $this->item($post);
     }
 
     /**
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -143,13 +143,14 @@ class PostsController extends ApiController
     public function update(PostRequest $request, Post $post)
     {
         // Only allow an admin/staff or the user who owns the post to update.
-        if (is_staff_user() || auth()->id() === $post->northstar_id) {
+        $this->authorize('update', $post);
+        // if (is_staff_user() || auth()->id() === $post->northstar_id) {
             $this->posts->update($post, $request->validated());
 
             return $this->item($post);
-        }
+        // }
 
-        throw new AuthorizationException('You don\'t have the correct role to update this post!');
+        // throw new AuthorizationException('You don\'t have the correct role to update this post!');
     }
 
     /**
--- a/app/Policies/PostPolicy.php
+++ b/app/Policies/PostPolicy.php
@@ -2,7 +2,7 @@
 
 namespace Rogue\Policies;
 
-use ;
+use Rogue\Models\Post;
 use Illuminate\Auth\Access\HandlesAuthorization;
 
 class PostPolicy
@@ -16,6 +16,16 @@ class PostPolicy
      */
     public function __construct()
     {
-        //
+    }
+
+    /**
+     * Determine if the given post can be updated by the user.
+     *
+     * @param  \App\User  $user
+     * @return bool
+     */
+    public function update(Post $post)
+    {
+        return is_staff_user() || auth()->id() === $post->northstar_id;
     }
 }
--- a/app/Policies/PostPolicy.php
+++ b/app/Policies/PostPolicy.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Rogue\Policies;
+
+use ;
+use Illuminate\Auth\Access\HandlesAuthorization;
+
+class PostPolicy
+{
+    use HandlesAuthorization;
+
+    /**
+     * Create a new policy instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        //
+    }
+}
--- a/app/Policies/PostPolicy.php
+++ b/app/Policies/PostPolicy.php
@@ -26,10 +26,12 @@ class PostPolicy
      * @param  Rogue\Models\Post $post
      * @return bool
      */
-    public function show(Authenticatable $user, Post $post)
+    public function show($user, Post $post)
     {
-        if ($post->status !== 'accepted') {
+        if ($post->status !== 'accepted' && $user !== null) {
             return is_staff_user() || $user->northstar_id === $post->northstar_id;
+        } elseif ($user === null && $post->status !== 'accepted') {
+            return false;
         }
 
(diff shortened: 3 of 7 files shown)


app/Policies/PostPolicy.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+    }
+
+    /**
+     * Determine if the given post can be updated by the user.
+     *
+     * @param  \App\User  $user
+     * @return bool
+     */
+    public function update(Post $post)
```
chloealee: I'm having some trouble here - when testing this out I am getting this error: 
```
"message": "Type error: Argument 1 passed to Rogue\\Policies\\PostPolicy::update() must be an instance of Rogue\\Models\\User, instance of DoSomething\\Gateway\\Server\\RemoteUser given, called in /home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php on line 481",
```
does this mean that we're not able to use Permission Policies if getting the `User` not from the User model?

chloealee: oh maybe I have to do something like [this](https://github.com/laravel/framework/commit/70f75255808ffc96275e6f2f356616dd2e163434#diff-961368895033e553787b301c3be0e17a)?

DFurnes: @chloealee Do you have a stack trace for that "must be an instance of User" error? It should just be looking for something `Authenticatable`, which `RemoteUser` is.

chloealee: @DFurnes here is the beginning of the stack trace! 
```
    "trace": [
        {
            "file": "/home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php",
            "line": 481,
            "function": "update",
            "class": "Rogue\\Policies\\PostPolicy",
            "type": "->"
        },
        {
            "file": "/home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php",
            "line": 338,
            "function": "Illuminate\\Auth\\Access\\{closure}",
            "class": "Illuminate\\Auth\\Access\\Gate",
            "type": "->"
        },
        {
            "file": "/home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php",
            "line": 313,
            "function": "callAuthCallback",
            "class": "Illuminate\\Auth\\Access\\Gate",
            "type": "->"
        },
        {
            "file": "/home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php",
            "line": 281,
            "function": "raw",
            "class": "Illuminate\\Auth\\Access\\Gate",
            "type": "->"
        },
        {
            "file": "/home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Foundation/Auth/Access/AuthorizesRequests.php",
            "line": 23,
            "function": "authorize",
            "class": "Illuminate\\Auth\\Access\\Gate",
            "type": "->"
        },
        {
            "file": "/home/vagrant/DoSomething.org/rogue/app/Http/Controllers/PostsController.php",
            "line": 146,
            "function": "authorize",
            "class": "Rogue\\Http\\Controllers\\Controller",
            "type": "->"
        },
        {
            "function": "update",
            "class": "Rogue\\Http\\Controllers\\PostsController",
            "type": "->"
        },
        {
            "file": "/home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Routing/Controller.php",
            "line": 54,
            "function": "call_user_func_array"
        },
        {
            "file": "/home/vagrant/DoSomething.org/rogue/vendor/laravel/framework/src/Illuminate/Routing/ControllerDispatcher.php",
            "line": 45,
            "function": "callAction",
            "class": "Illuminate\\Routing\\Controller",
            "type": "->"
        },
```

DFurnes: Interesting, I just pulled down and I'm getting a different error! I see this:

> Type error: Argument 1 passed to Rogue\Policies\PostPolicy::update() must be an instance of Rogue\Models\Post, instance of DoSomething\Gateway\Server\RemoteUser given

If I update your policy to have the first argument as the `Authenticatable` user, it works!

```php

<?php

use Illuminate\Contracts\Auth\Authenticatable;
// ...

class PostPolicy
{
    // ...

    public function update(Authenticatable $user, Post $post)
    {
        return is_staff_user() || auth()->id() === $post->northstar_id;
    }
}
```

DFurnes: You can also then update your conditional to check the passed `$user`'s ID against the post (updated in my code snippet above). It'd probably be a good idea to inline the checks in `is_staff_user()` into this policy then so they can also use the provided user instance.

chloealee:
@DFurnes love the idea of permission policies! However, as I'm building this out, besides for organization purposes, what is the real difference using policies here (since we still have to write it in the controller and just have another file)?

chloealee:
some of the tests aren't passing for the `show` policy but all seem ok for the `update` policy. When I run the tests, I'm dropping a `dd('show policy');` and `dd('update policy');` in each methods and it reaching the `update` policy but not the `show` policy. I don't think I did anything differently to set these two up so am confused why these aren't working? Works when I test out manually!

app/Policies/PostPolicy.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Determine if the given post can be seen by the user.
+     *
+     * @param  Illuminate\Contracts\Auth\Authenticatable $user
+     * @param  Rogue\Models\Post $post
+     * @return bool
+     */
+    public function show(Authenticatable $user, Post $post)
+    {
+        if ($post->status != 'accepted') {
```
sbsmith86: I think we usually do literal comparisons with `!==`

chloealee: ah thanks @sbsmith86 ! Updated :)

chloealee:
ah ok - figured it out! Looks like the Post model factory doesn't create a `northstar_id` so when I did `$post->northstar_id` it was `null`! Updated the model factory :)

app/Http/Controllers/PostsController.php
```diff
@@ -120,15 +120,18 @@ public function store(PostRequest $request)
      */
     public function show(Post $post)
     {
-        // Only allow an admin or the user who owns the post to see their own unapproved posts.
-        if ($post->status != 'accepted') {
-            if (is_staff_user() || auth()->id() === $post->northstar_id) {
+        // If there is no user (anonymous request), only return a post if it is approved.
+        if (auth()->id() === null) {
```
chloealee: Since this is an anonymous request, there is no `$user` being sent to the policy and it seems to break and never reach the `show` policy logic. 

The only way I could get this to work (kind of - still having trouble having the `AuthorizationException` being thrown, getting a `Undefined variable: errors (View: /home/vagrant/DoSomething.org/rogue/resources/views/layouts/master.blade.php) (View: /home/vagrant/DoSomething.org/rogue/resources/views/layouts/master.blade.php)` error when the post's `status` is `rejected` or `pending`) was to do something like this but this seems confusing to add this logic in addition to the policy. 

Any ideas as to how to send an anonymous user to the policy?

chloealee: do we want to do something like [this](https://laravel.io/forum/07-28-2014-force-special-user-instance-for-anonymous-user)?

chloealee: Note: looks like this is broken with the current code as well (throws the same error for an anonymous request for a `pending` or `rejected` post).

sbsmith86: @chloealee could we do something like this within the policy? https://laracasts.com/discuss/channels/laravel/user-laravel-policies-with-anonymous-users?page=1

sbsmith86: Like using that `before()` method to decide what to do with the anon user?

DFurnes: Hmm, that's frustrating! I'm surprised there isn't better support for that in the framework.

> I tried that but it seems like it doesn't even reach the `before()` method if there is no `$user` :-/

Yeah, it looks like Laravel [short-circuits and return `false` here](https://github.com/laravel/framework/blob/39165ddea22cf837a34a8b01e4974cebb5c1af6d/src/Illuminate/Auth/Access/Gate.php#L299-L301) if there's no user. I wonder if the cleanest solution would be to override the `Gate` class (by binding a custom implementation in our AppServiceProvider) and override `raw` to skip that conditional?

chloealee: @DFurnes I think that's a great idea! That's what I was thinking too after reading [this post](https://stackoverflow.com/questions/35317909/authorization-and-policies-for-guest-users) on Stackoverflow - I think they are suggesting something similar?

chloealee: @DFurnes that sounds good to me! How would we override the `Gate` class in the `AppServiceProvider`? I tried to add a `raw` function in [that file](https://github.com/DoSomething/rogue/blob/master/app/Providers/AppServiceProvider.php) but I don't think that is the correct way to do it?

DFurnes: We'd be making our own version of [the `Gate` contract](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Contracts/Auth/Access/Gate.php) (that extends the [default one](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Auth/Access/Gate.php), but overrides just that `raw` method). You can then [bind our custom implementation](https://laravel.com/docs/5.6/container#binding) in the app service provider.

chloealee: Thanks @DFurnes ! Sorry I'm having a hard time wrapping my head around this!  So do we just copy the [`Gate` contract](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Contracts/Auth/Access/Gate.php) and add a `raw` method there? Where should this file go? I don't think we want it to be in `vendor/laravel/framework/src/Illuminate/Contracts/Auth/Access` since it is custom?

sbsmith86: I _think_ you can put it in any sensible place. maybe a `App/Contract/Auth/` directory. And then in the binding that dave pointed to, you specify which file to point to.

DFurnes: Yep, I think `app/Auth/Gate.php` is a fine place for this!

DFurnes: > So do we just copy the [`Gate` contract](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Contracts/Auth/Access/Gate.php) and add a `raw` method there? Where should this file go? I don't think we want it to be in `vendor/laravel/framework/src/Illuminate/Contracts/Auth/Access` since it is custom?

We'll be making a new implementation of that same contract! The contract describes what the class can do, and it's what allows Laravel to let us provide our own implementation (since it knows it must have the same methods, then).

So we want to make a new `Gate` implementation (let's say in `app/Auth/Gate.php`) that implements that contract, extends the default (so we get all the existing behavior), and then just overrides that one method to work a little differently. So like this:

```php
<?php

namespace Rogue\Auth;

use Illuminate\Auth\Access\Gate;
use Illuminate\Contracts\Auth\Access\Gate as GateContract;


class Gate extends Gate implements GateContract
{
    protected function raw($ability, $arguments = [])
    {
        // copy-paste the implementation from Laravel's Gate, but without that
        // first conditional that we don't want!
    }
}

```

chloealee: this is super helpful, thanks @DFurnes and @sbsmith86 ! When binding, I think I would do something like this: 
```
$this->app->bind(
            'Illuminate\Contracts\Auth\Access\Gate',
            'Rogue\Auth\Gate'
        );
```
I'm getting an error though - is this not the way to bind? I'll push up what I've got now!

app/Auth/CustomGate.php
```diff
@@ -0,0 +1,36 @@
...
+use Illuminate\Auth\Access\Gate;
+use Illuminate\Contracts\Auth\Access\Gate as GateContract;
+
+class CustomGate extends Gate implements GateContract
+{
+    protected function raw($ability, $arguments = [])
+    {
+        if (! $user = $this->resolveUser()) {
+            // return false;
+            $user = new User;
```
chloealee: I think we need to return a `$user` here so the rest of the code works, but we probably want to set a `$user` from `Illuminate\Contracts\Auth\Authenticatable;` and not `Rogue\Models\User` - is this correct? I don't know if this is something we can do though... 

When I try to do set a `$user` from `Illuminate\Contracts\Auth\Authenticatable`, I get this error: 
```
"Cannot instantiate interface Illuminate\Contracts\Auth\Authenticatable"
``` 
and when I try to set a `$user` from `Rogue\Models\User`, I get: 
```
"Undefined variable: errors (View: /home/vagrant/DoSomething.org/rogue/resources/views/layouts/master.blade.php) (View: /home/vagrant/DoSomething.org/rogue/resou ▶"
```

chloealee: @DFurnes it doesn't like that either :-/ Got this error: 

```
"Type error: Argument 1 passed to Rogue\Policies\PostPolicy::show() must be an instance of Illuminate\Contracts\Auth\Authenticatable, null given, called in /home ▶"
```

chloealee:
Thanks for all of the help @DFurnes and @sbsmith86 ! Now all tests are passing except when you try to get a `rejected` or `pending` post as an anon user, I get this error:
```
Undefined variable: errors (View: /home/vagrant/DoSomething.org/rogue/resources/views/layouts/master.blade.php) (View: /home/vagrant/DoSomething.org/rogue/resources/views/layouts/master.blade.php)
```
This is also happening in current code (before switching to policies) but I just didn't notice until now! Trying to wrangle this issue out now.

chloealee:
> I'm also curious if it makes sense to add a middleware to always imply the Accept: application/json on all /api routes (since it's been a source of little mistakes in the past, and is a little tedious to have to set when testing a route via Postman/Paw/cURL...)

@DFurnes this is a great idea! I'm having trouble finding where the `guard:api` middleware is defined to add this in...

app/Providers/AuthServiceProvider.php
```diff
@@ -13,6 +15,7 @@ class AuthServiceProvider extends ServiceProvider
      */
     protected $policies = [
         'Rogue\Model' => 'Rogue\Policies\ModelPolicy',
```
DFurnes: Can remove this example line now.

app/Auth/CustomGate.php
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Rogue\Auth;
+
+use Illuminate\Support\Arr;
+use Illuminate\Auth\Access\Gate;
+use Illuminate\Contracts\Auth\Access\Gate as GateContract;
+
+class CustomGate extends Gate implements GateContract
+{
+    protected function raw($ability, $arguments = [])
```
DFurnes: It'd be nice to add a docblock explaining what we've overridden here:

```
    /**
     * Get the raw result from the authorization callback. We've overridden this
     * method to allow users to be `null` so policies can apply to anonymous users.
     *
     * @param  string  $ability
     * @param  array|mixed  $arguments
     * @return mixed
     */
```

chloealee: ah great call!

DFurnes:
One more thought - it'd be nice to pull "view" logic into policies as well:

```php
/**
  * Determine if the full post should be displayed.
  *
  * @param  Illuminate\Contracts\Auth\Authenticatable $user
  * @return bool
  */
public function viewAll($user, Post $post)
{
    return is_staff_user() || $user->northstar_id === $post->northstar_id;
}
```

And then in our transformers:

```php
if (Gate::allows('viewAll', $post) {
    $response['tags'] = $post->tagSlugs();
    $response['source'] = $post->source;
    // ....
}
```

Not sure if there's a better name for this permission, up to you!

DFurnes:
The `Gate` facade is just another way of accessing model policies (and is [behind the scenes](https://github.com/laravel/framework/blob/e501d88440ddcd629a65137a627ab0b21aab0bad/src/Illuminate/Foundation/Auth/Access/AuthorizesRequests.php) what the controller's `authorize` method uses)! In both cases, it reads the mapping you've defined in `AuthServiceProvider` to determine which policy to use to authorize that record.

PR #707, opened 2018-05-22
Drops remote_addr column in Posts table and returns standard IP address for all posts

chloealee:
#### What's this PR do?
- For GDPR, we don't want to store IP addresses anymore in Rogue. This PR drops the `remote_addr` column in `posts` table, doesn't look for the IP address when creating a post, and returns standard IP address for all posts (`000.000.00.0`) to not break anything. 
- This also updates documentation and Model Factories so tests pass. 

cc @sheyd will this affect anything with data? 

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157484136) Pivotal Card.

#### How to deploy


app/Models/Post.php
```diff
@@ -195,7 +195,6 @@ public function toBlinkPayload()
             'caption' => $this->text,
             'text' => $this->text,
             'status' => $this->status,
-            'remote_addr' => $this->remote_addr,
```
katiecrane: I think this spot and the Quasar one below are where we might want to return `000.000.00.0` too

chloealee: ah you're totally right @katiecrane thank you!

docs/endpoints/Legacy/Two/posts.md
```diff
@@ -120,7 +120,7 @@ POST /api/v2/posts
   - **source**: (string).
     Where the post was submitted from.
   - **remote_addr**: (string).
-    IP address of where the post is submitted from. 
+    IP address of where the post is submitted from. Will be `000.000.00.00` for all posts in compliance with GDPR.
```
katiecrane: I think it's safe to remove "IP address of where the post is submitted from." since that will no longer be true!

tests/Http/Legacy/Two/PostApiTest.php
```diff
@@ -41,7 +41,7 @@ public function testCreatingAPostAndSignup()
             'num_participants' => null,
             'caption'          => $caption,
             'source'           => 'phpunit',
-            'remote_addr'      => $this->faker->ipv4,
+            'remote_addr'      => '000.000.00.00',
```
katiecrane: Will we still be getting sent `remote_addr` in requests?

chloealee: ah you're right! I got mixed up and thought this was the response - thanks, @katiecrane !

chloealee:
@sheyd changed these values to `0.0.0.0`! Is this OK to merge with this update?

chloealee:
Great, thanks @sheyd !

PR #708, opened 2018-05-23
Adds "view" logic to post policy

chloealee:
#### What's this PR do?
Adds "view" logic to Post Policy (that should have been added in #702 !) 

- Only admins / owners should be able to see a Post's `tags`, `source`, `source_details`, `remote_addr`, and `details` via the API.

I noticed that for `v2`, anyone gets `tags` and `source` - should we update this as well? Since we don't use access tokens though I'm not sure if this is possible? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156700437) Pivotal Card and finishes up #702 

#### How to deploy

### DIFF
--- a/app/Http/Transformers/Legacy/Two/PostTransformer.php
+++ b/app/Http/Transformers/Legacy/Two/PostTransformer.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Http\Transformers\Legacy\Two;
 
+use Gate;
 use Carbon\Carbon;
 use Rogue\Models\Post;
 use League\Fractal\TransformerAbstract;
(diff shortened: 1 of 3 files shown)


DFurnes:
Yeah, we'd have to be careful. I know Ashes queries by tags, but might not need to read them per post.

chloealee:
@DFurnes should I do this as apart of this PR or maybe make a card for it to tackle and make sure it doesn't break anything?

DFurnes:
I'm good with making a card in the backlog! Maybe we'll just de-commision v2 by then... 😏 🔥

chloealee:
Cool [here's](https://www.pivotaltracker.com/n/projects/2019429/stories/157818005) the card for reference! :)

PR #709, opened 2018-05-23
Anonymizes data in Rogue so that we're compliant with GDPR

chloealee:
#### What's this PR do?
This PR anonymizes data in Rogue so that we're compliant with GDPR:
- Changes all EU member signups' `why_participated` to be `EU Member. Removed because of GDPR`.
- Changes all EU member posts' `text` to be `EU Member. Removed because of GDPR`.
- Deletes all EU member posts' image URLs from s3 and changes them to `NULL` in the database. 
- Tags all EU member posts as "Hide In Gallery" 
- All of the above is changed for all records, even if they have been soft deleted. 

👀 
Did I miss anything that we need to do here? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157709055) Pivotal Card.

#### How to deploy


app/Console/Kernel.php
```diff
@@ -25,6 +25,7 @@ class Kernel extends ConsoleKernel
         Commands\ReviewSignup::class,
         Commands\SendToQuasar::class,
         Commands\ForceDeleteTestRecords::class,
+        Commands\GDPRComplianceCommand::class,
```
DFurnes: Totally unrelated to this pull request, but we can have Laravel auto-load these (see [Northstar](https://github.com/DoSomething/northstar/blob/742e3de1d46621926b540cd2052456e08d500e70/app/Console/Kernel.php#L20-L29)'s kernel).

app/Console/Commands/GDPRComplianceCommand.php
```diff
@@ -0,0 +1,108 @@
...
+        $temp = tempnam('temp', 'command_csv');
+        file_put_contents($temp, fopen($this->argument('path'), 'r'));
+
+        // Load users from the CSV
+        $users_csv = Reader::createFromPath($temp, 'r');
+        $users_csv->setHeaderOffset(0);
+        $users = $users_csv->getRecords();
+
+        // Anonymize user data.
+        $this->info('rogue:GDPRcompliance: Anonymizing data...');
```
DFurnes: Nitpicky, let's use `rogue:gdpr` here like our other log lines.

chloealee: ah yes - I shortened it and forgot to change there! Thanks for pointing that out!

app/Console/Kernel.php
chloealee: Yes, I love this!!

PR #710, opened 2018-05-24
Fixes bug to actually delete from s3 when a post is deleted and adds fastly purge to GDPR script

chloealee:
#### What's this PR do?
Fixes bug to actually delete from s3 when a post is deleted and adds fastly purge to GDPR script in #709 

Delete a post (via API or by running GDPR script) - the url should also be deleted from s3! 

@DoSomething/team-bleed since this wasn't working before, do we need to do any work to delete images for posts that have been soft deleted? Is this even possible to find the file names again? 

#### Relevant tickets
More work to fix [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157709055) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Services/AWS.php
+++ b/app/Services/AWS.php
@@ -119,6 +119,7 @@ class AWS
         // any exception that is thrown while trying to delete and just returns true.
         // see: \Illuminate\Filesystem\FilesystemAdapter::delete().
         // So we check if the file exists first and then try to delete it.
+        dd('sending ' . $path . ' to storage');
         if (Storage::exists($path)) {
             $success = Storage::delete($path);
         } else {
--- a/app/Services/AWS.php
+++ b/app/Services/AWS.php
@@ -119,7 +119,7 @@ class AWS
         // any exception that is thrown while trying to delete and just returns true.
         // see: \Illuminate\Filesystem\FilesystemAdapter::delete().
         // So we check if the file exists first and then try to delete it.
-        dd('sending ' . $path . ' to storage');
+        dump('sending to exists ' . $path);
         if (Storage::exists($path)) {
             $success = Storage::delete($path);
         } else {
--- a/app/Services/AWS.php
+++ b/app/Services/AWS.php
@@ -120,7 +120,6 @@ class AWS
         // any exception that is thrown while trying to delete and just returns true.
         // see: \Illuminate\Filesystem\FilesystemAdapter::delete().
         // So we check if the file exists first and then try to delete it.
-        Log::info('sending to exists ' . $path);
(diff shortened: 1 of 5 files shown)


app/Services/AWS.php
```diff
@@ -119,6 +119,7 @@ public function deleteImage($path)
         // any exception that is thrown while trying to delete and just returns true.
         // see: \Illuminate\Filesystem\FilesystemAdapter::delete().
         // So we check if the file exists first and then try to delete it.
+        dd('sending ' . $path . ' to storage');
         if (Storage::exists($path)) {
```
chloealee: I'm seeing something strange here - I did this `dd` along with a `dd` in `namespace Illuminate\Filesystem\Filesystem.php` which `Storage::exists()` sends to. 

I expect to get the `dd('sending ' . $path . ' to storage');` message first when I hit the `DELETE` endpoint but instead, I am getting the `dd` message I placed in `Filesystem.php` with this `$path`: `"/home/vagrant/DoSomething.org/rogue/bootstrap/cache/services.php"`

I don't know why it is hitting that before the `dd('sending ' . $path . ' to storage');` - @sbsmith86 is this the caching you were mentioning perhaps?

chloealee: It seems like this is getting called before anything in the function `deleteImage` but I'm not sure where this is getting called? cc @DoSomething/team-bleed

katiecrane: @chloealee maybe try changing that dd in `Filesystem.php` to a log message in case that is getting hit on something you're not expecting, then you can still see the info on the hit you were expecting

chloealee: oh that is a great idea, @katiecrane ! So I did `dump` instead and got this:
```
"exists /home/vagrant/DoSomething.org/rogue/bootstrap/cache/services.php"
"exists /home/vagrant/DoSomething.org/rogue/bootstrap/cache/routes.php"
"exists /home/vagrant/DoSomething.org/rogue/bootstrap/cache/routes.php"
"sending to exists 198-923df5957838355206574f72d5520f0f-1527198297.jpeg"
{"success":{"code":200,"message":"Post deleted."}}
```

Seems like it doesn't actually go to `exists` function from `deleteImage()`?

chloealee:
I think I just need to mock the fastly purge to get this test to pass! Getting this error: 
```
"message": "Server error: `POST https://api.fastly.com/service//purge/post-152` resulted in a `500 Internal Server Error` response:\n{\"msg\":\"An error occurred while connecting to the fastly API, please try your request again.\",\"detail\":\"There has been a (truncated...)\n",
```

PR #711, opened 2018-05-25
Command to update signups and posts based on arguments passed in

chloealee:
#### What's this PR do?
- Updates `MakeSourceSms.php` command to be more general: 
  - Takes in three arguments to update signups AND posts by: 
    - `target`: The name of the field to update
    - `targetOldValue`: The value to search the target with.
    - `targetNewValue`: The value to update the target with.
    - (e.g. find all signups with the source (`target`) that equals `NULL` (`targetOldValue`) and update it to `phoenix-ashes` (`targetNewValue`).
- This script will allow us to update all signups and posts with the source values of `NULL` to `phoenix-ashes` since this is where they came from. This will make data more clean and accurate!

- Run the script (e.g. `php artisan rogue:updatefield --target=source --targetOldValue=NULL --targetNewValue=phoenix-ashes`)
- Make sure all previously `NULL` values for `source` are now `phoenix-ashes`

#### Relevant tickets
Fixes [https://www.pivotaltracker.com/n/projects/2019429/stories/155778318](this) Pivotal Card.

#### How to deploy


chloealee:
@jamjensen do we also want the posts with the `source` `NULL` to be phoenix-ashes? If so, I think this script will work! If not, I'd lean towards breaking this script in two different ones - one specific for posts and one specific for signups.

jamjensen:
@chloealee that would be incredibly helpful!

chloealee:
thanks for the quick response @jamjensen ! 

This PR is ready for review then, @DoSomething/team-bleed !

app/Console/Commands/UpdateSignupAndPostField.php
```diff
@@ -0,0 +1,104 @@
...
+            $this->error('No target old value specified.');
+
+            return;
+        }
+
+        if (! $this->option('targetNewValue')) {
+            $this->error('No target new value specified.');
+
+            return;
+        }
```
katiecrane: If these are all required I wonder if they should be regular arguments instead of options so that we don't have to do these checks ourselves?

app/Console/Commands/UpdateSignupAndPostField.php
```diff
@@ -0,0 +1,104 @@
...
+            }
+        });
+
+        // Log that updating signups are finished
+        info('rogue:updatefield: Finished updating signups!');
+
+        // Start updating posts
+        info('rogue:updatefield: Starting to update posts!');
+
+        // Get all posts that have "sms-mobilecommons" set as their source and update to "sms"
```
katiecrane: outdated comment!

chloealee: oops, thanks @katiecrane !

app/Console/Commands/UpdateSignupAndPostField.php
```diff
@@ -0,0 +1,104 @@
...
+        Post::withTrashed()->where($targetField, $targetOldValue)->chunkById(100, function ($posts) use ($targetField, $targetNewValue) {
```
katiecrane: What happens if we are updating a field that only lives on `signups` but not on `posts`? Does this error out?

chloealee: @katiecrane that is a good question! Yes - we'd _only_ want to use this if we're updating the exact same `target`, searching for the exact same `targetOldValue` and replacing with the exact same `targetNewValue`. This script works for this particular case but maybe we should just separate them into two different scripts right now (one for signups and one for posts)?

app/Console/Commands/UpdateSignupAndPostField.php
chloealee: ah I think you're right! I will update that now :)

app/Console/Commands/UpdateSignupAndPostField.php
katiecrane: @chloealee hmm I wonder if it could make sense to here use an option for which models to deal with? Seems easier to run just one command if you DO want to update something that lives on both. Could have like `--signups --posts` for if both should be update? What do you think, is that overkill?

chloealee: oooh @katiecrane I really like that idea - smart!! I'll work on that too in addition to making the other fields arguments instead of options. Thanks!!

chloealee:
@katiecrane this is ready for another look now with arguments and options to tell which tables we want to update (signups and/or posts)!

PR #712, opened 2018-05-29
Updates WithReviewing Component to use v3 endpoints 

chloealee:
#### What's this PR do?
- Updates WithReviewing Component to use v3 endpoints 
- Adds `api/v3/events` endpoint, documentation, and tests 
- Moves `ApiController` to base level of `Controllers` folder instead of in `Legacy` folder

- All tests should still pass! 

- The only endpoint that doesn't use v3 is the [endpoint to rotate an image](https://github.com/DoSomething/rogue/blob/master/resources/assets/components/WithReviewing/index.js#L211) - I don't think we necessarily need this in v3 but let me know if I'm wrong! @DoSomething/team-bleed 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156670788) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Legacy/Two/EventController.php
+++ b/app/Http/Controllers/Legacy/Two/EventController.php
@@ -4,6 +4,7 @@ namespace Rogue\Http\Controllers\Legacy\Two;
 
 use Rogue\Models\Event;
 use Illuminate\Http\Request;
+use Rogue\Http\Controllers\ApiController;
 use Rogue\Http\Transformers\Legacy\Two\EventTransformer;
 
 class EventController extends ApiController
--- a/app/Http/Controllers/Legacy/Two/EventController.php
+++ b/app/Http/Controllers/Legacy/Two/EventController.php
@@ -4,7 +4,6 @@ namespace Rogue\Http\Controllers\Legacy\Two;
 
 use Rogue\Models\Event;
 use Illuminate\Http\Request;
-use Rogue\Http\Controllers\ApiController;
 use Rogue\Http\Transformers\Legacy\Two\EventTransformer;
 
 class EventController extends ApiController
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -7,9 +7,9 @@ use Illuminate\Http\Request;
 use Rogue\Managers\PostManager;
 use Rogue\Http\Requests\PostRequest;
 use Rogue\Repositories\SignupRepository;
+use Rogue\Http\Controllers\ApiController;
 use Rogue\Http\Transformers\PostTransformer;
 use Rogue\Http\Controllers\Traits\FiltersRequests;
-use Rogue\Http\Controllers\Legacy\Two\ApiController;
 
 class PostsController extends ApiController
 {
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -7,7 +7,6 @@ use Illuminate\Http\Request;
 use Rogue\Managers\PostManager;
 use Rogue\Http\Requests\PostRequest;
 use Rogue\Repositories\SignupRepository;
-use Rogue\Http\Controllers\ApiController;
 use Rogue\Http\Transformers\PostTransformer;
 use Rogue\Http\Controllers\Traits\FiltersRequests;
 
--- a/app/Http/Controllers/ApiController.php
+++ b/app/Http/Controllers/ApiController.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Rogue\Http\Controllers\Legacy\Two;
+namespace Rogue\Http\Controllers;
 
 use Rogue\Http\Controllers\Traits\FiltersRequests;
 use Rogue\Http\Controllers\Controller as BaseController;
--- a/app/Http/Controllers/EventsController.php
+++ b/app/Http/Controllers/EventsController.php
@@ -4,7 +4,6 @@ namespace Rogue\Http\Controllers;
 
 use Rogue\Models\Event;
 use Illuminate\Http\Request;
-use Rogue\Http\Controllers\ApiController;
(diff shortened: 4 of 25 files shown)


app/Http/Controllers/PostsController.php
```diff
@@ -9,7 +9,6 @@
 use Rogue\Repositories\SignupRepository;
 use Rogue\Http\Transformers\PostTransformer;
 use Rogue\Http\Controllers\Traits\FiltersRequests;
-use Rogue\Http\Controllers\Legacy\Two\ApiController;
```
katiecrane: Does this still need to include this at its new location or no since they're in the same directory now?

chloealee: I don't think it does since they're in the same directory now!

app/Http/Controllers/Legacy/Two/EventController.php
```diff
@@ -4,6 +4,7 @@
 
 use Rogue\Models\Event;
 use Illuminate\Http\Request;
+use Rogue\Http\Controllers\ApiController;
```
sbsmith86: Above it looks like you moved `Legacy/Two/ApiController` to just `/ApiController` and then here you use the new `/ApiController` in a v2 API route. I'm wondering why we wouldn't be using `Legacy/Two/ApiController` in `/Legacy/Two/EventController.php`?

chloealee: hm you're right - I didn't notice that actually! Thanks for pointing that out, @sbsmith86 !

The only different between the `Event` and `Events` Controllers is the different middleware. I think it makes sense to delete the `/Legacy/Two/EventController.php`and leave the `Events` Controller and have everything that used `/Legacy/Two/EventController.php` to use the `Events` Controller at the root of the Controller directory?

sbsmith86: I think that anything that lives under `Legacy/Two` should use things that live under `Legacy/Two` 

I think that will make it easier to kill because it will really be silo'ed.

For example, here we are using root level `ApiController` in a `Legacy/Two` classes so that starts to get confusing to follow.

chloealee: that makes sense to me too! The reason why I moved `Legacy/Two/ApiController` to the root level of the Controllers is because v3 controllers also use `Legacy/Two/ApiController`. I thought it made more sense for this shared file to live at the root so when we do get rid of the `Legacy/Two` classes, we don't have to move the `ApiController` then. 

Is this ok for that reason or should I create two separate `ApiControllers`, similar to the `Events/Event` Controllers that live both under the Legacy folder and root folders?

sbsmith86: I see, that makes sense. I think my preference would be to have a `Legacy/Two/ApiController` and a `/ApiController`.

v2 stuff would use  `Legacy/Two/ApiController`
v3 stuff would use `/ApiController` 

This way any new development could happen in `/ApiController` without v2 stuff being affected.

chloealee: ah yes that makes a ton of sense to do it that way - thanks @sbsmith86 ! I'll update now :)

chloealee:
@sbsmith86 this is ready for another look! I think NS is down so that is why the tests aren't passing but should be ready for review (and won't merge until NS is back up and all is green!)

sbsmith86:
@chloealee shouldn't we be mocking any API calls so even if NS is down, we should not actually be making the call?

DFurnes:
Hmm, yeah. We rely on Northstar (QA, I think?) for Dusk tests since it doesn't have good built-in mocking, but unit tests should all have external requests mocked. If they aren't, now might be a good time to go one-by-one and mock them!

chloealee:
ohhh interesting @sbsmith86 and @DFurnes ! My local currently points to Northstar Thor because I was having issues with user info. on QA before. 

When I do `dd($response, $data);` [here](https://github.com/DoSomething/gateway/blob/master/src/NorthstarOAuthProvider.php#L136), I get the following: 
```
.GuzzleHttp\Psr7\Response {#2269
  -reasonPhrase: "Internal Server Error"
  -statusCode: 500
  -headers: array:5 [
    "Server" => array:1 [
      0 => "nginx/1.4.6 (Ubuntu)"
    ]
    "Content-Type" => array:1 [
      0 => "application/json"
    ]
    "Transfer-Encoding" => array:1 [
      0 => "chunked"
    ]
    "Cache-Control" => array:1 [
      0 => "no-cache, private"
    ]
    "Date" => array:1 [
      0 => "Mon, 04 Jun 2018 16:03:22 GMT"
    ]
  ]
  -headerNames: array:5 [
    "server" => "Server"
    "content-type" => "Content-Type"
    "transfer-encoding" => "Transfer-Encoding"
    "cache-control" => "Cache-Control"
    "date" => "Date"
  ]
  -protocol: "1.1"
  -stream: GuzzleHttp\Psr7\Stream {#2267
    -stream: stream resource {@1254
      wrapper_type: "PHP"
      stream_type: "TEMP"
      mode: "w+b"
      unread_bytes: 0
      seekable: true
      uri: "php://temp"
      options: []
    }
    -size: 121
    -seekable: true
    -readable: true
    -writable: true
    -uri: "php://temp"
    -customMetadata: []
  }
}
array:1 [
  "error" => array:2 [
    "code" => 500
    "message" => "Looks like something went wrong. We've noted the problem and will try to get it fixed!"
  ]
]
```

Does this mean it isn't actually mocking?

DFurnes:
Yep, good by me! Let's definitely prioritize that for this upcoming sprint, flaky tests are no good!

chloealee:
Great, thanks @DFurnes ! [Here](https://www.pivotaltracker.com/n/projects/2019429/stories/158083551) is a card for this! Once I get a an official review of this, I'll merge!

PR #713, opened 2018-05-30
adds logic to only log if signup or post id is divisible evenly by logfreq option

chloealee:
#### What's this PR do?
Adds logic to `rogue:updatefield` command to only log if signup or post id is divisible evenly by `logfreq` option.

👀 

There are 1,796,212 signups alone with the `source` equal to `NULL` - this is going to make us go over our Papertrail limit if we log each record that is being updated! 

#### Relevant tickets
Updates #711 

#### How to deploy

### DIFF
--- a/app/Console/Commands/UpdateSignupAndOrPostField.php
+++ b/app/Console/Commands/UpdateSignupAndOrPostField.php
@@ -15,7 +15,7 @@ class UpdateSignupAndOrPostField extends Command
      *
      * @var string
      */
-    protected $signature = 'rogue:updatefield {target} {targetOldValue} {targetNewValue} {--signups} {--posts}';
+    protected $signature = 'rogue:updatefield {target} {targetOldValue} {targetNewValue} {--signups} {--posts} {--logfreq=1}';
 
     /**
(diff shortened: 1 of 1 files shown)


app/Console/Commands/UpdateSignupAndOrPostField.php
```diff
@@ -15,7 +15,7 @@ class UpdateSignupAndOrPostField extends Command
      *
      * @var string
      */
-    protected $signature = 'rogue:updatefield {target} {targetOldValue} {targetNewValue} {--signups} {--posts}';
+    protected $signature = 'rogue:updatefield {target} {targetOldValue} {targetNewValue} {--signups} {--posts} {--logfreq=1}';
```
katiecrane: @chloealee I think the default should be something higher so that the logs don't get overloaded if it gets run without this option set. Maybe 10000?

chloealee: ah i was planning on having that as the default but then when actually running it doing something like `10000` but making that the default makes sense to me - thanks @katiecrane !

PR #714, opened 2018-05-30
Updates to add $logfreq to scope

chloealee:
#### What's this PR do?
Updates to add `$logfreq` to scope to the `UpdateSignupAndOrPostField` command.

👀 

#### Relevant tickets
Fixes mistake in #713 

#### How to deploy

### DIFF
--- a/app/Console/Commands/UpdateSignupAndOrPostField.php
+++ b/app/Console/Commands/UpdateSignupAndOrPostField.php
@@ -72,7 +72,7 @@ class UpdateSignupAndOrPostField extends Command
             info('rogue:updatefield: Starting to update signups!');
 
             // Get all signups that have "targetOldValue" set as their target and update to "targetNewValue"
-            Signup::withTrashed()->where($targetField, $targetOldValue)->chunkById(100, function ($signups) use ($targetField, $targetNewValue) {
+            Signup::withTrashed()->where($targetField, $targetOldValue)->chunkById(100, function ($signups) use ($targetField, $targetNewValue, $logfreq) {
                 foreach ($signups as $signup) {
                     if ($signup->id % $logfreq == 0) {
                         info('rogue:updatefield: changing ' . $targetField . ' to ' . $targetNewValue . ' for signup ' . $signup->id);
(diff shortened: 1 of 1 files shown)


chloealee:
@katiecrane another thought - I'm thinking that [this log](https://github.com/DoSomething/rogue/blob/master/app/Jobs/SendSignupToQuasar.php#L62) for signups to Quasar and [this log](https://github.com/DoSomething/rogue/blob/master/app/Jobs/SendPostToQuasar.php#L62) to posts to Quasar will clog the logs in Papertrail as well :-/ Did you run into this before?

katiecrane:
@chloealee ah yes great catch! Yeah I have run into that before, there is a second argument that you can provide when the job is dispatched, but I'm not sure that's helpful in this case since the job gets dispatched automatically and not part of the command 🤔

chloealee:
Finished [conversation in Slack](https://dosomething.slack.com/archives/C0YGXUE01/p1527698031000221) and decided to run on thor to see how much it used, make an estimate of how much space it will take in Papertrail, and then ask to increase Papertrail usage just when script is running.

chloealee:
Actually @katiecrane I just had another idea for this! Perhaps instead of sending to the `[SignupManager](https://github.com/DoSomething/rogue/blob/master/app/Managers/SignupManager.php#L65)`/`[PostManager](https://github.com/DoSomething/rogue/blob/master/app/Managers/PostManager.php#L77)` to update, I send to the `[SignupRepository](https://github.com/DoSomething/rogue/blob/master/app/Repositories/SignupRepository.php#L39)`/`P[ostRepository](https://github.com/DoSomething/rogue/blob/master/app/Repositories/PostRepository.php#L127)` to update records. 

Then, I can manually send to Quasar via the jobs ([SendSIgnupToQuasar](https://github.com/DoSomething/rogue/blob/master/app/Jobs/SendSignupToQuasar.php#L35) and [SendPostToQuasar](https://github.com/DoSomething/rogue/blob/master/app/Jobs/SendPostToQuasar.php#L35)) and turn the logging off as the second argument? 

The only problem is that it probably won't log any at all and that might be concerning?

sbsmith86:
@chloealee @katiecrane I think we should try and be disciplined about using those Manager classes so that we don't have a bunch of one-off cases. If we need to turn off logging in those classes, let's do that there (add a param or something to turn logging off)

chloealee:
@sbsmith86 ah that makes sense to me! So instead of increasing Papertrail for running the script, do you think it would be better to add a param to turn logging off when the Signup/Post is being sent to Quasar and have the script pass that through?

sbsmith86:
For example, `PostManager@update($post, $data)`

Could be updated to have another param (default to on)

`public function update($post, $data, $logging = true)`

sbsmith86:
@chloealee I think it is a worthwhile update since we run into this a lot.

chloealee:
@sbsmith86 ok sounds good!

PR #715, opened 2018-05-30
Removes type-hint so non-user apps can use update post policy method

chloealee:
#### What's this PR do?
Removes `Authenticatable` type-hint so non-user apps can use update post policy method.

👀 

We need to remove this so Chompy can finish updating! 

#### Relevant tickets
Slack thread [here](https://dosomething.slack.com/archives/C0YGXUE01/p1527701337000521).

#### How to deploy

### DIFF
--- a/app/Policies/PostPolicy.php
+++ b/app/Policies/PostPolicy.php
@@ -3,7 +3,6 @@
 namespace Rogue\Policies;
 
 use Rogue\Models\Post;
-use Illuminate\Contracts\Auth\Authenticatable;
 use Illuminate\Auth\Access\HandlesAuthorization;
 
 class PostPolicy
--- a/app/Policies/PostPolicy.php
+++ b/app/Policies/PostPolicy.php
@@ -21,7 +21,7 @@ class PostPolicy
     /**
      * Determine if the full post should be displayed.
      *
-     * @param  $user
+     * @param  Illuminate\Contracts\Auth\Authenticatable $user
      * @param  Rogue\Models\Post $post
      * @return bool
      */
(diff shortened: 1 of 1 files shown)


sbsmith86:
looks good. We might need to write a test for this case. Maybe write up a card for the test, but let's get this out the door.

chloealee:
@sbsmith86 great call - card [here](https://www.pivotaltracker.com/n/projects/2019429/stories/157977525)!

app/Policies/PostPolicy.php
```diff
@@ -22,7 +21,7 @@ public function __construct()
     /**
      * Determine if the full post should be displayed.
      *
-     * @param  Illuminate\Contracts\Auth\Authenticatable $user
+     * @param  $user
```
DFurnes: Let's leave the docblock since it's handy to know what this will be (even though PHP won't enforce it). When we upgrade Rogue to use PHP 7.1 at some point, this would be a great use-case for [nullable types](http://php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types), e.g:

```php
public function viewAll(?Authenticatable $user, Post $post) { // ...
```

app/Policies/PostPolicy.php
```diff
@@ -56,11 +55,11 @@ public function show($user, Post $post)
...
      * @param  Rogue\Models\Post $post
      * @return bool
      */
-    public function update(Authenticatable $user, Post $post)
+    public function update($user, Post $post)
     {
         return is_staff_user() || $user->northstar_id === $post->northstar_id;
```
DFurnes: I think we'll want to check if user exists here before comparing IDs, and pull the `is_staff_user` up into a separate condition before that. Otherwise I think this might fail if `$user` is null.

```php
if (is_staff_user()) {
    return true;
}

return $user && $user->northstar_id === $post->northstar_id;
```

DFurnes: Although this might not be a big deal since we have this endpoint protected by the `auth` middleware... still, can't hurt to be defensive about it!

chloealee: agreed - updated!

PR #716, opened 2018-05-30
updates Signup/PostManger to have option to not log when sending to Quasar

chloealee:
#### What's this PR do?
- updates Signup/PostManger to have option to not log when sending to Quasar
- updates `UpdateSignupAndOrPostField` command to not log when updating Signup/Post in Quasar

👀 

Goes along with [this discussion](https://github.com/DoSomething/rogue/pull/714#issuecomment-393259907).

#### How to deploy

### DIFF
--- a/app/Console/Commands/UpdateSignupAndOrPostField.php
+++ b/app/Console/Commands/UpdateSignupAndOrPostField.php
@@ -79,6 +79,7 @@ class UpdateSignupAndOrPostField extends Command
                     }
 
                     // Update signup but don't log that it was sent to Quasar in interest of taking up too much space in Papertrail.
+                    // @TODO: when running updates for smaller amount of records, do we want to turn logs back on?
                     $this->signups->update($signup, [$targetField => $targetNewValue], false);
                 }
             });
@@ -99,6 +100,7 @@ class UpdateSignupAndOrPostField extends Command
                     }
 
                     // Update post but don't log that it was sent to Quasar in interest of taking up too much space in Papertrail.
+                    // @TODO: when running updates for smaller amount of records, do we want to turn logs back on?
                     $this->posts->update($post, [$targetField => $targetNewValue], false);
                 }
             });
--- a/app/Console/Commands/UpdateSignupAndOrPostField.php
+++ b/app/Console/Commands/UpdateSignupAndOrPostField.php
@@ -78,7 +78,8 @@ class UpdateSignupAndOrPostField extends Command
                         info('rogue:updatefield: changing ' . $targetField . ' to ' . $targetNewValue . ' for signup ' . $signup->id);
                     }
 
-                    $this->signups->update($signup, [$targetField => $targetNewValue]);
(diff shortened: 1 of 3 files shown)


app/Managers/PostManager.php
```diff
@@ -72,9 +72,10 @@ public function create($data, $signupId, $authenticatedUserRole = null)
      *
      * @param \Rogue\Models\Post $post
      * @param array $data
+     * @param bool $log
      * @return \Rogue\Models\Post
      */
-    public function update($post, $data)
+    public function update($post, $data, $log)
```
sbsmith86: I think we should give this a default value. `$log = true`

chloealee: ah right - forgot that!

chloealee:
added one last thing (`@TODO`) - ready for another look! :)

app/Console/Commands/UpdateSignupAndOrPostField.php
```diff
@@ -78,7 +78,9 @@ public function handle()
                         info('rogue:updatefield: changing ' . $targetField . ' to ' . $targetNewValue . ' for signup ' . $signup->id);
                     }
 
-                    $this->signups->update($signup, [$targetField => $targetNewValue]);
+                    // Update signup but don't log that it was sent to Quasar in interest of taking up too much space in Papertrail.
+                    // @TODO: when running updates for smaller amount of records, do we want to turn logs back on?
```
sbsmith86: You can take this param in as a flag in the command, so you can set the option when you run the command.

sbsmith86: `$log = $this->option('log');`

chloealee: oh great call, @sbsmith86 ! Just updated to that ! :)

app/Console/Commands/UpdateSignupAndOrPostField.php
```diff
@@ -15,7 +15,7 @@ class UpdateSignupAndOrPostField extends Command
      *
      * @var string
      */
-    protected $signature = 'rogue:updatefield {target} {targetOldValue} {targetNewValue} {--signups} {--posts} {--logfreq=10000}';
+    protected $signature = 'rogue:updatefield {target} {targetOldValue} {targetNewValue} {--signups} {--posts} {--logfreq=10000} {--log}';
```
sbsmith86: we should set a default for the `log` option as well. probably just have it on by default.

chloealee: hm @sbsmith86 I'm not sure how to set the default to `true` by default - didn't seem to work when doing `{--log=true}`, `{--log=TRUE}`, or `{--log=1}`.

sbsmith86: Ack maybe we don't need it since it is just an on/off flag! Sorry. this looks great.

chloealee: No worries - thank you for the quick review! :)

PR #718, opened 2018-05-31
adds file validation to v2 post request

chloealee:
#### What's this PR do?
adds file validation to v2 post request

👀 

Will this break anything by adding the `image` to the validation? 

#### Relevant tickets
Fixes errors we were seeing in [Papertrail](https://dosomething.slack.com/archives/C0YGXUE01/p1527729205000183)

#### How to deploy

### DIFF
--- a/app/Http/Requests/Legacy/Two/PostRequest.php
+++ b/app/Http/Requests/Legacy/Two/PostRequest.php
@@ -31,7 +31,7 @@ class PostRequest extends Request
             'caption' => 'nullable|string',
             'status' => 'in:pending,accepted,rejected',
             'source' => 'nullable|string',
-            // 'file' => 'string', // @TODO - should do some better validation around files.
+            'file' => 'image|dimensions:min_width=400,min_height=400,max_width=5000,max_height_4000',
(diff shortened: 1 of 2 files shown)


app/Http/Requests/Legacy/Two/PostRequest.php
```diff
@@ -31,7 +31,7 @@ public function rules()
             'caption' => 'nullable|string',
             'status' => 'in:pending,accepted,rejected',
             'source' => 'nullable|string',
-            // 'file' => 'string', // @TODO - should do some better validation around files.
+            'file' => 'image|dimensions:min_width=400,min_height=400,max_width=5000,max_height_4000',
```
sbsmith86: haha that `@TODO` is damning.

DFurnes: Does the `image` and `dimensions` validator work if we're uploading a data URL string here (instead of a legit file, like we do on v3)? I think that's why we originally didn't do this.

chloealee: @DFurnes tested with a base64 string and validation worked!

PR #719, opened 2018-05-31
removes image validation from v2 posts

chloealee:
#### What's this PR do?
removes image validation from v2 posts in case an image is sent in as base64 string

👀 

#### Relevant tickets
Fixes mistake in #718 

#### How to deploy

### DIFF
--- a/app/Http/Requests/Legacy/Two/PostRequest.php
+++ b/app/Http/Requests/Legacy/Two/PostRequest.php
@@ -31,7 +31,7 @@ class PostRequest extends Request
             'caption' => 'nullable|string',
             'status' => 'in:pending,accepted,rejected',
             'source' => 'nullable|string',
-            'file' => 'image|dimensions:min_width=400,min_height=400,max_width=5000,max_height_4000',
+            'file' => 'dimensions:min_width=400,min_height=400,max_width=5000,max_height_4000',
             'type' => 'string|in:photo,voter-reg,text',
             'action' => 'string',
         ];


chloealee:
@DFurnes so sorry - you were right about this! I tested incorrectly in the last PR but confirmed what you said on thor. When removing `image` validation, all works as we want to!

app/Http/Requests/Legacy/Two/PostRequest.php
```diff
@@ -31,7 +31,7 @@ public function rules()
             'caption' => 'nullable|string',
             'status' => 'in:pending,accepted,rejected',
             'source' => 'nullable|string',
-            'file' => 'image|dimensions:min_width=400,min_height=400,max_width=5000,max_height_4000',
+            'file' => 'dimensions:min_width=400,min_height=400,max_width=5000,max_height_4000',
```
DFurnes: I think the `dimensions` rule requires a multi-part file upload as well. Can you double-check this works?

chloealee: @DFurnes yes - just double checked! :) 
screen shot 2018-05-31 at 2 02 56 pm

chloealee: ah you're right @DFurnes - it gives the same error for a string with valid dimensions :-/ Should I just remove this? I think that we should be ok to wait til Phoenix moves to the v3 endpoints since it was a one time occurrence?

DFurnes: Yeah, I think that's fine. 👍

chloealee:
updated - thanks for your patience with me on this @DFurnes !

PR #720, opened 2018-05-31
Bulk Accept Post Command

chloealee:
#### What's this PR do?
- Adds the `bulk` tag
- Writes script to bulk review and tag posts constrained by the provided parameters

- To bulk approve all `text` posts under a campaign and tag as `Bulk` and `Hide In Gallery` run: `php artisan rogue:bulkreviewposts {campaign_id} pending accepted --type=text --logfreq=1000 --tag=bulk --tag=hide-in-gallery`
- All text posts in that campaign should be `accepted` and tagged as `Bulk` and `Hide In Gallery`
  - Make sure _only_ `text` posts are updated! 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/157777047) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Console/Commands/BulkReviewPosts.php
+++ b/app/Console/Commands/BulkReviewPosts.php
@@ -62,15 +62,16 @@ class BulkReviewPosts extends Command
         $log = $this->option('log');
         $tags = $this->option('tag') ?? null;
 
-        $query = (new Post)->newQuery();
-        $query = $query->where('campaign_id', $this->argument('campaign'));
-        $query = $query->where('status', $this->argument('oldStatus'));
+        $postQuery = Post::where([
+            ['campaign_id', $this->argument('campaign')],
+            ['status', $this->argument('oldStatus')],
+        ]);
 
         if ($postType !== null) {
-            $query = $query->where('type', $postType);
+           $postQuery = $postQuery->where('type', $postType);
         }
 
-        $posts = $query->get();
+        $posts  = $postQuery->get();
 
         if ($posts->isNotEmpty()) {
             foreach ($posts as $post) {
--- a/app/Console/Commands/BulkReviewPosts.php
+++ b/app/Console/Commands/BulkReviewPosts.php
@@ -68,10 +68,10 @@ class BulkReviewPosts extends Command
         ]);
 
         if ($postType !== null) {
-           $postQuery = $postQuery->where('type', $postType);
+            $postQuery = $postQuery->where('type', $postType);
         }
 
-        $posts  = $postQuery->get();
+        $posts = $postQuery->get();
 
         if ($posts->isNotEmpty()) {
             foreach ($posts as $post) {
--- a/app/Console/Commands/BulkReviewPosts.php
+++ b/app/Console/Commands/BulkReviewPosts.php
@@ -80,8 +80,8 @@ class BulkReviewPosts extends Command
                 }
 
                 // If the $log flag is included when the command is run, logging will occur in the Post Manager for each post.
-                $this->posts->update($post, ['status' => $this->argument('newStatus')], $log);
-
+                $this->posts->update($post, ['status' => $this->argument('newStatus')], $log
+            );
                 foreach ($tags as $tag) {
                     $post->tag($tag);
                 }
--- a/app/Console/Commands/BulkReviewPosts.php
+++ b/app/Console/Commands/BulkReviewPosts.php
@@ -18,8 +18,8 @@ class BulkReviewPosts extends Command
                             {oldStatus}
                             {newStatus}
                             {--type= : Filter by the type of post}
-                            {--logfreq=1000}
-                            {--log}
+                            {--logfreq=1000: Flag to decide how often this command will log when a post is being updated}
+                            {--log : Flag to allow logging in Post Manager}
                             {--tag=* : Tag(s) slug to tag post by. e.g. hide-in-gallery}';
 
     /**
@@ -76,9 +76,10 @@ class BulkReviewPosts extends Command
         if ($posts->isNotEmpty()) {
             foreach ($posts as $post) {
                 if ($post->id % $logfreq == 0) {
-                    info('rogue:bulkreviewposts: accepting and tagging post: ' . $post->id);
+                    info('rogue:bulkreviewposts: updating: ' . $post->id);
                 }
 
+                // If the $log flag is included when the command is run, logging will occur in the Post Manager for each post.
                 $this->posts->update($post, ['status' => $this->argument('newStatus')], $log);
 
                 foreach ($tags as $tag) {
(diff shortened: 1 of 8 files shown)


sbsmith86:
I wonder if we should allow for an optional status or tag to use to be passed to the command. So like if you wanted to reject posts under a campaign, you could easily do that? Not sure how much we would use that but could be nice for it to be more flexible.

chloealee:
@sbsmith86 I love that idea - updated!

app/Console/Commands/BulkReviewPosts.php
```diff
@@ -0,0 +1,90 @@
...
+use Rogue\Managers\PostManager;
+
+class BulkReviewPosts extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'rogue:bulkreviewposts {campaign} {oldStatus} {newStatus} {--type= : Filter by the type of post} {--logfreq=1000} {--log} {--tag=* : Tag(s) slug to tag post by. e.g. hide-in-gallery}';
```
sbsmith86: This is totally nitpicky, but it might be easier to read this on multiple lines: 

```
protected $signature = 'email:send
                        {user : The ID of the user}
                        {--queue= : Whether the job should be queued}';
```

app/Console/Commands/BulkReviewPosts.php
```diff
@@ -0,0 +1,90 @@
...
+    public function handle()
+    {
+        info('rogue:bulkreviewposts: Starting to bulk accept posts!');
+
+        $postType = $this->option('type') ?? null;
+        $logfreq = $this->option('logfreq');
+        $log = $this->option('log');
+        $tags = $this->option('tag') ?? null;
+
+        $query = (new Post)->newQuery();
```
sbsmith86: I think you can simplify this by querying the model directly.

```
$postQuery = Post::where([
    ['campaign_id', '=', $this->argument('campaign')],
    ['status', '<>',  $this->argument('oldStatus'),
]);

if ($postType !== null) {
   $postQuery  = $query->where('type', $postType);
}

$posts  = $postQuery->get();
```

app/Console/Commands/BulkReviewPosts.php
```diff
@@ -0,0 +1,90 @@
...
+        if ($postType !== null) {
+            $query = $query->where('type', $postType);
+        }
+
+        $posts = $query->get();
+
+        if ($posts->isNotEmpty()) {
+            foreach ($posts as $post) {
+                if ($post->id % $logfreq == 0) {
+                    info('rogue:bulkreviewposts: accepting and tagging post: ' . $post->id);
```
sbsmith86: shouldn't this log message be hidden under the `$log` flag?

app/Console/Commands/BulkReviewPosts.php
```diff
@@ -0,0 +1,90 @@
...
+                }
+
+                $this->posts->update($post, ['status' => $this->argument('newStatus')], $log);
+
+                foreach ($tags as $tag) {
+                    $tagArray = explode('-', $tag);
+                    $tag = implode(' ', $tagArray);
+                    $post->tag(ucwords($tag));
```
sbsmith86: Are we not able to tag things with the machine name like `$post->tag("hide-in-gallery")`?

app/Console/Commands/BulkReviewPosts.php
chloealee: ah yes so much cleaner!

app/Console/Commands/BulkReviewPosts.php
chloealee: ooo so much cleaner @sbsmith86 ! I didn't know you could do both `Post::where` and `$query->where` together! 

One q - I've never seen `<>` before and I think it means `!=` - in this case, would we want the below instead? Not sure if I'm not understanding something!

```
$postQuery = Post::where([
    ['campaign_id', '=', $this->argument('campaign')],
    ['status', '=',  $this->argument('oldStatus'),
]);

if ($postType !== null) {
   $postQuery  = $query->where('type', $postType);
}

$posts  = $postQuery->get();
```

app/Console/Commands/BulkReviewPosts.php
chloealee: @sbsmith86 great q! This is specific to logging in the command and `info('rogue:bulkreviewposts: accepting and tagging post: ' . $post->id);` which is controlled by `$logfreq`. 

The `$log` flag sends to the `PostManager` and hides logging that is done there.

app/Console/Commands/BulkReviewPosts.php
chloealee: @sbsmith86 we're not able to do that :-/ Doing that creates a tag with both the `tag_name` and `tag_slug` `hide-in-gallery`

app/Console/Commands/BulkReviewPosts.php
sbsmith86: oh yes yes, i was copying from the example and missed that!

app/Console/Commands/BulkReviewPosts.php
sbsmith86: ok! can you clarify that in a comment? Also maybe in the command signature. It's a little confusing to read.

chloealee: yes good call! Updated :)

app/Console/Commands/BulkReviewPosts.php
sbsmith86: I think that is something we could make more flexible in `Post::tag($tagName)`

We can check if the parameter passed in has `-` or not and then decide how to handle it there?

My thought is that it would be nice not have to do that lift when using the function (especially as we use it more). Instead we can accept either the slug or the name and handle it nicely in one place.

chloealee: that is a good idea - just updated! :)

app/Models/Post.php
```diff
@@ -264,8 +264,12 @@ public function getQuantity()
     /**
      * Apply the given tag to this post.
      */
-    public function tag($tagName)
+    public function tag($tag)
```
sbsmith86: I think this might break other uses of this where we don't pass in a slug.

I think we should actually check if we got a string with `-` in it, which indicates it is a slug and handle it accordingly. If it does not have `-` then we need to create the slug and handle it accordingly. 

We want to be able to call the function two different ways `tag("hide-in-gallery")` or `tag("Hide in Gallery")`

chloealee: I tested this by sending `new-tag` and `New Tag` and both seem to work!

sbsmith86: ah yes yes. I guess it is still odd to me to do the explode()/implode() logic when you don't need to.

chloealee: totally ! And doesn't hurt to check - updated this!

PR #724, opened 2018-06-05
Update to BulkReviewPosts script

chloealee:
#### What's this PR do?
- When running the script on thor, I was getting this error: 
```thor.ERROR: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '470884-1' for key 'PRIMARY' (SQL: insert into `post_tag` (`created_at`, `post_id`, `tag_id`, `updated_at`) values (2018-06-05 17:53:51, 470884, 1, 2018-06-05 17:53:51)) {"exception":"[object] (Illuminate\\Database\\QueryException(code: 23000): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '470884-1' for key 'PRIMARY' (SQL: insert into `post_tag` (`created_at`, `post_id`, `tag_id`, `updated_at`) values (2018-06-05 17:53:51, 470884, 1, 2018-06-05 17:53:51)) at /app/vendor/laravel/framework/src/Illuminate/Database/Connection.php:664, Doctrine\\DBAL\\Driver\\PDOException(code: 23000): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '470884-1' for key 'PRIMARY' at /app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:107, PDOException(code: 23000): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '470884-1' for key 'PRIMARY' at /app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:105)"} {"user_id":null,"client_id":null,"request_id":null}```
- This script fixes that to first check if the tag exists before re tagging to avoid duplicate entry error. 

👀 

#### Relevant tickets
Updates script in #720 

#### How to deploy

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -275,11 +275,11 @@ class Post extends Model
         // Capitalize each word in the tag.
         $tagName = ucwords($tag);
 
-        $tag = Tag::firstOrCreate(['tag_name' => $tagName], ['tag_slug' => str_slug($tagName, '-')]);
-
         // Only tag if the tag doesn't exist on the post yet.
         // Otherwise, an integrity constraint violation / duplicate entry error will be thrown.
         if (! $this->tagNames()->contains($tagName)) {
+            $tag = Tag::firstOrCreate(['tag_name' => $tagName], ['tag_slug' => str_slug($tagName, '-')]);
+
             $this->tags()->attach($tag);
 
             // Update timestamps on the Post when adding a tag
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -277,12 +277,16 @@ class Post extends Model
 
         $tag = Tag::firstOrCreate(['tag_name' => $tagName], ['tag_slug' => str_slug($tagName, '-')]);
 
-        $this->tags()->attach($tag);
+        // Only tag if the tag doesn't exist on the post yet.
+        // Otherwise, a Integrity constraint violation / duplicate entry error will be thrown.
+        if (! in_array($tagName, $this->tagNames()->toArray())) {
+            $this->tags()->attach($tag);
 
-        // Update timestamps on the Post when adding a tag
-        $this->touch();
+            // Update timestamps on the Post when adding a tag
+            $this->touch();
 
-        event(new PostTagged($this, $tag));
+            event(new PostTagged($this, $tag));
+        }
 
         return $this;
     }
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -278,8 +278,8 @@ class Post extends Model
         $tag = Tag::firstOrCreate(['tag_name' => $tagName], ['tag_slug' => str_slug($tagName, '-')]);
 
         // Only tag if the tag doesn't exist on the post yet.
-        // Otherwise, a Integrity constraint violation / duplicate entry error will be thrown.
-        if (! in_array($tagName, $this->tagNames()->toArray())) {
+        // Otherwise, an integrity constraint violation / duplicate entry error will be thrown.
+        if (! $this->tagNames()->contains($tagName)) {
             $this->tags()->attach($tag);
 
             // Update timestamps on the Post when adding a tag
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: app/Models/Post.php [hunk 1]
```diff
@@ -277,12 +277,16 @@ public function tag($tag)
 
         $tag = Tag::firstOrCreate(['tag_name' => $tagName], ['tag_slug' => str_slug($tagName, '-')]);
 
-        $this->tags()->attach($tag);
+        // Only tag if the tag doesn't exist on the post yet.
```
sbsmith86: Shouldn't this toggle tags, so if a post is already tagged with something, when we make the tag request it removes the tag instead of adding (and vice versa)?

chloealee: @sbsmith86 the endpoint [toggles](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/TagsController.php#L57-L62) but this [`tag`](https://github.com/DoSomething/rogue/blob/master/app/Models/Post.php#L267) function doesn't actually work like that  - there is another function [`untag`](https://github.com/DoSomething/rogue/blob/master/app/Models/Post.php#L293).

sbsmith86: ahhh ok 👍

app/Models/Post.php
```diff
@@ -277,12 +277,16 @@ public function tag($tag)
...
+        // Otherwise, an integrity constraint violation / duplicate entry error will be thrown.
+        if (! $this->tagNames()->contains($tagName)) {
```
chloealee: for some reason, adding this check here breaks the tests but I'm not quiet sure why! Still trying to dig into this...

sbsmith86: does `tagNames()` return a collection or an array? I'm wondering if the `contains()` method works here?

chloealee: a collection! @sbsmith86

sbsmith86: Also do you see a more verbose error in the logs when the test fails?

sbsmith86: hmmm I think I need to pull down this branch to check it out.

chloealee: ty! I'll keep digging too - this is a 👻 -y one!

sbsmith86: also what happens if you ``dd$this->tagNames()->contains($tagName)`` before the if- statement.

chloealee: @sbsmith86 doing that returns `false` (and `dd(! $this->tagNames()->contains($tagName));` returns `true`!).

chloealee: wait - this is ringing a bell that I think I had trouble with before and could only get tests to pass by tagging via the API? I'm gonna do some digging through :slack: and :github: to see what the solution was!

chloealee: I found [the comment](https://github.com/DoSomething/rogue/pull/697#issuecomment-389298124) I was thinking of but I don't think this actually has anything to do with what's going on here :-/

chloealee: Even tagging with the API is breaking: 
```
$this->withAdminAccessToken()->postJson('api/v3/posts/' . $post->id . '/tags', [
             'tag_name' => 'Good Submission',
        ]);

        $this->withAdminAccessToken()->postJson('api/v3/posts/' . $post->id . '/tags', [
             'tag_name' => 'Tag To Delete',
        ]);
``` 
and it seems as it is just this new line that the test doesn't like 🤔 
`        // if (! $this->tagNames()->contains($tagName)) {`

tests/Http/Legacy/Web/TagsTest.php
```diff
@@ -81,6 +81,7 @@ public function testAddMultipleTagsAndDeleteOne()
         $post = factory(Post::class)->create();
         $post->tag('Good Submission');
         $post->tag('Tag To Delete');
+        $post = $post->fresh();
 
         // Make sure both tags actually exist
         $this->assertContains('Good Submission', $post->tagNames());
```
chloealee: yay I finally got the tests to pass! I'm not sure why but doing `$this->assertContains('Good Submission', $post->fresh()->tagNames());` didn't seem to work like it did below with `assertNotContains` so to keep things consistent, change to get a fresh version of the post always before any assertions.

PR #725, opened 2018-06-07
Adds log param whether or not to log when post is sent to Customer.io and when post is tagged

chloealee:
#### What's this PR do?
Adds log param whether or not to log when post is sent to Customer.io and when post is tagged

👀 

When running on Thor, I realized this was logging each time when it was sent to Customer.io. We probably don't want this and want to turn on/off with `log` flag. 

Also, it logged each time the post was tagged. This PR also hides this behind a flag so the logs don't clog!

#### How to deploy

### DIFF
--- a/app/Jobs/SendPostToCustomerIo.php
+++ b/app/Jobs/SendPostToCustomerIo.php
@@ -22,13 +22,21 @@ class SendPostToCustomerIo implements ShouldQueue
     protected $post;
 
     /**
+     * Whether or not to log that this Post was sent to Quasar.
+     *
+     * @var bool
+     */
+    protected $log;
+
+    /**
      * Create a new job instance.
      *
      * @return void
      */
-    public function __construct(Post $post)
+    public function __construct(Post $post, $log = true)
     {
         $this->post = $post;
+        $this->log = $log;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


chloealee:
Another thing that I am noticing is that every time a post is tagged in this script, it is logged: 
`Jun 07 10:18:46 rogue-thor app/text.8081: [2018-06-07 14:18:46] thor.INFO: post_tagged {"id":490031,"tag":"bulk"} {"user_id":null,"client_id":null,"request_id":null} 
`

In this case, each post triggers two `post_tagged` log entries. Should we also hide this behind a flag whether or not to log these events?

sbsmith86:
yea i think so!

chloealee:
@sbsmith86 updated to also send param to turn logging on/off when tagging a post!

app/Jobs/SendPostToCustomerIo.php
```diff
@@ -21,14 +21,22 @@ class SendPostToCustomerIo implements ShouldQueue
      */
     protected $post;
 
+    /**
+     * Whether or not to log that this Post was sent to Quasar.
```
sbsmith86: So below it looks like we are logging that we sent to blink, and then here we say that we are logging if we are sending to quasar, and the Job implies we are sending to CustomerIO. It is sorta confusing, because I know things go through blink and then to other platforms, but I think we should be consistent. I guess this should just be "sending to customer.io" since that is where it ends up?

chloealee: ah yes! Thanks for catching that! Updated :)

PR #726, opened 2018-06-07
changes type option to mandatory argument in BulkReviewPosts command 

chloealee:
#### What's this PR do?
changes `type` option to mandatory argument in `BulkReviewPosts` command 

👀 

For some reason, this query wasn't only looking for the `text` type when we ran in Thor (but worked as intended on local, even when I dumped out the thor db to my local). For extra safety measure, we decided to change the `type` option to a mandatory argument in the command.

#### Relevant tickets
Updates command made in #720 

#### How to deploy

### DIFF
--- a/app/Console/Commands/BulkReviewPosts.php
+++ b/app/Console/Commands/BulkReviewPosts.php
@@ -61,13 +61,11 @@ class BulkReviewPosts extends Command
         $log = $this->option('log');
         $tags = $this->option('tag') ?? null;
 
-        $postQuery = Post::where([
+        $posts = Post::where([
             ['campaign_id', $this->argument('campaign')],
             ['status', $this->argument('oldStatus')],
             ['type', $this->argument('type')],
-        ]);
-
-        $posts = $postQuery->get();
+        ])->get();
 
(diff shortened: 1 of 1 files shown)


app/Console/Commands/BulkReviewPosts.php
```diff
@@ -57,20 +57,16 @@ public function handle()
...
-        $postType = $this->option('type') ?? null;
         $logfreq = $this->option('logfreq');
         $log = $this->option('log');
         $tags = $this->option('tag') ?? null;
 
         $postQuery = Post::where([
             ['campaign_id', $this->argument('campaign')],
             ['status', $this->argument('oldStatus')],
+            ['type', $this->argument('type')],
         ]);
```
sbsmith86: you should be able to just do `->get()` here instead of in a separate statement. 
```
         $postQuery = Post::where([
             ['campaign_id', $this->argument('campaign')],
             ['status', $this->argument('oldStatus')],
            ['type', $this->argument('type')],
         ])->get();

```

chloealee: oh right! Updated :)

PR #728, opened 2018-06-08
Disable eager relations in BulkReviewPosts command

chloealee:
#### What's this PR do?
Adds `setEagerLoads` to query to disable eager relations in `BulkReviewPosts` command. 

👀 

When I ran this on Prod, I was getting this error:
```
Jun 08 07:41:31 rogue-prod app/run.5993: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /app/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php on line 701
Jun 08 07:41:31 rogue-prod app/run.5993: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /app/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php on line 122
```
@DFurnes suggested to add this to clear this! More context [here](https://dosomething.slack.com/archives/C0YGXUE01/p1528469204000469).

#### Relevant tickets
Updates command created in #720 

#### How to deploy

### DIFF
--- a/app/Console/Commands/BulkReviewPosts.php
+++ b/app/Console/Commands/BulkReviewPosts.php
@@ -61,11 +61,11 @@ class BulkReviewPosts extends Command
         $log = $this->option('log');
         $tags = $this->option('tag') ?? null;
 
-        $posts = Post::where([
-            ['campaign_id', $this->argument('campaign')],
-            ['status', $this->argument('oldStatus')],
-            ['type', $this->argument('type')],
-        ])->get();
+        $posts = Post::setEagerLoads([])
+            ->where('campaign_id', $this->argument('campaign'))
+            ->where('status', $this->argument('oldStatus'))
+            ->where('type', $this->argument('type'))
+            ->get();
 
         if ($posts->isNotEmpty()) {
             foreach ($posts as $post) {


app/Console/Commands/BulkReviewPosts.php
```diff
@@ -61,11 +61,11 @@ public function handle()
         $log = $this->option('log');
         $tags = $this->option('tag') ?? null;
 
-        $posts = Post::where([
-            ['campaign_id', $this->argument('campaign')],
-            ['status', $this->argument('oldStatus')],
-            ['type', $this->argument('type')],
-        ])->get();
+        $posts = Post::setEagerLoads([])
```
sbsmith86: what does this do?

chloealee:
@sbsmith86 when I ran the script on Prod on Friday, I was getting the error above (weird it didn't trigger on Thor!). @DFurnes suggested that I try to [disable the eager loading](https://stackoverflow.com/questions/34052056/disable-eager-relations/34052468#34052468) of the relations since that might have been causing this error (if a signup had many posts). 

One thing I'm still confused is that the query _should_ be only querying posts but perhaps there is something under the hood that I'm not understanding that it trying to load the signups as well because of this relationship?

sbsmith86:
@chloealee ah weird

If I am reading the stackoverflow thread correctly, the `$with` property triggers the eager load here: https://github.com/DoSomething/rogue/blob/master/app/Models/Post.php#L25

But I also wonder if this has to do with images?

chloealee:
@sbsmith86 ahhh I see that makes sense! I'll give this a whirl again on thor and see what happens 🤞

chloealee:
@sbsmith86 I think you were right that it might be a different error because I just tried to run again on thor and got the same error this time:
`
Jun 11 18:53:05 rogue-thor app/run.3340: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /app/vendor/laravel/framework/src/Illuminate/Queue/CallQueuedHandler.php on line 42 
Jun 11 18:53:05 rogue-thor app/run.3340: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 73728 bytes) in /app/vendor/composer/ClassLoader.php on line 444 
Jun 11 18:53:07 rogue-thor heroku/run.3340: Process exited with status 255 
`
Also, this should only be updating `text` posts - I'm not sure if that has anything to do with it but it shouldn't be pulling posts with photos so not sure if it is an image thing? 

I'm also not sure why this didn't happen before because the query was also pulling photo `posts` before and it didn't happen on thor - not just querying `text` posts seems to make this upset for some reason - not sure if that is relevant but trying to think of the differences!

chloealee:
@DoSomething/team-bleed I tried to run this again on thor with `--logfreq=1000` instead of `logfreq=100` and got another slightly different memory error:

`
Jun 13 11:44:36 rogue-thor app/run.5645:  PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /app/vendor/laravel/framework/src/Illuminate/Support/Str.php on line 102 
Jun 13 11:44:36 rogue-thor app/run.5645:  PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 32768 bytes) in /app/vendor/symfony/debug/Exception/FatalErrorException.php on line 1 
Jun 13 11:44:36 rogue-thor heroku/run.5645:  State changed from up to complete 
Jun 13 11:44:36 rogue-thor heroku/run.5645:  Process exited with status 255 
`

Seems like it is always a memory error but happens in different parts of the app?

chloealee:
Does this make sense to edit `/etc/php5/fpm/php.ini` as they suggest in [this post](https://stackoverflow.com/questions/34864524/allowed-memory-size-of-536870912-bytes-exhausted-in-laravel)?

sbsmith86:
Before we do that, lets do some more logging....do you know which post it fails at?

chloealee:
@sbsmith86 it seems to immediately fail as far as I can tell? 

`
Jun 13 11:44:13 rogue-thor app/api:  Starting process with command `php artisan rogue:bulkreviewposts 8064 pending accepted text --logfreq=1000 --tag=bulk --tag=hide-in-gallery` by user clee@dosomething.org 
Jun 13 11:44:29 rogue-thor heroku/run.5645:  Starting process with command `php artisan rogue:bulkreviewposts 8064 pending accepted text --logfreq=1000 --tag=bulk --tag=hide-in-gallery` 
Jun 13 11:44:29 rogue-thor heroku/run.5645:  State changed from starting to up 
Jun 13 11:44:34 rogue-thor app/run.5645:  [2018-06-13 15:44:33] thor.INFO: rogue:bulkreviewposts: Starting to bulk review posts! [] {"user_id":null,"client_id":null,"request_id":null} 
Jun 13 11:44:36 rogue-thor app/run.5645:  PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /app/vendor/laravel/framework/src/Illuminate/Support/Str.php on line 102 
Jun 13 11:44:36 rogue-thor app/run.5645:  PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 32768 bytes) in /app/vendor/symfony/debug/Exception/FatalErrorException.php on line 1 
Jun 13 11:44:36 rogue-thor heroku/run.5645:  State changed from up to complete 
Jun 13 11:44:36 rogue-thor heroku/run.5645:  Process exited with status 255 
`

Should I add more logging into [the script](https://github.com/DoSomething/rogue/blob/master/app/Console/Commands/BulkReviewPosts.php) to get a better sense where exactly it is failing (e.g. in the query, after the query, etc)?

sbsmith86:
I would definitely try to narrow down the problem more to see where the issue is coming from. 

That being said, if we do need more memory for some reason, you can [do this in the script](https://davidwalsh.name/increase-php-memory-limit-ini_set)

sbsmith86:
Also if you run the same query against thor in sequel pro, does it take a long time?

chloealee:
oh cool, thanks for that @sbsmith86 ! 

That is the thing that is kinda weirding me out - the query is so quick in sequel pro and also isn't a very complicated one!

sbsmith86:
interesting. For debugging purposes, can you limit the query to like 1 or 10 posts and see if we get stuff back and also log the post objects so we can see what it is trying to grab?

PR #734, opened 2018-06-13
logging

chloealee:
#### What's this PR do?
Adds more log messages to debug memory error we're seeing in thor. 

👀 

More of the convo. [here](https://github.com/DoSomething/rogue/pull/728#issuecomment-396412879)! 

#### Relevant tickets
Continuation of debugging from #728 

#### How to deploy

### DIFF
--- a/app/Console/Commands/BulkReviewPosts.php
+++ b/app/Console/Commands/BulkReviewPosts.php
@@ -64,7 +64,7 @@ class BulkReviewPosts extends Command
         Post::where([
             ['campaign_id', $this->argument('campaign')],
             ['status', $this->argument('oldStatus')],
-            ['type', $this->argument('type')]
+            ['type', $this->argument('type')],
         ])->chunk(100, function ($posts) {
             if ($posts->isNotEmpty()) {
                 foreach ($posts as $post) {
--- a/app/Console/Commands/BulkReviewPosts.php
+++ b/app/Console/Commands/BulkReviewPosts.php
@@ -60,12 +60,14 @@ class BulkReviewPosts extends Command
         $logfreq = $this->option('logfreq');
         $log = $this->option('log');
         $tags = $this->option('tag') ?? null;
-
+        info('rogue:bulkreviewposts: Beginning to query');
         $posts = Post::setEagerLoads([])
             ->where('campaign_id', $this->argument('campaign'))
             ->where('status', $this->argument('oldStatus'))
             ->where('type', $this->argument('type'))
             ->get();
+
+        info('rogue:bulkreviewposts: Posts result from query: ' . $posts);
 
         if ($posts->isNotEmpty()) {
             foreach ($posts as $post) {
--- a/app/Console/Commands/BulkReviewPosts.php
+++ b/app/Console/Commands/BulkReviewPosts.php
@@ -60,34 +60,31 @@ class BulkReviewPosts extends Command
         $logfreq = $this->option('logfreq');
         $log = $this->option('log');
         $tags = $this->option('tag') ?? null;
-        info('rogue:bulkreviewposts: Beginning to query');
-        $posts = Post::setEagerLoads([])
-            ->where('campaign_id', $this->argument('campaign'))
-            ->where('status', $this->argument('oldStatus'))
-            ->where('type', $this->argument('type'))
-            ->limit(10)
-            ->get();
 
-        info('rogue:bulkreviewposts: First post result from query: ' . $posts->first());
+        Post::where([
+            ['campaign_id', $this->argument('campaign')],
+            ['status', $this->argument('oldStatus')],
+            ['type', $this->argument('type')]
+        ])->chunk(100, function ($posts) {
+            if ($posts->isNotEmpty()) {
+                foreach ($posts as $post) {
+                    if ($post->id % $logfreq == 0) {
+                        info('rogue:bulkreviewposts: updating: ' . $post->id);
+                    }
 
-        if ($posts->isNotEmpty()) {
-            foreach ($posts as $post) {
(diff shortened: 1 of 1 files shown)


app/Console/Commands/BulkReviewPosts.php
```diff
@@ -60,13 +60,15 @@ public function handle()
...
         $tags = $this->option('tag') ?? null;
-
+        info('rogue:bulkreviewposts: Beginning to query');
         $posts = Post::setEagerLoads([])
             ->where('campaign_id', $this->argument('campaign'))
             ->where('status', $this->argument('oldStatus'))
             ->where('type', $this->argument('type'))
             ->get();
 
+        info('rogue:bulkreviewposts: Posts result from query: ' . $posts);
```
sbsmith86: let's just log the first result `->first()`

sbsmith86: also can you limit the query to 10 posts (instead of all of them) has any issues?

chloealee: ah yes forgot that! Updated :)

DFurnes:
@chloealee It looks like this query returns quite a lot of results (53,742 rows) on production. It might be a good idea to [chunk the results or use a cursor](https://laravel.com/docs/5.6/eloquent#chunking-results) so we don't need to load all those rows into memory at once.

chloealee:
ohhh that is a great idea @DFurnes ! If we use `chunk` as I added, can we limit the query and dump out things like @sbsmith86 suggested before?

sbsmith86:
@chloealee ooo right right chunking! might be worth seeing if that is all that was needed to deal with the memory issue.

DFurnes:
It might be best to follow @sbsmith86's suggestions first since chunking the query will mess with those things. Once you're confident that we've narrowed it down though, I do think chunking is the right call here.

app/Console/Commands/BulkReviewPosts.php
```diff
@@ -61,30 +61,30 @@ public function handle()
         $log = $this->option('log');
         $tags = $this->option('tag') ?? null;
 
-        $posts = Post::setEagerLoads([])
-            ->where('campaign_id', $this->argument('campaign'))
-            ->where('status', $this->argument('oldStatus'))
-            ->where('type', $this->argument('type'))
-            ->get();
+        Post::where([
+            ['campaign_id', $this->argument('campaign')],
```
DFurnes: This is subjective, but I think the way you had the query before (as chained methods) reads a lil' nicer.

chloealee:
ok! So should I revert back to the commit before with the extra log messages and then try chunking?

app/Console/Commands/BulkReviewPosts.php
chloealee: @DFurnes totally ! I just wasn't sure the syntax for using `chunk` with the chained methods?

DFurnes: Same, just instead of using `->get()` to grab the entire result set, call `->chunk(...)` to chunk it!

chloealee:
@sbsmith86 @DFurnes I changed back to more logging and dumping out the first post and if that works, will update again to use the chunking method (@mirie and I chatted about chunking yesterday, oof I totally forgot to add that!)!

chloealee:
lmk if you meant something different though and not by the above!

chloealee:
@sbsmith86 @DFurnes I think that was it and hopefully using `chunk` should work! I just ran on thor and it was successful with the limit of 10 records! 

```
Jun 13 16:11:45 rogue-thor app/run.4976: [2018-06-13 20:11:44] thor.INFO: rogue:bulkreviewposts: Starting to bulk review posts! [] {"user_id":null,"client_id":null,"request_id":null} 
Jun 13 16:11:45 rogue-thor app/run.4976: [2018-06-13 20:11:44] thor.INFO: rogue:bulkreviewposts: Beginning to query [] {"user_id":null,"client_id":null,"request_id":null} 
Jun 13 16:11:45 rogue-thor app/run.4976: [2018-06-13 20:11:44] thor.INFO: rogue:bulkreviewposts: First post result from query: {"id":467915,"signup_id":9162432,"campaign_id":"8064","northstar_id":"594acff2a0bfad6dd410fbbf","type":"text","action":"default","quantity":null,"url":null,"text":"finding the confidence to stick up for myself","status":"pending","source":"sms","source_details":null,"details":null,"created_at":"2018-05-10 17:09:32","updated_at":"2018-05-10 17:09:32","deleted_at":null,"reactions_count":0} [] {"user_id":null,"client_id":null,"request_id":null} 
Jun 13 16:11:47 rogue-thor heroku/run.4976: State changed from starting to up 
Jun 13 16:11:47 rogue-thor heroku/run.4976: Process exited with status 0 
Jun 13 16:11:47 rogue-thor heroku/run.4976: State changed from up to complete 
Jun 13 16:11:54 rogue-thor app/run.4976: [2018-06-13 20:11:47] thor.INFO: rogue:bulkreviewposts: ALL DONE! [] {"user_id":null,"client_id":null,"request_id":null} 
Jun 13 16:12:16 rogue-thor app/heroku-redis: source=REDIS sample#active-connections=1 sample#load-avg-1m=0.13 sample#load-avg-5m=0.155 sample#load-avg-15m=0.12 sample#read-iops=0 sample#write-iops=0 sample#memory-total=15664360kB sample#memory-free=13104248kB sample#memory-cached=641888kB sample#memory-redis=2306848bytes sample#hit-rate=0.6769 sample#evicted-keys=0 
```

sbsmith86:
Cool! let's try chunking the query

PR #738, opened 2018-06-19
Delete posts

sbsmith86:
#### What's this PR do?

This PR adds a command to our arsenal that will let us delete posts as we normally do via the command line, ex:

`php artisan rogue:deleteposts 33 34 35` 

The command above will soft delete posts 33, 34, and 35 and send the deletions to quasar so the data is in sync. 

👀 

This came from [a bug that we found in chompy](https://github.com/DoSomething/chompy/issues/20) where there was a set of signups that got duplicate posts made. This was limited to one of the imports we did on 5/15/18. We are still not quite sure what cause this, but we know that these duplicate records need to be removed. 

This command will allow me to just delete those duplicate posts by their post id and for that deletion event to be sent to the data team so things stay in sync. 

#### Relevant tickets
Fixes 🐛 in :chompy:

#### How to deploy

### DIFF
--- a/app/Console/Commands/DeletePosts.php
+++ b/app/Console/Commands/DeletePosts.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Rogue\Console\Commands;
+
+use Illuminate\Console\Command;
+use Rogue\Models\Post;
+
+class DeletePosts extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'rogue:deleteposts
+                            {ids* : A space-separated list of post ids to delete}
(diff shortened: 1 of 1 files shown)


app/Console/Commands/DeletePosts.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $ids = $this->argument('ids');
```
katiecrane: @sbsmith86 not strictly necessary especially since this soft-deletes, but might be nice to start with an "Are you sure?" message since there is some deletion involved.

sbsmith86: Ooo a prompt! I can add that.

PR #740, opened 2018-06-20
Members tags

deguzmankevin:
#### What's this PR do?
This PR makes all posts with "good" tags (good quote, good for sponsor, etc) get sent to the #badass-members slack channel.

:eyes:

There was also a "gotcha" where the same post with multiple "good" tags would get sent to Slack multiple times, but this takes care of it

Note: probably not the most optimal solution, had some trouble with the straightforward solutions I thought of (adding a boolean to the post model for whether or not it had already been posted to slack, or caching the post id) so I may come back sometime to implement one of those instead now that I know how to debug :stuck_out_tongue:

#### Relevant tickets
Finishes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/156352613) pivotal card

#### How to deploy


app/Listeners/SendTaggedNotification.php
```diff
@@ -36,7 +36,7 @@ public function handle(PostTagged $event)
             ]);
         }
 
-        if ($tag->tag_slug === 'good-for-storytelling') {
+        if (! ($post->hasGoodTag()) && in_array($tag->tag_slug, ['good-for-storytelling', 'good-submission', 'good-for-sponsor', 'good-quote'])) {
```
chloealee: good catch @deguzmankevin ! I was wondering about this too when we were talking about notifying Slack for other tags as well. I love this solution so it will only notify the channel once!

app/Models/Post.php
```diff
@@ -368,4 +368,20 @@ public function scopeWithHiddenPosts($query)
...
+
+    /**
+     * Return whether or not a post already has a "good" tag.
+     *
+     * @return bool
+     */
+    public function hasGoodTag()
+    {
+        foreach ($this->tagSlugs() as $tagslug) {
+            if (in_array($tagslug, ['good-for-storytelling', 'good-submission', 'good-for-sponsor', 'good-quote'])) {
```
sbsmith86: It might be nice to have a property on the `Post` model that returns the allowed tags so that you don't have to define this array in multiple places

deguzmankevin: ah yeah great idea! I knew it was repetitive but couldn't think of how to not repeat myself

PR #744, opened 2018-06-22
adds source_details column to the signups table

chloealee:
#### What's this PR do?
adds `source_details` column to the signups table

👀 

In Ashes, we grab info from the `source` query string in the url and currently set that to `source`. However, `source` ideally should always be a "trusted" value. Instead, we should save this to `source_details` column but it doesn't exist! It exists in the `posts` table so we should add it to the `signups` to stay consistent and to capture this query string source. 

#### Relevant tickets
Needed in order to merge [this PR](https://github.com/DoSomething/phoenix/pull/7546). 

#### How to deploy


DFurnes:
Should we add this to transformers (everywhere? only v3?), and notify Storm so they can ingest?

chloealee:
@DFurnes that's a great idea ! I added to both v2 and v3 transformers. Do you think I should only add to v3 because we can control permission in v3 but not v2? Not sure how sensitive the data in the `source_details` column is so might be ok to expose in v2?

katiecrane:
I have a question! In the linked Phoenix PR, it still looks like the information is getting sent over as `source`, but I expected it to be `source_details`. Can you explain to me how this gets used?

DFurnes:
> @DFurnes that's a great idea ! I added to both v2 and v3 transformers. Do you think I should only add to v3 because we can control permission in v3 but not v2? Not sure how sensitive the data in the `source_details` column is so might be ok to expose in v2? 

I think it should be fine to add to v2! Let's also add it to our Customer.io & Quasar queue payloads (pending confirmation from Storm that adding that field won't break anything).

chloealee:
@katiecrane ah great catch! You're totally right - I haven't updated the Ashes PR yet! I am going to merge this and also update the Ashes PR to be `source_details` instead and test!

app/Models/Signup.php
```diff
@@ -141,6 +141,7 @@ public function toBlinkPayload()
             'campaign_run_id' => (string) $this->campaign_run_id,
             'quantity' => $quantity,
             'why_participated' => $this->why_participated,
+            'source_details' => $this->source_details,
```
chloealee: hello @DoSomething/storm @sheyd ! We are adding a new column to the `signups` table called `source_details`. This column will record things like where the signup was created. 

For some context, this came out of [this comment](https://github.com/DoSomething/phoenix/pull/7546/files#r197483566), where we are updating the `source_details` from `phoenix-web` to `phoenix-ashes`. 

We'd like for Quasar and Customer.io to ingest this data - but want to make sure this won't break anything. Can you confirm this will be ok?

sheyd:
@chloealee What's the timeline on rolling this out to production?

app/Models/Signup.php
sheyd: This won't break any current data pipelines, but we need to update the consumer on Quasar to take into the account for the field updates and update the schema in Quasar. What's the timeline for rolling this out to production?

chloealee: good to know, thanks @sheyd ! I was going to merge once you gave the thumbs up - should I wait until the updates are made on the Quasar side? If it won't break anything, maybe I can merge this work and then wait until the consumer is updated on the Quasar side to run [this script](https://github.com/DoSomething/rogue/blob/master/app/Console/Commands/UpdateSignupAndOrPostField.php) to update the source to `phoenix-ashes` instead of `phoenix-web`?

sheyd: @chloealee I'm working on the Quasar side updates right now. One catch I just talked through with @DFurnes was if this field also exists on `posts`. And it does, but it's not currently being sent to Blink or Quasar. Can we update these methods: https://github.com/DoSomething/rogue/blob/master/app/Models/Post.php#L179-L247 to send those `source_details` along as well with this PR? I'll include those updates on Quasar side as well, and we can "pull the trigger" around the same time :)

chloealee: @sheyd ah yes! Thanks for flagging that - just updated to also send a post's `source_details` to Customer.io and Quasar as well!

chloealee:
cool thanks for the speedy review @DFurnes ! Is it ok to merge this before @sheyd 's changes?

DFurnes:
Yes, I believe so!

PR #745, opened 2018-06-27
Saves source_details to v2 posts and v2 & v3 signups endpoints

chloealee:
#### What's this PR do?
Saves source_details when creating a post via v2 posts endpoint and when creating a signup via v2 & v3 signups endpoints.

Also, updates `source_details` column in the `signups` table to be type `VARCHAR` instead of `text` and `nullable` to mimic the `source_details` in the `posts` table for consistency. 

**Question** - this migration is just on thor and my local so I thought instead of creating a new migration, it'd be better to just update this one. On my local, I deleted the migration in the `migrations` table and the `source_details` column and re-ran the migration. Is it ok to do this for thor too the same way? 

👀 

Related to this ticket: #744. This PR actually saves this info. into these new columns! 

#### Relevant tickets
This is needed in order for [this PR](https://github.com/DoSomething/phoenix/pull/7546) to work in Ashes.

#### How to deploy

### DIFF
--- a/database/migrations/2018_06_22_160305_add_source_details_column_to_signup_table.php
+++ b/database/migrations/2018_06_22_160305_add_source_details_column_to_signup_table.php
@@ -14,7 +14,7 @@ class AddSourceDetailsColumnToSignupTable extends Migration
     public function up()
     {
         Schema::table('signups', function (Blueprint $table) {
-            $table->text('source_details')->after('source')->comment('Extra details about the signup source, like referral source from url query string');
+            $table->string('source_details')->after('source')->nullable()->comment('Extra details about the signup source, like referral source from url query string');
         });
     }
 
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -30,7 +30,7 @@ class Signup extends Model
      *
      * @var array
      */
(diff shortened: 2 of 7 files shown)


database/migrations/2018_06_22_160305_add_source_details_column_to_signup_table.php
```diff
@@ -14,7 +14,7 @@ class AddSourceDetailsColumnToSignupTable extends Migration
     public function up()
     {
         Schema::table('signups', function (Blueprint $table) {
-            $table->text('source_details')->after('source')->comment('Extra details about the signup source, like referral source from url query string');
+            $table->string('source_details')->after('source')->nullable()->comment('Extra details about the signup source, like referral source from url query string');
```
DFurnes: This needs to be a new migration if you want to make a change to this column! Once a migration has run, it doesn't get run again (unless you run `php artisan migrate:rollback` to undo it, like when testing on your local).

chloealee: @DFurnes ah ok! So make this a new migration instead of just editing here? 

Not sure if you saw the above but here's what I was thinking:
>  this migration is just on thor and my local so I thought instead of creating a new migration, it'd be better to just update this one. On my local, I deleted the migration in the migrations table and the source_details column and re-ran the migration. Is it ok to do this for thor too the same way?

chloealee: ok sounds good to me, thanks @DFurnes !

PR #746, opened 2018-06-27
Hide Voter Reg Posts in Rogue 

chloealee:
#### What's this PR do?
- Adds [include parameters](https://fractal.thephpleague.com/transformers/#include-parameters) to add ability to filter posts by type. 
- Only returns `photo` and `text` posts in Rogue admin. 
- Updates documentation. 
- Adds unit tests.

- All tests should pass! 
- Voter reg posts should not appear in Rogue admin.

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/158011726) Pivotal Card. 

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Legacy/Web/SignupsController.php
+++ b/app/Http/Controllers/Legacy/Web/SignupsController.php
@@ -55,7 +55,7 @@ class SignupsController extends Controller
                 'signup_id' => $signup->id,
                 // @TODO - We could probably grab campaign and user info from API as well.
                 'campaign' => $campaign,
-                'user' => $user->toArray(),
+                // 'user' => $user->toArray(),
             ]);
     }
 }
--- a/app/Http/Controllers/Legacy/Web/SignupsController.php
+++ b/app/Http/Controllers/Legacy/Web/SignupsController.php
@@ -55,7 +55,7 @@ class SignupsController extends Controller
                 'signup_id' => $signup->id,
                 // @TODO - We could probably grab campaign and user info from API as well.
                 'campaign' => $campaign,
-                // 'user' => $user->toArray(),
+                'user' => $user->toArray(),
             ]);
     }
 }
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -80,6 +80,9 @@ class PostsController extends ApiController
         // Only return posts tagged "Hide In Gallery" if staff user or if is owner of the post.
         $query = $query->withHiddenPosts();
 
+        // Do not return any voter-reg posts.
+        $query = $query->where('type', '!=', 'voter-reg');
+
         // If tag param is passed, only return posts that have that tag.
         if (array_has($filters, 'tag')) {
             $query = $query->withTag($filters['tag']);
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -80,9 +80,6 @@ class PostsController extends ApiController
         // Only return posts tagged "Hide In Gallery" if staff user or if is owner of the post.
         $query = $query->withHiddenPosts();
 
-        // Do not return any voter-reg posts.
-        $query = $query->where('type', '!=', 'voter-reg');
-
         // If tag param is passed, only return posts that have that tag.
         if (array_has($filters, 'tag')) {
             $query = $query->withTag($filters['tag']);
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -6,6 +6,7 @@ use Rogue\Models\Signup;
 use League\Fractal\ParamBag;
 use Rogue\Services\Registrar;
 use League\Fractal\TransformerAbstract;
+use Illuminate\Validation\ValidationException;
 
 class SignupTransformer extends TransformerAbstract
 {
@@ -62,28 +63,22 @@ class SignupTransformer extends TransformerAbstract
      */
     public function includePosts(Signup $signup, ParamBag $params = null)
     {
-        if ($params === null) {
-            // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
-            $post = $signup->visiblePosts;
+        // Validate `type` parameter, if provided:
+        if ($params) {
+            $usedParams = array_keys(iterator_to_array($params));
 
-            return $this->collection($post, new PostTransformer);
+            if ($invalidParams = array_diff($usedParams, $this->validParams)) {
+                throw ValidationException::withMessages([
+                    'include' => 'Parameters must be one of: ' . implode(',', $this->validParams),
+                ]);
+            }
         }
 
-        // Optional params validation
-        $usedParams = array_keys(iterator_to_array($params));
-
-        if ($invalidParams = array_diff($usedParams, $this->validParams)) {
-            throw new \Exception(sprintf(
-                'Invalid param(s): "%s". Valid param(s): "%s"',
-                implode(',', $usedParams),
-                implode(',', $this->validParams)
-            ));
-        }
+        $type = $params ? $params->get('type') : null;
 
         // Only allow an admin or the user who owns the signup to see the signup's unapproved posts by type.
-        $post = $signup->visiblePosts($params->get('type'));
-
-        return $this->collection($post, new PostTransformer);
+        // $post = $signup->visiblePosts($params->get('type'));
+        return $this->collection($signup->visiblePosts($type), new PostTransformer);
     }
 
     /**
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -3,6 +3,7 @@
 namespace Rogue\Http\Transformers;
 
 use Rogue\Models\Signup;
+use League\Fractal\ParamBag;
 use Rogue\Services\Registrar;
 use League\Fractal\TransformerAbstract;
 
@@ -16,6 +17,13 @@ class SignupTransformer extends TransformerAbstract
     protected $availableIncludes = [
         'posts', 'user', 'accepted_quantity',
     ];
+
+    /**
+     * List of params possible to include
+     *
+     * @var array
+     */
+    protected $validParams = ['type'];
 
     /**
      * Transform resource data.
@@ -49,12 +57,33 @@ class SignupTransformer extends TransformerAbstract
      * Include posts
      *
      * @param \Rogue\Models\Signup $signup
+     * @param \League\Fractal\ParamBag|null
      * @return \League\Fractal\Resource\Collection
      */
-    public function includePosts(Signup $signup)
+    public function includePosts(Signup $signup, ParamBag $params = null)
     {
-        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
-        $post = $signup->visiblePosts;
+        if ($params === null) {
+            // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
+            $post = $signup->visiblePosts($params['type'][0]);
+
+            return $this->collection($post, new PostTransformer);
+        }
+
+        // Optional params validation
+        $usedParams = array_keys(iterator_to_array($params));
+
+        if ($invalidParams = array_diff($usedParams, $this->validParams)) {
+            throw new \Exception(sprintf(
+                'Invalid param(s): "%s". Valid param(s): "%s"',
+                implode(',', $usedParams),
+                implode(',', $this->validParams)
+            ));
+        }
+
+        $type = $params->get('type')[0];
+
+        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts by type.
+        $post = $signup->visiblePosts($type);
 
         return $this->collection($post, new PostTransformer);
     }
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -72,13 +72,11 @@ class SignupTransformer extends TransformerAbstract
                     'include' => 'Parameters must be one of: ' . implode(',', $this->validParams),
                 ]);
             }
+
+            return $this->collection($signup->visiblePosts($params->get('type')), new PostTransformer);
+        } else {
+            return $this->collection($signup->visiblePosts, new PostTransformer);
         }
-
-        $type = $params ? $params->get('type') : null;
-
-        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts by type.
-        // $post = $signup->visiblePosts($params->get('type'));
-        return $this->collection($signup->visiblePosts($type), new PostTransformer);
     }
 
     /**
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -3,10 +3,8 @@
 namespace Rogue\Http\Transformers;
 
 use Rogue\Models\Signup;
-// use League\Fractal\ParamBag;
 use Rogue\Services\Registrar;
 use League\Fractal\TransformerAbstract;
-// use Illuminate\Validation\ValidationException;
 
 class SignupTransformer extends TransformerAbstract
 {
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -3,10 +3,10 @@
 namespace Rogue\Http\Transformers;
 
 use Rogue\Models\Signup;
-use League\Fractal\ParamBag;
+// use League\Fractal\ParamBag;
 use Rogue\Services\Registrar;
 use League\Fractal\TransformerAbstract;
-use Illuminate\Validation\ValidationException;
+// use Illuminate\Validation\ValidationException;
 
 class SignupTransformer extends TransformerAbstract
 {
@@ -18,13 +18,6 @@ class SignupTransformer extends TransformerAbstract
     protected $availableIncludes = [
         'posts', 'user', 'accepted_quantity',
     ];
-
-    /**
-     * List of params possible to include
-     *
-     * @var array
-     */
-    protected $validParams = ['type'];
 
     /**
      * Transform resource data.
@@ -58,25 +51,14 @@ class SignupTransformer extends TransformerAbstract
      * Include posts
      *
      * @param \Rogue\Models\Signup $signup
-     * @param \League\Fractal\ParamBag|null
      * @return \League\Fractal\Resource\Collection
      */
-    public function includePosts(Signup $signup, ParamBag $params = null)
+    public function includePosts(Signup $signup)
     {
-        // Validate `type` parameter, if provided:
-        if ($params) {
-            $usedParams = array_keys(iterator_to_array($params));
+        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
+        $post = $signup->visiblePosts;
 
-            if ($invalidParams = array_diff($usedParams, $this->validParams)) {
-                throw ValidationException::withMessages([
-                    'include' => 'Parameters must be one of: ' . implode(',', $this->validParams),
-                ]);
-            }
-        }
-
-        $type = $params ? $params->get('type') : null;
-
-        return $this->collection($signup->visiblePosts($type), new PostTransformer);
+        return $this->collection($post, new PostTransformer);
     }
 
     /**
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -72,11 +72,11 @@ class SignupTransformer extends TransformerAbstract
                     'include' => 'Parameters must be one of: ' . implode(',', $this->validParams),
                 ]);
             }
-
-            return $this->collection($signup->visiblePosts($params->get('type')), new PostTransformer);
-        } else {
-            return $this->collection($signup->visiblePosts, new PostTransformer);
         }
+
+        $type = $params ? $params->get('type') : null;
+
+        return $this->collection($signup->visiblePosts($type), new PostTransformer);
     }
 
     /**
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -64,7 +64,7 @@ class SignupTransformer extends TransformerAbstract
     {
         if ($params === null) {
             // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
-            $post = $signup->visiblePosts($params['type'][0]);
+            $post = $signup->visiblePosts;
 
             return $this->collection($post, new PostTransformer);
         }
@@ -80,10 +80,8 @@ class SignupTransformer extends TransformerAbstract
             ));
         }
 
-        $type = $params->get('type')[0];
-
         // Only allow an admin or the user who owns the signup to see the signup's unapproved posts by type.
-        $post = $signup->visiblePosts($type);
+        $post = $signup->visiblePosts($params->get('type'));
 
         return $this->collection($post, new PostTransformer);
     }
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -71,13 +71,16 @@ class Signup extends Model
      */
     public function visiblePosts()
     {
+        $query = $this->hasMany(Post::class);
+
         if (! is_staff_user()) {
-            return $this->hasMany(Post::class)->where('status', '=', 'accepted')
-                                              ->orWhere('northstar_id', auth()->id())
-                                              ->with('tags');
+            $query->where(function($query) {
+                $query->where('status', 'accepted')
+                    ->orWhere('northstar_id', auth()->id());
+            });
         }
 
-        return $this->hasMany(Post::class)->with('tags');
+        return $query;
     }
 
     /**
@@ -221,21 +224,16 @@ class Signup extends Model
     public function scopeWithVisiblePosts($query, $types = null)
     {
         return $query->with(['visiblePosts' => function ($query) use ($types) {
-
-            if ($types) {
-                // $query->whereIn('type', $types);
-
-                if (count($types) >1 ) {
-                    $query->where(function ($query) use ($types) {
-                        foreach ($types as $type) {
-                            $query->orWhere('type', $type);
-                        }
-                    });
-                } else {
-                    $query->where('type', $types[0]);
-                }
+            if (! is_staff_user()) {
+                $query->where(function($query) {
+                    $query->where('status', 'accepted')
+                        ->orWhere('northstar_id', auth()->id());
+                });
             }
 
+            if ($types) {
+                $query->whereIn('type', $types);
+            }
         }]);
     }
 }
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -80,19 +80,20 @@ class Signup extends Model
             }
 
             if (count($type) > 1) {
-                return $this->hasMany(Post::class)->where('status', '=', 'accepted')
+                return $this->hasMany(Post::class)->whereIn('type', array_values($type))
+                                                  ->where('status', '=', 'accepted')
                                                   ->orWhere('northstar_id', auth()->id())
-                                                  ->whereIn('type', array_values($type))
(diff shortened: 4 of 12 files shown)


app/Http/Controllers/Legacy/Web/SignupsController.php
```diff
@@ -55,7 +55,7 @@ public function show($id)
                 'signup_id' => $signup->id,
                 // @TODO - We could probably grab campaign and user info from API as well.
                 'campaign' => $campaign,
-                'user' => $user->toArray(),
+                // 'user' => $user->toArray(),
```
chloealee: this will be commented back in, just needed to comment this out on my local bc I have some weird settings!

resources/assets/components/Signup/index.js
```diff
@@ -486,7 +486,7 @@ class Signup extends React.Component {
               title="Meta"
               details={{
                 'Signup ID': signup.id,
-                'Northstar ID': user.id,
+                // 'Northstar ID': user.id,
```
chloealee: same with this will be commented in!

app/Http/Controllers/PostsController.php
```diff
@@ -80,6 +80,9 @@ public function index(Request $request)
         // Only return posts tagged "Hide In Gallery" if staff user or if is owner of the post.
         $query = $query->withHiddenPosts();
 
+        // Do not return any voter-reg posts.
+        $query = $query->where('type', '!=', 'voter-reg');
```
chloealee: @DoSomething/team-bleed we probably want to hide this behind a query param, correct? 

I'm thinking that we want the API to expose posts that have the `type` `voter-reg` but Rogue will use the query param to hide these posts in Rogue Admin?

chloealee:
@DoSomething/team-bleed I'm getting a little stumped on how to update the `/v3/signups/{signup}?include=posts` URL to hide posts that have the `type` `voter-reg`. The `?include=posts` part comes from [Fractal](https://github.com/thephpleague/fractal/blob/master/src/Manager.php#L163) and as far as I can tell, we'll need to update the code in Fractal. Does anyone have any other thoughts on how to get around this?

DFurnes:
@chloealee The post relationship should be loaded [here](https://github.com/DoSomething/rogue/blob/33863aea42873a50518f77fc5cbbdae31314b117/app/Http/Transformers/SignupTransformer.php#L48-L60), so you can either adjust that `visiblePosts` relationship or make a new relationship that excludes "special" types of posts.

chloealee:
@DFurnes ah that makes a lot more sense! Is there anyway to tell from OAuth if the request is coming internally from the Rogue app? Trying to think about how to add this logic in since I think we still want to expose `voter-reg` posts to admins via the API?

DFurnes:
I think it'd probably be cleanest to handle this in the request, rather than by inspecting who made it. It looks like Fractal has [support for include parameters](https://fractal.thephpleague.com/transformers/#include-parameters), like `?include=posts(photo,text)`, which looks like it could be a very clean solution here. Another hackier solution would be to create a separate `?include=reviewablePosts` that excludes things that shouldn't show up in Rogue.

app/Models/Signup.php
```diff
@@ -65,21 +65,41 @@ public function posts()
...
 
-        return $this->hasMany(Post::class)->with('tags');
+        if ($type === null) {
+            return $this->hasMany(Post::class)->with('tags');
+        }
+
+
+        if (count($type) > 1) {
+            return $this->hasMany(Post::class)->whereIn('type', array_values($type))
+                                          ->with('tags');
```
chloealee: @DFurnes @katiecrane still need to clean up code but have a quick q! 

I am trying to incorporate @DFurnes 's [great suggestion](https://github.com/DoSomething/rogue/pull/746#issuecomment-401134852) and fidgeting with getting something like this `api/v3/signups?include=posts:type(text|photo)` to work to exclude voter reg posts. Instead of iterating through the params, (in this case, text and photo), I thought that I could use [`whereIn`](https://laravel.com/docs/5.5/queries) instead but am having a hard time getting it to work. 

My question - _should_ `whereIn` be able to work in this case? If so, I'll fidget around more but if not I'll go towards the route of looping through those types !

katiecrane: @chloealee I would have expected it to work but that doesn't mean it should haha, looks like in other places we do kind of the same thing with `id` instead of `type`. What does this return that's wrong?

DFurnes: Yeah interesting, I would have thought this would work too!

DFurnes: It'd be interesting to see what SQL query this is getting compiled to, from [`/__clockwork`](http://rogue.test/__clockwork)

chloealee: @katiecrane it just doesn't return any posts when there are definitely associated `photo` and `text` posts! 

@DFurnes oh that is so neat I've never used that before - thank you! Those queries look very strange to me! Looks like it isn't factoring in `type` at all if I'm reading correctly?
screen shot 2018-07-05 at 5 07 42 pm

chloealee: ooo I think I got this to work - by adding the `->get();` at the end!

app/Models/Signup.php
```diff
@@ -65,21 +65,62 @@ public function posts()
...
+                                                  ->with('tags')
+                                                  ->get();
+                // return $this->hasMany(Post::class)->where(function ($q) use ($type) {
+                //     $q->whereIn('type', array_values($type))
+                //       ->where('status', '=', 'accepted')
+                //       ->orWhere('northstar_id', auth()->id())
+                //       ->with('tags')
+                //       ->get();
+                // });
+
```
chloealee: Everything is working except when using include params as an owner (e.g. `api/v3/signups?include=posts:type(text|photo)` still returns `voter-reg` posts. I tried to do what is commented out but didn't return any posts. I'm wondering if there is a certain order I'd have to use `whereIn`, `where`, and `orWhere` when using them all together?

app/Models/Signup.php
```diff
@@ -65,21 +65,66 @@ public function posts()
...
+                                                  ->orWhere('northstar_id', auth()->id())
+                                                  ->with('tags');
+            }
+
+            return $this->hasMany(Post::class)->whereIn('type', array_values($type))
+                                              ->where('status', '=', 'accepted')
                                               ->orWhere('northstar_id', auth()->id())
-                                              ->with('tags');
+                                              ->with('tags')
+                                              ->get();
```
chloealee: Everything is working except when using include params as an owner (e.g. api/v3/signups?include=posts:type(text|photo) still returns voter-reg posts. I tried to do what is commented out below but none returned the correct response either. I'm wondering if there is a certain order or way I'd have to use whereIn, where, and orWhere when using them all together? @DoSomething/team-bleed

app/Models/Signup.php
DFurnes: @chloealee What query is this making? I'm thinking you might need to use [parameter grouping](https://laravel.com/docs/5.6/queries#parameter-grouping) so that these conditions get applied with parenthesis in the right place...

chloealee: @DFurnes I tried to do that (see things I've tried commented out!) but didn't work. Maybe I'm doing it incorrectly?

app/Models/Signup.php
```diff
@@ -65,21 +65,66 @@ public function posts()
...
+            // });
+
+            // return $this->hasMany(Post::class)->whereIn('type', array_values($type))
+            //                                   ->where(function($q) {
+            //                                     $q->where('status', '=', 'accepted')
+            //                                       ->orWhere('northstar_id', auth()->id());
+            //                                   })
+            //                                   ->with('tags')
+            //                                   ->get();
+
```
chloealee: @DFurnes here are some examples of how I tried to do the parameter grouping! Still didn't work though and returned voter-reg posts when hitting `api/v3/signups?include=posts:type(text|photo)` :(

chloealee: omg I am an idiot - when testing I must have made a mistake - this actually _does_ work!!

chloealee:
@DFurnes I don't _think_ we need to change any of the settings in wercker since it doesn't have the `NORTHSTAR_URL` setting but could something there have changed that are making these tests fail? Everything is passing locally!

DFurnes:
Yeah, that's strange. Could you try pushing up a `$response->dump()` in by one of those failures and push it up to see if we can find out what those 500s are saying?

chloealee:
that's a great idea! I'll do that now, @DFurnes !

chloealee:
@DFurnes interesting - a [type error](https://app.wercker.com/dosomething/rogue/runs/build/5b43c8a40bf2c200015c88c6?step=5b43c8de63c37f0001a2e79f)?

DFurnes:
We really need to stop these tests from hitting Northstar! Given where that error is coming from, I'm thinking it's either a wrong Redirect URL (and the response is one of those `Client authentication failed` errors) or we're still pointing at https://northstar-thor.dosomething.org.

chloealee:
@DFurnes ah yes - that is a [ticket](https://www.pivotaltracker.com/n/projects/2019429/stories/158083551) that I am supposed to get to in this sprint but haven't gotten to yet!

chloealee:
You're right - `NORTHSTAR_URL` in dosomething-rogue-dev wasn't updated though (I was looking at the wrong environment before) so just did that !

chloealee:
@DFurnes sorry to keep bothering you on this! I've updated the `NORTHSTAR_URL`s in heroku to all be pointing at the new identity urls - is that the correct place to be updating this? The tests still seem to be failing :-/ For `dosomething-rogue-dev` I put `https://identity-qa.dosomething.org` because `identity-dev.dosomething.org` wasn't working but still doesn't seem to get the tests to work. These just started failing yesterday :-/

DFurnes:
Tests run in Wercker, so they'd be reading from the [environment variables](https://app.wercker.com/dosomething/rogue/environment) there. Heroku hosts our websites for dev, QA, and production so the environment variables there affect each of those specific applications.

Let me dig into this a bit and get back to you!

chloealee:
ah ok, thanks @DFurnes ! I just made sure that the `NORTHSTAR_AUTH_SECRET` and `NORTHSTAR_CLIENT_SECRET` were correct for the wercker env variables corresponding to https://identity-qa.dosomething.org. Doesn't look like there is a place for `NORTHSTAR_URL` though. Thanks!!

DFurnes:
Ah, yep [we default `NORTHSTAR_URL` to https://identity-dev.dosomething.org](https://github.com/DoSomething/rogue/blob/6a5c41fead1a79c15dd2f4801e38ef5b40e3fb79/.env.example#L52) so that'd probably explain why this is not working! I set `NORTHSTAR_URL` on Wercker and re-triggered a build. Let's see if that helps.

DFurnes:
Still some test failures but looks like the Northstar 500s [cleared up](https://app.wercker.com/dosomething/rogue/runs/build/5b44c7a40bf2c200015cece1?step=5b44c7ddde6bc30001abf96f)!

```
PHPUnit 6.5.8 by Sebastian Bergmann and contributors.

........................................I......................  63 / 129 ( 48%)
..............I.....I.................FF......................I 126 / 129 ( 97%)
.I.                                                             129 / 129 (100%)

Time: 12.41 seconds, Memory: 36.00MB

There were 2 failures:

1) Tests\Http\SignupTest::testSignupIndexWithIncludedUserAsAdmin
Failed asserting that true matches expected false.

/pipeline/source/tests/Http/SignupTest.php:502

2) Tests\Http\SignupTest::testSignupIndexWithIncludedUserAsOwner
Failed asserting that true matches expected false.

/pipeline/source/tests/Http/SignupTest.php:522

FAILURES!
Tests: 129, Assertions: 1951, Failures: 2, Incomplete: 5.
```

chloealee:
ah bingo - thanks so much @DFurnes !!

DFurnes:
> Changed wercker env back to https://identity-dev.dosomething.org with corresponding secrets because the tests were failing since those users didn't exist in identity-qa. Hope that is ok! 

Yes, that sounds perfect!

app/Http/Transformers/SignupTransformer.php
```diff
@@ -48,12 +56,31 @@ public function transform(Signup $signup)
...
+            $post = $signup->visiblePosts;
+
+            return $this->collection($post, new PostTransformer);
+        }
+
+        // Optional params validation
+        $usedParams = array_keys(iterator_to_array($params));
+
+        if ($invalidParams = array_diff($usedParams, $this->validParams)) {
+            throw new \Exception(sprintf(
```
DFurnes: This should probably be a `ValidationException` so it gets formatted as a 422:

```php
use Illuminate\Validation\ValidationException;

// ...

throw ValidationException::withMessages([
    'include' => 'Parameters must be one of: ' . implode(',', $this->validParams),
]);
```

app/Http/Transformers/SignupTransformer.php
```diff
@@ -48,12 +56,31 @@ public function transform(Signup $signup)
...
+                'Invalid param(s): "%s". Valid param(s): "%s"',
+                implode(',', $usedParams),
+                implode(',', $this->validParams)
+            ));
+        }
+
+        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts by type.
+        $post = $signup->visiblePosts($params->get('type'));
```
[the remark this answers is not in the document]
DFurnes: I'd also like to double-check that we're okay calling `visiblePosts` as a method here, since usually relationships are accessed as a property. Let's make sure we're not accidentally adding additional SQL queries!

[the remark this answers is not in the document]
chloealee: ah yes so much cleaner!

app/Http/Transformers/SignupTransformer.php
chloealee: ah great call - updated! :)

app/Http/Transformers/SignupTransformer.php
```diff
@@ -49,14 +58,25 @@ public function transform(Signup $signup)
...
+            if ($invalidParams = array_diff($usedParams, $this->validParams)) {
+                throw ValidationException::withMessages([
+                    'include' => 'Parameters must be one of: ' . implode(',', $this->validParams),
+                ]);
+            }
 
-        return $this->collection($post, new PostTransformer);
+            return $this->collection($signup->visiblePosts($params->get('type')), new PostTransformer);
+        } else {
+            return $this->collection($signup->visiblePosts, new PostTransformer);
```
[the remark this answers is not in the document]
DFurnes: Hmm, what error do you get when you pass `null` to that function? Also were you able to see if there was any difference in what SQL queries are made when you use the relationship property `$signup->visiblePosts` vs. the function `$signup->visiblePosts()`?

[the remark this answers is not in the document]
chloealee: @DFurnes when I ass `null` to the function, no posts are returned for some reason 🤔 
> Also were you able to see if there was any difference in what SQL queries are made when you use the relationship property $signup->visiblePosts vs. the function $signup->visiblePosts()?
That is interesting! Not quite sure how to see this? However, if I `dd($this->collection($signup->visiblePosts, new PostTransformer));`
I get: 
`League\Fractal\Resource\Collection {#1938
  #data: Illuminate\Database\Eloquent\Collection {#1958
    #items: array:1 [
      0 => Rogue\Models\Post {#1902
...`
`

but with `dd($this->collection($signup->visiblePosts(null), new PostTransformer));` I can't scroll all the way up but here's the beginning of what I can see:
`
                  #verbosity: 32
                  #verbosityMap: array:5 [ …5]
                  -application: Illuminate\Console\Application {#52 …18}
                  -processTitle: null
                  -aliases: []
                  -definition: Symfony\Component\Console\Input\InputDefinition {#691 …6}
                  -help: null
                  -ignoreValidationErrors: false
                  -applicationDefinitionMerged: false
                  -applicationDefinitionMergedWithArgs: false
                  -code: null
                  -synopsis: []
                  -usages: []
                  -helperSet: Symfony\Component\Console\Helper\HelperSet {#507 …2}
                  -name: "queue:failed"
                  -hidden: false
                  -description: "List all of the failed queue jobs"
                }
                "command.queue.flush" => Illuminate\Queue\Console\FlushFailedCommand {#692
                  #name: "queue:flush"
                  #description: "Flush all of the failed queue jobs"
                  #laravel: Illuminate\Foundation\Application {#565}
                  #input: null
                  #output: null
                  #signature: null
                  #hidden: false
                  #verbosity: 32
                  #verbosityMap: array:5 [ …5]
                  -application: Illuminate\Console\Application {#52 …18}
                  -processTitle: null
                  -aliases: []
                  -definition: Symfony\Component\Console\Input\InputDefinition {#693 …6}
                  -help: null
                  -ignoreValidationErrors: false
                  -applicationDefinitionMerged: false
                  -applicationDefinitionMergedWithArgs: false
                  -code: null
                  -synopsis: []
                  -usages: []
                  -helperSet: Symfony\Component\Console\Helper\HelperSet {#507 …2}
                  -name: "queue:flush"
`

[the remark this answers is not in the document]
chloealee: @DFurnes oof I figured it out! I was missing the `->get();` for the queries `if ($type === null)` 🤦‍♀️ !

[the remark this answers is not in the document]
DFurnes: Be careful calling `->get()`! That tells Eloquent to run the query then and there, so by using it inside the relationship we're accidentally making a bunch of extra queries.

To see what database queries are being made, I'd recommend using Clockwork (accessible at [`/__clockwork`](http://rogue.test/__clockwork)). Just keep that open in a tab while you make requests on your local Postman, and then you can go back and inspect what happened each time:

screen shot 2018-07-19 at 5 05 38 pm

I think we can still do a little work here - the plain `v3/signups` index makes one SQL query, but once we add the `?include=posts` query parameter it jumps to 61!! 😱 Instead of scoping the query on each individual signup here, let's use the [`scopeVisiblePosts`](https://github.com/DoSomething/rogue/blob/0446b280a57bc8a2e92f6730cbf86d09aa303b21/app/Models/Signup.php#L215-L228) scope in the controller to [eager-load](https://laravel.com/docs/5.6/eloquent-relationships#eager-loading) these before they hit the transformer.

This is a bit messy, but something to work off of - down to 4 SQL queries!

```diff
diff --git a/app/Http/Controllers/SignupsController.php b/app/Http/Controllers/SignupsController.php
index 693eaa9c..1da966fa 100644
--- a/app/Http/Controllers/SignupsController.php
+++ b/app/Http/Controllers/SignupsController.php
@@ -91,8 +91,13 @@ class SignupsController extends ApiController
         $query = $this->filter($query, $filters, Signup::$indexes);
 
         // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
-        if ($request->query('include') === 'posts') {
-            $query = $query->withVisiblePosts();
+        if (starts_with($request->query('include'), 'posts')) {
+            $types = (new \League\Fractal\Manager)
+                ->parseIncludes($request->query('include'))
+                ->getIncludeParams('posts')
+                ->get('type');
+
+            $query = $query->withVisiblePosts($types);
         }
 
         $orderBy = $request->query('orderBy');
diff --git a/app/Http/Transformers/SignupTransformer.php b/app/Http/Transformers/SignupTransformer.php
index a27c6437..72e3621d 100644
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -74,9 +74,7 @@ class SignupTransformer extends TransformerAbstract
             }
         }
 
-        $type = $params ? $params->get('type') : null;
-
-        return $this->collection($signup->visiblePosts($type), new PostTransformer);
+        return $this->collection($signup->posts, new PostTransformer);
     }
 
     /**
@@ -235,13 +241,15 @@ class Signup extends Model
      *
      * @return \Illuminate\Database\Eloquent\Builder
      */
-    public function scopeWithVisiblePosts($query)
+    public function scopeWithVisiblePosts($query, $type)
     {
-        if (! is_staff_user()) {
-            return $query->with(['posts' => function ($query) {
-                 $query->where('status', 'accepted')
-                ->orWhere('northstar_id', auth()->id());
-            }]);
-        }
+        return $query->with(['posts' => function ($query) use ($type) {
+            $query->whereIn('type', $type);
+
+            if (! is_staff_user()) {
+               $query->where('status', 'accepted')
+                    ->orWhere('northstar_id', auth()->id());
+            }
+        }, 'posts.tags']);
     }
 }
```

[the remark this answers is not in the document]
chloealee: @DFurnes ahh that is so interesting, thank you for sharing! You mentioned clockwork to me before but I forgot about it so thank you for reminding me! Very helpful in terms of performance :) 

Also, didn't realize that about using `->get()` - I've always wondered! It is interesting though because when I did the same exact code in the previous commit before the most recent, without `->get()`, I couldn't get the results that I wanted. 

I went about the way that you suggested above (so smart!) and it seems that I am running into the same problem - the query isn't working and isn't actually sorting by type (you can see this with some of the failed tests). Do you have any ideas about this?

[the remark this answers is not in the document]
DFurnes: Try swapping to use the plain `$signup->posts` relation in the transformer... that seems to get things properly filtering on my local. It seems like using `$signup->visiblePosts` still "forces" Laravel to re-fetch the relation.

```diff
diff --git a/app/Http/Transformers/SignupTransformer.php b/app/Http/Transformers/SignupTransformer.php
index 9569f92e..31919adb 100644
--- a/app/Http/Transformers/SignupTransformer.php
+++ b/app/Http/Transformers/SignupTransformer.php
@@ -55,10 +55,7 @@ class SignupTransformer extends TransformerAbstract
      */
     public function includePosts(Signup $signup)
     {
-        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
-        $post = $signup->visiblePosts;
-
-        return $this->collection($post, new PostTransformer);
+        return $this->collection($signup->posts, new PostTransformer);
     }
```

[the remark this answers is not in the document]
DFurnes: Actually wait - better yet, let's change how we eager-load in `scopeWithVisiblePosts`! That should let us keep using `$signup->visiblePosts` like we want to in `SignupTransformer`.

```diff
diff --git a/app/Models/Signup.php b/app/Models/Signup.php
index 82b20baf..662b64cc 100644
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -219,7 +219,7 @@ class Signup extends Model
      */
     public function scopeWithVisiblePosts($query, $types = null)
     {
-        return $query->with(['posts' => function ($query) use ($types) {
+        return $query->with(['visiblePosts' => function ($query) use ($types) {

             if ($types) {
                 $query->whereIn('type', $types);
```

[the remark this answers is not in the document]
chloealee: @DFurnes updated to that and still not passing / working :(

app/Models/Signup.php
```diff
@@ -213,17 +212,23 @@ public function getAcceptedQuantity()
...
+                $query->whereIn('type', $types);
+            }
+
+            if (! is_staff_user()) {
+               $query->where('status', 'accepted')
+                    ->orWhere('northstar_id', auth()->id());
+            }
+        }, 'posts.tags']);
     }
 }
```
chloealee: @DFurnes I'm wondering if this isn't working because this also has a query in `visiblePosts` and it is overriding something? Do we need to have `scopeWithVisiblePosts` and `visiblePosts`? Seems like to me they're doing very similar things and might be too much to have both?

chloealee: oh wait, I remember now why! You explained this before!
> Instead of scoping the query on each individual signup here, let's use the scopeVisiblePosts scope in the controller to eager-load these before they hit the transformer.

I _think_ the queries are still somehow cancelling each other out though? I tried to removed 
```
if (! is_staff_user()) {
  $query->where('status', 'accepted')
               ->orWhere('northstar_id', auth()->id());
}
```

out of `scopeWithVisiblePosts` since it should already do that in `visiblePosts` but still not getting the correct response 🤔

app/Models/Signup.php
```diff
@@ -213,17 +213,29 @@ public function getAcceptedQuantity()
...
+                    $query->where(function ($query) use ($types) {
+                        foreach ($types as $type) {
+                            $query->orWhere('type', $type);
+                        }
+                    });
+                } else {
+                    $query->where('type', $types[0]);
+                }
+            }
+
```
chloealee: @DFurnes also tried this way too but still not returning posts with the matching `types` I'm sending in. I checked clockwork and it does look like the queries are potentially overriding the `type` match (if the order goes from oldest query at the top and latest query at the bottom of the list)?

DFurnes: The queries look fine to me at first glance. First, we get the matching signups & then we make one more query for each of the relationships we're included (posts, reactions, and tags):

screen shot 2018-07-23 at 4 43 18 pm

DFurnes: Our problem is somewhere in that first "get posts for each signup" query, where it's not properly scoping things to the provided posts.

chloealee: @DFurnes I don't _think_ this matters but is it because it isn't doing `posts.type` and just doing `type` in the query?

chloealee: actually I don't think so because `status` is just there alone 🤔

DFurnes: I think our problem is this first `status = 'accepted' or northstar_id IS NULL` in here, which is coming from the `visiblePosts` relationship:

```sql
SELECT `posts`.*, (SELECT count(*) FROM `reactions` WHERE `posts`.`id` = `reactions`.`post_id` and `reactions`.`deleted_at` IS NULL) as `reactions_count` FROM `posts` WHERE (`status` = 'accepted' or `northstar_id` IS NULL and `posts`.`signup_id` in ('1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '20', '21') and `status` = 'accepted' and `type` in ('voter-reg', 'text')) and `posts`.`deleted_at` IS NULL
```

When I update that relationship to check similarly to our eager load method in 6b5313a, things look like they start working (both in the failing test & manual testing with Paw).

That leaves me with just one failing test:

```
1) Tests\Http\SignupTest::testSignupIndexWithIncludedPostsAndParamsAsAdmin
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'text'
+'photo'

/home/vagrant/Sites/rogue/tests/Http/SignupTest.php:381
```

chloealee: ah interesting, @DFurnes ! Thanks for this!! So does this have to re-check in the `scopeWithVisiblePosts` after it gets the results from `visiblePosts`? 

Re: the failing test, that is a funny one too! When I run that by itself, it seems to pass but only failed with running all tests together. In the wercker build it seemed to pass ok?

chloealee:
yay @DFurnes your fix did it! Is it ok to merge this? I'll test on thor tonight but will wait to push to prod tomorrow if all looks good!

app/Models/Signup.php
DFurnes: After running it a few more times, I think that test may just be flaky based on timing (since we're checking specifically the first post is type `text` and the second one is type `photo`). Instead, let's check the number of responses we get with different filters (updated in b82192a6).

app/Models/Signup.php
```diff
@@ -213,17 +216,24 @@ public function getAcceptedQuantity()
...
+    public function scopeWithVisiblePosts($query, $types = null)
     {
-        if (! is_staff_user()) {
-            return $query->with(['posts' => function ($query) {
-                $query->where('status', 'accepted')
-                ->orWhere('northstar_id', auth()->id());
-            }]);
-        }
+        return $query->with(['visiblePosts' => function ($query) use ($types) {
+            if (! is_staff_user()) {
```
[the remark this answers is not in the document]
DFurnes: I removed that in ca8c914 & tests still seem to be passing! 🍃

chloealee:
awesome, @DFurnes ! Removed those comments out of the `SignupTransformer`! Thanks for pairing with me on this!!!

PR #747, opened 2018-07-03
Update environments in setup command/default environment.

DFurnes:
#### What's this PR do?
Just updating our default environment to have the correct Northstar & Aurora URLs.

👀

N/A

#### Relevant tickets
References DoSomething/devops#383 and DoSomething/devops#388.

#### How to deploy

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -49,10 +49,10 @@ SQS_SECRET_ACCESS_KEY=
 SQS_PREFIX=
 SQS_DEFAULT_QUEUE=
 
-NORTHSTAR_URL=https://northstar-qa.dosomething.org
-NORTHSTAR_AUTH_ID=
+NORTHSTAR_URL=https://identity-dev.dosomething.org
+NORTHSTAR_AUTH_ID=dev-oauth
 NORTHSTAR_AUTH_SECRET=
-NORTHSTAR_CLIENT_ID=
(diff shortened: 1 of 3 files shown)


.env.example
```diff
@@ -49,10 +49,10 @@ SQS_SECRET_ACCESS_KEY=
 SQS_PREFIX=
 SQS_DEFAULT_QUEUE=
 
-NORTHSTAR_URL=https://northstar-qa.dosomething.org
-NORTHSTAR_AUTH_ID=
+NORTHSTAR_URL=https://identity-dev.dosomething.org
+NORTHSTAR_AUTH_ID=dev-oauth
 NORTHSTAR_AUTH_SECRET=
-NORTHSTAR_CLIENT_ID=
+NORTHSTAR_CLIENT_ID=dev-machine
```
chloealee: @DFurnes random q! Why is this `importer-client` in chompy but `dev-machine` here? I would have expected something related to Rogue?

DFurnes: We should be using `dev-machine` on our locals (so that we can audit which environment requests or changes are coming from), and `importer-client` on Chompy (prod?).

chloealee: oooh ok good to know! I've been trying to rebuild my local chompy and still haven't gotten it to work so I'll do that this time and maybe that will do that trick. Thanks for explaining!

PR #748, opened 2018-07-10
adds share-social as a valid post type

chloealee:
#### What's this PR do?
- Adds share-social as a valid post type
- Adds tests to make sure you can create a `voter-reg` and `share-social` post.
- Adds a test to make sure you can't create a post if it isn't one of the following types: `voter-reg`, `share-social`, `photo`, `text`
- Adds `assertPostStructure` helper function in Post tests to DRY up code.

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/158870385) Pivotal Card.

#### How to deploy

### DIFF
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -14,6 +14,39 @@ use Illuminate\Http\UploadedFile;
 class PostTest extends TestCase
 {
     /**
+     * Helper function to assert post structure
+     *
+     */
+    public function assertPostStructure($response)
+    {
+        return $response->assertJsonStructure([
+            'data' => [
+                'id',
+                'signup_id',
+                'northstar_id',
+                'type',
+                'action',
+                'media' => [
+                    'url',
+                    'original_image_url',
+                    'text',
+                ],
+                'quantity',
+                'tags' => [],
+                'reactions' => [
+                    'reacted',
+                    'total',
+                ],
+                'status',
+                'details',
+                'source',
+                'remote_addr',
+                'created_at',
+                'updated_at',
+            ],
+        ]);
+    }
+    /**
      * Test that a POST request to /posts creates a new
      * post and signup, if it doesn't already exist.
      *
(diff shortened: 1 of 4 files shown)


katiecrane:
@chloealee Do you think we need a test for this? I know we have `testCreatingATextPost` and `testCreatingAPhotoPost`

chloealee:
cc @DoSomething/storm does this have any implications for Quasar if we start allowing the post type `share-social`?

chloealee:
@katiecrane great call - added! I also added a test that makes sure you can't create a post that isn't `text`, `voter-reg`, `photo`, or `share-social`.

tests/Http/PostTest.php
```diff
@@ -229,6 +229,169 @@ public function testCreatingATextPost()
...
+
+        // Mock the Blink API call.
+        $this->mock(Blink::class)->shouldReceive('userSignupPost');
+
+        // Create the post with an invalid type (not in text, photo, voter-reg, share-social).
+        $response = $this->withAccessToken($signup->northstar_id)->postJson('api/v3/posts', [
+            'northstar_id'     => $signup->northstar_id,
+            'campaign_id'      => $signup->campaign_id,
+            'campaign_run_id'  => $signup->campaign_run_id,
+            'type'             => 'social-share',
```
DFurnes: Oh man, this took me a while to realize what was off! 😂 A very real-world test!

tests/Http/PostTest.php
```diff
@@ -229,6 +229,169 @@ public function testCreatingATextPost()
...
+            'type'             => 'share-social',
+            'action'           => 'test-action',
+            'quantity'         => $quantity,
+            'text'             => $text,
+            'details'          => json_encode($details),
+        ]);
+
+        $response->assertStatus(201);
+
+        $response->assertJsonStructure([
```
DFurnes: I wonder if we should make a helper since this same structure is repeated a few places... e.g.

```php
public function assertPostStructure($response)
{
    return $response->assertJsonStructure([
        'data' => [
            'id',
            ...
        ],
    ]);
}

// ...

public function testCreatingAVoterRegPost()
{
    // ...
    $response->assertStatus(201);
    $this->assertPostStructure($response);
    // ...
}

public function testCreatingAShareSocialPost()
{
    // ...
    $response->assertStatus(201);
    $this->assertPostStructure($response);
    // ...
}
```

tests/Http/PostTest.php
chloealee: haha @DFurnes I kept wanting to write `social-share` instead of `share-social` when I was coding for this ticket so thought it would be a tricky test! :)

tests/Http/PostTest.php
chloealee: ooo i like this, @DFurnes ! Updated!

PR #749, opened 2018-07-11
Mocks all NS Api Calls in Unit Tests

chloealee:
#### What's this PR do?
Mocks all NS Api Calls in Unit Tests.
  - Adds helper function `mockNorthstarApiCallAndReturnUser()` to mock API call to NS and return user info. 

All tests should pass! 

We should not be making actual calls to any outside API in unit tests - tests were sometimes failing not because the Rogue code was broken but because other systems (NS) was down or configuration was wrong. This defeats the purpose of unit tests actually testing the Rogue code and logic since it isn't isolating just Rogue code but tests other apps as well. 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/158083551) Pivotal Card. 

#### How to deploy

### DIFF
--- a/tests/CreatesApplication.php
+++ b/tests/CreatesApplication.php
@@ -2,19 +2,10 @@
 
 namespace Tests;
 
-use DoSomething\Gateway\Northstar;
 use Illuminate\Contracts\Console\Kernel;
-use DoSomething\Gateway\Resources\NorthstarUser;
 
 trait CreatesApplication
 {
-    /**
-     * The Northstar API client mock.
-     *
-     * @var \Mockery\MockInterface
-     */
-    protected $northstarMock;
-
     /**
      * The Faker generator, for creating test data.
      *
@@ -32,21 +23,6 @@ trait CreatesApplication
         $this->app = require __DIR__.'/../bootstrap/app.php';
 
         $this->app->make(Kernel::class)->bootstrap();
-
-        $this->northstarMock = $this->mock(Northstar::class);
-        $this->northstarMock->shouldReceive('asClient')->andReturnSelf();
-        $this->northstarMock->shouldReceive('asUser')->andReturnSelf();
-        $this->northstarMock->shouldReceive('refreshIfExpired')->andReturnSelf();
-        $this->northstarMock->shouldReceive('getUser')->andReturnUsing(function ($type, $id) {
-            return new NorthstarUser([
-                    'id' => $id ? $id : $this->faker->northstar_id,
-                    'first_name' => $this->faker->firstName,
-                    'last_name' => $this->faker->lastName,
-                    'birthdate' => $this->faker->date,
-                    'email' => $this->faker->email,
-                    'mobile' => $this->faker->phoneNumber,
-                ]);
-        });
 
         return $this->app;
     }
--- a/tests/CreatesApplication.php
+++ b/tests/CreatesApplication.php
@@ -29,11 +29,11 @@ trait CreatesApplication
      */
     public function createApplication()
     {
-        $app = require __DIR__.'/../bootstrap/app.php';
+        $this->app = require __DIR__.'/../bootstrap/app.php';
 
-        $app->make(Kernel::class)->bootstrap();
+        $this->app->make(Kernel::class)->bootstrap();
 
-        $this->northstarMock = $this->mock(\DoSomething\Gateway\Northstar::class);
+        $this->northstarMock = $this->mock(Northstar::class);
         $this->northstarMock->shouldReceive('asClient')->andReturnSelf();
         $this->northstarMock->shouldReceive('getUser')->andReturnUsing(function ($type, $id) {
             return new NorthstarUser([
@@ -46,6 +46,6 @@ trait CreatesApplication
                 ]);
         });
 
-        return $app;
+        return $this->app;
     }
 }
--- a/tests/CreatesApplication.php
+++ b/tests/CreatesApplication.php
@@ -39,7 +39,7 @@ trait CreatesApplication
         $this->northstarMock->shouldReceive('refreshIfExpired')->andReturnSelf();
         $this->northstarMock->shouldReceive('getUser')->andReturnUsing(function ($type, $id) {
             return new NorthstarUser([
-                    'id' => $this->faker->northstar_id,
+                    'id' => $id ? $id : $this->faker->northstar_id,
                     'first_name' => $this->faker->firstName,
                     'last_name' => $this->faker->lastName,
                     'birthdate' => $this->faker->date,
--- a/tests/CreatesApplication.php
+++ b/tests/CreatesApplication.php
@@ -2,10 +2,26 @@
 
 namespace Tests;
 
+use DoSomething\Gateway\Northstar;
 use Illuminate\Contracts\Console\Kernel;
+use DoSomething\Gateway\Resources\NorthstarUser;
 
 trait CreatesApplication
 {
+    /**
+     * The Northstar API client mock.
+     *
+     * @var \Mockery\MockInterface
+     */
+    protected $northstarMock;
+
+    /**
+     * The Faker generator, for creating test data.
+     *
+     * @var \Faker\Generator
+     */
+    protected $faker;
+
     /**
      * Creates the application.
      *
@@ -16,6 +32,19 @@ trait CreatesApplication
         $app = require __DIR__.'/../bootstrap/app.php';
 
         $app->make(Kernel::class)->bootstrap();
+
+        $this->northstarMock = $this->mock(\DoSomething\Gateway\Northstar::class);
+        $this->northstarMock->shouldReceive('asClient')->andReturnSelf();
+        $this->northstarMock->shouldReceive('getUser')->andReturnUsing(function ($type, $id) {
+            return new NorthstarUser([
+                    'id' => $this->faker->northstar_id,
+                    'first_name' => $this->faker->firstName,
+                    'last_name' => $this->faker->lastName,
+                    'birthdate' => $this->faker->date,
+                    'email' => $this->faker->email,
+                    'mobile' => $this->faker->phoneNumber,
+                ]);
+        });
 
         return $app;
     }
--- a/tests/CreatesApplication.php
+++ b/tests/CreatesApplication.php
@@ -35,6 +35,8 @@ trait CreatesApplication
 
         $this->northstarMock = $this->mock(Northstar::class);
         $this->northstarMock->shouldReceive('asClient')->andReturnSelf();
+        $this->northstarMock->shouldReceive('asUser')->andReturnSelf();
+        $this->northstarMock->shouldReceive('refreshIfExpired')->andReturnSelf();
         $this->northstarMock->shouldReceive('getUser')->andReturnUsing(function ($type, $id) {
             return new NorthstarUser([
                     'id' => $this->faker->northstar_id,
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -8,6 +8,7 @@ use Rogue\Models\User;
 use Rogue\Models\Signup;
 use Rogue\Models\Reaction;
 use Rogue\Services\Fastly;
+use Rogue\Services\Registrar;
 use DoSomething\Gateway\Blink;
 use Illuminate\Http\UploadedFile;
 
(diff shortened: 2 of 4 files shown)


tests/Http/SignupTest.php
```diff
@@ -6,10 +6,38 @@
...
+    /**
+     * Helper function to mock NS API calls and receive a user.
+     */
+    public function mockNorthstarApiCallAndReturnUser($northstarId = null)
+    {
+        // Mock the Northstar API calls.
+        $this->mock(Registrar::class)
+            ->shouldReceive('find')
+            ->andReturnUsing(function($northstarId) {
+                $mockNorthstarUser = new \StdClass;
```
DFurnes: Let's make this a [`NorthstarUser`](https://github.com/DoSomething/gateway/blob/15d3201b904d34d4220193938c0d4dce067f32dc/src/Resources/NorthstarUser.php) instead of a `StdClass`, in case we type-check this anywhere:

```php
->andReturnUsing(function($northstarId) {
    return new NorthstarUser([
        'id' => $northstarId ? $northstarId : $this->faker->northstar_id,
        'first_name' => $this->faker->firstName,
        // ...
    ]);
}
```

tests/Http/PostTest.php
```diff
@@ -1102,6 +1134,10 @@ public function testCreatingVoterRegistrationPost()
         // Mock the Blink API call.
         $this->mock(Blink::class)->shouldReceive('userSignupPost');
 
+        // Mock the Northstar API calls.
+            $this->mock(Registrar::class)
```
DFurnes: Rather than mocking the `Registrar`, which has some Rogue-specific logic in it, let's mock the [`Northstar`](https://github.com/DoSomething/gateway/blob/15d3201b904d34d4220193938c0d4dce067f32dc/src/Northstar.php) client itself. We can set up a global mock in `createApplication` [like we do in Northstar](https://github.com/DoSomething/northstar/blob/e0eac7d94ec1b2c74345de61ac9a7ead1990ab11/tests/CreatesApplication.php#L46-L48):

```php
/**
 * Creates the application.
 *
 * @return \Illuminate\Foundation\Application
 */
public function createApplication()
{
    // ...

    $this->northstarMock = $this->mock(\DoSomething\Gateway\Northstar::class);

    $this->northstarMock->shouldReceive('asClient')->andReturnSelf();
    $this-> northstarMock->shouldReceive('getUser')->andReturnUsing(function ($type, $id) {
        return new NorthstarUser([
            // ...
        ]);
    });

```

DFurnes: This also reminds me of [this pull request](https://github.com/DoSomething/gateway/pull/106), which I was pretty excited about back in the day.

chloealee: Ah this is great, thanks @DFurnes ! What would the `$type` param be here? Also, how would `createApplication()` receive the `$id` (or `$northstarId`)?

tests/Http/SignupTest.php
```diff
@@ -6,10 +6,32 @@
...
 
 class SignupTest extends TestCase
 {
+    /**
+     * Helper function to mock NS API calls and receive a user.
+     */
+    public function mockNorthstarApiCallAndReturnUser($northstarId = null)
+    {
+        // Mock the Northstar API calls.
+        $this->mock(Registrar::class)
```
chloealee: @DFurnes should this also mock the `Northstar` client instead of `Registrar`?

DFurnes: Ideally, yes! We always want our mocks to run as close as possible to the thing we're trying to mock (which in this case is the external HTTP call). That way, if there are bugs that pop up in the Registrar itself we'll still catch them in our unit tests.

tests/Http/PostTest.php
DFurnes: The `$type` parameter is what kind of ID is being passed, so in this case [we pass `id`](https://github.com/DoSomething/rogue/blob/e1c4e3cb89df7b4d57809d422c5517768aa61c82/app/Services/Registrar.php#L33) and an ID, but we could theoretically also pass `'email'` or `'mobile'` (I don't think we do in Rogue, though, so not really necessary to cover in our mock.)

DFurnes: > Also, how would createApplication() receive the $id (or $northstarId)?

It doesn't! The `createApplication` method is run by PHPUnit before each test case, and it configures a "fresh" instance of the Laravel app and sets up mocks or other things like that. 

The mock we've configured here for `Northstar` will then use this callback instead whenever some code calls `getUser` on it. (We're basically saying "Hey Laravel, whenever you're asked for `Northstar`, use this instead. And when this thing is asked to `getUser`, here's what it should do.")

tests/Http/SignupTest.php
chloealee: Makes sense, thanks for explaining, @DFurnes !

tests/Http/PostTest.php
chloealee: ahh I see, thank you for this explanation!!

tests/Http/SignupTest.php
```diff
@@ -6,10 +6,33 @@
...
 
 class SignupTest extends TestCase
 {
+    /**
+     * Helper function to mock NS API calls and receive a user.
+     */
+    public function mockNorthstarApiCallAndReturnUser($northstarId = null)
+    {
+        // Mock the Northstar API calls.
+        $this->mock(Northstar::class)
```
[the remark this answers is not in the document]
DFurnes: Hmm, it looks like `$this->app` is null? Can you try changing the `$app` in `createApplication` to `$this->app`?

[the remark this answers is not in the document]
chloealee: @DFurnes that helped! But now it seems like a lot of other tests (event tests, web tests) are breaking. Here's an example of the message for the `testDeletingAPost()` from `Tests\Http\Legacy\Web`:
```
array:5 [
  "message" => "Method Mockery_0_DoSomething_Gateway_Northstar::asUser() does not exist on this mock object"
  "exception" => "BadMethodCallException"
  "file" => "/home/vagrant/DoSomething.org/rogue/vendor/mockery/mockery/library/Mockery/Loader/EvalLoader.php(16) : eval()'d code"
  "line" => 709
  "trace" => array:72 [
    0 => array:5 [
      "file" => "/home/vagrant/DoSomething.org/rogue/vendor/mockery/mockery/library/Mockery/Loader/EvalLoader.php(16) : eval()'d code"
      "line" => 1014
      "function" => "_mockery_handleMethodCall"
      "class" => "Mockery_0_DoSomething_Gateway_Northstar"
      "t
```

I don't think that these tests should be hitting NS though?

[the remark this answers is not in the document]
DFurnes: That error is saying we haven't added that method to our mock, which we haven't! Let's add it here and see what we get after that:

```diff
public function createApplication()
{
    $this->app = require __DIR__.'/../bootstrap/app.php';
    $this->app->make(Kernel::class)->bootstrap();
    $this->northstarMock = $this->mock(Northstar::class);
    $this->northstarMock->shouldReceive('asClient')->andReturnSelf();
+   $this->northstarMock->shouldReceive('asUser')->andReturnSelf();
    $this->northstarMock->shouldReceive('getUser')->andReturnUsing(function ($type, $id) {
```

tests/CreatesApplication.php
```diff
@@ -2,21 +2,52 @@
...
     public function createApplication()
     {
-        $app = require __DIR__.'/../bootstrap/app.php';
+        $this->app = require __DIR__.'/../bootstrap/app.php';
+
+        $this->app->make(Kernel::class)->bootstrap();
 
-        $app->make(Kernel::class)->bootstrap();
+        $this->northstarMock = $this->mock(Northstar::class);
+        $this->northstarMock->shouldReceive('asClient')->andReturnSelf();
```
chloealee: @DFurnes I'm still getting a lot of these errors for Post Tests and Signup Tests:
```
"message" => "Method Mockery_0_DoSomething_Gateway_Northstar::asClient() does not exist on this mock object"
```
however, it does exist here. Do I need to call this in a special way or anything from the `PostTest.php` / `SignupTest.php` files?

DFurnes: Since we're calling `$this->mock(Northstar::class)` in each of these tests we're rewriting the mock created here. Instead, let's use the existing mock (we can remove all of these in `PostTest` and `SignupTest` and use the base mock we've created here).

If we ever want a one-off expectation applied to this mock, instead of `$this->mock(Northstar::class)`, we can just overwrite the existing mock at `$this->northstarMock`.

chloealee: OHH I get it now! Thank you @DFurnes ! All is updated :)

chloealee:
@DFurnes sorry for all the q's! The unit tests are all passing but now the Dusk tests seem to be failing! Getting this error:
```
1) Tests\Browser\HomeTest::testHomepage
Error: Call to undefined method Tests\Browser\HomeTest::mock()

/pipeline/source/tests/CreatesApplication.php:36
/pipeline/source/vendor/laravel/framework/src/Illuminate/Foundation/Testing/TestCase.php:91
/pipeline/source/vendor/laravel/framework/src/Illuminate/Foundation/Testing/TestCase.php:68
/pipeline/source/vendor/laravel/dusk/src/TestCase.php:27
```
Again, not sure why this page / test would break since it shouldn't be trying to hit NS?

DFurnes:
Oh, right. We use the same `CreateApplication` trait in both PHPUnit & Dusk tests... can we move this configuration into [`configureMocks`](https://github.com/DoSomething/rogue/blob/9d68e5fd550d3cd2f0686c17663c4fa6d9cb0f1a/tests/WithMocks.php#L18-L33) instead? That should only be run for PHPUnit tests.

chloealee:
@DFurnes ah yes - updated! :)

tests/Http/PostTest.php
```diff
@@ -66,6 +66,9 @@ public function testCreatingAPostAndSignup()
             ->shouldReceive('userSignup')
             ->shouldReceive('userSignupPost');
 
+        // Mock the Northstar API calls.
+        $this->northstarMock;
```
DFurnes: These lines aren't doing anything and can be removed (`$this->northstarMock` is a property, not a method).

tests/Http/SignupTest.php
```diff
@@ -382,6 +382,9 @@ public function testSignupIndexWithIncludedUserAsAdmin()
         $post = factory(Post::class)->create();
         $signup = $post->signup;
 
+        // Mock the Northstar API calls.
+        $this->northstarMock;
```
DFurnes: These lines aren't doing anything & can be removed, [like in `PostTest`](
https://github.com/DoSomething/rogue/pull/749/files/76985add32c27f1f95beb0a1e1ed7fb9645488bd#diff-ec1f9ba9f2cdaf675bcdfdd5136f75f).

tests/Http/PostTest.php
```diff
@@ -1102,6 +1102,10 @@ public function testCreatingVoterRegistrationPost()
         // Mock the Blink API call.
         $this->mock(Blink::class)->shouldReceive('userSignupPost');
 
+        // Mock the Northstar API calls.
+            $this->mock(Northstar::class)
```
DFurnes: It isn't necessary to re-mock Northstar here since we handle all calls to `getUser` in the base mock.

tests/Http/SignupTest.php
chloealee: ah shoot sorry about that - my bad! 🤦‍♀️ Updated now!

tests/WithMocks.php
```diff
@@ -30,6 +32,22 @@ public function configureMocks()
...
         $this->faker->addProvider(new \FakerCampaignId($this->faker));
+
+        // Northstar Mock
+        $this->northstarMock = $this->mock(Northstar::class);
+        $this->northstarMock->shouldReceive('asClient')->andReturnSelf();
+        $this->northstarMock->shouldReceive('asUser')->andReturnSelf();
+        $this->northstarMock->shouldReceive('refreshIfExpired')->andReturnSelf();
+        $this->northstarMock->shouldReceive('getUser')->andReturnUsing(function ($type, $id) {
+            return new NorthstarUser([
+                    'id' => $id ? $id : $this->faker->northstar_id,
```
DFurnes: Hmm, I think we should make sure this _is_ an ID here before setting it (e.g. if `getUser('email', '...')` was called, we'd still set it as `id` with how this is set up right now). Let's instead check if `$type == 'id'`.

```php
'id' => $type == 'id' ? $id : $this->faker->northstar_id,
```

chloealee: ah that is super smart, thanks for flagging @DFurnes ! Updated that :)

PR #752, opened 2018-07-30
fixes GET /signups/{signup} to parse by included post types

chloealee:
#### What's this PR do?
- Updates `GET /signups/{signup}` to be able to parse by included post types when hitting something like `api/v3/signups/1?include=posts:type(voter-reg)` to only return `voter-reg` type posts.
- Updates tests and documentation.

Hit that endpoint - only posts with the type `voter-reg` should be returned! 

This is needed to Hide Voter Reg Posts in Rogue (work started in #746) and to hide these posts on the signup page in Rogue admin

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/158011726) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Http/Controllers/SignupsController.php
+++ b/app/Http/Controllers/SignupsController.php
@@ -124,6 +124,17 @@ class SignupsController extends ApiController
      */
     public function show(Request $request, Signup $signup)
     {
+        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
+        if (starts_with($request->query('include'), 'posts')) {
+            $types = (new \League\Fractal\Manager)
+                ->parseIncludes($request->query('include'))
+                ->getIncludeParams('posts');
+
+            $types = $types ? $types->get('type') : null;
+
+            $signup = $signup->withVisiblePosts($types)->first();
+        }
+
         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
     }
 
--- a/app/Http/Controllers/SignupsController.php
+++ b/app/Http/Controllers/SignupsController.php
@@ -132,7 +132,7 @@ class SignupsController extends ApiController
 
             $types = $types ? $types->get('type') : null;
 
-            $signup = $signup->withVisiblePosts($types)->first();
+            $signup = $signup->withVisiblePosts($types)->get()->keyBy('id')[$signup->id];
         }
 
         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
--- a/app/Http/Controllers/SignupsController.php
+++ b/app/Http/Controllers/SignupsController.php
@@ -132,7 +132,11 @@ class SignupsController extends ApiController
 
             $types = $types ? $types->get('type') : null;
 
-            $signup = $signup->withVisiblePosts($types)->get()->keyBy('id')[$signup->id];
+            $signup->load(['visiblePosts' => function ($query) use ($types) {
+                if ($types) {
+                    $query->whereIn('type', $types);
+                }
+            }]);
         }
 
         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/SignupsController.php
```diff
@@ -124,6 +124,17 @@ public function index(Request $request)
...
     {
+        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
+        if (starts_with($request->query('include'), 'posts')) {
+            $types = (new \League\Fractal\Manager)
+                ->parseIncludes($request->query('include'))
+                ->getIncludeParams('posts');
+
+            $types = $types ? $types->get('type') : null;
+
+            $signup = $signup->withVisiblePosts($types)->first();
```
DFurnes: This `->first()` is going to cause us trouble I think - this is equivalent to calling `Signup::first();` so it'll always rewrite `$signup` as the first signup in the database, rather than whatever was passed from the route we're on.

DFurnes: It's a little tricky that the test suite is still passing here - let's edit one of the `v3/signup/{id}` tests to seed a few more signups and grab one in the middle so we let our test suite catch an error like this.

chloealee: ohhh I thought that `first()` called the first response back! Thanks for flagging that @DFurnes ! I was testing with the first signup so that actually makes a lot of sense. 

If I just do `$signup->withVisiblePosts($types)` I get this error:
```
  "message": "Type error: Argument 1 passed to Rogue\\Http\\Transformers\\SignupTransformer::transform() must be an instance of Rogue\\Models\\Signup, instance of Illuminate\\Database\\Eloquent\\Builder given, called in /home/vagrant/DoSomething.org/rogue/vendor/league/fractal/src/Scope.php on line 338",
    "exception": "Symfony\\Component\\Debug\\Exception\\FatalThrowableError",
``` 
I'm not sure how to access this scope on the `$signup` without passing a query (instead of just the signup)?

chloealee: oh wait I think I found a fix! :)

chloealee: @DFurnes also added new tests to test this logic as well! Thanks for flagging that! Ready for another 👀 when you have a sec!

app/Http/Controllers/SignupsController.php
```diff
@@ -124,6 +124,17 @@ public function index(Request $request)
...
     {
+        // Only allow an admin or the user who owns the signup to see the signup's unapproved posts.
+        if (starts_with($request->query('include'), 'posts')) {
+            $types = (new \League\Fractal\Manager)
+                ->parseIncludes($request->query('include'))
+                ->getIncludeParams('posts');
+
+            $types = $types ? $types->get('type') : null;
+
+            $signup = $signup->withVisiblePosts($types)->get()->keyBy('id')[$signup->id];
```
DFurnes: This seems suspect! It looks like we're loading all signups into memory, running a transformation on that array, and then grabbing an index. This gets our test passing, but would probably choke in production (we have a lot of signups!!)

DFurnes: How about something like this, where we load the relationship onto the model (using `load`) and then customize it w/ a closure to limit it by the `type` column (similar to how we handled eager-loading before):

```diff
             $types = $types ? $types->get('type') : null;

-            $signup = $signup->withVisiblePosts($types)->get()->keyBy('id')[$signup->id];
+            $signup->load(['visiblePosts' => function ($query) use ($types) {
+                if ($types) {
+                    $query->whereIn('type', $types);
+                }
+            }]);
         }

         return $this->item($signup, 200, [], $this->transformer, $request->query('include'));
```

chloealee: ah that is so true - I need to keep that top of mind when using queries and large datasets for performance! Updated with what you said - thanks @DFurnes !

PR #753, opened 2018-07-31
Limit post siblings to only return up to 10 siblings 

chloealee:
#### What's this PR do?
Limits post siblings to only return up to 10 siblings and updates documentation.

👀 

Inboxes were not loading because there were some posts that had many siblings and the page would time out. Now, we only return up to 10 post siblings so this doesn't happen! 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/159406186) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -85,21 +85,6 @@ class PostsController extends ApiController
             $query = $query->withTag($filters['tag']);
         }
 
-        // If there is a limit set on siblings, include in query.
-        if ($request->query('include')) {
-            foreach ($request->query('include') as $key => $value) {
-                if ((strpos($value, 'siblings') !== false) && (strpos($value, 'limit') !== false)) {
-
-                    $limit = (new \League\Fractal\Manager)
-                            ->parseIncludes($value)
-                            ->getIncludeParams('siblings');
-
-                    $limit = $limit ? (int) $limit->get('limit')[0] : null;
-
-                    $query = $query->withPostLimit($limit);
-                }
-            }
-        }
         return $this->paginatedCollection($query, $request);
     }
 
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -85,6 +85,21 @@ class PostsController extends ApiController
             $query = $query->withTag($filters['tag']);
         }
 
+        // If there is a limit set on siblings, include in query.
+        if ($request->query('include')) {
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/PostsController.php
```diff
@@ -85,6 +85,21 @@ public function index(Request $request)
...
+            foreach ($request->query('include') as $key => $value) {
+                if ((strpos($value, 'siblings') !== false) && (strpos($value, 'limit') !== false)) {
+
+                    $limit = (new \League\Fractal\Manager)
+                            ->parseIncludes($value)
+                            ->getIncludeParams('siblings');
+
+                    $limit = $limit ? (int) $limit->get('limit')[0] : null;
+
+                    $query = $query->withPostLimit($limit);
```
chloealee: @DFurnes working on [setting a limit when grabbing a post's siblings](https://www.pivotaltracker.com/n/projects/2019429/stories/159406186) so this doesn't break the inbox! 

I'm having trouble getting this to work and wanted a quick gut check to get your thoughts on doing it this way (closely following how we did the work here to parse by post types [here](https://github.com/DoSomething/rogue/pull/752)). 

Is this a reasonable way to do this?

app/Models/Post.php
```diff
@@ -393,4 +393,19 @@ public function hasGoodTag()
...
+    * Scope a query to only return X amount of siblings, determined by limit param.
+    *
+    * @param int $limit
+    * @return \Illuminate\Database\Eloquent\Builder
+    */
+    public function scopeWithPostLimit($query, $limit = null)
+    {
+        return $query->with(['siblings' => function ($query) use ($limit) {
+            if ($limit) {
+                $query->take($limit);
```
chloealee: tried using both `take()` and `limit()` and both seem to return 0 results when there should be siblings :-/

app/Http/Controllers/PostsController.php
DFurnes: TBH, I'm not sure this needs to be configurable per request. If you want, just adding a `->take(10);` to the [siblings relationship](https://github.com/DoSomething/rogue/blob/6455f3c742b68c89703d1bb740135ec6f8ce14eb/app/Models/Post.php#L173-L179) should do the trick without too much added complexity!

chloealee: ah that makes total sense I was definitely overthinking! Updated and also the documentation! :)

PR #756, opened 2018-08-06
Eager load signups on the v3 posts index

chloealee:
#### What's this PR do?
- We want to eagerly load the signup on the v3 posts index for performance - some inboxes were timing out! 
- Adds `has_include` filter to see if a certain include value is present (and deals with how an include filter can be submitted). 

👀 

@DFurnes noticed that there is a lot of time being spent on the pivot relationship and we want to see if not returning the reactions helps. If so, we'll update the API to exclude this in the inbox. More info. [here](https://dosomething.slack.com/archives/C0YGXUE01/p1533574191000185?thread_ts=1533569414.000231&cid=C0YGXUE01).

#### Relevant tickets
https://www.pivotaltracker.com/n/projects/2019429/stories/159506797

#### How to deploy

### DIFF
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -70,8 +70,8 @@ class PostsController extends ApiController
     {
         $query = $this->newQuery(Post::class)
             ->orderBy('created_at', 'desc');
-
-        if ($request->query('include') && in_array('signup', $request->query('include'))) {
+            dd(has_include($request, 'signup'));
+        if ($request->query('include') && has_include($request, 'signup')) {
             // Eagerly load the `signup` relationship.
             $query->with('signup');
         }
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -69,7 +69,7 @@ class PostsController extends ApiController
     public function index(Request $request)
     {
         $query = $this->newQuery(Post::class)
-            ->orderBy('created_at', 'desc');
+            ->orderBy('created_at', 'desc')
             // Eagerly load the `signup` relationship.
             ->with('signup');
 
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -87,7 +87,6 @@ class PostsController extends ApiController
             $query = $query->withTag($filters['tag']);
         }
 
-
         return $this->paginatedCollection($query, $request);
     }
 
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -71,7 +71,7 @@ class PostsController extends ApiController
         $query = $this->newQuery(Post::class)
             ->orderBy('created_at', 'desc');
 
-        if (in_array('signup', $request->query('include'))) {
+        if ($request->query('include') && in_array('signup', $request->query('include'))) {
             // Eagerly load the `signup` relationship.
             $query->with('signup');
         }
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -70,7 +70,7 @@ class PostsController extends ApiController
     {
         $query = $this->newQuery(Post::class)
             ->orderBy('created_at', 'desc');
-            dd(has_include($request, 'signup'));
+
         if ($request->query('include') && has_include($request, 'signup')) {
             // Eagerly load the `signup` relationship.
             $query->with('signup');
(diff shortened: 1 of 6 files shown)


app/Http/Transformers/PostTransformer.php
```diff
@@ -42,10 +42,10 @@ public function transform(Post $post)
                 'text' => $post->text,
             ],
             'quantity' => $post->quantity,
-            'reactions' => [
-                'reacted' => ! empty($post->reaction),
-                'total' => $post->reactions_count,
-            ],
+            // 'reactions' => [
```
DFurnes: So we don't end up with errors elsewhere when we test this, let's just return `false` and `0` for each of these.

app/Models/Post.php
```diff
@@ -22,7 +22,7 @@ class Post extends Model
      * Always load a user's own reaction,
      * if they're logged-in.
      */
-    protected $with = ['reaction', 'tags'];
+    protected $with = ['tags'];
```
DFurnes: Let's remove tags too (and same deal, just return `[]` for that field in the transformer).

DFurnes:
@chloealee Once you've made those changes, you can [deploy this branch to QA](https://dashboard.heroku.com/apps/dosomething-rogue-qa/deploy/github) here without merging this PR into `master`. May be worth dropping a heads-up in #deploys as well so folks know to expect that reactions will be temporarily unavailable on QA as well!

chloealee:
@DFurnes so cool! So do I just click on the `reactions` branch and manually deploy like below? 
screen shot 2018-08-06 at 1 32 25 pm

DFurnes:
Yup! Just make sure to replace `$post->tagSlugs()` with `[]` in the transformer first so we're not still accidentally making that database query.

chloealee:
@DFurnes oh right, thank you! Updated :)

chloealee:
@DFurnes I updated this to eagerly load the signup to see if that helps since we were seeing some funky issues with the signup yesterday and changed everything else back! I'd love to deploy just this branch to QA like we did yesterday but wanted to get your approval first!

chloealee:
yay @DFurnes this worked for loading the inbox on Thor! Can you review?

app/Models/Post.php
```diff
@@ -22,7 +22,7 @@ class Post extends Model
      * Always load a user's own reaction,
      * if they're logged-in.
      */
-    protected $with = ['reaction', 'tags'];
+    // protected $with = ['reaction', 'tags'];
```
DFurnes: Let's add these back in!

chloealee: ah yes you beat me to that! Just updated that :)

app/Http/Controllers/PostsController.php
```diff
@@ -69,7 +69,9 @@ public function __construct(PostManager $posts, SignupManager $signups, PostTran
     public function index(Request $request)
     {
         $query = $this->newQuery(Post::class)
-            ->orderBy('created_at', 'desc');
+            ->orderBy('created_at', 'desc')
+            // Eagerly load the `signup` relationship.
+            ->with('signup');
```
DFurnes: We only want to eager-load this if `?include=signup` right?

chloealee: ah yes! Thanks for catching that, @DFurnes ! I updated and deployed that to QA to test that it still worked and looks good! :)

app/Http/Controllers/PostsController.php
```diff
@@ -71,6 +71,11 @@ public function index(Request $request)
         $query = $this->newQuery(Post::class)
             ->orderBy('created_at', 'desc');
 
+        if ($request->query('include') && in_array('signup', $request->query('include'))) {
```
DFurnes: This will break if someone provides a single include, e.g. `?include=signup`. Since this can be provided as a single include, a comma-separated list of includes, or `?include[]=` array format, let's make a `has_include($request, 'signup')` helper that can handle all these cases.

chloealee: @DFurnes super smart thanks flagging! Updated :)

app/Http/Controllers/PostsController.php
```diff
@@ -71,6 +71,11 @@ public function index(Request $request)
         $query = $this->newQuery(Post::class)
             ->orderBy('created_at', 'desc');
 
+        if ($request->query('include') && has_include($request, 'signup')) {
```
DFurnes: Let's embed this `$request->query('include')` check in the helper as well, so this is just `if (has_include(...))`!

chloealee: oh yes that is nice, thanks @DFurnes !

PR #757, opened 2018-08-08
Checks that we're storing created_at in correct format when creating a Post

chloealee:
#### What's this PR do?
adds check to make sure we're storing `created_at` in correct format in the database.

👀 

#### Relevant tickets
[This](https://www.pivotaltracker.com/n/projects/2019429/stories/156165836) Pivotal Card and [this](https://github.com/DoSomething/chompy/pull/38) PR in Chompy.

#### How to deploy

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -103,7 +103,7 @@ class PostRepository
 
             // If there is a created_at property, fill this in (e.g. if created_at is sent when creating a record with the importer app).
             if (isset($data['created_at'])) {
-                $post->created_at = $data['created_at'];
+                $post->created_at = strtotime($data['created_at']);
             }
         }
 


app/Repositories/PostRepository.php
```diff
@@ -103,7 +103,7 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
 
             // If there is a created_at property, fill this in (e.g. if created_at is sent when creating a record with the importer app).
             if (isset($data['created_at'])) {
-                $post->created_at = $data['created_at'];
+                $post->created_at = strtotime($data['created_at']);
```
katiecrane: @chloealee won't this be a Unix timestamp instead of `YYYY-MM-DD HH:MM:SS`?

chloealee: @katiecrane great q! This does get stored in `created_at` section in `2018-06-15 10:39:00` format though - I think it has to do with the logic when a post is saved and what settings we have for this.

katiecrane: @chloealee oh that's cool!! if that's the case, why can't Chompy send over `strtotime` as well?

chloealee: @katiecrane that's a good q! It is because of [this](https://github.com/DoSomething/rogue/blob/master/app/Http/Requests/PostRequest.php#L40) validation here in Rogue where we validate any incoming `created_at`s to be a `date`.

katiecrane: @chloealee hmm and strings from `strtodate` aren't valid for that? I would have thought they would be based on the documentation here but maybe I'm reading it wrong: https://laravel.com/docs/5.5/validation#rule-date

chloealee: @katiecrane yeah you're right - it does seem like that! I tried in my Postman and via Chompy to send through the `created_at` time as something like `1528482420` and `"1528482420"` (in case it thought it was an integer, but I get this response: 

```
{
    "message": "The given data was invalid.",
    "errors": {
        "created_at": [
            "The created at is not a valid date."
        ]
    }
}
```

katiecrane: @chloealee Weird! I guess maybe the docs are misleading then, good 🕵️‍♀️

katiecrane: OHHHHH that makes total sense!! That's for explaining!

chloealee: I was confused too - thanks for bringing to my attention! :)

PR #760, opened 2018-08-16
Auto-accept share-social posts when they are created 

chloealee:
#### What's this PR do?
- Auto-accepts `share-social` posts when they are created (unless an admin sets a custom `status` for the post). 
- Updates/adds tests. 
- Updates documentation. 

👀 

#### Relevant tickets
Fixes the first task in [this](https://www.pivotaltracker.com/n/projects/2019429/stories/159629479) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -87,11 +87,13 @@ class PostRepository
             'action' => isset($data['action']) ? $data['action'] : null,
             'url' => $fileUrl,
             'text' => isset($data['text']) ? $data['text'] : null,
-            'status' => 'pending',
             'source' => token()->client(),
             'source_details' => isset($data['source_details']) ? $data['source_details'] : null,
             'details' => isset($data['details']) ? $data['details'] : null,
         ]);
+
+        // If this is a share-social type post, auto-accept.
+        $post->status = $post->type === 'share-social' ? 'accepted' : 'pending';
 
         $isAdmin = auth()->user() && auth()->user()->role === 'admin';
         $hasAdminScope = in_array('admin', token()->scopes());
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -103,12 +103,6 @@ class PostRepository
             // If the admin sets a custom status, set this status.
             if (isset($data['status'])) {
                 $post->status = $data['status'];
-            // If the admin doesn't set a custom status and it is a share-social post, auto-accept it.
-            } elseif (! isset($data['status']) && $post->type === 'share-social') {
-                $post->status = 'accepted';
-            // Otherwise, the status should be pending.
-            } else {
-                $post->status = 'pending';
             }
 
             $post->source = isset($data['source']) ? $data['source'] : token()->client();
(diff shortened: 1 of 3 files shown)


docs/endpoints/posts.md
```diff
@@ -158,7 +158,7 @@ POST /api/v3/posts
 - **campaign_run_id**: (int)
   The drupal campaign run node id of the campaign that the user's post is associated with.
 - **type**: (string) required.
-  The type of post submitted. Must be one of the following types: photo, voter-reg, text, share-social.
+  The type of post submitted. Must be one of the following types: `photo`, `voter-reg`, `text`, `share-social`. `share-social` posts will be auto-accepted unless an admin sets a custom `status`.
```
katiecrane: I'm not sure what this part means about an admin setting a custom status? Looks like it will be `accepted` no matter what if it comes in with type `share-social`

chloealee: @katiecrane great q! If an admin sends along a custom `status` when creating the post, the 
 custom `status` will be overridden [here](https://github.com/DoSomething/rogue/blob/master/app/Repositories/PostRepository.php#L101). Now looking at it more closely though, it seems like it will get set to `pending` if they don't set a custom status so we don't want that either! I'll update this now!

katiecrane: ah I see now! yeah good catch, does look like it would be overwritten to `pending` ⚾️

chloealee: @katiecrane thanks for catching this! I updated the logic and also added a few more tests to make sure that if an admin is creating a `share-social` post with a custom `status`, then it will take that status instead of auto-accepting.

app/Repositories/PostRepository.php
```diff
@@ -87,18 +87,30 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
...
+            // If the admin sets a custom status, set this status.
+            if (isset($data['status'])) {
+                $post->status = $data['status'];
+            // If the admin doesn't set a custom status and it is a share-social post, auto-accept it.
+            } elseif (! isset($data['status']) && $post->type === 'share-social') {
+                $post->status = 'accepted';
+            // Otherwise, the status should be pending.
+            } else {
+                $post->status = 'pending';
+            }
```
katiecrane: I'm a little confused here. Doesn't the `share-social` situation get taken care of above? And don't we want the status to be `pending` even if `$isAdmin || $hasAdminScope` is `false`?

chloealee: >Doesn't the share-social situation get taken care of above? 

This block deals with the following logic: (I'm not sure how often this will happen since admins probably can't upload `share-social` posts for users) if an admin sets a custom status for a `share-social` post, accept that status instead of auto-accepting like we would normally do. If we didn't update this, the old logic would change the `status` back to `pending` after auto-accepting if the admin doesn't set a custom status.

>And don't we want the status to be pending even if $isAdmin || $hasAdminScope is false

We want to auto-accept `share-social` posts when they are created!

@katiecrane  lmk if that still doesn't make sense or if you have more q's!

PR #761, opened 2018-08-28
Adds ability to filter by `created_at` for the v3/posts index 

chloealee:
#### What's this PR do?
Adds ability to filter by `created_at` for the v3/posts index.
 
👀 
Try to use `api/v3/posts?filter[created_at]=2018-08-13 15:10:03` and only posts that were created at `2018-08-13 15:10:03` should be returned. 
 
We want to update the [Import Facebook Share Posts](https://github.com/DoSomething/chompy/blob/master/app/Jobs/ImportFacebookSharePosts.php)  job  in Chompy to first check if a post exists in Rogue before creating it. 

In order to make sure that the job doesn't create duplicate posts, we will search by `campaign_id`, `northstar_id`, `action`, and `created_at` date. This adds the ability to do so. 

#### Relevant tickets
Related to [this](https://www.pivotaltracker.com/n/projects/2019429/stories/160102193) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -53,7 +53,7 @@ class Post extends Model
      *
      * @var array
      */
-    public static $indexes = ['id', 'signup_id', 'campaign_id', 'type', 'action', 'northstar_id', 'status'];
+    public static $indexes = ['id', 'signup_id', 'campaign_id', 'type', 'action', 'northstar_id', 'status', 'created_at'];
 
     /**
      * The tags prefixed with 'good' that will send a post to Slack.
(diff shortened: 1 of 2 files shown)


katiecrane:
@chloealee Just curious! Is this functionality that we'd want in Rogue for other cases? Wondering if there are other benefits of this over having Chompy get posts by `northstar_id`, `action`, `campaign_id` and then do the timestamp checking itself. I guess it's easier overall to do it this way at least?

chloealee:
@katiecrane great q! So actually, we'll be finding posts by all of those - `northstar_id`, `action`, `campaign_id`, AND `created_at` - we'll need all of these fields to be sure that there aren't any dupes (a user can have many `share-social` posts with the same `action` under the same campaign so that is why we also need the `created_at`.

chloealee:
@katiecrane oh I see what you're saying now - interesting! We def can do that if we don't want to add this to Rogue? I am leaning more towards this way so we don't add more complexity to the jobs in Chompy and have it take longer?

katiecrane:
@chloealee totally works for me! was pure curiosity :)

PR #762, opened 2018-08-29
creates script to force delete posts by type and source

chloealee:
#### What's this PR do?
- creates script to force delete posts by type and source

- Run this query before running script to see how many posts we start with that have the `type` = `share-social` and created by chompy: 
```
SELECT COUNT(*) FROM posts where type = 'share-social' and source = 'importer-client'; 
```
- Run the script: `php artisan rogue:forceDeletePosts share-social importer-client`
- Run the query again and it should be `0`

- We started importing Historic FB shares from chompy on prod but realized there was a bug and it created dupes! 
- We now plan to get rid of all the dupes and re-start the script. 
- This script takes in the arguments `type` and `source` since this is what we need in this case but can be modified in the future if we need it. 
- I decided to not expand on [this command](https://github.com/DoSomething/rogue/blob/master/app/Console/Commands/DeletePosts.php) that soft deletes posts based on ID because it seems like, although similar, have different case scenarios for what data is available to us. 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/160102193) Pivotal card.

#### How to deploy

### DIFF
--- a/app/Console/Commands/ForceDeletePosts.php
+++ b/app/Console/Commands/ForceDeletePosts.php
@@ -4,6 +4,7 @@ namespace Rogue\Console\Commands;
 
 use Rogue\Models\Post;
 use Illuminate\Console\Command;
+use Rogue\Managers\PostManager;
 
 class ForceDeletePosts extends Command
 {
@@ -28,8 +29,11 @@ class ForceDeletePosts extends Command
      *
      * @return void
      */
-    public function __construct()
+    public function __construct(PostManager $postManager)
     {
+        // Post Manager Instance.
+        $this->postManager = $postManager;
+
         parent::__construct();
     }
 
@@ -42,8 +46,18 @@ class ForceDeletePosts extends Command
     {
         info('Starting to force delete posts by type and source.');
 
-        Post::where('type', $this->argument('type'))->where('source', $this->argument('source'))->forceDelete();
+        $posts = Post::where('type', $this->argument('type'))->where('source', $this->argument('source'))->get();
 
-        info('All posts have been force deleted.');
+        if ($posts->isNotEmpty()) {
+            $posts->map(function ($post, $key) {
+                $this->postManager->destroy($post->id);
+
+                $post->forceDelete();
+            });
+
+            info('Posts Deleted!');
+        } else {
+            info('No Posts found');
+        }
     }
 }
--- a/app/Console/Commands/ForceDeletePosts.php
+++ b/app/Console/Commands/ForceDeletePosts.php
@@ -44,9 +44,9 @@ class ForceDeletePosts extends Command
      */
     public function handle()
     {
-        if ($this->confirm('"Are you sure you want to delete posts of type ' . $this->argument('type') . 'from ' . $this->argument('source'))) {
+        if ($this->confirm('Are you sure you want to delete posts of type ' . $this->argument('type') . 'from ' . $this->argument('source'))) {
 
-            info('Starting to force delete posts by type and source.');
+            info('rogue:forceDeletePosts: Starting to force delete posts by type and source.');
 
             $posts = Post::where('type', $this->argument('type'))->where('source', $this->argument('source'))->get();
 
@@ -57,9 +57,9 @@ class ForceDeletePosts extends Command
                     $post->forceDelete();
                 });
 
-                info('Posts Deleted!');
+                info('rogue:forceDeletePosts: Posts Deleted!');
             } else {
-                info('No Posts found');
+                info('rogue:forceDeletePosts: No Posts found');
             }
         }
     }
--- a/app/Console/Commands/ForceDeletePosts.php
+++ b/app/Console/Commands/ForceDeletePosts.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Rogue\Console\Commands;
+
+use Rogue\Models\Post;
+use Illuminate\Console\Command;
+
+class ForceDeletePosts extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'rogue:forceDeletePosts
+                            {type}
+                            {source}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Force deletes posts in Rogue by type and source';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        info('Starting to force delete posts by type and source.');
+
+        Post::where('type', $this->argument('type'))->where('source', $this->argument('source'))->forceDelete();
+
+        info('All posts have been force deleted.');
+    }
+}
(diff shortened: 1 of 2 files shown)


app/Repositories/PostRepository.php
```diff
@@ -163,12 +163,14 @@ public function destroy($postId)
...
+            // Delete the image file from AWS.
+            $this->aws->deleteImage($post->url);
 
-        // Set the url of the post to null.
-        $post->url = null;
-        $post->save();
+            // Set the url of the post to null.
+            $post->url = null;
+            $post->save();
+        }
```
katiecrane: This is a really good catch!!!!

app/Console/Commands/ForceDeletePosts.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        info('Starting to force delete posts by type and source.');
```
katiecrane: Before logging, I think it'd be nice to have this ask the user for confirmation since we're wiping stuff out, something like: "Are you sure you want to delete posts of type {type} from {source}?" There are implementation details [here](https://laravel.com/docs/5.6/artisan#command-io) under "Asking For Confirmation."

chloealee: i love this, thanks for sharing @katiecrane !! Updated :)

app/Console/Commands/ForceDeletePosts.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        if ($this->confirm('"Are you sure you want to delete posts of type ' . $this->argument('type') . 'from ' . $this->argument('source'))) {
```
katiecrane: I think you have an extra `"` here!

app/Console/Commands/ForceDeletePosts.php
```diff
@@ -0,0 +1,66 @@
...
+
+            info('Starting to force delete posts by type and source.');
```
katiecrane: I think in other commands we prefix the log messages with the command like `rogue:forceDeletePosts`, what are your thoughts on that?

app/Console/Commands/ForceDeletePosts.php
```diff
@@ -0,0 +1,66 @@
...
+
+            $posts = Post::where('type', $this->argument('type'))->where('source', $this->argument('source'))->get();
+
+            if ($posts->isNotEmpty()) {
+                $posts->map(function ($post, $key) {
```
katiecrane: Not a suggestion, I'm just personally curious why you went with `map` instead of a `forEach`?

app/Console/Commands/ForceDeletePosts.php
chloealee: oops !

app/Console/Commands/ForceDeletePosts.php
chloealee: yes great catch - love this for consistency! Updated :)

app/Console/Commands/ForceDeletePosts.php
chloealee: great q! TBH, I was just doing what Shae had done [here](https://github.com/DoSomething/rogue/blob/master/app/Console/Commands/DeletePosts.php#L51) :)

app/Console/Commands/ForceDeletePosts.php
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct(PostManager $postManager)
+    {
+        // Post Manager Instance.
+        $this->postManager = $postManager;
```
DFurnes: Nitpicky, but if we're going to document this let's add the property & docblock (rather than creating implicitly)!

```php
/**
 * The post manager.
 * @var PostManager
 */
protected $postManager;
```

app/Console/Commands/ForceDeletePosts.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        if ($this->confirm('Are you sure you want to delete posts of type ' . $this->argument('type') . 'from ' . $this->argument('source'))) {
```
DFurnes: It'd be nice to write this as a guard clause so we follow the [golden path](https://en.wikipedia.org/wiki/Happy_path) (rather than indenting the rest of this function)!

```php
if (! $this->confirm('Are you sure you want to delete posts of type ' . $this->argument('type') . 'from ' . $this->argument('source'))) {
    return;
}

info('rogue:forceDeletePosts: Starting to force delete posts by type and source.');
$posts = Post::where(...);
```

app/Console/Commands/ForceDeletePosts.php
```diff
@@ -0,0 +1,66 @@
...
+
+            info('rogue:forceDeletePosts: Starting to force delete posts by type and source.');
+
+            $posts = Post::where('type', $this->argument('type'))->where('source', $this->argument('source'))->get();
+
+            if ($posts->isNotEmpty()) {
```
DFurnes: Same thought with using a guard expression here (since no posts being found is not the expected "path").

app/Console/Commands/ForceDeletePosts.php
DFurnes: Yeah, this should use [`forEach`](https://laravel.com/docs/5.6/collections#method-forEach) since we're not returning anything. You'd use [`map`](https://laravel.com/docs/5.6/collections#method-map) to iterate over a collection & modify each element (and so since we're not returning anything, this line technically makes and returns a new empty collection).

app/Console/Commands/ForceDeletePosts.php
chloealee: ohhh I love this @DFurnes ! I've never heard of the golden/happy path but I love this idea and is a lot easier to read. Thanks for sharing! Updated :)

app/Console/Commands/ForceDeletePosts.php
```diff
@@ -0,0 +1,73 @@
...
+        if (! $this->confirm('Are you sure you want to delete posts of type ' . $this->argument('type') . 'from ' . $this->argument('source'))) {
+            return;
+        }
+
+        info('rogue:forceDeletePosts: Starting to force delete posts by type and source.');
+
+        $posts = Post::where('type', $this->argument('type'))->where('source', $this->argument('source'))->get();
+
+        if ($posts->isNotEmpty()) {
+            $posts->map(function ($post, $key) {
```
DFurnes: Let's swap this to use `forEach` like @katiecrane suggested!

chloealee: ok! @DFurnes just curious - why is it better to use a `forEach` statement here instead of `map`?

DFurnes: This should use [`forEach`](https://laravel.com/docs/5.6/collections#method-forEach) since we're not returning anything. You'd use [`map`](https://laravel.com/docs/5.6/collections#method-map) to iterate over a collection & modify each element (and so since we're not returning anything, this line technically makes and returns a new empty collection).

chloealee: ah I see - good to know, thanks! Updated :)

PR #764, opened 2018-08-31
adds chunking to force delete posts script

chloealee:
#### What's this PR do?
Adds `chunk` to force delete posts script to save memory when dealing with a lot of records. 

👀 

When running on thor, there were 35,550 records that needed to be force deleted. This cause a memory issue: 
screen shot 2018-08-31 at 10 37 25 am

Using `chunk` should help this problem. 

#### Relevant tickets
Updates script created in #762 

#### How to deploy

### DIFF
--- a/app/Console/Commands/ForceDeletePosts.php
+++ b/app/Console/Commands/ForceDeletePosts.php
@@ -52,18 +52,21 @@ class ForceDeletePosts extends Command
     {
         info('rogue:forceDeletePosts: Starting to force delete posts by type and source.');
 
-        $posts = Post::where('type', $this->argument('type'))->where('source', $this->argument('source'))->get();
+        Post::where('type', $this->argument('type'))
+            ->where('source', $this->argument('source'))
+            ->chunk(100, function ($posts) {
+                if ($posts->isNotEmpty()) {
+                    foreach ($posts as $post) {
+                        $this->postManager->destroy($post->id);
 
-        if ($posts->isNotEmpty()) {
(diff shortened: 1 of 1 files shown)


app/Managers/PostManager.php
```diff
@@ -135,7 +135,7 @@ public function destroy($postId)
         $trashed = $this->repository->destroy($postId);
 
         if ($trashed) {
-            $this->fastly->purgeKey('post-'.$postId);
+            // $this->fastly->purgeKey('post-'.$postId);
```
DFurnes: This probably shouldn't be commented, right?

chloealee: oops yes - commented out for testing on local ! Thanks for catching that! Commented back in :)

app/Console/Commands/ForceDeletePosts.php
```diff
@@ -52,18 +52,21 @@ public function handle()
...
+                    }
 
-                $post->forceDelete();
-            }
-
-            info('rogue:forceDeletePosts: Posts Deleted!');
-        } else {
-            info('rogue:forceDeletePosts: No Posts found');
-        }
+                    info('rogue:forceDeletePosts: Posts Deleted!');
```
katiecrane: Maybe this should say "100 posts deleted" or something like that? Since this logs after a chunk is done.

chloealee: ah that is a great idea! Updated :)

PR #771, opened 2018-10-09
Increases image throttling to allow 30 images per minute

chloealee:
#### What's this PR do?
Increases image throttling to allow 30 images per minute.

👀 

Papertrail Rate-Limited Requests alerts were very noisy because of this threshold. @DFurnes had a [great idea](https://dosomething.slack.com/archives/C0YGXUE01/p1539096182000100?thread_ts=1539091395.000100&cid=C0YGXUE01) to increase this throttle to avoid false alarms. 

#### How to deploy


DFurnes:
Oh, just need to update the feature test for this before merging. And let's keep an eye on whether this negatively impacts application performance over the next few weeks in New Relic.

PR #773, opened 2018-10-10
Adds new(ish) tags to filter bar

chloealee:
#### What's this PR do?
Adds new(ish) tags to filter bar: 
  - Irrelevant 
  - Inappropriate
  - Unrealistic Quantity
  - Test
  - Incomplete Action 

- Are these all showing up in the filter bar? 
- When clicked, does the filter bar filter correctly? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/154973569) Pivotal Card.

#### How to deploy

### DIFF
--- a/resources/assets/components/CampaignSingle/index.js
+++ b/resources/assets/components/CampaignSingle/index.js
@@ -55,6 +55,26 @@ class CampaignSingle extends React.Component {
             label: 'Bulk',
             active: false,
           },
+          irrelevant: {
+            label: 'Irrelevant',
+            active: false,
+          },
+          inappropriate: {
+            label: 'Inappropriate',
+            active: false,
+          },
+          'unrealistic-quantity': {
+            label: 'Unrealistic Quantity',
+            active: false,
+          },
+          test: {
+            label: 'Test',
+            active: false,
+          },
+          'incomplete-action': {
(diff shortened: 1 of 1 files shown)


resources/assets/components/CampaignSingle/index.js
```diff
@@ -55,6 +55,26 @@ class CampaignSingle extends React.Component {
...
           },
+          irrelevant: {
+            label: 'Irrelevant',
+            active: false,
+          },
+          inappropriate: {
+            label: 'Inappropriate',
+            active: false,
+          },
+          'unrealistic-quantity': {
```
chloealee: this is weird to me but I think this is part of the automatic styling to have `'` when there is more than one word in the tag! It wouldn't let me add `'` for the single words!

katiecrane: this is strange!!

DFurnes: Yep, this is how Prettier formats objects (no quotes unless necessary). In most cases, we use "plain" words for keys (like `label` and `active` above) and so quotes are just noise, but `unrealistic-quantity` would be a syntax error without the quotes.

PR #776, opened 2018-10-25
Backend work for Rogue to create campaign IDs

chloealee:
#### What's this PR do?
- Fleshes out backend work for Rogue to create campaign IDs and be the main source of truth for campaigns (once we migrate everything from Ashes to Rogue!). 
  - Adds new `campaigns` table 
  - Renames CampaignsController to AshesCampaignsController whenever it is grabbing campaign data from Ashes
  - Builds out new Campaign Model and Campaign Controller:
    - New Campaign Controller that will be used for all new CRUD activity. 
      - This includes the following new routes:
        - `GET /api/v3/campaigns` index 
        - `GET /api/v3/campaigns/{campaign}`
        - `POST /campaigns`
        - `PATCH /campaigns/{campaign}`
        - `DELETE /campaigns/{campaign}`
        - _Note:_ I decided not to put the last three routes in `api/v3` because they should only be used internally in Rogue admin, for now? 
  - Adds new Campaign Transformer 
  - Updates documentation 
  - Adds tests for each of these new endpoints 
- **Notes:**
  - Place holders for now to redirect to campaign pages for frontend PR! 
  - Right now, these endpoints will only reveal/manipulate data that is in the Rogue database, not legacy campaigns. 
  - Rogue admin is still hooked up to getting campaign data from Ashes. We'll deal with the transition to Rogue as main source for campaign data and to get off Ashes in another PR I assume @ngjo ? 

Does the structure of the api routes/controller make sense? Anything I missed? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/161257914) Pivotal Card.

#### How to deploy


app/Http/Controllers/CampaignsController.php
```diff
@@ -0,0 +1,108 @@
...
+     *
+     */
+    public function __construct()
+    {
+        $this->transformer = new CampaignTransformer;
+
+        // $this->middleware('auth');
+        // $this->middleware('role:admin,staff', ['only' => ['store', 'update', 'destroy']]);
+        // $this->middleware('scopes:write', ['only' => ['store', 'update', 'destroy']]);
+    }
```
chloealee: I don't think we need the `auth` middleware but we should require that someone is an `admin`/`staff` and has the `write` scope to be able to create a campaign, update a campaign, or delete a campaign. However, I keep getting `403` status code and the `You don't have the correct role to do that!` message when I comment back in lines 25 and 26 but I checked my token to confirm the role is `admin` and has the `write` scope... is there something I'm missing?

app/Http/Controllers/CampaignsController.php
```diff
@@ -0,0 +1,108 @@
...
+        if (! $campaign) {
+            $campaign = new Campaign;
+
+            $campaign->internal_title = $request['internal_title'];
+            $campaign->start_date = $request['start_date'];
+            $campaign->end_date = $request['end_date'];
+            $campaign->save();
+        }
+
+        // @TODO: return redirect()->route('campaigns.show', $campaign->id);
```
chloealee: I don't think it is necessary but just want to double check - do we want to log anything when a campaign is created, updated or deleted?

app/Http/Controllers/CampaignsController.php
katiecrane: @chloealee what does `auth` do? Makes sure someone is logged in?

chloealee: oh yes that is what it is @katiecrane ! Ok so I think we do need `auth` then but only for `store`, `update`, and `destroy`, same with the other middleware but still can't get this to work :-/ Now I'm getting `"message": "Unauthenticated."`

ngjo:
> Rogue admin is still hooked up to getting campaign data from Ashes. We'll deal with the transition to Rogue as main source for campaign data and to get off Ashes in another PR I assume @ngjo ?

Yes! Here's the card for that here https://www.pivotaltracker.com/story/show/161494931 -- great flag!

database/factories/ModelFactory.php
```diff
@@ -95,3 +96,14 @@
 $factory->defineAs(User::class, 'admin', function () use ($factory) {
     return array_merge($factory->raw(User::class), ['role' => 'admin']);
 });
+
+// Campaign Factory
+$factory->define(Campaign::class, function (Generator $faker) {
+
+    return [
+        'id' => $faker->randomNumber(4),
```
chloealee: decided to use this instead of `$faker->addProvider(new FakerCampaignId($faker));` because those only have 3 campaigns. We can add more but I figured this is more likely of a case to use a random 4 digit number instead when we create new campaigns in Rogue.

chloealee: @katiecrane ooo I'm sure this is possible! But at the same time, maybe I'll just change to `randomNumber()`? I think the number of digits the number is is sorta arbitrary because AFAIK, we don't have any validation/requirements about a campaign id before 4 digits?

katiecrane: @chloealee as far as I know that is correct!

chloealee: @katiecrane updated! :)

app/Http/Controllers/CampaignsController.php
katiecrane: @chloealee were you able to actually hit the endpoint though?

app/Http/Controllers/CampaignsController.php
```diff
@@ -0,0 +1,108 @@
...
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Rogue\Models\Campaign  $campaign
+     * @return \Illuminate\Http\Response
+     */
+    public function update(Request $request, Campaign $campaign)
+    {
+        $campaign->update($request->all());
+
+        // @TODO: return redirect()->route('campaigns.show', $campaign->id);
```
katiecrane: @chloealee you got a `@TODO:` here and in the next function, are these to-done?

app/Http/Controllers/CampaignsController.php
chloealee: @katiecrane if I comment out the middleware, the logic works in Postman! I'm not sure if this is sufficient but I was planning on fully testing this when I build out the front end and via a "Create Campaign" button.

database/factories/ModelFactory.php
```diff
@@ -95,3 +96,14 @@
 $factory->defineAs(User::class, 'admin', function () use ($factory) {
     return array_merge($factory->raw(User::class), ['role' => 'admin']);
 });
+
+// Campaign Factory
+$factory->define(Campaign::class, function (Generator $faker) {
+
+    return [
+        'id' => $faker->randomNumber,
+        'internal_title' => $faker->sentence(),
```
katiecrane: can't wait to see what it comes up with for these

app/Http/Controllers/CampaignsController.php
chloealee: @katiecrane this is a `@TODO` to be fleshed out when I do the front end stuff !

katiecrane:
@chloealee I think it'd be nice to have a seeder to create a few campaigns for local development!

app/Http/Controllers/CampaignsController.php
katiecrane: @chloealee hmm in the meantime would that block a prod deploy?

chloealee: @katiecrane I'm don't think so because nothing is really using it yet? But this can totally also wait to be deployed with the front end when that is done too if that makes more sense and so we can totally test this works with 💯 percent confidence!

chloealee:
@katiecrane great idea! Just added the seeder to create 10 campaigns!

app/Http/Controllers/CampaignsController.php
katiecrane: @chloealee you're gonna need it to do the front end part though! I think it's okay to leave a note in here about what might not be working/needs to be tested since this whole campaigns thing is a WIP, as long as we can still deploy to prod

app/Http/Controllers/CampaignsController.php
katiecrane: @chloealee hmm not if we don't do that for posts and signups I don't think?

app/Http/Controllers/CampaignsController.php
katiecrane: @chloealee AH gotcha!

tests/Http/CampaignTest.php
```diff
@@ -0,0 +1,132 @@
...
+        $this->actingAsAdmin()->postJson('campaigns', [
+            'internal_title' => $firstCampaignTitle,
+            'start_date' => $this->faker->dateTime,
+            'end_date' => $this->faker->dateTime,
+        ]);
+
+        $response = $this->getJson('api/v3/campaigns');
+        $decodedResponse = $response->decodeResponseJson();
+
+        $this->assertEquals(1, $decodedResponse['meta']['pagination']['count']);
```
katiecrane: @chloealee oooo I know you were having trouble hitting the endpoints but do these tests pass?

chloealee: @katiecrane they do!

routes/web.php
```diff
@@ -20,6 +20,10 @@
 $router->get('campaigns', 'Legacy\Web\CampaignsController@index');
 $router->get('campaigns/{id}/inbox', 'Legacy\Web\CampaignsController@showInbox');
 $router->get('campaigns/{id}', 'Legacy\Web\CampaignsController@showCampaign')->name('campaigns.show');
+// Create, update, delete campaigns via Rogue.
+$router->post('campaigns', 'CampaignsController@store');
+$router->patch('campaigns/{campaign}', 'CampaignsController@update');
+$router->delete('campaigns/{campaign}', 'CampaignsController@destroy');
```
katiecrane: @chloealee will we want `get` here?

database/migrations/2018_10_23_203754_create_campaigns_table.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('campaigns', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('internal_title')->comment('Campaign title used internally.');
```
katiecrane: @chloealee I saw elsewhere that you required this to be unique, should that be enforced here as well?

app/Http/Controllers/CampaignsController.php
chloealee: @katiecrane we do for [signups](https://github.com/DoSomething/rogue/blob/master/app/Managers/SignupManager.php#L53) and [posts](https://github.com/DoSomething/rogue/blob/master/app/Managers/PostManager.php#L65) but I feel like that is more so to make sure everything is working ok through the ecosystem (from Phoenix->Rogue etc). 

For campaigns, I'm not sure this is necessary to log since we can see in the database the timestamps if a campaign is created?

katiecrane: @chloealee eh, my thought is that there are way less campaigns than signups so couldn't hurt to log this information

tests/Http/CampaignTest.php
katiecrane: @chloealee that makes it seem like something is just up with postman to me!

routes/web.php
chloealee: @katiecrane good q! The `GET /api/v3/campaigns` and `GET /api/v3/campaigns/:campaign_id` routes are all in `routes/api.php`!

tests/Http/CampaignTest.php
chloealee: @katiecrane yes, agreed! :)

database/migrations/2018_10_23_203754_create_campaigns_table.php
katiecrane: @chloealee yes!

chloealee: @katiecrane oh great call ! Updated :)

database/factories/ModelFactory.php
chloealee: @katiecrane also you're right - this is needed when we seed the `campaigns` table with random data because we probably don't want the seeder to create campaigns with the id under `9000` for when we backfill campaign data from Ashes and in case we want the entire database in our local !

routes/web.php
DFurnes: Can we put these in a separate controller inside [`app/Http/Controllers/Legacy/Web`](https://github.com/DoSomething/rogue/tree/master/app/Http/Controllers/Legacy/Web)? I don't think it will actually cause trouble, since the guard is applied by the middleware group here, but it's a little disorienting that some of the methods in the controller use OAuth and some use HTTP sessions.

database/migrations/2018_10_23_203754_create_campaigns_table.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('campaigns', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('internal_title')->unique()->comment('Campaign title used internally.');
```
DFurnes: I don't think we need to have a unique constraint on this field (at least, not ensured by the database).

app/Http/Controllers/CampaignsController.php
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Update the specified resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Rogue\Models\Campaign  $campaign
+     * @return \Illuminate\Http\Response
+     */
+    public function update(Request $request, Campaign $campaign)
+    {
+        $campaign->update($request->all());
```
DFurnes: Let's add some validation above this line, like we do for the `store` method!

app/Http/Controllers/CampaignsController.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Display the specified resource.
+     *
+     * @param  \Rogue\Models\Campaign  $campaign
+     * @return \Illuminate\Http\Response
+     */
+    public function show(Campaign $campaign)
+    {
+        return $this->item($campaign, 200, [], $this->transformer);
```
DFurnes: You can omit these parameters, since they're all the defaults!

```suggestion
        return $this->item($campaign);
```

app/Http/Controllers/CampaignsController.php
```diff
@@ -0,0 +1,118 @@
...
+    public function store(Request $request)
+    {
+        $this->validate($request, [
+            'internal_title' => 'required|string',
+            'start_date' => 'required|date',
+            'end_date' => 'required|date',
+        ]);
+
+        // Check to see if the campaign exists before creating one.
+        $campaign = Campaign::where([
```
DFurnes: You can handle this with a validation rule! On `internal_title`, above:

```php
        $this->validate($request, [
```

app/Http/Controllers/CampaignsController.php
```diff
@@ -0,0 +1,118 @@
...
+            'internal_title' => $request['internal_title'],
+        ])->first();
+
+        // If there is no campaign with that title, create one.
+        if (! $campaign) {
+            $campaign = new Campaign;
```
DFurnes: Instead of making a new class & manually applying the fields, let's use `Campaign::create` & mass-assignment:

```
$campaign = Campaign::create($request->all());
```

app/Http/Controllers/CampaignsController.php
```diff
@@ -0,0 +1,118 @@
...
+     * Create a controller instance.
+     *
+     */
+    public function __construct()
+    {
+        $this->transformer = new CampaignTransformer;
+
+        $this->middleware('auth', ['only' => ['store', 'update', 'destroy']]);
+        $this->middleware('role:admin,staff', ['only' => ['store', 'update', 'destroy']]);
+        $this->middleware('scopes:write', ['only' => ['store', 'update', 'destroy']]);
```
DFurnes: Oh, this is another good argument for breaking the web methods into a separate controller actually. We only want to check scopes on API endpoints, since a plain HTTP session doesn't have any scopes to check!

database/factories/ModelFactory.php
DFurnes: Oh man, wouldn't it be fun to make a Markov chain generator based on real campaign names? 😂

routes/web.php
chloealee: @DFurnes ah ok! Since there already is a `CampaignsController` in `app/Http/Controllers/Legacy/Web` with an `index` and `showCampaign` method, should I create a new one and call it `CampaignsController` and maybe rename the existing `CampaignsController` to `AshesCampaignsController`?

DFurnes: @chloealee I think that's a good plan!

database/migrations/2018_10_23_203754_create_campaigns_table.php
chloealee: @katiecrane had suggested this - strong thoughts on keeping this in, @katiecrane ?

app/Http/Controllers/CampaignsController.php
chloealee: oh cool! So if it already exists, will it throw a validation error?

DFurnes: Yep!

app/Models/Campaign.php
```diff
@@ -0,0 +1,33 @@
...
+     * @var array
+     */
+    protected $dates = ['deleted_at'];
+
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var array
+     */
+    protected $fillable = ['id', 'internal_title', 'start_date', 'end_date', 'created_at', 'updated_at'];
```
DFurnes: Another thing that just occurred to me. Let's not have all these mass-assignable:

```suggestion
    protected $fillable = ['internal_title', 'start_date', 'end_date'];
```

If we want to edit other ones (say, setting a specific ID in an import script), we can do that by manual assignment.

app/Http/Controllers/CampaignsController.php
chloealee: very cool!

app/Http/Controllers/CampaignsController.php
chloealee: oh right! Deleted the `scopes` middleware.

database/migrations/2018_10_23_203754_create_campaigns_table.php
chloealee: @katiecrane thanks for the input and totally agree! I'll take it out since neither of us having strong feelings cc @DFurnes

chloealee: cool that makes sense, I'll just keep it removed then! :)

app/Models/Campaign.php
```diff
@@ -0,0 +1,33 @@
...
+     * @var array
+     */
+    protected $dates = ['deleted_at'];
+
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var array
+     */
+    protected $fillable = ['internal_title', 'start_date', 'end_date', 'created_at', 'updated_at'];
```
DFurnes: Let's not make `created_at` or `updated_at` fillable here (since they're set automatically by Eloquent). We can set them explicitly (e.g. `$campaign->created_at = …`) if we want to keep these when importing campaigns via a script.

chloealee: oh oops - thanks @DFurnes ! Updated that!

docs/endpoints/campaigns.md
```diff
@@ -0,0 +1,75 @@
...
+        {
+            "id": 9000,
+            "internal_title": "Test campaign",
+            "start_date": "2018-10-10 00:00:00",
+            "end_date": "2018-12-10 00:00:00",
+            "created_at": {
+                "date": "2018-10-24 19:02:06.000000",
+                "timezone_type": 3,
+                "timezone": "UTC"
+            },
```
aaronschachter: The date formats in this example are inconsistent. We started pullling this data over in https://github.com/DoSomething/gambit-content/pull/1105 --  I've noticed that the current `end_date` property is returned as an object, not a string. We've previously gotten burned on the Gambit side from an [unannounced format change to end date](https://github.com/DoSomething/gambit-content/pull/1097/files).

Does it make sense to return these dates as strings instead of objects, to keep consistent with formatting of date properties in the other endpoints?

chloealee: PR [here](https://github.com/DoSomething/rogue/pull/803)!

PR #777, opened 2018-11-05
Campaign creation frontend

chloealee:
#### What's this PR do?
- Adds front end to creating campaign ids in Rogue. 
  - Adds "Campaign IDs" to the nav bar. When you click on that, it brings you to `/campaign-ids`:
     
screen shot 2018-11-05 at 6 15 04 pm
  - Click on "NEW CAMPAIGN ID" and it brings you to /campaign-ids/create`:
screen shot 2018-11-05 at 6 16 22 pm
  - Fill the form out to create a new campaign. Validation rules:
    - You cannot create a campaign that has an `internal_title` already in the database. This is also required. 
    - `start_date` is required and must be a valid date. 
    - `end_date` is not required but if it is, it must be a valid date. Also, `end_date` must be after `start_date`
  - Once you create the campaign, it redirects you to `campaign-ids/{campaign}` of the newly created campaign:
screen shot 2018-11-05 at 6 19 18 pm
  - From there, you can click on "Edit this campaign" and it will take you to `campaign-ids/{campaign}/edit`: 
screen shot 2018-11-05 at 6 20 30 pm
    - Same validation rules as the create page. 
  - Finally, when you go back to the `/campaign-ids` page, your campaign should be at the bottom of the table with all the correct information! 

- Do the screenshots above look like the one below? Anything I'm missing? 
campaign ids 1
campaign ids 2
campaign ids 3

@DoSomething/team-bleed two things:

1. The mocks have `Campaign IDs` replacing `Import CSV` in the nav bar. Should I also kill the ImportController and all routes/code associated with this? Wanted to double check before 🔪 ! 
2. This PR doesn't include front end testing but I think it would be a good idea to add this. @ngjo is it ok if I tackle in another PR since this one is quite large already? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/161258192) Pivotal Card.

#### How to deploy


katiecrane:
> 1. The mocks have Campaign IDs replacing Import CSV in the nav bar. Should I also kill the ImportController and all routes/code associated with this? Wanted to double check before 🔪 !

I think yes but in a separate PR!

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -78,4 +97,66 @@ public function destroy(Campaign $campaign)
...
+
+    /**
+     * Helper function to format start and end dates for frontend.
+     *
+     * @param $campaign
+     * @param $end_date_placeholder
+     */
+    public function formatStartEndDates($campaign, $end_date_placeholder)
+    {
+        $campaign->start_date = date("m/d/Y", strtotime($campaign->start_date));
```
katiecrane: I'm wondering if the piece of this function where we get the display date should live on the model? We have something similar with [getting a user's display name in Northstar](https://github.com/DoSomething/northstar/blob/master/app/Models/User.php#L335-L349)

ngjo:
> This PR doesn't include front end testing but I think it would be a good idea to add this. @ngjo is it ok if I tackle in another PR since this one is quite large already?

I think that's super smart, @chloealee ! Feel free to break up the code into however many PRs you need. And in terms of tests, I would definitely defer to you on including critical tests as a part of the definition of "done" for the work. Nice to have tests we can create backlog cards for and prioritize those later on tho!

chloealee:
great, thanks @ngjo! I added front end tests as a task on the [card](https://www.pivotaltracker.com/n/projects/2019429/stories/161258192) and then a [separate card](https://www.pivotaltracker.com/n/projects/2019429/stories/161754358) to kill Importer stuff since that isn't as pressing!

app/Http/Controllers/Legacy/Web/CampaignsController.php
chloealee: I love that idea @katiecrane - updated!

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -18,56 +19,74 @@ public function __construct()
         $this->middleware('role:admin,staff', ['only' => ['store', 'update', 'destroy']]);
     }
 
+    /**
+     * Show overview of campaigns and their IDs.
+     */
+    public function index()
+    {
+        $campaigns = Campaign::all();
```
DFurnes: We probably want to [paginate](https://laravel.com/docs/5.7/pagination) this index page.

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -18,56 +19,74 @@ public function __construct()
...
+
+        // Format start and end dates how we want them to be viewed.
+        foreach ($campaigns as $campaign) {
+            $campaign = $this->formatStartEndDates($campaign, '');
```
DFurnes: This should be done in the template (no need to iterate over and modify the models in-memory here).

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -18,56 +19,74 @@ public function __construct()
...
         $this->validate($request, [
             'internal_title' => 'required|string|unique:campaigns',
             'start_date' => 'required|date',
-            'end_date' => 'required|date',
+            'end_date' => 'nullable|date|after:start_date',
         ]);
 
-        $campaign = Campaign::create($request->all());
+        // Change dates to YYYY-MM-DD format so it will save in the database.
+        $campaignDetails = $this->formatDatesForDatabase($request->all());
```
DFurnes: This should be done with a [date mutator](https://laravel.com/docs/5.7/eloquent-mutators#date-mutators) on the model.

app/Models/Campaign.php
```diff
@@ -30,4 +30,20 @@ public function signup()
     {
         return $this->hasMany(Signup::class);
     }
+
+    /**
+     * Get the display start date for the campaign.
+     */
+    public function displayStartDate()
+    {
+        return date("m/d/Y", strtotime($this->start_date));
```
DFurnes: Dates should be returned by Eloquent as Carbon objects (assuming they're set as [dates](https://laravel.com/docs/5.7/eloquent-mutators#date-mutators) on the model), so you don't need a separate method to handle this. Instead, in the Blade template:

```blade
<td class="table__cell">{{ $campaign->start_date->format('m/d/Y') }}</td>
<td class="table__cell">{{ $campaign->end_date ? $campaign->end_date('m/d/Y') : '–' }}</td>
```

database/migrations/2018_11_05_164652_make_end_date_nullable_in_campaigns_table.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('campaigns', function (Blueprint $table) {
+            $table->dateTime('end_date')->comment('The end date of the campaign.');
```
DFurnes: This "down" migration should always undo the change in the "up" migration:

```suggestion
            $table->dateTime('end_date')->nullable(false)->change();
```

routes/web.php
```diff
@@ -21,9 +21,14 @@
 $router->get('campaigns/{id}/inbox', 'Legacy\Web\AshesCampaignsController@showInbox');
 $router->get('campaigns/{id}', 'Legacy\Web\AshesCampaignsController@showCampaign')->name('campaigns.show');
 // Create, update, delete campaigns via Rogue.
-$router->post('campaigns', 'Legacy\Web\CampaignsController@store');
-$router->patch('campaigns/{campaign}', 'Legacy\Web\CampaignsController@update');
-$router->delete('campaigns/{campaign}', 'Legacy\Web\CampaignsController@destroy');
+// @TODO: rename these routes (to just /campaigns) when we get off Ashes.
+$router->get('campaign-ids/create', ['as' => 'campaign_id.create', 'uses' => 'Legacy\Web\CampaignsController@create']);
```
DFurnes: These can all be consolidated into a resource route:

```php
$router->resource('campaign-ids', 'Legacy\Web\CampaignsController');
```

resources/views/pages/campaigns_show.blade.php
```diff
@@ -0,0 +1,29 @@
...
+
+                <h3>End Date</h3>
+                <p>{{ $campaign->end_date }}</p>
+            </div>
+            <div class="container__block -narrow">
+                <a class="secondary" href="{{ route('campaign_id.edit', [ $campaign ]) }}">Edit this campaign</a>
+                <p class="footnote">
+                    Last updated: {{ $campaign->updated_at->format('F d, Y g:ia') }}<br />
+                    Created: {{ $campaign->created_at->format('F d, Y g:ia') }}
+                </p>
```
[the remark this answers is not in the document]
DFurnes: Oh yeah, that's tripped folks up in Aurora a couple times. Let's display these in EST here:

```blade
Created: {{ $campaign->created_at->setTimezone('EST')->format('F d, Y g:ia') }}
```

app/Models/Campaign.php
chloealee: @DFurnes ah I tried that before but didn't work but I think it might be because when I call `format`, it says I can't call that method on a string. However, the `start_date` and `end_date` are both type `DATETIME` in the database - should these be changed to `DATE`? Not sure if that would make a difference though?

DFurnes: `DATETIME` and `TIME` are column types in SQL (so how they're persisted in the database, not their type in PHP). Unless you set those fields as [dates](https://laravel.com/docs/5.7/eloquent-mutators#date-mutators) on the model, Eloquent will still provide them (and any other columns) to you as strings regardless of the database schema.

routes/web.php
DFurnes: I believe because `campaign-id` isn't the actual model name, you'll need to [explicitly bind it on the model](https://laravel.com/docs/5.7/routing#route-model-binding):

```php
/**
 * Get the route key for the model.
 *
 * @return string
 */
public function getRouteKeyName()
{
    return 'campaign-ids';
}
```

app/Models/Campaign.php
chloealee: OH i see what you're saying now - updated!

app/Http/Controllers/Legacy/Web/CampaignsController.php
chloealee: @DFurnes not sure what you mean by this - do you mean setting it in a different format to in `$dateFormat`?

DFurnes: You don't need to manually iterate over the request and convert to `Carbon` instances here. If you set the `$dates` property on the `Campaign` model to include `start_date` and `end_date`, this will be handled automatically. In `Campaign.php`:

```php
class Campaign
{
    /**
     * The attributes that should be mutated to dates.
     *
     * @var array
     */
    protected $dates = [
        'created_at',
        'updated_at',
        'start_date',
        'end_date',
    ];
```

That will remove the need for the `formatDatesForDatabase` method in this controller, and the need for the `formatStartEndDates` method (by allowing you to directly call `format()` on the Carbon instances for those properties).

routes/web.php
```diff
@@ -21,9 +21,16 @@
 $router->get('campaigns/{id}/inbox', 'Legacy\Web\AshesCampaignsController@showInbox');
 $router->get('campaigns/{id}', 'Legacy\Web\AshesCampaignsController@showCampaign')->name('campaigns.show');
 // Create, update, delete campaigns via Rogue.
-$router->post('campaigns', 'Legacy\Web\CampaignsController@store');
-$router->patch('campaigns/{campaign}', 'Legacy\Web\CampaignsController@update');
-$router->delete('campaigns/{campaign}', 'Legacy\Web\CampaignsController@destroy');
+// @TODO: rename these routes (to just /campaigns) when we get off Ashes.
+$router->resource('campaign-ids', 'Legacy\Web\CampaignsController', [
+    'names' => [
```
DFurnes: You shouldn't need to manually set names for each of these routes, since the `resource` helper should assign these same defaults for you. This can just be:

```php
$router->resource('campaign-ids', 'Legacy\Web\CampaignsController');
```

app/Http/Controllers/Legacy/Web/CampaignsController.php
chloealee: ah great call @DFurnes ! Updated this to have pagination (25 per page for now). Here's what it looks like (limited to 5 so you could see the pagination). 

screen shot 2018-11-06 at 5 09 27 pm

DFurnes: Woo, looking nice! If you want, you can grab Forge-style pagination from Aurora [by copying over this template](https://github.com/DoSomething/aurora/blob/master/resources/views/vendor/pagination/default.blade.php) (which I believe Laravel should auto-register if put in the same path).

app/Http/Controllers/Legacy/Web/CampaignsController.php
chloealee: @DFurnes thanks for explaining more! That is what I thought you meant but when I change that and just send `$request->all()` to `Campaign::create($`request->all()`, it gives me this error: 
screen shot 2018-11-06 at 5 13 29 pm

routes/web.php
chloealee: oh duh - thanks @DFurnes ! It wasn't working when I did that at first and added the names array because I forgot to change the references to the names elsewhere to what they're actually supposed to be! This is updated :)

app/Http/Controllers/Legacy/Web/CampaignsController.php
DFurnes: Oh, right! Frustratingly, Laravel supports auto-serializing a couple of formats (like UNIX timestamps and `YYYY-MM-DD`), but not `MM/DD/YYYY`. We built an ["arbitrary date" mutator](https://github.com/DoSomething/northstar/blob/3e37877184882d280f3574ebd09893f6a931c32e/app/Models/User.php#L223-L252) for this purpose in Northstar that'd make sense to snag here.

database/migrations/2018_11_05_164652_make_end_date_nullable_in_campaigns_table.php
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('campaigns', function (Blueprint $table) {
+            $table->dateTime('end_date')->nullable(false)->change();
+    }
```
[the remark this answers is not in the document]
chloealee: yikes thank you for catching that!

chloealee:
@DoSomething/team-bleed I can't figure out the error from the tests that `The table is empty.` - has anyone seen this before or know why this is happening? Seems like the campaign isn't getting stored in the database via the tests? But I don't think anything has changed so not sure why this is failing or why it isn't in the table.

app/Http/Controllers/Legacy/Web/CampaignsController.php
DFurnes: @chloealee You can handle this with a mutator, like we do for the `last_messaged_at` field in Northstar in the [code sample](
https://github.com/DoSomething/northstar/blob/3e37877184882d280f3574ebd09893f6a931c32e/app/Models/User.php#L223-L252) I linked above! So copy over that helper method to the model, and then you can use it for each field. For example, for `start_date`:

```php
    /**
     * Mutator for setting the start_date field.
     *
     * @param string|Carbon $value
     */
    public function setStartDateAttribute($value)
    {
        $this->setArbitraryDateString('start_date', $value);
    }
```

chloealee: sry @DFurnes I'm still confused how this works! How does `setStartDateAttribute` get called to mutate the date? Is that just some Laravel magic?

chloealee: (but it is updated! and works)

DFurnes: Yep, that's an [Eloquent mutator](https://laravel.com/docs/5.7/eloquent-mutators#defining-a-mutator)! When setting a field, Laravel will check whether there's a mutator with a matching name before saving the variable.

chloealee: very cool - that is exactly what I was looking for in the docs - thanks so much @DFurnes !

chloealee:
omg I totally goofed and see that I didn't update the new routes in tests which is why they were failing! They are updated now and I think I covered everything for one final look @DoSomething/team-bleed !

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -78,4 +86,32 @@ public function destroy(Campaign $campaign)
 
         // @TODO: redirect to campaign deleted page
     }
+
+    /**
+     * Create a new campaign.
+     */
+    public function create()
```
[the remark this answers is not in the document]
chloealee: not nitpicky at all - totally makes sense! Updated :)

ngjo:
I can't remember, on https://user-images.githubusercontent.com/9019452/48032681-96371800-e126-11e8-9932-fc8a4707052d.png, are we restricting access/editing to this page to only Admins? Or is the existing text explicit enough that they SHOULDN'T do it if they're not an admin? Do we think we should have folks requesting in #dev-phoenix since it's related to the Phoenix campaign process?

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -18,56 +19,63 @@ public function __construct()
         $this->middleware('role:admin,staff', ['only' => ['store', 'update', 'destroy']]);
```
DFurnes: Re: Jen's comment, let's lock down this whole controller to authenticated staff/admin users (so normal users couldn't log in and view our campaign ID list).

DFurnes:
> I can't remember, on https://user-images.githubusercontent.com/9019452/48032681-96371800-e126-11e8-9932-fc8a4707052d.png, are we restricting access/editing to this page to only Admins? Or is the existing text explicit enough that they SHOULDN'T do it if they're not an admin? Do we think we should have folks requesting in #dev-phoenix since it's related to the Phoenix campaign process? 

Yeah, I think locking create/update/delete endpoints down to only admins (for now) is a good call.

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -78,4 +86,32 @@ public function destroy(Campaign $campaign)
...
+    }
+
+    /**
+     * Show a specific campaign page.
+     *
+     * @param  \Rogue\Models\Campaign  $campaign
+     */
+    public function show(Campaign $campaign)
+    {
+        return view('pages.campaigns_show')->with('campaign', $campaign);
```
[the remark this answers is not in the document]
chloealee: Ah yes, that is way less confusing! Updated :)

app/Http/Controllers/Legacy/Web/CampaignsController.php
chloealee: Great catch @ngjo ! Updated here and the documentation!

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -16,7 +16,7 @@ class CampaignsController extends Controller
     public function __construct()
     {
         $this->middleware('auth', ['only' => ['store', 'update', 'destroy']]);
-        $this->middleware('role:admin,staff', ['only' => ['store', 'update', 'destroy']]);
+        $this->middleware('role:admin,staff');
```
DFurnes: @chloealee Can you add the `'auth'` middleware across the board as well?

DFurnes: I think we want all staffers & admins to be able to view these but only allow admins to edit… so:

```php
$this->middleware('auth');
$this->middleware('role:admin,staff', ['only' => ['index', 'show']]);
$this->middleware('role:admin', ['only' => ['create', 'store', 'edit', 'update', 'destroy']]);

```

PR #780, opened 2018-11-13
Adds cause to campaign creation in Rogue 

chloealee:
#### What's this PR do?
- Adds cause to campaign creation in Rogue: 
  - adds `cause` column in the `campaigns` table.
    
screen shot 2018-11-13 at 3 45 18 pm

  - adds a cause drop down to Rogue frontend when creating and editing a campaign. 
    
screen shot 2018-11-13 at 4 07 45 pm

  - adds cause field to index page and show views. 
    
screen shot 2018-11-13 at 3 46 25 pm

  - adds backend work to categorize campaign cause when creating and editing a campaign. 
- Updates tests. 
- Frontend enhancement: 
  - keeps old values in the create campaign form if validation fails (previously, it deleted the values in the form when submitted when validation failed and error message would pop up - user would have to re-write title, cause, start/end date). 
  
👀 

#### Relevant tickets
Fixes 1st and 4th task in [this](https://www.pivotaltracker.com/n/projects/2019429/stories/161604438) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Legacy/Web/CampaignsController.php
+++ b/app/Http/Controllers/Legacy/Web/CampaignsController.php
@@ -47,6 +47,7 @@ class CampaignsController extends Controller
     {
         $this->validate($request, [
             'internal_title' => 'required|string|unique:campaigns',
+            'cause' => 'required|string',
             'start_date' => 'required|date',
             'end_date' => 'nullable|date|after:start_date',
         ]);
@@ -89,6 +90,7 @@ class CampaignsController extends Controller
     {
         $this->validate($request, [
             'internal_title' => 'string',
+            'cause' => 'string',
             'start_date' => 'date',
             'end_date' => 'nullable|date|after:start_date',
         ]);
--- a/app/Http/Controllers/Legacy/Web/CampaignsController.php
+++ b/app/Http/Controllers/Legacy/Web/CampaignsController.php
@@ -109,7 +109,10 @@ class CampaignsController extends Controller
             'Violence',
         ];
 
-        return view('campaign-ids.edit')->with('campaign', $campaign)->with('causes', $causes);
+        return view('campaign-ids.edit', [
+            'campaign' => $campaign,
+            'causes' => $causes
+        ]);
     }
 
     /**
--- a/app/Http/Controllers/Legacy/Web/CampaignsController.php
+++ b/app/Http/Controllers/Legacy/Web/CampaignsController.php
@@ -36,6 +36,7 @@ class CampaignsController extends Controller
     public function create()
     {
         $causes = [
+            'No Cause',
             'Animals',
             'Bullying',
             'Disasters',
@@ -49,7 +50,6 @@ class CampaignsController extends Controller
             'Relationships',
             'Sex',
             'Violence',
-            'No Cause',
         ];
 
         return view('campaign-ids.create')->with('causes', $causes);
@@ -95,6 +95,7 @@ class CampaignsController extends Controller
     public function edit(Campaign $campaign)
     {
         $causes = [
+            'No Cause',
             'Animals',
             'Bullying',
             'Disasters',
@@ -108,7 +109,6 @@ class CampaignsController extends Controller
             'Relationships',
             'Sex',
             'Violence',
-            'No Cause',
         ];
 
         return view('campaign-ids.edit')->with('campaign', $campaign)->with('causes', $causes);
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -35,7 +35,24 @@ public function index()
      */
     public function create()
     {
-        return view('campaign-ids.create');
+        $causes = [
+            'No Cause',
```
chloealee: I put `No Cause` as the first since this will be shown as default selected in the dropdown list - if it were any other campaign, I'd be worried people would skip this step and if so, campaigns would be incorrectly labeled as `Animals` for example and that cause would have many false labels. Would this cause the same problem though if they ignore this drop down and select `No Cause` (`No Cause` would have a lot of false labeling)? Not sure if I'm explaining this clearly...

katiecrane: @chloealee makes sense! do we actually want to allow campaigns to not have causes? Should we make this field required and have a `-` placeholder?

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -47,6 +64,7 @@ public function store(Request $request)
     {
         $this->validate($request, [
             'internal_title' => 'required|string|unique:campaigns',
+            'cause' => 'required|string',
```
chloealee: For now, I only put validation to make sure the cause is a string and is required since all causes we'd want come from the drop down menu that is hardcoded (there is no way to enter your own cause and no API route to write this in manually either). Because of this, I didn't think there was need for validation for the cause to be within specific parameters since the drop down values has that covered. 

I'll make a note to myself though to check the values in Ashes to make sure there aren't any additional cause values besides what is in the array above so we have consistent data when we migrate.

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -76,7 +94,24 @@ public function show(Campaign $campaign)
...
+            'Homelessness',
+            'Mental Health',
+            'Physical Health',
+            'Poverty',
+            'Relationships',
+            'Sex',
+            'Violence',
+        ];
+
+        return view('campaign-ids.edit')->with('campaign', $campaign)->with('causes', $causes);
```
katiecrane: Might want to go with [the first option here](https://laravel.com/docs/5.5/views#passing-data-to-views) to pass an array instead of having two `with`s

app/Http/Controllers/Legacy/Web/CampaignsController.php
katiecrane: @chloealee Is there a way to make the `No cause` choice blank so that it wouldn't register as a string and we wouldn't have to worry about folks submitting `No cause`? Or maybe we actually don't need to worry about it so much since can change the cause later and only a small number of people will be creating campaigns in Rogue

app/Http/Controllers/Legacy/Web/CampaignsController.php
chloealee: @katiecrane right now in Rogue admin (prod) at the bottom, there is a section that has `No Cause` with all the campaigns that don't have a cause! I guess in Ashes their `cause` would be `null` but I was thinking of just giving it a `No Cause` as the cause which is silly haha I think you're right that we should just have `-` as a placeholder but it has to be required (which it is now!) instead of `No Cause`. This is updated now to have a `Select the cause` placeholder! 

screen shot 2018-11-13 at 4 07 45 pm

app/Http/Controllers/Legacy/Web/CampaignsController.php
chloealee: @katiecrane great call! Updated that here! https://github.com/DoSomething/rogue/pull/780#discussion_r233221919

app/Http/Controllers/Legacy/Web/CampaignsController.php
katiecrane: @chloealee this looks awesome!!! Oooo interesting yeah I think maybe `null` for their cause would be best?

chloealee: @katiecrane right now, this is required so they must choose something from the dropdown otherwise a validation error pops up saying cause is required!

PR #783, opened 2018-11-21
Command to migrate legacy campaigns from Ashes -> Rogue

chloealee:
#### What's this PR do?
- Updates to `campaigns` table: 
  - adds `secondary_causes` column
  - adds `campaign_run_Id` column 
  - adds `internal_doc` column
  - makes fields nullable 
- Creates script to import legacy campaigns from Ashes to Rogue based on rules outlined [here](https://docs.google.com/document/d/1_B1499L9xhJQfaYDa2-Sjsuj3IRtLKb_1qYLvNAXsbI/edit#).
- Creates test to make sure script is working as expected! 

- Did I capture all of the logic correctly? 
- Is there anything I am missing? 
- Feel free to ping me if you'd like to see a screen shot of the database and CSV to make sure all logic checks out! Another easy way to make sure this is correct is by looking at the sample CSV that the test imports and the logic there. 
- Finally, happy to share the SQL query @DFurnes and I came up to for the CSV export to also make sure that is correct since that is what this logic is based on! 

#### Relevant tickets
Fixes task 3 in [this](https://www.pivotaltracker.com/n/projects/2019429/stories/161604438) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Legacy/Web/CampaignsController.php
+++ b/app/Http/Controllers/Legacy/Web/CampaignsController.php
@@ -5,6 +5,7 @@ namespace Rogue\Http\Controllers\Legacy\Web;
 use Carbon\Carbon;
 use Rogue\Models\Campaign;
 use Illuminate\Http\Request;
+use Illuminate\Validation\Rule;
 use Rogue\Http\Controllers\Controller;
 
 class CampaignsController extends Controller
@@ -125,7 +126,7 @@ class CampaignsController extends Controller
     public function update(Request $request, Campaign $campaign)
     {
         $this->validate($request, [
-            'internal_title' => 'string',
+            'internal_title' => [Rule::unique('campaigns')->ignore($campaign->id)],
             'cause' => 'string',
             'impact_doc' => 'url',
             'start_date' => 'date',


chloealee:
One note - there is only one campaign that I found in the export that has `NULL` as the `start_date`. Since we don't want `start_date` to be `nullable` in Rogue and this is only one campaign, when we do the export from Ashes, can we just manually put its `created_at` date for the `start_date`?

app/Console/Commands/ImportAshesCampaigns.php
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+
+            // See if the campaign exists
+            $existing_campaign = Campaign::where('id', $legacy_campaign['campaign_id'])->orWhere('id', $legacy_campaign['run_id'])->first();
+
+            // Create the campaign if there isn't one already
+            if (! $existing_campaign) {
+                // If there is no campaign_id, set the campaign_run_id to the id.
+                if (is_null($legacy_campaign['campaign_id'])) {
```
katiecrane: Does this mean that in the csv runs will not have the `campaign_id` column populated?

app/Console/Commands/ImportAshesCampaigns.php
```diff
@@ -0,0 +1,115 @@
...
+                    $campaign = $this->createCampaign($legacy_campaign['run_id'], $legacy_campaign);
+                    // If the campaign_id does not equal the next record's campaign_id or the previous record's campaign id, this is either the latest run or there is only one campaign run, so keep the original campaign_id as the id in Rogue.
+                } else if ($legacy_campaigns_csv->fetchOne($iterator) && $legacy_campaign['campaign_id'] != $legacy_campaigns_csv->fetchOne($iterator)['campaign_id'] && $legacy_campaign['campaign_id'] != $legacy_campaigns_csv->fetchOne($iterator - 1)) {
```
katiecrane: This might be easier to look at if this logic was extracted into a `shouldUseCampaignId` function? But I guess it might still all end up on one line. What do you think @chloealee ?

app/Console/Commands/ImportAshesCampaigns.php
chloealee: @katiecrane great q! It will have the `campaign_id` column! However, after running the query, some campaigns didn't have this populated so wanted to account for these edge cases. Happy to share the CSV I'm working with too if that is helpful!

katiecrane: OH or does it mean that only CURRENT runs will have the `campaign_id populated>

tests/Console/ImportLegacyCampaignsCommandTest.php
```diff
@@ -0,0 +1,97 @@
...
+use Carbon\Carbon;
+use Tests\TestCase;
+use Rogue\Models\Campaign;
+
+class ImportLegacyCampaignsCommandTest extends TestCase
+{
+    public function testImportingLegacyCampaigns()
+    {
+        // Run the legacy campaign import command
+        $this->artisan('rogue:ashescampaignsimport', ['path' => 'tests/Console/example-legacy-campaigns.csv']);
```
katiecrane: yesssss thank you for making tests for this!! 🙌

app/Console/Commands/ImportAshesCampaigns.php
katiecrane: @chloealee hmmmm does that mean that there are runs not attached to any campaign? I think seeing the CSV would help!

katiecrane: @chloealee ah okay I see, the first 7 rows are the ones with `null`. Are these actual campaigns? Seems like one of them is the DS app and 5 of them are in Spanish?

DFurnes: Yeah, these look like placeholder campaigns. I think it's safe to add them with the script, and we can delete later.

tests/Console/ImportLegacyCampaignsCommandTest.php
chloealee: 😄

app/Console/Commands/ImportAshesCampaigns.php
```diff
@@ -0,0 +1,115 @@
...
+        foreach ($legacy_campaigns as $iterator => $legacy_campaign) {
+            // Normalize all "NULL" values to null
+            foreach ($legacy_campaign as $key => $value) {
+                if ($value === "NULL") {
+                    $legacy_campaign[$key] = null;
+                }
+            }
+
+            // See if the campaign exists
+            $existing_campaign = Campaign::where('id', $legacy_campaign['campaign_id'])->orWhere('id', $legacy_campaign['run_id'])->first();
```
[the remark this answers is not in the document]
chloealee: @DFurnes great q! Yes, this should work because the way we write the query and export the results to a CSV from Ashes, it'll always be reverse order of old runs -> current like you said. 

Here's an example!

screen shot 2018-11-28 at 1 22 05 pm

app/Console/Commands/ImportAshesCampaigns.php
```diff
@@ -0,0 +1,118 @@
...
+            // Create the campaign if there isn't one already
+            if (! $existing_campaign) {
+                // If there is no campaign_id, set the campaign_run_id to the id.
+                if (is_null($legacy_campaign['campaign_id'])) {
+                    $campaign = $this->createCampaign($legacy_campaign['run_id'], $legacy_campaign);
+                    // If the campaign_id does not equal the next record's campaign_id or the previous record's campaign id, this is either the latest run or there is only one campaign run, so keep the original campaign_id as the id in Rogue.
+                } else if ($legacy_campaigns_csv->fetchOne($iterator) &&
+                           $legacy_campaign['campaign_id'] != $legacy_campaigns_csv->fetchOne($iterator)['campaign_id'] &&
+                           $legacy_campaign['campaign_id'] != $legacy_campaigns_csv->fetchOne($iterator - 1)) {
+
```
chloealee: @katiecrane @DFurnes I agree this is a little hard to read! Since it is just for this script, I'm not sure if it would make it that much easier to read if broken into a separate helper function but I played with the formatting to see if this would make it easier to read - lmk what you think!

app/Console/Commands/ImportAshesCampaigns.php
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        // Make a local copy of the CSV
+        $path = $this->argument('path');
+        info('rogue:legacycampaignimport: Loading in csv from ' . $path);
```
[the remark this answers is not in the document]
chloealee: @DFurnes not sure what you mean by that?

app/Console/Commands/ImportAshesCampaigns.php
chloealee: ok great, thanks @katiecrane !

app/Console/Commands/ImportAshesCampaigns.php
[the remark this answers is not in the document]
katiecrane: @chloealee that would be doing `$this->line('rogue:legacycampaignimport: Loading in csv from ' . $path)`. It'll print out in your terminal locally and when running on a one-off dyno in Heroku, it will print to the logs for that dyno (which is [something we figured out semi-recently](https://devcenter.heroku.com/articles/one-off-dynos#running-tasks-in-background)!)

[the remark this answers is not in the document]
katiecrane: I believe it will still log to PaperTrail

app/Console/Commands/ImportAshesCampaigns.php
```diff
@@ -0,0 +1,119 @@
...
+                    // If this is the last row in the CSV, it must either be the only run or last run, so use the campaign_id as the id in Rogue.
+                } else if (!$legacy_campaigns_csv->fetchOne($iterator)) {
+                    $campaign = $this->createCampaign($legacy_campaign['campaign_id'], $legacy_campaign);
+                    // Else use the campaign_run_id as the id in Rogue.
+                } else {
+                    $campaign = $this->createCampaign($legacy_campaign['run_id'], $legacy_campaign);
+                }
+
+                $this->line('rogue:legacycampaignimport: Created campaign ' . $campaign->internal_title);
+            }
```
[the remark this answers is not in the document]
katiecrane: @chloealee whatever you think will be the most useful works for me 😄

app/Console/Commands/ImportAshesCampaigns.php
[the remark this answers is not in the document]
chloealee: ooooo what are the benefits of logging this way?

app/Models/Campaign.php
```diff
@@ -21,7 +21,7 @@ class Campaign extends Model
      *
      * @var array
      */
-    protected $fillable = ['internal_title', 'cause', 'start_date', 'end_date'];
+    protected $fillable = ['id','internal_title', 'cause', 'secondary_causes', 'campaign_run_id', 'impact_doc', 'start_date', 'end_date', 'created_at', 'updated_at'];
```
DFurnes: Let's remove `id`, `created_at`, and `updated_at` from the fillable array & use `Campaign::forceCreate` in the script (since we don't want these to be mass-assignable elsewhere).

database/migrations/2018_11_16_200157_add_secondary_causes_run_id_impact_doc_and_make_causes_nullable_to_campaigns_table.php
```diff
@@ -0,0 +1,38 @@
...
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('campaigns', function (Blueprint $table) {
+            $table->string('secondary_causes')->index()->after('cause')->nullable()->comment('Secondary causes associated with legacy campaigns');
+            $table->integer('campaign_run_id')->index()->after('secondary_causes')->nullable()->comment('Campaign Run Id to reference when we run script to update signup and post ids after legacy campaign migration. This column can be deleted once migration and update script are complete.');
+            $table->string('impact_doc')->index()->after('campaign_run_id')->nullable()->comment('URL to proof of impact doc.');
```
DFurnes: We don't need an index on the `impact_doc` column since it won't be queried against.

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -64,6 +64,7 @@ public function store(Request $request)
         $this->validate($request, [
             'internal_title' => 'required|string|unique:campaigns',
             'cause' => 'required|string',
+            'impact_doc' => 'required|string',
```
DFurnes: Let's use the [`url` validator](https://laravel.com/docs/5.5/validation#rule-url) to enforce that this is a URL.

app/Http/Controllers/Legacy/Web/CampaignsController.php
```diff
@@ -126,6 +127,7 @@ public function update(Request $request, Campaign $campaign)
         $this->validate($request, [
             'internal_title' => 'string',
```
DFurnes: This is missing the `unique:campaigns` validation rule that we use when creating a new campaign.

DFurnes: You can include it here [by specifying a campaign ID to ignore for the uniqueness check](https://laravel.com/docs/5.5/validation#rule-unique), e.g:

```php
'internal_title' => [Rule::unique('campaigns')->ignore($campaign->id)],
```

app/Models/Campaign.php
chloealee: cool i didn't know about `forceCreate` ! Much better solution!

app/Http/Controllers/Legacy/Web/CampaignsController.php
chloealee: great call!

app/Http/Controllers/Legacy/Web/CampaignsController.php
chloealee: very cool, thanks @DFurnes !

PR #786, opened 2018-11-28
Adds indexes to posts and signups tables to prevent database lock for force delete test records script

chloealee:
#### What's this PR do?
- Adds indexes to `posts` and `signups` tables to prevent [database lock ](https://github.com/DoSomething/internal/issues/408#issuecomment-390422665) for force delete test records script
  - I didn't add `why_participated` as an index for in the `signups` table because I got this error when I tried to do that (since it is a text blob):
```
SQLSTATE[42000]: Syntax error or access violation: 1170 BLOB/TEXT column 'why_participated' used in key specification without a key length
```

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/161383417) Pivotal Card.

#### How to deploy

### DIFF
--- a/database/migrations/2018_11_28_162339_add_indexes_to_signups_and_posts_for_force_delete_test_records.php
+++ b/database/migrations/2018_11_28_162339_add_indexes_to_signups_and_posts_for_force_delete_test_records.php
@@ -1,0 +1,42 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class AddIndexesToSignupsAndPostsForForceDeleteTestRecords extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('signups', function (Blueprint $table) {
+            $table->index('source');
+        });
+
+        Schema::table('posts', function (Blueprint $table) {
+            $table->index('source');
+            $table->index('text');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('signups', function (Blueprint $table) {
+            $table->dropIndex('source');
+        });
+
+        Schema::table('posts', function (Blueprint $table) {
+            $table->dropIndex('source');
+            $table->dropIndex('text');
+        });
+    }
+}


database/migrations/2018_11_28_162339_add_indexes_to_signups_and_posts_for_force_delete_test_records.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function up()
+    {
+        Schema::table('signups', function (Blueprint $table) {
+            $table->index('source');
+        });
+
+        Schema::table('posts', function (Blueprint $table) {
+            $table->index('source');
+            $table->index('text');
```
DFurnes: I'm a little unsure about creating an index on the `text` column (especially if it's just for this script), since the content there is so varied and otherwise unlikely to be queried against. It might be better to think about lighter weight solutions, like perhaps using a `source_details` value to mark these.

DFurnes: If we do go forward with this, let's check how big this index is on QA & make sure that we don't see any performance hits.

chloealee: @DFurnes ah that makes sense! I'm not sure if we can use `source_details` column for this since it is a Ghost inspector test though so it is submitting posts via the front end with a specific caption for `text` and `participated`.

chloealee: @DFurnes is there a way for Phoenix to check, if it is a Ghost Inspector test, send along something in the  `source_details` so we can update [force delete test script](https://github.com/DoSomething/rogue/blob/master/app/Console/Commands/ForceDeleteTestRecords.php) to check for this instead of the `why_participated` and `text` columns?

chloealee: also, this is random but I noticed that the signup doesn't have a signup `source` for signup 9900736. Is there something unique happening with these 👻 tests that would change that? I see other signups have the correct sources.

chloealee: chatted with @mendelB [on Slack](https://dosomething.slack.com/archives/C3ASB4204/p1543434017258600) and Phoenix could append something in the `source_details` for any signup that has `why_participated_ghost_test` as the `why_participated` and for any post that has `caption_ghost_test` for the `text`. Then, we can update the command to search by the `source_details` instead of adding these indexes? Does that sound like a better idea @DFurnes ?

DFurnes: @chloealee @mendelB Can we set the `source_details` on Phoenix based on a query string, rather than hardcoding a magic `why_participated`? Say, if the user enters the campaign flow with `?utm_source=runscope` that `runscope` is sent as the `source_details`?

chloealee: oooh that is interesting, @DFurnes ! So for this to work, we'd have to: 
- update the runscope test to include the `?utm_source=runscope` when it runs the tests
- update Phoenix to send `runscope` in the `source_details` for anything with the utm param 
- update the job in Rogue to look for `source_details` instead of the `why_participated` and `text`?

mendelB: Cool, that does sound interesting! Would we want this to be a general rule for RBs on Phoenix, where we'd always set the stored `utm_source` value to `source_details`, or would this be a specific rule for when that `utm_source` is `runscope`?

mendelB: Awesome, ok, so I can add a Pivotal ticket to address bullet point two! @chloealee What's the time sensitivity on this work? I can chat with @weerd about fitting this into our last sprint.

chloealee: Thanks @DFurnes and @mendelB ! @mendelB I don't think this is too urgent and can probably wait until the next year since I know we're super focused on retiring ashes! cc'ing @ngjo though for her thoughts too!

ngjo: > I think it'd make sense across the board. It solves our specific problem here, but I believe this has also been a request from the marketing team (to find what signups/posts came from Fastweb traffic, for example).

+1 to what @DFurnes said ^ definitely has been a marketing request (and has come in the form of a very specific product opportunity assessment [here](https://docs.google.com/document/d/15w6O9Wy8n4rnXku9SoUBWEFZ0WeWCVpr-SayI7L5iCc/edit), but more broadly, they just want to have as much segmentation information as possible.

Given that it's also a product opportunity assessment and like @chloealee said, we're heads down in Ashes stuff, and it's the end of the year basically, I agree that I think it's fine to wait until next year!

chloealee: Thanks all!! I'm going to close this PR and excited to work on this in 2019!

PR #787, opened 2018-12-04
Script to update signup and post campaign ids after legacy campaign migration

chloealee:
#### What's this PR do?
- Adds script to update signup and post campaign ids after legacy campaign migration
  - Grabs all campaigns from the `campaigns` table
  - Gets all signups by `campaign_run_id` and updates the signup's `campaign_id`to the campaign's `id`
  - Gets all signup's posts and updates the post's `campaign_id` to the campaign's `id`
- Adds a test to make sure this logic is working as we expect 

👀 
Is this the most efficient way to do this?

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/162309728) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Console/Commands/UpdateSignupAndPostCampaignIds.php
+++ b/app/Console/Commands/UpdateSignupAndPostCampaignIds.php
@@ -2,7 +2,6 @@
 
 namespace Rogue\Console\Commands;
 
-use DB;
 use Rogue\Models\Post;
 use Rogue\Models\Signup;
 use Rogue\Models\Campaign;
@@ -15,7 +14,7 @@ class UpdateSignupAndPostCampaignIds extends Command
      *
      * @var string
      */
-    protected $signature = 'rogue:updatesignupandpostcampaignids';
+    protected $signature = 'rogue:updatesignupandpostcampaignids {--campaign= : The campaign_id to update signup and posts under}';
 
     /**
      * The console command description.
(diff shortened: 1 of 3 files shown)


app/Console/Commands/UpdateSignupAndPostCampaignIds.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->line('rogue:updatesignupandpostcampaignids: Starting script to update campaign ids!');
+
+
+        // Grab all of the campaigns in the campaigns table.
+        $query = (new Campaign)->newQuery();
+
+        if ($this->option('campaign')) {
+            $query = $query->where('id', $this->option('campaign'));
+        }
+
```
chloealee: added this option so we can test a small batch of campaigns before running the update on all signups and posts!

app/Console/Commands/UpdateSignupAndPostCampaignIds.php
```diff
@@ -0,0 +1,76 @@
...
+        $campaigns = $query->get();
+
+        // Update each signup and the signup's post(s) in each campaign with new campaign id.
+        foreach ($campaigns as $campaign) {
```
katiecrane: Would it be helpful to have any other logging through out this loop? I could see potentially logging out which campaign we are on being helpful

chloealee: ooo that is a great idea, thanks @katiecrane ! Updated! :)

tests/Console/UpdateSignupAndPostCampaignIdsCommandTest.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->artisan('rogue:updatesignupandpostcampaignids');
+
+        // Make sure the signup and posts were updated.
+        $this->assertDatabaseHas('signups', [
+            'id' => $signup->id,
+            'northstar_id' => $signup->northstar_id,
+            'campaign_id' => $campaign->id,
+            'campaign_run_id' => 1,
+            'why_participated' => $signup->why_participated,
+        ]);
```
katiecrane: Do we know for sure that the `signup` had a `campaign_id` different from `$campaign->id` to start with? I'm thinking if they happened to be the same this could be a false positive.

app/Console/Commands/UpdateSignupAndPostCampaignIds.php
```diff
@@ -0,0 +1,78 @@
...
+            $this->line('rogue:updatesignupandpostcampaignids: Updating signups/posts under campaign id: ' . $campaign->id);
+
+            $signups = Signup::where('campaign_run_id', $campaign->campaign_run_id)->get();
```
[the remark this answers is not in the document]
chloealee: ah this is so helpful, thanks @DFurnes !! I knew that I was doing something wrong by iterating but wasn't quite sure how do to this - thank you!! Updated :)

PR #788, opened 2018-12-04
Always update signup for latest (or empty) run, if it exists.

DFurnes:
#### What's this PR do?
This pull request updates Rogue's `POST v2/signups` and `POST v3/signups` to always update the latest run if it exists for a campaign, rather than making a new one where the `campaign_run_id` is null. This will allow us to stop sending Campaign Run IDs (and run the script in #787) without breaking this functionality for applications that rely on it.

I added a failing test in 3e5590a which demonstrates the issue (we create a new signup when we shouldn't), and you can see the issue is fixed in e51a989. This doesn't break any other tests, and I don't believe we rely on the prior unexpected behavior anywhere.

This was flagged by @weerd when testing Phoenix's new sans-runs signup logic.

#### Relevant tickets
Fixes [#162426572](https://www.pivotaltracker.com/story/show/162426572).

#### How to deploy

### DIFF
--- a/app/Http/Controllers/SignupsController.php
+++ b/app/Http/Controllers/SignupsController.php
@@ -65,7 +65,7 @@ class SignupsController extends ApiController
         $northstarId = getNorthstarId($request);
 
         // Check to see if the signup exists before creating one.
-        $signup = $this->signups->get($northstarId, $request['campaign_id'], $request['campaign_run_id']);
+        $signup = $this->signups->get($northstarId, $request['campaign_id']);
 
         $code = $signup ? 200 : 201;
 
--- a/app/Managers/Legacy/Two/SignupManager.php
+++ b/app/Managers/Legacy/Two/SignupManager.php
@@ -61,9 +61,9 @@ class SignupManager
      * @param  int $campaignRunId
      * @return \Rogue\Models\Signup|null
      */
-    public function get($northstarId, $campaignId, $campaignRunId)
+    public function get($northstarId, $campaignId)
     {
-        $signup = $this->signup->get($northstarId, $campaignId, $campaignRunId);
+        $signup = $this->signup->get($northstarId, $campaignId);
 
         return $signup;
     }
--- a/app/Managers/Legacy/Two/SignupManager.php
+++ b/app/Managers/Legacy/Two/SignupManager.php
@@ -58,7 +58,6 @@ class SignupManager
      *
      * @param  string $northstarId
      * @param  int $campaignId
-     * @param  int $campaignRunId
      * @return \Rogue\Models\Signup|null
      */
     public function get($northstarId, $campaignId)
--- a/app/Managers/SignupManager.php
+++ b/app/Managers/SignupManager.php
@@ -108,9 +108,9 @@ class SignupManager
      * @param  int $campaignRunId
      * @return \Rogue\Models\Signup|null
      */
-    public function get($northstarId, $campaignId, $campaignRunId)
+    public function get($northstarId, $campaignId)
     {
-        $signup = $this->signup->get($northstarId, $campaignId, $campaignRunId);
+        $signup = $this->signup->get($northstarId, $campaignId);
 
         return $signup;
     }
--- a/app/Managers/SignupManager.php
+++ b/app/Managers/SignupManager.php
@@ -105,7 +105,6 @@ class SignupManager
      *
      * @param  string $northstarId
      * @param  int $campaignId
-     * @param  int $campaignRunId
      * @return \Rogue\Models\Signup|null
      */
     public function get($northstarId, $campaignId)
--- a/app/Repositories/Legacy/Two/SignupRepository.php
+++ b/app/Repositories/Legacy/Two/SignupRepository.php
@@ -49,7 +49,6 @@ class SignupRepository
      *
      * @param  string $northstarId
      * @param  int $campaignId
-     * @param  int $campaignRunId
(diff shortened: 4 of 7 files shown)


app/Managers/Legacy/Two/SignupManager.php
```diff
@@ -61,9 +61,9 @@ public function create($data)
      * @param  int $campaignRunId
      * @return \Rogue\Models\Signup|null
```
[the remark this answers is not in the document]
DFurnes: Resolved in 3c39519.

tests/Http/SignupTest.php
```diff
@@ -90,9 +90,6 @@ public function testNotCreatingDuplicateSignups()
     {
         $signup = factory(Signup::class)->states('contentful')->create();
 
-        // Mock the Blink API call.
-        $this->mock(Blink::class)->shouldReceive('userSignup');
-
         $response = $this->withAccessToken($signup->northstar_id)->postJson('api/v3/signups', [
```
chloealee: just curious why you deleted here!

app/Managers/SignupManager.php
```diff
@@ -108,9 +108,9 @@ public function destroy($signupId)
      * @param  int $campaignRunId
```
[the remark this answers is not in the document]
DFurnes: Resolved in 3c39519.

app/Repositories/Legacy/Two/SignupRepository.php
```diff
@@ -52,13 +52,14 @@ public function create($data)
      * @param  int $campaignRunId
```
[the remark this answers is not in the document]
DFurnes: Resolved in 3c39519.

app/Repositories/SignupRepository.php
```diff
@@ -52,13 +52,14 @@ public function update($signup, $data)
      * @param  int $campaignRunId
```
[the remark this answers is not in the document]
DFurnes: Resolved in 3c39519.

tests/Http/SignupTest.php
DFurnes: We're not creating a signup, and so this method is never called. Mocking it is unnecessary, and potentially harmful since if we started sending `userSignup` events here due to a bug we'd never know it.

app/Repositories/Legacy/Two/SignupRepository.php
```diff
@@ -49,16 +49,16 @@ public function create($data)
...
      */
-    public function get($northstarId, $campaignId, $campaignRunId)
+    public function get($northstarId, $campaignId)
     {
         $signup = Signup::where([
             'northstar_id' => $northstarId,
             'campaign_id' => $campaignId,
-            'campaign_run_id' => $campaignRunId,
-        ])->first();
+        ])->orderByRaw('campaign_run_id IS NULL')
```
chloealee: ah interesting! So `orderByRaw` just makes sure that the `NULL` values are at the bottom?

tests/Http/SignupTest.php
chloealee: ohhh right because the signup already exists since we are creating it in the model factory - got it!

app/Repositories/Legacy/Two/SignupRepository.php
DFurnes: Oh! It should be the exact opposite, that was a mistake. Updated in c7deaa4.

The idea is that if we make new signups where `campaign_run_id` is null, we want those to be prioritized over old ones where this is a number. So the final query looks like kinda like this:

```sql
SELECT * FROM signups
WHERE northstar_id = "5543dfd6469c64ec7d8b46b3" AND campaign_id = 1144
ORDER BY campaign_run_id IS NULL DESC, `campaign_run_id` DESC LIMIT 1
```

DFurnes: @chloealee I added a new test for this in 9d794d2. Thanks for catching!

chloealee:
Ah got it - thanks for explaining and putting in that extra test! I hadn't seen `orderByRaw` before so was trying to understand! Looks great !!

app/Http/Controllers/SignupsController.php
```diff
@@ -65,7 +65,7 @@ public function store(Request $request)
         $northstarId = getNorthstarId($request);
 
         // Check to see if the signup exists before creating one.
-        $signup = $this->signups->get($northstarId, $request['campaign_id'], $request['campaign_run_id']);
+        $signup = $this->signups->get($northstarId, $request['campaign_id']);
```
aaronschachter: @DFurnes - I think this line may what's causing some unexpected behavior in Gambit, where [we're not seeing new signups from yes responses to the TYF broadcast when a user has signed up for the campaign in a previous run](https://dosomething.slack.com/archives/C1V0M6RPE/p1544051436007200).

DFurnes: Oh, shoot. You’re totally right. We can revert this PR this morning to resolve that problem.

I think our best course of action is to avoid removing runs from applications until we’ve run the script in Rogue, since otherwise we don’t have a reliable way to know whether or not a signup is new. (Or we could adjust this to find the latest run for _all_ signups on a given campaign ID, but I’m thinking that adds some unnecessary complexity and a chance we introduce other bugs.)

DFurnes: I’ll update the order-of-operations issue to reflect this. cc: @chloealee @weerd

PR #789, opened 2018-12-05
Updates GET /v3/posts and GET /v3/signups endpoints to be able to filter by source

chloealee:
#### What's this PR do?
- Updates GET /v3/posts and GET /v3/signups endpoints to be able to filter by source. 
- Updates documentation.

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/162448781) Pivotal Card. 

cc @aaronschachter 
#### How to deploy

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -53,7 +53,7 @@ class Post extends Model
      *
      * @var array
      */
-    public static $indexes = ['id', 'signup_id', 'campaign_id', 'type', 'action', 'northstar_id', 'status', 'created_at'];
+    public static $indexes = ['id', 'signup_id', 'campaign_id', 'type', 'action', 'northstar_id', 'status', 'created_at', 'source'];
 
     /**
      * The tags prefixed with 'good' that will send a post to Slack.
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -41,7 +41,7 @@ class Signup extends Model
      * @var array
      */
     public static $indexes = [
-        'campaign_id', 'campaign_run_id', 'updated_at', 'northstar_id', 'id', 'quantity',
+        'campaign_id', 'campaign_run_id', 'updated_at', 'northstar_id', 'id', 'quantity', 'source'
     ];
 
     /**
--- a/database/migrations/2018_12_05_155452_add_source_as_index_to_signups_and_posts_tables.php
+++ b/database/migrations/2018_12_05_155452_add_source_as_index_to_signups_and_posts_tables.php
@@ -1,0 +1,40 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class AddSourceAsIndexToSignupsAndPostsTables extends Migration
+{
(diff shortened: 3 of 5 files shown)


app/Models/Post.php
```diff
@@ -53,7 +53,7 @@ class Post extends Model
      *
      * @var array
      */
-    public static $indexes = ['id', 'signup_id', 'campaign_id', 'type', 'action', 'northstar_id', 'status', 'created_at'];
+    public static $indexes = ['id', 'signup_id', 'campaign_id', 'type', 'action', 'northstar_id', 'status', 'created_at', 'source'];
```
DFurnes: Did we already add a corresponding database index on this field?

chloealee: @DFurnes yeah - looks like this was already an index in the `posts` and `signups` tables!

DFurnes: Hmm, I don't see that index when looking at those tables on QA & am having a hard time finding where we set that in a migration. Can you point me in the right direction?

chloealee: oh you are totally right - sorry @DFurnes ! It was already set on my local weirdly so that is why I thought it was an index already... I'll add this migration in now!

DFurnes: Spooky! I'd recommend running `php artisan migrate:fresh --seed` if you haven't lately.

aaronschachter: I'm having trouble getting a `filter[type]` query parameter to work on a `GET /v3/posts` request... and can't tell if it's an error in on my end or a bug? I'm using the `gambit` OAuth client.

aaronschachter:
This question is a little out of scope of the PR too, but it seems the `meta` response in the documentation doesn't match what I see in a `GET /posts` request:
```
  meta: 
   { cursor: 
      { current: 1,
        prev: null,
        next: 'https://activity-qa.dosomething.org/api/v3/posts?orderBy=id%2Cdesc&limit=50&page=2',
        count: 50 } } }
```
Is total count is no longer available?

chloealee:
ah great q @aaronschachter ! Sorry I will update the docs because you're right, that is out of date! We don't show that anymore for v3 routes for performance enhancements made in [this PR](https://github.com/DoSomething/rogue/pull/531). Is this something you'll need? If so, should be pretty easy to pop back in!

app/Models/Post.php
chloealee: ah interesting, what type are you trying to filter by @aaronschachter ?

chloealee:
That sounds good to me @aaronschachter ! I can make a card for that. How urgent is this?

app/Models/Post.php
aaronschachter: 🤦‍♂️ This is working for me, sorry for a false flag! I sometimes forget to use the `filter[fieldName]` syntax 🥇

aaronschachter:
Totes non-urgent!

app/Models/Post.php
chloealee: it is a confusing one! Glad it is working ! :)

chloealee:
Card [here](https://www.pivotaltracker.com/n/projects/2019429)!

PR #794, opened 2018-12-05
Update Rogue Admin to read local campaign data.

DFurnes:
#### What's this PR do?
This pull request updates Rogue's admin interface to read local campaign data (rather than trying to ask Ashes or Phoenix for it). We should deploy this after we've run the "migrate campaigns" script on production (and before we update IDs on individual signups & posts).

screen shot 2018-12-05 at 6 23 08 pm

screen shot 2018-12-05 at 6 23 23 pm

screen shot 2018-12-05 at 6 23 05 pm

Since we don't have Staff Picks in Rogue, we'll maybe want to think about making some UX improvements to the main campaign overview page (say, prioritizing active campaigns or ones that have pending posts).

I still need to do some testing on this!

#### Relevant tickets
References [#161650678](https://www.pivotaltracker.com/story/show/161650678).

#### How to deploy


tests/Unit/Services/CampaignServiceTest.php
```diff
@@ -12,94 +12,6 @@
...
-
-        foreach ($testCampaigns as $key => $test) {
-            // Make sure the campaign was returned from Ashes.
-            $this->assertEquals($test, $campaigns[$key]);
-
-            // Make sure the campaigns got stored in cached.
-            $this->assertEquals($test, $cachedCampaigns[$test['id']]);
-        }
-    }
-
```
katiecrane: I guess we don't really need tests for this part now since we are literally just pulling straight from the Rogue database?

app/Services/CampaignService.php
```diff
@@ -49,38 +50,7 @@ public function __construct(Ashes $ashes, Phoenix $phoenix)
...
-            }
-
-            // Cache campaign for a day.
-            $this->cache->store($id, $campaign['data'], 1440);
-
-            $campaign = $campaign['data'];
-        }
-
-        return $campaign;
+        return Campaign::find($id);
```
katiecrane: that looks so much better!!!!

tests/Unit/Services/CampaignServiceTest.php
DFurnes: Yeah, that's what I was thinking. 💃

app/Services/CampaignService.php
DFurnes: Helllz yeah. 🎸

DFurnes:
This is looking good to me – I've tested the campaign overview, single-campaign view, campaign inbox, reviewing & tagging posts, and the edit quantity modal and everything seems to work as expected (with seed data). I think the next step is to merge this and test on QA.

PR #798, opened 2018-12-06
Remove ability to write or query campaign_run_id.

DFurnes:
#### What's this PR do?
This pull request removes the ability to create new posts or signups with `campaign_run_id` values or to filter on existing values (so `?filter[campaign_run_id]`, for example, is a no-op). We can make a follow-up pull request in a few sprints to remove the `campaign_run_id` field from the database and transformers once and for all.

I started by removing assertions that depended on `campaign_run_id`, then removed the ability to read or write that value from the application. Rather than add some short-lived tests to ensure we can't write this value anymore, let's just have a quick testing party after we deploy this.

The plan is to deploy this [once we've run the script](https://github.com/DoSomething/infrastructure/issues/61) to remove `campaign_run_id` values in production, so that even if Ashes, Phoenix, and Gambit aren't immediately updated to stop querying or writing this value, we can safely ignore it here.

#### Relevant tickets
References DoSomething/infrastructure#61.

#### How to deploy

### DIFF
--- a/resources/assets/components/SignupCard/index.js
+++ b/resources/assets/components/SignupCard/index.js
@@ -29,9 +29,6 @@ class SignupCard extends React.Component {
                 {campaign ? campaign.internal_title : signup.campaign_id}
               </h2>
               <h4 className="heading">Campaign ID: {signup.campaign_id}</h4>
-              <h4 className="heading">
-                Campaign Run ID: {signup.campaign_run_id}
-              </h4>
               {campaign ? (
                 <h4 className="heading">
                   Campaign Run Start Date: {campaign.start_date}
@@ -80,7 +77,6 @@ SignupCard.propTypes = {
     why_participated: PropTypes.string,
     signup_id: PropTypes.number,
     campaign_id: PropTypes.string,
-    campaign_run_id: PropTypes.number,
     quantity: PropTypes.number,
   }).isRequired,
   campaign: PropTypes.shape({
--- a/resources/assets/components/SignupCard/index.js
+++ b/resources/assets/components/SignupCard/index.js
@@ -7,26 +7,6 @@ import { RestApiClient } from '@dosomething/gateway';
 import { extractPostsFromSignups } from '../../helpers';
 
 class SignupCard extends React.Component {
-  /**
-   * Gets the campaign run start date.
-   *
-   * @param {Object} campaignRuns
-   * @param {String} campaignRunId
-   * @return {String}
-   */
-  getCampaignRunStartDate(campaignRuns, campaignRunId) {
-    for (const key in campaignRuns) {
-      if (campaignRuns[key].id == campaignRunId) {
-        const date = campaignRuns[key].start_date;
-
-        if (date) {
-          return campaignRuns[key].start_date.split(' ')[0];
-        }
-      }
-      return null;
-    }
-  }
-
   render() {
     const signup = this.props.signup;
     const campaign = this.props.campaign;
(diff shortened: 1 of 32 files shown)


chloealee:
I LOVE THIS PR SO MUCH!!!!!

tests/Http/PostTest.php
```diff
@@ -776,26 +761,20 @@ public function testPostsIndexAsAdminHiddenPosts()
      */
     public function testPostsIndexAsOwner()
     {
-        // Owners should only see accepted posts and their own pending/rejected posts.
-        $posts = factory(Post::class, 2)->create();
-        $rejectedPosts = factory(Post::class, 'rejected', 3)->create();
+        $userId = $this->faker->unique()->northstar_id;
+        factory(Post::class, 2)->create(['northstar_id' => $userId]);
+        factory(Post::class, 'rejected', 1)->create(['northstar_id' => $userId]);
```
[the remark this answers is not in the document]
chloealee: ah yes thank you for fixing this!!

tests/Http/Legacy/Two/ActivityApiTest.php
```diff
@@ -307,25 +304,25 @@ public function testQuantityOnActivityIndex()
...
     /**
-     * Test for retrieving a user's activity with northstar_id and campaign_run_id query params.
+     * Test for retrieving a user's activity with northstar_id and campaign_id query params.
      *
      * GET /activity?filter[]=
      * @return void
      */
     public function testActivityIndexWithNorthstarIdAndCampaignRunIdFilters()
     {
-        $signup = factory(Signup::class)->create(['northstar_id' => 17, 'campaign_run_id' => 143]);
```
katiecrane: I think we should change the title of this test to `testActivityIndexWithNorthstarIdAndCampaignIdFilters`

tests/Http/PostTest.php
```diff
@@ -776,26 +761,20 @@ public function testPostsIndexAsAdminHiddenPosts()
...
 
-        foreach ($rejectedPosts as $rejectedPost) {
-            $rejectedPost->northstar_id = $this->faker->unique()->northstar_id;
-            $rejectedPost->save();
-        }
 
-        $response = $this->withAccessToken($northstarId)->getJson('api/v3/posts');
-        $response->assertStatus(200);
-        $response->assertJsonCount(2, 'data');
+        // Owners should only see their own pending/accepted posts, and not
```
katiecrane: it might be more clear to say "Owners should be able to see their own posts of any status, but not pending or rejected posts from other users" since we are showing this user their own post which has been rejected

resources/assets/components/SignupCard/index.js
```diff
@@ -46,19 +26,12 @@ class SignupCard extends React.Component {
...
-                Campaign Run ID: {signup.campaign_run_id}
-              </h4>
               {campaign ? (
                 <h4 className="heading">
-                  Campaign Run Start Date:{' '}
-                  {this.getCampaignRunStartDate(
-                    campaign.campaign_runs.current,
-                    signup.campaign_run_id,
-                  )}
+                  Campaign Run Start Date: {campaign.start_date}
```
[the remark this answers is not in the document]
DFurnes: We sure do! Updated that in 5b408d3a.

tests/Http/Legacy/Two/ActivityApiTest.php
DFurnes: Ah, of course! Updated in 4227bd7.

tests/Http/PostTest.php
DFurnes: That reads a lot better! Updated in 4227bd7. 👍

katiecrane:
@DFurnes this is amazing!!!!

PR #799, opened 2018-12-07
Remove unused functionality.

DFurnes:
#### What's this PR do?
This pull request removes a lot of functionality that's no longer used:

screen shot 2018-12-07 at 3 46 23 pm

Here's the breakdown of what was removed:

↔️ Separates `AshesCampaignsController` (campaign activity overview & inbox) into standard CRUD-dy `CampaignsController` and `InboxController`, and renames campaign ID controller to `CampaignIdController` for clarity.

🔥 Removes web-only APIs that were once used for Rogue admin: `/posts`, `/reviews`, and `/tags`. These were replaced by Rogue's v3 API a while back, and are no longer in use.

🔥 Removes the `v2/events`, `v2/reviews`, `v2/tags`, and `v2/campaigns` routes, controllers, tests, and documentation since these were only used by Rogue's admin panel and have since been updated to use the corresponding v3 endpoints.

🔥 Removed `PagesController`, since these two methods can just use Laravel's `$router->view`.

🔥 Removes the now unused `Ashes` and `Phoenix` API clients & their config settings.

🔥 Removes Laravel's default scaffolded `PasswordController`, which is not used since this app doesn't manage passwords or password resets.

🔥 Removes the `/import` and `/export` routes, controllers, and UI elements since we no longer handle importing or exporting data in this app (that's done in Chompy & Looker)!

🔥 Removes old Capistrano config, since this application is deployed with Heroku now.

🔥 Removes some files that snuck into our repository – `test.js`, `test.csv`, and `Tag To Delete,`.

Let's double-check that these are indeed unused, and run through some good manual testing as well in case there's anywhere our tests aren't covering. Also there's no rush to deploy this, so let's wait until Monday!

This is a follow-up from some things I noticed when working on #798.

#### Relevant tickets
N/A

#### How to deploy


app/Http/Controllers/Legacy/Two/PostsController.php
```diff
@@ -3,13 +3,14 @@
...
-use Rogue\Http\Controllers\Traits\PostRequests;
+use Rogue\Http\Requests\Legacy\Two\PostRequest;
+use Rogue\Http\Controllers\Traits\FiltersRequests;
 use Rogue\Repositories\Legacy\Two\SignupRepository;
 use Rogue\Http\Transformers\Legacy\Two\PostTransformer;
 
 class PostsController extends ApiController
 {
-    use PostRequests;
+    use FiltersRequests;
```
chloealee: just curious why you got rid of PostRequests and moved this here! Just because it is easier to read and follow?

tests/Http/EventTest.php
```diff
@@ -119,11 +119,10 @@ public function testUpdatingPostEvent()
      */
     public function testDeletingPostViaApiEvent()
     {
-        // Create a post.
+        $this->mockTime('8/4/2017 18:02:00');
```
katiecrane: Don't we want this to be the time when the post is deleted? I thought that's why it was moved closer to the call to delete it.

katiecrane:
I love this PR and everything it stands for!

app/Http/Controllers/Legacy/Two/PostsController.php
DFurnes: Yup. This trait was only used in one place now (this controller), so didn't make sense to have extracted. I also prefer to be able to see what methods a controller supports by skimming the actual controller, and not have to jump into traits to see where something might be tucked away.

tests/Http/EventTest.php
DFurnes: It seemed like it was for both, but not 100% sure tbh! I can dig into what's up with this a bit more.

DFurnes: A ha, I hadn't updated the endpoint in this test, so it was posting to one I'd removed, 404ing, and not creating an event. Fixed this in 6fbccc8. Thanks for pushing me to look into this! 🎏

katiecrane: Oh hey!!! I was thinking that in this test we want to make sure it a post deleted event gets made so I thought we may want to change the time right before we delete so we can make sure we get an event upon deletion and aren't just seeing the one from creation?

DFurnes: Yeah, that's what was happening!

katiecrane: oh yup. confused myself when looking at your commit and read it as doing the opposite of what it did! those always get me!

routes/web.php
```diff
@@ -9,49 +9,32 @@
...
-$router->get('campaigns/{id}/inbox', 'Legacy\Web\AshesCampaignsController@showInbox');
-$router->get('campaigns/{id}', 'Legacy\Web\AshesCampaignsController@showCampaign')->name('campaigns.show');
-// Create, update, delete campaigns via Rogue.
-// @TODO: rename these routes (to just /campaigns) when we get off Ashes.
-$router->resource('campaign-ids', 'Legacy\Web\CampaignsController');
-
-// Exports
-$router->get('exports/{id}', 'Web\ExportController@show');
+$router->get('campaigns', 'Legacy\Web\CampaignsController@index');
+$router->get('campaigns/{id}', 'Legacy\Web\CampaignsController@show')->name('campaigns.show');
```
chloealee: yes!!! so much better then the awkward `/campaign-ids`!

chloealee:
This is the best PR ever!!! Rogue is dropping some lbs!

routes/web.php
```diff
@@ -9,49 +9,32 @@
...
 
-// Posts
-$router->post('posts', 'Legacy\Web\PostController@store');
-$router->get('posts', 'Legacy\Web\PostController@index');
-$router->delete('posts/{id}', 'Legacy\Web\PostController@destroy');
+// Create, update, delete campaigns via Rogue.
+// @TODO: Merge into CampaignsController, above.
+$router->resource('campaign-ids', 'Legacy\Web\CampaignIdsController');
```
katiecrane: I'm confused about how this bit works, can you walk me through it?

chloealee: oh wait I totally missed this it is still here! This makes sense though to me because it has to be different then showing the other campaign pages (overview, inbox) - sorry if I caused confusion!

DFurnes: Yeah, totally!

Right now, we have a `/campaigns` resource which has an "index" of campaigns & their activity, and a "show" resource with posts for a particular campaign. Then we have a separate `/campaign-ids` resource which also has campaign "index" and "show" pages, but only for information about the campaign itself.

Instead of having separate `campaign` and `campaign-id` resources, I think we could merge these overlapping views together and both simplify the taxonomy (no longer two things that refer to campaigns) and also make the admin flow a little more straightforward.

routes/web.php
chloealee: scratch this comment!

routes/web.php
DFurnes: I'm not sure when we'd get to that work, though, so maybe not the most applicable `TODO`! 🙃

app/Http/Controllers/Legacy/Web/InboxController.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace Rogue\Http\Controllers\Legacy\Web;
+
+use Rogue\Services\CampaignService;
+use Rogue\Http\Controllers\Controller;
+
+class InboxController extends Controller
```
katiecrane: Was the stuff in this file pulled out of `AshesCampaignsController`?

DFurnes: Yes! This was originally the `showInbox` method in `AshesCampaignsController`.

routes/web.php
katiecrane: AHHH I got it so this controller is only for the `campaign`s themselves? The other controller is for their activity? Could probably do some renaming down the line to make that a little less confusing but totally don't need to tackle now

DFurnes: Yeah, exactly! And yeah, it's definitely confusing right now!

DFurnes:
I snuck in one more tiny commit 3a0918b which removes three incomplete test assertions that I don't believe we're ever planning to revisit (since we no longer need that behavior for Quasar). 🐭

PR #803, opened 2018-12-10
Returns dates as a string in campaign transformer 

chloealee:
#### What's this PR do?
Returns dates as a string in campaign transformer 

Is the date returned as a string and not an object? 

#### Relevant tickets
Fixes issue in [this comment](https://github.com/DoSomething/rogue/pull/776#discussion_r239979680)

#### How to deploy

### DIFF
--- a/app/Http/Transformers/CampaignTransformer.php
+++ b/app/Http/Transformers/CampaignTransformer.php
@@ -19,7 +19,7 @@ class CampaignTransformer extends TransformerAbstract
             'id' => $campaign->id,
             'internal_title' => $campaign->internal_title,
             'start_date' => $campaign->start_date->toIso8601String(),
-            'end_date' => $campaign->end_date->toIso8601String(),
+            'end_date' => optional($campaign->end_date)->toIso8601String(),
             'created_at' => $campaign->created_at->toIso8601String(),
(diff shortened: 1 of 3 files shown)


app/Http/Transformers/CampaignTransformer.php
```diff
@@ -18,10 +18,10 @@ public function transform(Campaign $campaign)
         return [
             'id' => $campaign->id,
             'internal_title' => $campaign->internal_title,
-            'start_date' => $campaign->start_date,
-            'end_date' => $campaign->end_date,
-            'created_at' => $campaign->created_at,
-            'updated_at' => $campaign->updated_at,
+            'start_date' => $campaign->start_date->toIso8601String(),
+            'end_date' => $campaign->end_date->toIso8601String(),
```
DFurnes: Since `end_date` is optional, we should wrap this in [`optional()`](https://laravel.com/docs/5.7/helpers#method-optional).

```php
'end_date' => optional($campaign->end_date)->toIso8601String(),
```

Otherwise, we'll run into errors trying to call `->toIso8601String` on `null`.

chloealee: oh great call! Updated :)

PR #807, opened 2018-12-11
Fixes viewAll in PostPolicy to take client credentials into account

chloealee:
#### What's this PR do?
- Fixes `viewAll` in PostPolicy to take client credentials into account. 
  - This was returning `false` if a client with `admin` credentials tried to get the full `post` response and therefore didn't return the following for a client when it should have: 
```
$response['tags'] = $post->tagSlugs();
$response['source'] = $post->source;
$response['source_details'] = $post->source_details;
$response['remote_addr'] = '0.0.0.0';
$response['details'] = $post->details;
```
- Also adds a `SignupPolicy` so the `SignupTransformer` also uses the `Gate` for permission logic.
👀 

~~I thought about updating the Signup Transformer to also use the Gate but it seems like a little much for right now since this would be the only rule and is a little less intuitive then [how it is now](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/SignupTransformer.php#L38). If there are more rules in the future, I think we can add a `SignupPolicy` but not sure if it is needed now.~~ Added this in the last commit!

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/162548698) Pivotal Card.

cc @aaronschachter 

#### How to deploy

### DIFF
--- a/app/Policies/PostPolicy.php
+++ b/app/Policies/PostPolicy.php
@@ -4,9 +4,11 @@ namespace Rogue\Policies;
 
 use Rogue\Models\Post;
 use Illuminate\Auth\Access\HandlesAuthorization;
+use Rogue\Http\Controllers\Traits\AuthorizesWithToken;
 
 class PostPolicy
 {
+    use AuthorizesWithToken;
     use HandlesAuthorization;
 
     /**
@@ -27,18 +29,7 @@ class PostPolicy
      */
     public function viewAll($user, Post $post)
     {
-        // If this is a machine token, show full post:
-        if (token()->exists() && ! token()->id()) {
-          return true;
-        }
-
-        // If this is an anonymous request, only show public fields:
-        if ($user == null) {
-          return false;
-        }
-
-        // Otherwise, only allow staffers & post owner to see  full post:
-        return is_staff_user() || $user->northstar_id === $post->northstar_id;
+        return $this->allowOwnerStaffOrMachine($user, $post);
     }
 
     /**
--- a/app/Policies/PostPolicy.php
+++ b/app/Policies/PostPolicy.php
@@ -8,8 +8,7 @@ use Rogue\Http\Controllers\Traits\AuthorizesWithToken;
 
 class PostPolicy
 {
-    use AuthorizesWithToken;
-    use HandlesAuthorization;
+    use AuthorizesWithToken, HandlesAuthorization;
 
     /**
      * Create a new policy instance.
--- a/app/Policies/PostPolicy.php
+++ b/app/Policies/PostPolicy.php
@@ -27,11 +27,17 @@ class PostPolicy
      */
     public function viewAll($user, Post $post)
     {
-        // If there is no user and it is not a client with admin credentials, return false.
-        if ($user === null && is_staff_user() === false) {
-            return false;
+        // If this is a machine token, show full post:
+        if (token()->exists() && ! token()->id()) {
+          return true;
         }
 
+        // If this is an anonymous request, only show public fields:
+        if ($user == null) {
+          return false;
+        }
+
+        // Otherwise, only allow staffers & post owner to see  full post:
         return is_staff_user() || $user->northstar_id === $post->northstar_id;
     }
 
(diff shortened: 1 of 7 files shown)


app/Policies/PostPolicy.php
```diff
@@ -27,7 +27,8 @@ public function __construct()
      */
     public function viewAll($user, Post $post)
     {
-        if ($user === null) {
+        // If there is no user and it is not a client with admin credentials, return false.
+        if ($user === null && is_staff_user() === false) {
```
katiecrane: Do you still need the `$user === null` part?

chloealee: @katiecrane great q! I think we want that in case it is an anonymous user.

katiecrane: @chloealee wouldn't an anonymous user also not be a staff user?

chloealee: oh that is a good point, @katiecrane ! I wonder though if `$user->northstar_id === $post->northstar_id;` will break if there is no `$user` to try to get the `northstar_id` from

DFurnes: Yeah, it might be good to separate these into two different cases for readability.

DFurnes: Oh, although that doesn't work since then we'd prevent users from seeing their own info. Bleh.

DFurnes:
> I thought about updating the Signup Transformer to also use the Gate but it seems like a little much for right now since this would be the only rule and is a little less intuitive then [how it is now](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/SignupTransformer.php#L38). If there are more rules in the future, I think we can add a `SignupPolicy` but not sure if it is needed now.

I'd love to use policies for both of these at some point in the future, for consistency and discoverability. To reduce code duplication, we could have a shared `AuthorizesWithToken` trait, e.g:

```php
trait AuthorizesWithToken
{
    public function allowOwnerStaffOrMachine($user, Model $model)
    {
        // ...this logic
    }
}

class PostPolicy
{
    public function viewAll($user, Post $post)
    {
        return $this->allowOwnerOrStaff($user, $post)
    }
}
```

app/Policies/PostPolicy.php
chloealee: ah this is a lot more readable, thanks @DFurnes ! I just updated this :)

Just want to double check - is it safe to assume that all machines should have the `admin` scope, right? I see that they all have it now but just want to be sure this is safe to assume in the future too.

DFurnes: I'm thinking it makes sense to remove that scope in the future, since it's confusing and possible to misuse. For Rogue, I think allowing any machine client to access full profiles is a-okay.

chloealee:
@DFurnes that makes sense to me - cleaned up and added so that the `SignupTransformer` now also uses the Gate!

app/helpers.php
```diff
@@ -98,8 +98,7 @@ function is_staff_user()
 {
     $isStaffUser = token()->exists() && in_array(token()->role, ['admin', 'staff']);
     $isStaffClient = token()->exists() && in_array('admin', token()->scopes());
-
-    if ($isStaffUser || $isStaffClient) {
+ if ($isStaffUser || $isStaffClient) {
```
DFurnes: Indentation looks a little off here! 👾

app/Providers/AuthServiceProvider.php
```diff
@@ -15,6 +17,7 @@ class AuthServiceProvider extends ServiceProvider
      */
     protected $policies = [
         Post::class => PostPolicy::class,
+        Signup::class =>SignupPolicy::class,
```
DFurnes: Let's add a space after the `=>` on this line! 👾

app/Policies/PostPolicy.php
```diff
@@ -4,9 +4,11 @@
 
 use Rogue\Models\Post;
 use Illuminate\Auth\Access\HandlesAuthorization;
+use Rogue\Http\Controllers\Traits\AuthorizesWithToken;
 
 class PostPolicy
 {
+    use AuthorizesWithToken;
     use HandlesAuthorization;
```
DFurnes: Totally subjective, but if you want you can combine these traits into one line:

```php
use AuthorizesWithToken, HandlesAuthorization;
```

PR #810, opened 2018-12-12
Fixes fatal issue on the campaign "show" page.

DFurnes:
#### What's this PR do?
This pull request fixes up a [fatal error](https://dosomething.slack.com/archives/C0YGXUE01/p1544633480025200) when viewing imported campaigns without an `impact_doc` set, and adds a "Campaign ID" field for a lil' better admin UX.

👀

This was dying because we'd send `null` through `url()` and that would return the `UrlGenerator` instance rather than a string – since this is already a fully formed URL, I just removed that helper! I've also added a fallback so we show a `–` here if it's empty.

#### Relevant tickets
References this [Slack conversation](https://dosomething.slack.com/archives/C0YGXUE01/p1544633480025200).

### DIFF
--- a/resources/views/campaign-ids/show.blade.php
+++ b/resources/views/campaign-ids/show.blade.php
@@ -14,7 +14,11 @@
                 <p>{{ $campaign->cause }}</p>
 
                 <h3>Proof of Impact</h3>
-                <p><a href="{{ url($campaign->impact_doc) }}" target="_blank">{{ $campaign->impact_doc }}</a></p>
+                @if($campaign->impact_doc)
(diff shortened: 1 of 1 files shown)


resources/views/campaign-ids/show.blade.php
```diff
@@ -10,11 +10,18 @@
                 <h3>Internal Campaign Name</h3>
                 <p>{{ $campaign->internal_title }}</p>
 
+                <h3>Campaign ID</h3>
+                <p>{{ $campaign->id }}</p>
+
```
chloealee: this is great! This actually was a suggestion from when we went through with Adam that this would be a nice to have - thank you for adding!

PR #812, opened 2018-12-12
Doesn't pass $signup to SignupTransformer if it has been soft deleted when hitting GET /posts?include=signup

chloealee:
#### What's this PR do?
- If there is no `$signup` (it has been soft deleted and we don't have cascading deletes yet) when hitting the` /posts?include=signup`, don't call `SignupTransformer`
- Lots of StyleCi cleanup!

- Test endpoint that doesn't work in [this](https://www.pivotaltracker.com/n/projects/2019429/stories/158442898) Pivotal card and it should work! 

We're getting a lot of Papertrail errors for this and we want to prevent false alarms!
```
Jul 05 14:19:32 rogue-prod app/web.1: [05-Jul-2018 21:19:32 UTC] [2018-07-05 21:19:32] production.ERROR: Type error: Argument 1 passed to Rogue\Http\Transformers\Legacy\Two\SignupTransformer::transform() must be an instance of Rogue\Models\Signup, null given, called in /app/vendor/league/fractal/src/Scope.php on line 338 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Type error: Argument 1 passed to Rogue\\Http\\Transformers\\Legacy\\Two\\SignupTransformer::transform() must be an instance of Rogue\\Models\\Signup, null given, called in /app/vendor/league/fractal/src/Scope.php on line 338 at /app/app/Http/Transformers/Legacy/Two/SignupTransformer.p…
``` 

#### Relevant tickets
Fixes these Pivotal Cards:
https://www.pivotaltracker.com/n/projects/2019429/stories/158870350
https://www.pivotaltracker.com/n/projects/2019429/stories/158442898

#### How to deploy

### DIFF
--- a/app/Http/Transformers/Legacy/Two/PostTransformer.php
+++ b/app/Http/Transformers/Legacy/Two/PostTransformer.php
@@ -67,6 +67,10 @@ class PostTransformer extends TransformerAbstract
      */
     public function includeSignup(Post $post)
     {
+        if (! $post->signup) {
+            return;
+        }
+
         // Don't include posts but include the user.
         $transformer = (new SignupTransformer)->setDefaultIncludes(['user']);
 
--- a/app/Console/Commands/ImportAshesCampaigns.php
+++ b/app/Console/Commands/ImportAshesCampaigns.php
@@ -73,8 +73,8 @@ class ImportAshesCampaigns extends Command
                     $campaign = $this->createCampaign($legacy_campaign['run_id'], $legacy_campaign);
                 // If the campaign_id does not equal the next record's campaign_id or the previous record's campaign id, this is either the latest run or there is only one campaign run, so keep the original campaign_id as the id in Rogue.
                 } elseif ($legacy_campaigns_csv->fetchOne($iterator) &&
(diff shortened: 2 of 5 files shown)


app/Http/Transformers/Legacy/Two/PostTransformer.php
```diff
@@ -67,6 +67,10 @@ public function transform(Post $post)
      */
     public function includeSignup(Post $post)
     {
+        if (! $post->signup) {
+            return;
```
chloealee: I feel like there should be a more elegant response to indicate that there is no signup included with the post because it has been deleted but when I trie dto do a custom message in a simple array, I'd get this message: 
```
Exception: Invalid return value from League\Fractal\TransformerAbstract::includeSignup(). Expected League\Fractal\Resource\ResourceAbstract, received array. in file /home/vagrant/DoSomething.org/rogue/vendor/league/fractal/src/TransformerAbstract.php on line 185
```

DFurnes: I think returning `null` here is great!

PR #816, opened 2019-01-02
Adds arbitrary limit to when grabbing campaigns to avoid pagination

chloealee:
#### What's this PR do?
Adds arbitrary limit to when grabbing campaigns to avoid pagination. I updated some things in [this](https://github.com/DoSomething/rogue/pull/814/files) PR to fix the bug to grab campaign names. However, after testing, it seemed like only some campaign names were showing up on the user page. This is because of pagination! The front end would make a call to `/campaigns` and filter by campaign id according to a user's signup. If a user had many signups, only the first 20 campaign names would show up. This adds a limit of `1000` to hopefully do a quick fix for this. 

In the future, I think we should add something like `limit=all` to the API to avoid this and return all records instead of a paginated collection.

Do all the campaign names in the user profile page in [this](https://www.pivotaltracker.com/n/projects/2019429/stories/162742881) Pivotal Card show up? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/162742881) Pivotal Card. 
Related to work down in #814 

#### How to deploy

### DIFF
--- a/resources/assets/components/UserOverview/index.js
+++ b/resources/assets/components/UserOverview/index.js
@@ -72,6 +72,8 @@ class UserOverview extends React.Component {
         filter: {
           id: ids.join(),
         },
+        //@TODO: when we update TransformsRequest.php to include limit=all,
+        // update this API request to use that instead of abritrary 1000.
         limit: 1000,
       })
       .then(json =>
--- a/resources/assets/components/UserOverview/index.js
+++ b/resources/assets/components/UserOverview/index.js
@@ -72,6 +72,7 @@ class UserOverview extends React.Component {
         filter: {
           id: ids.join(),
         },
+        limit: 1000,
       })
       .then(json =>
         this.setState({
--- a/resources/assets/components/UserOverview/index.js
+++ b/resources/assets/components/UserOverview/index.js
@@ -72,9 +72,10 @@ class UserOverview extends React.Component {
         filter: {
           id: ids.join(),
         },
-        //@TODO: when we update TransformsRequest.php to include limit=all,
-        // update this API request to use that instead of abritrary 1000.
-        limit: 1000,
+        // @TODO: update this to paginate calls to smaller batches or use GraphQL
+        // to be able to request only the subset of campaign fields we actually use for the UI
+        // (to make sure performance isn't hit).
+        limit: 100,
       })
       .then(json =>
         this.setState({


resources/assets/components/UserOverview/index.js
```diff
@@ -72,6 +72,9 @@ class UserOverview extends React.Component {
         filter: {
           id: ids.join(),
         },
+        //@TODO: when we update TransformsRequest.php to include limit=all,
+        // update this API request to use that instead of abritrary 1000.
+        limit: 1000,
```
DFurnes: We [cap this limit at `100`](https://github.com/DoSomething/rogue/blob/db848e7114f4ead73c89d960e0b93d1445df3b1f/app/Http/Controllers/Traits/TransformsRequests.php#L116-L120), and would likely want to continue doing that since allowing more than that number of items in a response could lead to pretty unwieldy request size & performance issues.

DFurnes: If we do need to get this number of items all at once, I'd recommend paginating the calls (into multiple smaller batches) & maybe looking into using GraphQL to be able to request only the subset of campaign fields we actually use for the UI (since as the number of things on a campaign grows over time, this will get less and less performant).

chloealee: @DFurnes ah thanks for linking to [this](https://github.com/DoSomething/rogue/blob/db848e7114f4ead73c89d960e0b93d1445df3b1f/app/Http/Controllers/Traits/TransformsRequests.php#L116-L120)! Is this cap at `100` the number of pages or the number of objects? 

The issue here is the API response only returns 20 campaigns in a page so only 20 of the campaigns on the user's page have the correct campaign title while the rest have the campaign number instead of the title.

DFurnes: That's 100 items per response. If you need more than 100 per page, you should make multiple paginated calls to request all the items you need.

chloealee: ah gotcha - thanks! So for this quick fix, can is it ok to change this to be `limit: 100` and then create a new ticket to what you said in [this](https://github.com/DoSomething/rogue/pull/816#discussion_r244875162) comment?

DFurnes: Yup, that sounds good to me. 👍

chloealee: cool this is updated and ready for review! The Pivotal Card is [here](https://www.pivotaltracker.com/n/projects/2019429/stories/162987212)!

PR #817, opened 2019-01-03
adds check to see if auth()->user() after or conditional in PostRepository

chloealee:
#### What's this PR do?
adds check to see if `auth()->user()` after `||` conditional in PostRepository to make sure it exists before checking to avoid errors we saw [here](https://dosomething.slack.com/archives/C0YGXUE01/p1546537309014400)

👀 

#### Relevant tickets
Fixes work done in #815 

#### How to deploy

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -95,7 +95,7 @@ class PostRepository
         // If this is a share-social type post, auto-accept.
         $post->status = $post->type === 'share-social' ? 'accepted' : 'pending';
 
-        $isAdminOrStaff = auth()->user() && auth()->user()->role === 'admin' || auth()->user()->role === 'staff';
+        $isAdminOrStaff = auth()->user() && auth()->user()->role === 'admin' || auth()->user() && auth()->user()->role === 'staff';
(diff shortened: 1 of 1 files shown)


app/Repositories/PostRepository.php
```diff
@@ -95,7 +95,7 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
         // If this is a share-social type post, auto-accept.
         $post->status = $post->type === 'share-social' ? 'accepted' : 'pending';
 
-        $isAdminOrStaff = auth()->user() && auth()->user()->role === 'admin' || auth()->user()->role === 'staff';
+        $isAdminOrStaff = auth()->user() && auth()->user()->role === 'admin' || auth()->user() && auth()->user()->role === 'staff';
```
DFurnes: Gah, tricky! It might read nicely to combine these into an `is_array` check:

```php
$isAdminOrStaff = auth()->user() && in_array(auth()->user()->role, ['admin', 'staff'])
```

chloealee: thanks @DFurnes ! I like this way better!

PR #818, opened 2019-01-04
Fixes bug that shows 11 pending posts for Grab the mic run 1 when the inbox is empty

chloealee:
#### What's this PR do?
- Fixes bug that shows 11 pending posts for Grab the mic run 1 when the inbox is empty.
- This is happening because the query isn't omitting signups that are deleted - there was a deleted signup that had pending posts and the pending count thus returned 11 because it didn't omit these posts.
- This PR updates the `getPostTotals` query to remove any signups that have been deleted.

👀 
On prod, is the Grab the mic run one pending count 0? 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/162968074) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Services/CampaignService.php
+++ b/app/Services/CampaignService.php
@@ -70,6 +70,7 @@ class CampaignService
                     DB::raw('SUM(case when posts.status = "rejected" && posts.deleted_at is null then 1 else 0 end) as rejected_count'))
                 ->where('signups.campaign_id', '=', $campaign['id'])
                 ->groupBy('signups.campaign_id')
+                ->whereNull('signups.deleted_at')
                 ->first();
     }
 


app/Services/CampaignService.php
```diff
@@ -70,6 +70,7 @@ public function getPostTotals($campaign)
                     DB::raw('SUM(case when posts.status = "rejected" && posts.deleted_at is null then 1 else 0 end) as rejected_count'))
                 ->where('signups.campaign_id', '=', $campaign['id'])
                 ->groupBy('signups.campaign_id')
+                ->whereNull('signups.deleted_at')
```
DFurnes: Do we have an index on this column? If not, that might be why this is timing out for you.

PR #821, opened 2019-01-11
Makes sure pending count query only counts photo and text posts

chloealee:
#### What's this PR do?
Updates `getPostTotals` query to only count `photo` and `text` posts so pending count is accurate and bug #818 is properly fixed (was incorrect about why this was happening in #818, discovered in [this comment](https://github.com/DoSomething/rogue/pull/820/files#issuecomment-453563430)). 

👀
On prod, is the Grab the mic run one pending count 0?

Note: if we ever start showing accepted or rejected count in Rogue admin, we will want to rework this query since this will only count accepted/rejected photo and post counts. 

Also, there are 11 `voter-reg` posts that are `pending` - I believe this is a bug since they should all be `accepted`. We should run an update script to accept these (this is probably the real fix for the bug if we don't want to update this query).

#### Relevant tickets
Fixes issue in [this](https://www.pivotaltracker.com/n/projects/2019429/stories/162968074) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Services/CampaignService.php
+++ b/app/Services/CampaignService.php
@@ -68,6 +68,7 @@ class CampaignService
                     DB::raw('SUM(status = "pending") as pending_count'),
                     DB::raw('SUM(status = "rejected") as rejected_count'))
                 ->whereIn('status', ['accepted', 'pending', 'rejected'])
+                ->whereIn('type', ['photo', 'text'])
                 ->where('campaign_id', '=', $campaign['id'])
                 ->whereNull('deleted_at')
                 ->first();
--- a/app/Services/CampaignService.php
+++ b/app/Services/CampaignService.php
@@ -68,6 +68,7 @@ class CampaignService
                     DB::raw('SUM(status = "pending") as pending_count'),
                     DB::raw('SUM(status = "rejected") as rejected_count'))
                 ->whereIn('status', ['accepted', 'pending', 'rejected'])
+                // @TODO: if we ever want to expose other post types in Rogue to review/count, update this.
                 ->whereIn('type', ['photo', 'text'])
                 ->where('campaign_id', '=', $campaign['id'])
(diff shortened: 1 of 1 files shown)


katiecrane:
> Also, there are 11 voter-reg posts that are pending - I believe this is a bug since they should all be accepted. We should run an update script to accept these (this is probably the real fix for the bug if we don't want to update this query).

So strange! I don't think these should be `accepted` or `pending`, but should be one of the statuses listed here:
https://github.com/DoSomething/chompy/wiki/How-Rock-The-Vote-(RTV)-records-are-imported#status-hierarchy

chloealee:
@katiecrane ohohoh you're so right! Thanks for flagging that! 

@jng should I create a card of the 11 records and to find a way to update these to the statuses they should be instead of `pending`?

app/Services/CampaignService.php
```diff
@@ -68,6 +68,8 @@ public function getPostTotals($campaign)
                     DB::raw('SUM(status = "pending") as pending_count'),
                     DB::raw('SUM(status = "rejected") as rejected_count'))
                 ->whereIn('status', ['accepted', 'pending', 'rejected'])
+                // @TODO: if we ever want to expose other post types in Rogue to review/count, update this.
+                ->whereIn('type', ['photo', 'text'])
```
DFurnes: I wonder if this makes sense to expose in a static `$reviewable` property on the post model?

chloealee: ooo yes I love this idea! Would this be something like [`scopeWithTag`](https://github.com/DoSomething/rogue/blob/master/app/Models/Post.php#L335-L340)?

DFurnes: Oh yeah, I had been thinking just a public static property that you could grab here (like `Post::$reviewable`) but a query scope makes even better sense!

PR #822, opened 2019-01-15
Adds whereReviewable Query Scope

chloealee:
#### What's this PR do?
- Adds a `whereReviewable` query scope to the Post model - this only includes `photo` and `text` posts since those are the only types of posts that are currently reviewable in Rogue admin. 
- Updates `getPendingPostTotals` to run more efficiently and use `whereReviewable` query scope

👀 

Suggested in [this comment](https://github.com/DoSomething/rogue/pull/821#discussion_r247269939).

#### Relevant tickets
Related to #821 

#### How to deploy


app/Services/CampaignService.php
```diff
@@ -62,17 +63,16 @@ public function getCampaignIdsFromSignups()
...
-                ->whereIn('type', ['photo', 'text'])
-                ->where('campaign_id', '=', $campaign['id'])
-                ->whereNull('deleted_at')
-                ->first();
+        return Post::select(
+                'campaign_id',
+                DB::raw('SUM(status = "accepted") as accepted_count'),
+                DB::raw('SUM(status = "pending") as pending_count'),
+                DB::raw('SUM(status = "rejected") as rejected_count'))
+            ->whereIn('status', ['accepted', 'pending', 'rejected'])
```
katiecrane: I know this wasn't changed in this PR, but why do we need this line?

app/Models/Post.php
```diff
@@ -171,7 +171,7 @@ public function getMediaPath()
     }
 
     /**
-     * Get the reactions associated with this post.
+     * Get the sibilngs associated with this post.
```
DFurnes: Typo! `sibilngs` → `siblings`

app/Services/CampaignService.php
```diff
@@ -62,17 +63,16 @@ public function getCampaignIdsFromSignups()
...
+            ->reviewable()
+            ->where('campaign_id', '=', $campaign['id'])
+            ->whereNull('deleted_at')
```
DFurnes: Can you double-check if this is necessary now? By default, Eloquent [should exclude](https://laravel.com/docs/5.7/eloquent#querying-soft-deleted-models) soft-deleted records.

app/Services/CampaignService.php
DFurnes: I'm also interested in this!

katiecrane: I'm curious if it was to exclude `voter_reg` posts since the new scope will take care of that 🔭

app/Models/Post.php
chloealee: oops thank you!

app/Services/CampaignService.php
chloealee: @katiecrane great q! I was confused by this when I was working on this too! But this is to exclude the `voter-reg` statuses :)

app/Services/CampaignService.php
chloealee: @DFurnes ah you're right! Just got rid of it :)

app/Services/CampaignService.php
```diff
@@ -89,6 +88,7 @@ public function getPendingPostTotals($campaigns)
                 ->leftJoin('posts', 'signups.id', '=', 'posts.signup_id')
                 ->selectRaw('signups.campaign_id, count(posts.id) as pending_count')
                 ->where('status', '=', 'pending')
+                ->whereIn('type', ['photo', 'text'])
```
chloealee: @DoSomething/team-bleed I think we also have to update here so that these numbers are the same on the `/campaigns` page and `/campaigns/:campaign_id`

app/Services/CampaignService.php
chloealee: oh wait - I think I lied because the query is specifically counting `accepted`, `rejected`, and `pending` - thanks for flagging and deleting this now!

app/Services/CampaignService.php
DFurnes: Can we update this query to run on the `posts` table as well (like we did for `getPostTotals` above)? I think that'd probably be another big performance win for this interface & should be a pretty straightforward change.

chloealee: ah yes I was going to do that but when I first looked thought we needed the join so didn't but realize now we totally don't! thanks @DFurnes , updated!

PR #823, opened 2019-01-16
Adds actions table - part of Post Metadata work

chloealee:
#### What's this PR do?
Adds the `actions` table for Post Metadata work (second thing on checklist [here](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/32)).

👀 
Are all columns accurate? Did I miss any columns? Are there any missing or unnecessary indexes?  

#### Relevant tickets

#### How to deploy

### DIFF
--- a/database/migrations/2019_01_16_174349_add_actions_table.php
+++ b/database/migrations/2019_01_16_174349_add_actions_table.php
@@ -35,6 +35,6 @@ class AddActionsTable extends Migration
      */
     public function down()
     {
-        Schema::drop('tagging_tagged');
+        Schema::drop('actions');
     }
 }
--- a/database/migrations/2019_01_16_174349_add_actions_table.php
+++ b/database/migrations/2019_01_16_174349_add_actions_table.php
@@ -16,12 +16,12 @@ class AddActionsTable extends Migration
         Schema::create('actions', function (Blueprint $table) {
             $table->increments('id');
             $table->string('name', 255)->index()->comment('The action name of the post.');
-            $table->integer('campaign_id')->index()->comment('The campaign id the post is associated to.');
+            $table->unsignedInteger('campaign_id')->index()->comment('The campaign id the post is associated to.');
+            $table->foreign('campaign_id')->references('id')->on('campaigns')->index()->comment('The campaign id the post is associated to.');
             $table->string('post_type', 255)->index()->comment('The type of post the post is (e.g. photo, text, voter-reg, etc).');
             $table->boolean('reportback')->comment('Whether or not the post should count as a reportback.');
             $table->boolean('civic_action')->comment('Whether or not the post should count as a civic action.');
             $table->boolean('scholarship_entry')->comment('Whether or not the post should count as a scholarship entry.');
-            $table->string('legacy_action_name', 255)->index()->nullable()->comment('The action from the posts table pre post metadata.');
             $table->boolean('active')->comment('Whether or not the action is active.');
             $table->string('noun', 255)->nullable()->comment('The noun of the post (e.g. jeans).');
             $table->string('verb', 255)->nullable()->comment('The verb of of the post (e.g. donated).');
--- a/database/migrations/2019_01_16_174349_add_actions_table.php
+++ b/database/migrations/2019_01_16_174349_add_actions_table.php
@@ -1,0 +1,40 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class AddActionsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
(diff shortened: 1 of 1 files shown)


database/migrations/2019_01_16_174349_add_actions_table.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        Schema::create('actions', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('name', 255)->index()->comment('The action name of the post.');
+            $table->integer('campaign_id')->index()->comment('The campaign id the post is associated to.');
+            $table->string('post_type', 255)->index()->comment('The type of post the post is (e.g. photo, text, voter-reg, etc).');
+            $table->boolean('reportback')->comment('Whether or not the post should count as a reportback.');
+            $table->boolean('civic_action')->comment('Whether or not the post should count as a civic action.');
+            $table->boolean('scholarship_entry')->comment('Whether or not the post should count as a scholarship entry.');
+            $table->string('legacy_action_name', 255)->index()->nullable()->comment('The action from the posts table pre post metadata.');
```
chloealee: @DoSomething/team-bleed I'm not sure if we actually need this column? I'm thinking when we do the backfill, we can move what was in the `action` column in the `posts` table to just be the `name` in this table? More in this convo [here]( https://dosomething.slack.com/archives/C2BLZGBPH/p1547590807031100?thread_ts=1547579556.028900&cid=C2BLZGBPH).

katiecrane: @chloealee yeah I think it makes sense to fill that information into the `name` column!

database/migrations/2019_01_16_174349_add_actions_table.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::drop('tagging_tagged');
```
katiecrane: Update to `actions` 🔍 😃

chloealee: oh whoops! Thanks for catching that @katiecrane ! :)

database/migrations/2019_01_16_174349_add_actions_table.php
DFurnes: That sounds reasonable. We'll just want to be careful not to mess w/ those names until we stop relying on them!

database/migrations/2019_01_16_174349_add_actions_table.php
```diff
@@ -0,0 +1,40 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('actions', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('name', 255)->index()->comment('The action name of the post.');
+            $table->integer('campaign_id')->index()->comment('The campaign id the post is associated to.');
+            $table->string('post_type', 255)->index()->comment('The type of post the post is (e.g. photo, text, voter-reg, etc).');
```
DFurnes: Do we want to support actions that could have multiple post types? (For example, the old Transfomers campaign where you could reportback on how you beat bullying using either a photo or a text post?)

chloealee: ohh interesting, @DFurnes ! I think I am in the same boat as @katiecrane and think of that as two separate actions?

DFurnes: I don't have strong opinions either way, but let's make sure whatever we decide matches with the way the campaigns team will want to use this!

chloealee: Great, point! Thanks for flagging and I will write in my notes to run by campaigns !

database/migrations/2019_01_16_174349_add_actions_table.php
```diff
@@ -0,0 +1,42 @@
...
+            $table->foreign('campaign_id')->references('id')->on('campaigns')->index()->comment('The campaign id the post is associated to.');
+            $table->string('post_type', 255)->index()->comment('The type of post the post is (e.g. photo, text, voter-reg, etc).');
+            $table->boolean('reportback')->comment('Whether or not the post should count as a reportback.');
+            $table->boolean('civic_action')->comment('Whether or not the post should count as a civic action.');
+            $table->boolean('scholarship_entry')->comment('Whether or not the post should count as a scholarship entry.');
+            $table->boolean('active')->comment('Whether or not the action is active.');
+            $table->string('noun', 255)->nullable()->comment('The noun of the post (e.g. jeans).');
+            $table->string('verb', 255)->nullable()->comment('The verb of of the post (e.g. donated).');
+            $table->timestamps();
+            $table->softDeletes();
```
chloealee: @DoSomething/team-bleed forgot to add `timestamps` and `softDeletes` to this table!

katiecrane: ah good catch!!

PR #824, opened 2019-01-16
Creates/Updates Models for Post Metadata work

chloealee:
#### What's this PR do?
- Updates models and `posts` tables for Post Metadata work:
  - Creates an `Action` model and adds the following relations:
    - belongs to a campaign
    - has many posts 
  - Adds "has many actions" relation to `Campaign` model
  - Adds "has one action" relation to the Post model

(Third checkbox in [this](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/32) discussion.)

Did I get the relations right between the models? 

#### Relevant tickets

#### How to deploy


app/Models/Action.php
```diff
@@ -0,0 +1,51 @@
...
+class Action extends Model
+{
+    use SoftDeletes;
+
+    /**
+     * The attributes that should be mutated to dates.
+     *
+     * @var array
+     */
+    protected $dates = ['deleted_at', 'start_date', 'end_date'];
```
DFurnes: Is `deleted_at` something that's actually visible on the model, or just used behind the scenes for soft deletes?

chloealee: ah great call @DFurnes ! Deleted this whole thing bc there is no `start_date` or `end_date` for this - was copied over by mistake :)

DFurnes: Ahhh I didn't even notice that part!! 🙃

app/Console/Commands/ImportPrePostMetaDataActions.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace Rogue\Console\Commands;
+
+use League\Csv\Reader;
+use Rogue\Models\Action;
+use Illuminate\Console\Command;
+
+class ImportPrePostMetaDataActions extends Command
```
DFurnes: I think this file snuck in! 👻

chloealee: eek you're right ! just removed :)

app/Models/Action.php
DFurnes: @chloealee I think this snuck back in when getting rid of that other sneaky file!

PR #825, opened 2019-01-18
Backfill Commands for Post Metadata and Update to Posts table with action_id

chloealee:
#### What's this PR do?
- Adds a command, `rogue:prepostmetadataactionsimport`, to import backfill of actions to the `actions` table from a CSV (pulled distinct actions from DB and had campaigns fill in missing data). 
- Adds `action_id` column to the the `posts` table. 
- Adds a command, `rogue:updatepostswithactionids`, to update all posts to have their corresponding `action_id`s.
- Ran `composer update`.

Does this logic make sense? 

#### Relevant tickets

4th check box in [Post Metadata Tech Brief](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/32)

#### How to deploy


actions_backfill_test.csv
```diff
@@ -0,0 +1,74 @@
...
+,,,,,,,,,,
+,,,,,,,,,,
+,,,,,,,,,,
+,,,,,,,,,,
+,,,,,,,,,,
+,,,,,,,,,,
+,,,,,,,,,,
+,,,,,,,,,,
+,,,,,,,,,,
+,,,,,,,,,,
```
chloealee: will delete this file once this is approved!

katiecrane: @chloealee why does this need to be deleted?

app/Console/Commands/UpdatePostsWithActionIds.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        $this->line('rogue:updatepostswithactionids: Staring script to add action_ids to posts!');
+
+        // Grab all of the actions in the actions table.
+        $actions = Action::all();
+
+        // For each action, find all of its posts and add the action_id to each of them.
+        foreach ($actions as $action) {
+            DB::table('posts')
+                ->where('action', $action->name)
```
katiecrane: Won't this be `default` for most `posts`? I thought on the import these would have something more like an internal title, so I wouldn't have thought we'd be finding matches here.

app/Console/Commands/ImportPrePostMetaDataActions.php
```diff
@@ -0,0 +1,82 @@
...
+        // Import each backfill action
+        $this->line('rogue:prepostmetadataactionsimport: Loading in csv from ' . $path);
+
+        foreach ($backfill_actions as $backfill_action) {
+            // See if the action exists
+            $existing_action = Action::where('campaign_id', $backfill_action['campaign_id'])->where('name', $backfill_action['action'])->first();
+
+            // Create the action if there isn't one already
+            if (! $existing_action) {
+                $action = Action::forceCreate([
```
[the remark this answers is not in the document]
chloealee: ah great call!

actions_backfill_test.csv
chloealee: @katiecrane this is just a test file that I'm working with on my local!

app/Console/Commands/UpdatePostsWithActionIds.php
chloealee: @katiecrane yes, most are `default` but there are others too! Some have something like `action-NA`, `action-1`, `affirmation`, `june-2018-turbovote`, etc.

katiecrane: @chloealee won't only the ones with an exact match on the action name end up with an `action_id` though?

chloealee: @katiecrane yup! Which we will import from the csv that we pulled that got distinct actions from the posts in the database. I don't think I'm understanding what you're saying here?

katiecrane: @chloealee ah sorry for being confusing! I thought we wanted to give every post an `action_id` with this script, but it seems to that posts with `action` of `default` will NOT get an `action_id` here?

chloealee: @katiecrane don't apologize, this is a little confusing! 

I _think_ I understand what you're asking but sorry if this is just repeating what you already know! So we have pulled a CSV that looks like this (dummy data):
screen shot 2019-01-18 at 5 43 08 pm

We first import all of these into the `actions` table so there will be may actions with the name `default` but if they are under different campaigns and/or have different post types, then it will be a different action and have a different `action_id`. 

Then, we run this part of the script and go through each action and find all the posts that the action has (determined if they have the same `campaign_id`, `name`/`action`, and post `type`. Then we'll add that `action_id` to those posts. 

Does that make sense?

katiecrane: @chloealee YES OMG THANK YOU!!! I didn't look at your test csv carefully enough, I didn't realize action titles would be `default` when we import!

chloealee: ok yay glad it makes sense!! It could have very well been that I did something wrong too bc I started writing this script way differently initially and I feel like I've gotten mixed up during this! Thanks for confirming logic makes sense! :)

katiecrane: Thanks for walking me through it @chloealee !

PR #827, opened 2019-01-25
Updates logic to include Post Metadata action when creating a post 

chloealee:
#### What's this PR do?
This PR updates the logic around CRUD Posts to include Post Metadata (adding the `action_id`). Everything should still work for all other systems and no one else should see the changes! 
- When a post is created, it grabs the `action_id` from the payload. If this is not included in the payload, it goes into the database and looks for this `action_id`, found by querying the `actions` table by `campaign_id`, `post_type`, and `name` (all info that should come in with the post request). 
- This PR also updates the Post Transformer to include the `action_id` and sends along the `action_id` to Customer.io and Quasar as well. 
- Updates documentation.
- Updates tests and adds seeders to get them to pass. 
- Adds a new test to make sure all logic still works, even when an `action_id` it not passed when creating a post via `POST v3/api/posts`.
- Finally, this PR includes tests for different scenarios.

👀 
Did I miss anything? Does everything still work as it did before?

#### Relevant tickets

Part of 5th check box in [Post Metadata Tech Brief](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/32)

#### How to deploy


app/Repositories/PostRepository.php
```diff
@@ -77,6 +79,21 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
...
+        // Get the action_id either from the payload or the DB.
+        if (isset($data['action_id'])) {
+            $actionId = $data['action_id'];
+        } else {
+            $type = isset($data['type']) ? $data['type'] : 'photo';
+            $action = Action::where('campaign_id', $signup->campaign_id)->where('post_type', $type)->where('name', $data['action'])->first();
+
+            if (! $action) {
+                // Throw an exception that no action exists... should this be caught at an earlier point though?
+                throw new ModelNotFoundException('Action not found.');
```
chloealee: Is there a better exception that we should throw here? This might be too rigid - if an action doesn't exist in the database and anything gets out of sync (an action isn't created first for a post type and isn't added in Contentful for example) creating a post will throw an error and might cause some issues.

app/Http/Requests/PostRequest.php
```diff
@@ -29,7 +29,9 @@ public function rules()
             'campaign_id' => 'required|integer',
             'northstar_id' => 'nullable|objectid',
             'type' => 'required|string|in:photo,voter-reg,text,share-social',
+            // @TODO: eventually, deprecate action in the payload and make action_id required when all systems have been updated.
             'action' => 'required|string',
```
katiecrane: Is this the `name` that you referred to here?

> When a post is created, it grabs the action_id from the payload. If this is not included in the payload, it goes into the database and looks for this action_id, found by querying the actions table by campaign_id, post_type, and name (all info that should come in with the post request).

app/Models/Post.php
```diff
@@ -95,11 +95,11 @@ public function reactions()
     }
 
     /**
-     * Each post has one action.
+     * Each post belongs to an action.
      */
-    public function action()
+    public function actionModel()
```
katiecrane: Why does this have "Model" in the name? I noticed that the other relations do not have that

app/Repositories/PostRepository.php
katiecrane: @chloealee can you explain the syncing issues again? I'm not sure I follow

app/Repositories/PostRepository.php
```diff
@@ -77,6 +79,21 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
...
+            } else {
+                $actionId = $action->id;
```
katiecrane: I don't think this strictly needs to be in an `else` statement if you don't want it to be

database/factories/ModelFactory.php
```diff
@@ -119,3 +125,23 @@
         'end_date' => $faker->dateTimeBetween('-3 months', 'now'),
     ]);
 });
+
+// Action Factory
```
katiecrane: I love the idea of an "Action Factory" 💥 💥 💥

tests/Http/PostTest.php
```diff
@@ -111,6 +119,8 @@ public function testCreatingAPhotoPost()
         $why_participated = $this->faker->paragraph;
         $text = $this->faker->sentence;
         $details = ['source-detail' => 'broadcast-123', 'other' => 'other'];
+        $action = factory(Action::class)->create([
+            'campaign_id' => $signup->campaign_id, ]);
```
katiecrane: this formatting looks a little funky to me 😬

katiecrane:
@chloealee thank you so much for updating existing tests and writing a new one for this work!!!

app/Http/Requests/PostRequest.php
chloealee: Yes exactly!

app/Models/Post.php
chloealee: @katiecrane great q! We aren't going to remove the `action` id column yet from the `posts` table (will do this at a later date) so we have to name this relation something different to differentiate when we want to get the `action` column from the `$posts` table or the relation to get the post's `action`. e.g. to get the `action` column from the `posts` table, we'd do `$post->action` versus if we wanted to get the action model that the post belongs to, we'd do `$post->actionModel`.

app/Repositories/PostRepository.php
chloealee: @katiecrane of course! For now, we won't require that `action_id` is sent along in the payload when a post is being created via the `POST v3/api/posts` endpoint (we will do this later but not immediately to allow other systems to adjust). At the same time, we don't want anything to break and have all work exactly the same so no systems should really see a difference with this PR. 

To account for this, we first will try to get the `action_id` from the payload if it is sent. If it is not, we'll try to get the `action` from the `actions` table in Rogue with this query, which we should be able to do with the `name` of the action, the `campaign_id` and `post_type`. This is mostly for existing campaigns so people can still reportback! 

However, if a campaign lead forgets to create an `action` on a new campaign for a new type of post, this will break. It should work with all other existing campaigns since we will run a backfill to migrate all the existing actions to the `actions` table and update all existing posts to have the correct `action_id` (scripts in [this PR](https://github.com/DoSomething/rogue/pull/825). 

Not sure if this is the best way to handle this error though so trying to think through if there is a better way and will make it easier to debug and not lose any reportbacks...

app/Repositories/PostRepository.php
chloealee: ah! Do you mean it might read better as a ternary?

tests/Http/PostTest.php
chloealee: i know it is weird! I had it as
```
$action = factory(Action::class)->create(['campaign_id' => $signup->campaign_id]);
```
without the `,` but StyleCi wanted me to add that! So weird!!

app/Http/Requests/PostRequest.php
katiecrane: thanks!

app/Models/Post.php
katiecrane: OOOO that is very tricky!! thanks for explaining!

app/Repositories/PostRepository.php
katiecrane: I _think_ you can remove the `else` entirely

tests/Http/PostTest.php
katiecrane: hmm I wonder if something changed with our StyleCi settings?? This seems strange I can't imagine we'd really want it like this

app/Repositories/PostRepository.php
chloealee: oh you're totally right! Removed :)

tests/Http/PostTest.php
chloealee: ah wait - I just changed it to one line and it seemed to like it better! Thanks, @katiecrane !

app/Repositories/PostRepository.php
katiecrane: @chloealee Thank you for the easy to understand description! I think it makes sense to keep it this way, it sounds like we'd only ever be missing an action in the case where it is a new action that wasn't properly created, and in that case we should be throwing an error!

chloealee: Sounds good, thanks @katiecrane ! I'll remove that comment then now :)

PR #829, opened 2019-02-01
Backend work to create Actions

chloealee:
#### What's this PR do?
- Adds the ability to create actions in Rogue.
- Adds Actions API: 
  - `GET /api/v3/actions index`
  - `GET /api/v3/actions/{action}`
  - `POST /actions`
  - `PATCH /actions/{action}`
  - `DELETE /actions/{action}`
- Adds unit tests to test the above routes. 
- Updates documentation. 

👀 

#### Relevant tickets

Part of 5th check box in [Post Metadata Tech Brief](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/32)

#### How to deploy


app/Http/Controllers/Legacy/Web/CampaignIdsController.php
```diff
@@ -15,8 +15,7 @@ class CampaignIdsController extends Controller
     public function __construct()
     {
         $this->middleware('auth');
-        $this->middleware('role:admin,staff', ['only' => ['index', 'show']]);
-        $this->middleware('role:admin', ['only' => ['create', 'store', 'edit', 'update', 'destroy']]);
```
chloealee: I removed this because I think we want staff to also be able to do all of this, not sure why before it only allowed admins to do this.

app/Http/Controllers/CampaignsController.php
```diff
@@ -9,7 +9,7 @@
 class CampaignsController extends ApiController
 {
     /**
-     * @var Rogue\Http\Transformers\SignupTransformer;
+     * @var Rogue\Http\Transformers\CampaignTransformer;
```
katiecrane: good catch!

app/Http/Controllers/Legacy/Web/CampaignIdsController.php
katiecrane: @chloealee I think at one time we had planned to only let admins touch the campaigns stuff...that sounds familiar to me at least

app/Http/Controllers/Legacy/Web/ActionsController.php
```diff
@@ -0,0 +1,80 @@
...
+
+class ActionsController extends Controller
+{
+    /**
+     * Create a controller instance.
+     */
+    public function __construct()
+    {
+        $this->middleware('auth');
+        $this->middleware('role:admin,staff', ['only' => ['create', 'store', 'edit', 'update', 'destroy']]);
```
katiecrane: I thought we didn't want to allow folks to delete actions through the web interface?

chloealee: I think we don't want to make it obvious but from what I understand, an action can be edited/deleted until the campaign is live and then after that, they can't edit or delete it. Is that right @ngjo ?

app/Http/Controllers/Legacy/Web/CampaignIdsController.php
chloealee: @katiecrane ah yes you're totally right, thank you for jogging my memory on that!

tests/Http/ActionTest.php
```diff
@@ -0,0 +1,139 @@
...
+        ]);
+
+        // Make sure the action is persisted.
+        $this->assertDatabaseHas('actions', [
+            'name' => $actionName,
+        ]);
+    }
+
+    /**
+     * Test that a GET request to /api/v3/actions returns an index of all campaigns.
```
katiecrane: This might be too in the weeds but would this be better tested if there were actions assigned to multiple campaigns?

tests/Http/ActionTest.php
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * Test that a PATCH request to /actions/:action_id updates an action.
+     *
+     * PATCH /actions/:action_id
+     * @return void
+     */
+    public function testUpdatingAnAction()
+    {
+        // Create a campaign that the action will belong to.
+        $campaign = factory(Campaign::class)->create();
```
katiecrane: Do you need to create the campaign first? [In the action factory](https://github.com/DoSomething/rogue/blob/master/database/factories/ModelFactory.php#L137) I think it will still assign it a `campaign_id`

chloealee: @katiecrane great q! We do need to create a campaign bc I _think_ if we don't, then there are no campaigns in the database. So the foreign key restraint throws an error because there is no `id` in the `campaigns` table that matches with this action.

chloealee: super smart, @katiecrane ! Thanks for that - updated!

tests/Http/ActionTest.php
chloealee: also updated this!

tests/Http/ActionTest.php
```diff
@@ -0,0 +1,132 @@
...
+    public function testActionIndex()
+    {
+        // Create five actions.
+        factory(Action::class, 5)->create();
+
+        // Create a few more for a different campaign.
+        factory(Action::class, 3)->create();
+
+        // And one more for good measure.
+        factory(Action::class, 2)->create();
```
katiecrane: @chloealee does each grouping get the same `campaign_id` or will they all be different? I don't remember exactly how this works

chloealee: ohhh @katiecrane you're right! So I changed it back because even though it creates 5 actions at the same time, each action has a different `campaign_id`

katiecrane: ah right! because of the factory update! I love it!

PR #830, opened 2019-02-11
Actions frontend

chloealee:
#### What's this PR do?
- Updates front end to be able to add new actions (post metadata)
- Some updates were made to try to get this to be mostly in React but interest of time, only the following was done: 
  - `CampaignIdOverview` page is now in React
  - `CampaignIdSingle` page is now in React 
- What's not in React: 
  - campaign-id create form (still in `campaign-ids.create.blade.php`) 
  - campaign-id edit form (still in `campaign-ids.edit.blade.php`)
  - create action form (created in `actions.create.blade.php` because I was having trouble getting the modal to work in React. 
- New route: `campaign-ids/{id}/actions/create` for the action creation form. 

Screenshots:
- `/campaign-ids/{id}`:
screen shot 2019-02-13 at 1 40 21 pm
- From this page, you can click the "ADD ACTION" button and it will redirect to: `/campaign-ids/{id}/actions/create`:
screen shot 2019-02-13 at 1 41 58 pm
- From there, you can click the "CREATE ACTION" button and it will redirect to: `/campaign-ids/{id}` with the new action:
screen shot 2019-02-13 at 1 42 48 pm

Is everything still working properly? Anything look incorrect? Is the functionality logical? 

This is VERY MUCH an MVP for this! Things that still need to be done: 
- Paginate `/campaign-ids` page 
- Add module instead of `/campaign-ids/{id}/actions/create` page when creating a new action
- Front end tests 
- Ability to edit and delete actions from the front end (up until the campaign is live - after it launches, remove edit/delete functionality for actions). 
cc @ngjo lmk if this is ok for an MVP - I am happy to create extra cards so we can get back to this when there is more time! 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/163124096) Pivotal Card and almost all of Frontend checkbox in [tech spec](https://github.com/orgs/DoSomething/teams/team-bleed/discussions/32).

#### How to deploy


resources/assets/components/CampaignIdSingle/index.js
```diff
@@ -0,0 +1,156 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { map, isEmpty } from 'lodash';
+
+import './campaignidsingle.scss';
+
+import Action from '../Action';
+import UploaderModal from '../UploaderModal';
+// import CreateActionModal from '../CreateActionModal';
```
chloealee: @DFurnes @katiecrane I'm a little stuck trying to get the action modal to pop up when clicking the "ADD ACTION" button (for reference, [here](https://projects.invisionapp.com/share/DEQ8H2DXN2W#/screens/343746016) is the mock!). 

If I comment out this line, everything loads as it should on the page. But when this line is commented back in, it seems like it loads the `CreateActionModal` component before loading this page. I'm not sure why because the state should be `false` and not render this until it is called when the button is clicked but I'm thinking maybe something got out of sync and it thinks it should load the `CreateActionModal` component first? Is there anything that I'm missing here?

chloealee: hm another thing that is happening is that when this is commented in, `/campaign-ids` doesn't load either with this error: `Uncaught TypeError: Cannot convert undefined or null to object`. When it is commented out, everything is fine. This makes me further think that something with the sequence of components being loaded is off with this `CreateActionModel` but I can't see anything obvious why it is off!

katiecrane: @chloealee are you able to see the state and if `displayCreateActionModal` is actually set as `true` or `false` in the case where it's borked and loads that component too soon?

chloealee: @katiecrane I'm not! So when `import CreateActionModal from '../CreateActionModal';` in commented IN, and I `console.log` the `state`, it shows up as `null`. But when it is commented OUT and I `console.log` the `state`, `displayCreateActionModal` is `false`. So thats why I'm thinking for some reason it is skipping right over this page and going straight to the modal first?

katiecrane: weirddddd @chloealee can you try console logging something maybe in the constructor of each component and see if that shows anything? Maybe some higher up stuff too to try to see where it's getting stuck. You might need to use the "preserve logs" option if you end up getting redirected or something

chloealee: thanks @katiecrane !! Really appreciate you taking a look at this! It DOES seem to be going straight to the `CreateActionModal` when I `console.log` something in both the constructors :thinking_face: also still wondering why it breaks the `CampaignIdOverview` page when the `import CreateActionModal from '../CreateActionModal';` is commented in so maybe it is something higher that I'm not importing correctly or something?

katiecrane: hmmmm I wonder if the constructor gets hit when it is imported? and then something with that errors out so it never makes it farther in this file?? @chloealee wait is the `CampaignIdOverview` page a totally separate page?

chloealee: @katiecrane it is! That is why it is strange why it is even hit!

chloealee: is it ok to keep this in here to continue iterating on this when I have time? I feel like it is so close!

resources/assets/components/CampaignIdSingle/index.js
```diff
@@ -0,0 +1,146 @@
...
+  //   if (event) {
+  //     event.preventDefault();
+  //   }
+
+  //   this.setState({
+  //     displayCreateActionModal: false,
+  //     campaign: null,
+  //   });
+  // }
+
```
chloealee: same with the above - ok to leave commented out to come back to when there is time to try to add the modal again?

resources/assets/components/CreateActionModal/index.js
```diff
@@ -0,0 +1,69 @@
...
+      <div className="modal">
+        <a href="#" onClick={this.props.onClose} className="modal-close-button">
+          &times;
+        </a>
+        <div className="modal__block">
+          <h3>Add Post Metadata</h3>
+          <form
+          // className="post-form"
+          // onSubmit={this.handleOnSubmitForm}
+          // ref={form => (this.form = form)}
```
chloealee: and here as reference to come back to?

resources/views/actions/create.blade.php
```diff
@@ -0,0 +1,66 @@
...
+        <div class="wrapper">
+            <div class="container__block -narrow">
+                <h1>Add Post Metadata</h1>
+                <form method="post" enctype="application/x-www-form-urlencoded" action="{{ route('actions.store', ['campaign_id' => $campaignId]) }}">
+                {{ csrf_field()}}
+
+                    <div class="form-item">
+                        <label class="field-label">Action Name</label>
+                        <input type="text" name="name" class="text-field" placeholder="Name your action e.g. Teens for Jeans Photo Upload" value="{{old('internal_title') }}">
+                    </div>
```
chloealee: @ngjo @lkpttn is the below naming convention finalized? If so, I can add as the placeholder instead of "Name your action..."
```
CampaignAbbrev- YYYY Action Description
Example: SSD-2019 Bin
```

app/Http/Controllers/Legacy/Web/ActionsController.php
```diff
@@ -28,13 +46,16 @@ public function store(Request $request)
...
-            'reportback' => 'required|boolean',
-            'civic_action' => 'required|boolean',
-            'scholarship_entry' => 'required|boolean',
             'noun' => 'required|string',
             'verb' => 'required|string',
         ]);
 
+        // Checkbox values are only sent from the front end if they are checked.
+        // Assign checkbox values if sent from the front end or are 1 (sent from API).
+        $request['reportback'] = isset($request['reportback']) || $request['reportback'] === 1 ? 1 : 0;
```
katiecrane: Could the api send `true`? I don't think that is `===` to ` 1`

app/Http/Controllers/Legacy/Web/CampaignIdsController.php
```diff
@@ -23,9 +23,10 @@ public function __construct()
      */
     public function index()
     {
-        $campaigns = Campaign::orderBy('created_at', 'desc')->paginate(25);
+        $campaigns = Campaign::orderBy('created_at', 'desc')->get()->toArray();
+        // $campaigns = Campaign::paginate(25);
```
katiecrane: Can this be removed?

app/Http/Controllers/Legacy/Web/CampaignIdsController.php
```diff
@@ -76,13 +77,18 @@ public function store(Request $request)
     }
 
     /**
-     * Show a specific campaign page.
+     * Show a specific campaign page and it's actions.
```
katiecrane: 😬 "its"
(I am so sorry)

resources/assets/components/CampaignIdSingle/index.js
katiecrane: @chloealee yes I think that makes sense!!

resources/assets/components/CampaignIdSingle/index.js
katiecrane: I think that is fine for now!

app/Http/Controllers/Legacy/Web/CampaignIdsController.php
chloealee: oh yes sorry!

app/Http/Controllers/Legacy/Web/CampaignIdsController.php
chloealee: omg don't be sorry thank you for catching this! How embarrassing!

app/Http/Controllers/Legacy/Web/ActionsController.php
chloealee: @katiecrane it expects a boolean! Should it be `==` instead of `===`?

katiecrane: I believe `===` will hold it to the exact type and value of `1`. Since it's going to be a boolean could you use `$request['reportback']` itself? If the value as `false` it'd get evaluated to `false` right?

chloealee: @katiecrane you're totally right, changed to `==` instead! 

The reason why we can't just use `$request['reportback']` itself is that it will only get send in the payload from the form if it is checked. If it is not checked, `$request['reportback']` won't be in the payload at all, so that is why we have to do this weird logic in order for it to register unchecked values as well.

katiecrane: ah sorry @chloealee I was unclear! I mean something like this:
```
$request['scholarship_entry'] = isset($request['scholarship_entry']) || $request['scholarship_entry'] ? 1 : 0;
```

chloealee: @katiecrane ah thank you! I just tried that via the api and it didn't work :-/ `$request['scholarship_entry']` came in as `"0"` via Postman so since that was a truthy value, it saved `1` in the database incorrectly for it to count as a scholarship.

katiecrane: hmm that's so strange since we make sure that it is a boolean. since that is a string would it still come through as `true` in your updated way with `== 1`?

chloealee: it's weird that it is coming in as a string but in the request in Postman, I am putting `0` as the value, not `false`. So with the updated `==`, since it doesn't look for type, it works!

chloealee: oohhh I see what you're saying now! Ok got it working! I changed `||` to `&&` in this ternary and I think this makes more sense!

katiecrane: YES I LOVE THAT

PR #834, opened 2019-02-20
Adds ability for posts to be anonymous

chloealee:
#### What's this PR do?
Adds ability for posts to be anonymous:
  - Adds `anonymous` column to the `actions` table.
  - Removes the `active` column from the `actions` table since we are no longer using this.
  - Updates `PostTransformer` to only return `northstar_id` and `location` if the post is not anonymous or if the post's owner is hitting the `GET /api/v3/posts` endpoint.
  - Updates model factory and adds test. 
  - Updates documentation. 

- Does this logic make sense? 
- Is there any other info we want to hide for anonymous posts? 

#### Relevant tickets
Adds backend functionality for [this](https://www.pivotaltracker.com/n/projects/2019429/stories/164088666) Pivotal Card.

#### How to deploy


app/Http/Transformers/PostTransformer.php
```diff
@@ -48,12 +47,18 @@ public function transform(Post $post)
...
             ],
             'status' => $post->status,
-            'location' => $post->location,
             'created_at' => $post->created_at->toIso8601String(),
             'updated_at' => $post->updated_at->toIso8601String(),
         ];
 
-        if (Gate::allows('viewAll', $post)) {
+        // If the post isn't anonymous or if this is the owner of the post, return PII information.
+        if ((! filter_var($post->actionModel->anonymous, FILTER_VALIDATE_BOOLEAN) && ! is_null($post->actionModel->anonymous)) || $post->northstar_id === token()->id()) {
```
DFurnes: Instead of using `FILTER_VALIDATE_BOOLEAN`, let's [specify a casting rule](https://laravel.com/docs/5.7/eloquent-mutators#attribute-casting) for this on the Eloquent model.

DFurnes: Instead of hard-coding the authorization logic here, let's use the standard policy:

```diff
- $post->northstar_id === token()->id()
+ Gate::allows('viewAll', $post)
```

app/Http/Transformers/PostTransformer.php
```diff
@@ -48,12 +47,18 @@ public function transform(Post $post)
...
+            $response['northstar_id'] = $post->northstar_id;
+            $response['location'] = $post->location;
```
DFurnes: We don't want to hide `location` for anonymous posts, just the user ID.

app/Http/Transformers/PostTransformer.php
```diff
@@ -48,12 +47,18 @@ public function transform(Post $post)
...
+        }
+
+        // If the post isn't anonymous and the user has credientials to viewAll, return the following detais.
+        if (! filter_var($post->actionModel->anonymous, FILTER_VALIDATE_BOOLEAN) && Gate::allows('viewAll', $post)) {
```
DFurnes: These can continue to just rely on `Gate::allows('viewAll', $post)`.

database/factories/ModelFactory.php
```diff
@@ -140,6 +140,7 @@
         'reportback' => 1,
         'civic_action' => 1,
         'scholarship_entry' => 1,
+        'anonymous' => 0,
```
DFurnes: Once we add casts for these, let's just set these as `true` or `false` here rather than integers.

docs/endpoints/posts.md
```diff
@@ -14,6 +14,8 @@ Only admins and post owners will have `tags`, `source`, `remote_addr` (which wil
 
 Anonymous requests will only return accepted posts. Logged-in users can see accepted posts & any of their own pending or rejected posts. Staff can see anything!
 
+If a post is anonymous, only `northstar_id` and `location` will be returned for the post owner.
```
DFurnes: Can we re-word this to "the `northstar_id` field will only be returned for the post owner"? The current wording makes me think if a post is anonymous, I'd only get the `northstar_id` and `location` fields in the response.

DFurnes:
Just to clarify the requirements here, the _only_ unique thing we care about hiding for anonymous posts is the User ID of the person who made the post. Everything else can stay as-is.

DFurnes:
Can we also double-check (and add a test ensuring) that you can't see anonymous posts if you filter by ID? (For example, `v3/posts/?filter[northstar_id]=…` shouldn't show that user's anonymous posts).

app/Http/Transformers/PostTransformer.php
chloealee: oh i love this!

chloealee: Oh you're right! I didn't use [`Gate::allows('viewAll', $post)`](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/Traits/AuthorizesWithToken.php#L16-L19) here initially because I thought that if a machine (Phoenix) requested, it would still return the `northstar_id` but I forgot that this is using the user auth! Thanks for catching!

docs/endpoints/posts.md
chloealee: ah yes much clearer!

app/Http/Transformers/PostTransformer.php
chloealee: @DFurnes actually I think I need to change it back to `$post->northstar_id === token()->id()` because the `viewAll` gate allows staff to also see this information and we don't want that.

docs/endpoints/posts.md
```diff
@@ -14,6 +14,8 @@ Only admins and post owners will have `tags`, `source`, `remote_addr` (which wil
 
 Anonymous requests will only return accepted posts. Logged-in users can see accepted posts & any of their own pending or rejected posts. Staff can see anything!
 
+The `northstar_id` field will only be returned for the post owner.
```
DFurnes: This is only true if the action is anonymous, right? Can we reword this to:

> If the post's action is marked as "anonymous", the `northstar_id` field will only be returned for the owner.

tests/Http/ActionTest.php
```diff
@@ -28,6 +28,7 @@ public function testCreatingAnAction()
             'reportback' => 1,
             'civic_action' => 0,
             'scholarship_entry' => 1,
+            'anonymous' => 0,
```
DFurnes: Let's update these tests to send these as booleans as well.

tests/Http/PostTest.php
```diff
@@ -762,6 +762,43 @@ public function testPostsIndexAsNonAdminNonOwner()
...
+     * GET /api/v3/posts
+     * @return void
+     */
+    public function testPostsIndexWithAnonymousPosts()
+    {
+        // Create an accepted post.
+        $regularPost = factory(Post::class, 'accepted')->create();
+
+        // Create 1 post that is anonymous.
+        sleep(1);
```
DFurnes: Why is this `sleep(1)` needed here?

tests/Http/PostTest.php
```diff
@@ -762,6 +762,43 @@ public function testPostsIndexAsNonAdminNonOwner()
...
+        $anonymousPost = factory(Post::class, 'accepted')->create();
+        $anonymousPost->actionModel->anonymous = 1;
+        $anonymousPost->actionModel->save();
+
+        // Anonymously hit the endpoint and norhtstar_id should not be returned for the anonymous post (first post since it was most recently created).
```
DFurnes: `norhtstar_id` → `northstar_id`

tests/Http/PostTest.php
```diff
@@ -762,6 +762,43 @@ public function testPostsIndexAsNonAdminNonOwner()
...
+        $response = $this->getJson('api/v3/posts');
+        $response->assertStatus(200);
+        $this->assertArrayNotHasKey('northstar_id', $response->decodeResponseJson()['data'][0]);
+        $this->assertEquals($regularPost->northstar_id, $response->decodeResponseJson()['data'][1]['northstar_id']);
+
+        // Hit the endpoint with admin credientals and same results as above.
```
DFurnes: `credientals` → `credentials`

tests/Http/PostTest.php
```diff
@@ -762,6 +762,43 @@ public function testPostsIndexAsNonAdminNonOwner()
...
+        $response = $this->withAdminAccessToken()->getJson('api/v3/posts');
+        $response->assertStatus(200);
+        $this->assertArrayNotHasKey('northstar_id', $response->decodeResponseJson()['data'][0]);
```
DFurnes: Hmm, admins should be able to view Northstar IDs for these posts. Let's double-check the permission logic.

app/Http/Transformers/PostTransformer.php
```diff
@@ -48,11 +47,16 @@ public function transform(Post $post)
...
             ],
             'status' => $post->status,
-            'location' => $post->location,
             'created_at' => $post->created_at->toIso8601String(),
             'updated_at' => $post->updated_at->toIso8601String(),
+            'location' => $post->location,
         ];
 
+        // If the post isn't anonymous or if this is the owner of the post, return PII information.
+        if ((! $post->actionModel->anonymous && ! is_null($post->actionModel->anonymous)) || Gate::allows('viewAll', $post) {
```
DFurnes: I think this conditional can still be simplified a bit:

```php
if ((! $post->actionModel->anonymous || Gate::allows('viewAll', $post)) {
```

DFurnes: Can you double-check that we're not ending up with an [N+1 query problem](https://secure.phabricator.com/book/phabcontrib/article/n_plus_one/) by querying a property on the `actionModel` relationship here? I think we might want to set the post model to [eager-load](https://laravel.com/docs/5.7/eloquent-relationships#eager-loading) the action relationship [here](https://github.com/DoSomething/rogue/blob/d12bb7f7e9ea622f5672342d53859dbe74b63d80/app/Models/Post.php#L21-L25).

app/Http/Transformers/PostTransformer.php
DFurnes: It's okay for staff to be able to see this!

chloealee: oh okay! I'll update the tests for that then :)

docs/endpoints/posts.md
chloealee: ah shoot yes, I was going too fast yesterday! Updated :)

tests/Http/PostTest.php
chloealee: @DFurnes if `sleep(1)` isn't used, both the `$regularPost` and `$anonymousPost` are created at the same time bc the test goes so fast! I want to deliberately make sure the `$anonymousPost` is created after the `$regularPost` to make sure that when we hit the index, the `$anonymousPost` is the first in the array since it returns records in descending order.

tests/Http/PostTest.php
chloealee: oops!

tests/Http/PostTest.php
chloealee: updated this test to instead check that another logged in user cannot see `northstar_id` and adding another line to assure that admins can still see this!

app/Http/Transformers/PostTransformer.php
chloealee: @DFurnes the reason I put `! is_null($post->actionModel->anonymous)` is because I'm not sure if after we run this, existing records will have `NULL` for the `anonymous` column.  If they are `NULL`, I think we'd want to check this?

DFurnes: That should be okay, because we want the default `null` to work the same way as if `anonymous` was set to false. In both cases (`! false` and `! null`) this evaluates to `true`.

tests/Http/PostTest.php
DFurnes: It _should_ still fall back to ID for ordering if two posts have the same `created_at`, I think. If ou remove this, do the tests become flaky?

We want to avoid using `sleep()` in our test suite whenever possible, since if we over-use it it'd significantly slow things down. (We have 137 tests that run in ~20 seconds now... imagine if each of those had a `sleep(1)` in them! 😨)

app/Http/Transformers/PostTransformer.php
chloealee: ah super smart, it was having the N+1 query problem! Added this!

tests/Http/PostTest.php
chloealee: @DFurnes yeah I've tried to remove this and then the tests fail :-/

app/Http/Transformers/PostTransformer.php
chloealee: oh true! Updated :)

tests/Http/PostTest.php
DFurnes: Okay, bummer! We can use the `mockTime` method in that case ([example](https://github.com/DoSomething/rogue/blob/1cf8256dcc6426e8b8f8ee4205e6d3dd740a0584/tests/Unit/Models/ReactionModelTest.php#L20-L28)).

chloealee: oh right I forgot about that! Updated! :)

tests/Http/PostTest.php
```diff
@@ -762,6 +762,70 @@ public function testPostsIndexAsNonAdminNonOwner()
...
+        $this->assertArrayNotHasKey('northstar_id', $response->decodeResponseJson()['data']);
+
+        // Hit the endpoint as the owner of the post and you should be able to see northstar_id for anonymous post.
+        $response = $this->withAccessToken($anonymousPost->northstar_id)->getJson('api/v3/posts');
+        $response->assertStatus(200);
+        $this->assertEquals($anonymousPost->northstar_id, $response->decodeResponseJson()['data'][0]['northstar_id']);
+        $this->assertEquals($regularPost->northstar_id, $response->decodeResponseJson()['data'][1]['northstar_id']);
+
+        // Hit the endpoint with filter[northstar_id] and should have same results as above.
+        $response = $this->withAccessToken($anonymousPost->northstar_id)->getJson('api/v3/posts?filter[northstar_id]=' . $anonymousPost->northstar_id);
```
DFurnes: This one's a bit tricker – the `northstar_id` on the post is technically exposed by the fact that it showed up on a `?filter[northstar_id]=…` response (since anything returned by this query would _have_ to have that user ID). Ideally, we'll want to update the index route to specifically exclude "anonymous" posts if querying for a user that's not you.

chloealee: ah great catch! 

Updated the tests and I tried to add it to the `whereVisible` query scope but it is still being returned when it shouldn't be. Using clockwork, it looks like the query is correct yet I'm not sure why it is still returned. Am I missing something? 

screen shot 2019-02-21 at 3 34 37 pm

chloealee: Moved this to a separate query scope and got it to work! Ready for review!

chloealee: oh wait - this is omitting the entire thing from the index without this filter - need to update this! I'll let you know when it is ready for review :)

chloealee: ok NOW this is ready! :)

app/Http/Transformers/PostTransformer.php
```diff
@@ -53,6 +52,11 @@ public function transform(Post $post)
             'updated_at' => $post->updated_at->toIso8601String(),
         ];
 
+        // If the post isn't anonymous or if this is the owner of the post, return PII information.
```
DFurnes: Sorry, can we re-word this? I think saying "PII" here is a little misleading (since we already lock down most personally-identifiable information here and in Northstar based on the user's privileges).

> If this post is for an anonymous action (and viewer is not owner/staff), hide the user ID.

chloealee: yes great call!

PR #836, opened 2019-02-20
Fix some bugs with Action IDs.

DFurnes:
#### What's this PR do?
This pull request fixes some things that caught me off guard when working in Phoenix:

🗺 Instead of throwing a 404 error if an action isn't created, throw a standard 422. (Throwing a 404 is confusing because it indicates that the `v3/posts` API resource doesn't exist, which is not true).

🌱 Add a `default` action to the model seeder so these exist for the dev environment (to match up with what we set in Contentful), and fixes issue where we'd create a fresh action for _every_ single seeded post.

📄 Finally, adds a default "text" action and posts so we can use the Text Submission Action on these seeded campaigns without doing any more manual setup steps.

👀

N/A

#### Relevant tickets
References [this Slack conversation](https://dosomething.slack.com/archives/C3ASB4204/p1550695423014100).

#### How to deploy

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -84,11 +84,11 @@ class PostRepository
             $actionId = $data['action_id'];
         } else {
             $type = isset($data['type']) ? $data['type'] : 'photo';
-            $action = Action::where('campaign_id', $signup->campaign_id)->where('post_type', $type)->where('name', $data['action'])->first();
-
-            if (! $action) {
-                throw new ModelNotFoundException('Action not found.');
-            }
+            $action = Action::where([
+                'campaign_id' => $signup->campaign_id,
+                'post_type' => $type,
+                'name' => $data['action'],
+            ])->firstOrFail();
 
             $actionId = $action->id;
         }
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -9,7 +9,6 @@ use Rogue\Models\Review;
 use Rogue\Models\Signup;
 use Rogue\Services\Registrar;
 use Intervention\Image\Facades\Image;
-use Illuminate\Database\Eloquent\ModelNotFoundException;
 
 class PostRepository
 {
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -53,6 +53,35 @@ $factory->define(Post::class, function (Generator $faker) {
     ];
 });
 
+// @TODO: These should all extend a photo-less "base" post, instead.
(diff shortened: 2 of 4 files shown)


app/Repositories/PostRepository.php
```diff
@@ -84,11 +83,11 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
...
-            $action = Action::where('campaign_id', $signup->campaign_id)->where('post_type', $type)->where('name', $data['action'])->first();
-
-            if (! $action) {
-                throw new ModelNotFoundException('Action not found.');
-            }
+            $action = Action::where([
+                'campaign_id' => $signup->campaign_id,
+                'post_type' => $type,
+                'name' => $data['action'],
+            ])->firstOrFail();
```
weerd: 👌

database/factories/ModelFactory.php
```diff
@@ -53,6 +53,35 @@
     ];
 });
 
+// @TODO: These should all extend a photo-less "base" post, instead.
```
aaronschachter: @DFurnes I've been curious about how this would look, would you be able to provide some pseudo-code? I'm thinking it'd help DRY the cleanup work I've been thinking about, for only [setting a Post school ID if its action has its `collect_school_id` field set to `true`.](https://github.com/DoSomething/rogue/blob/d77c9b879e168deaf5118962c2da4fabcb2c4d75/database/factories/ModelFactory.php#L87)

DFurnes: Oooh, this was so long ago... 🙈 

I think my thought process here was that we don't need an image or `url` on the `text` posts, but since this "extends" the base `Post` factory above, it's still doing the work to generate a fake image for them anyways. So we're sorta making a photo post and then "undoing" it in the `text` version.

Instead, we could maybe have a base `Post` factory with the common elements (IDs, location, source, whatnot) and then a specific `$factory->defineAs(Post::class, 'photo', …` that adds the photo-specific fields).

aaronschachter: Gotcha, was thinking it'd be using the post type as a [state](https://laravel.com/docs/5.8/database-testing#factory-states) like you mention, but wanted to be sure. So then if we wanted different [states for the post status as we do currently,](https://github.com/DoSomething/rogue/blob/d77c9b879e168deaf5118962c2da4fabcb2c4d75/database/factories/ModelFactory.php#L94) we could potentially have states like `text-accepted`, `text-rejected`, `photo-accepted` ?

DFurnes: Yep, I think that's how we'd have to handle that.

PR #837, opened 2019-02-22
Removes `campaign_id` requirement when creating a signup or post when there's an `action_id`

chloealee:
#### What's this PR do?
- Removes `campaign_id` requirement when creating a signup or post when there's an `action_id`.
- Adds test to create a signup and post without `campaign_id` but has `action_id`.
- Fixes tests to use `$action->name` and `$action->post_type` instead of custom string for `action` when creating the post. 
- Deletes `testCreatingAVoterRegPost` because this was a duplicate test.
- Updates documentation. 

cc @rapala61 and @aaronschachter on this too since we were discussing this in [Slack](https://dosomething.slack.com/archives/C0YGXUE01/p1551291369075400?thread_ts=1551285205.065200&cid=C0YGXUE01)! 

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/164121341)Pivotal Card

#### How to deploy

### DIFF
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -3,6 +3,7 @@
 namespace Rogue\Http\Controllers;
 
 use Rogue\Models\Post;
+use Rogue\Models\Campaign;
 use Illuminate\Http\Request;
 use Rogue\Managers\PostManager;
 use Rogue\Managers\SignupManager;
@@ -109,7 +110,7 @@ class PostsController extends ApiController
         $northstarId = getNorthstarId($request);
 
         // Get the campaign id from the request by campaign_id or action_id.
-        $campaignId = $request['campaign_id'] ? $request['campaign_id'] : get_campaign_by_action_id($request['action_id'])->id;
+        $campaignId = $request['campaign_id'] ? $request['campaign_id'] : Campaign::fromActionId($request['action_id'])->id;
 
         $signup = $this->signups->get($northstarId, $campaignId);
 
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -114,7 +114,7 @@ class PostsController extends ApiController
         $signup = $this->signups->get($northstarId, $campaignId);
 
         if (! $signup) {
-            $signup = $this->signups->create($request->all(), $northstarId);
+            $signup = $this->signups->create($request->all(), $northstarId, $campaignId);
         }
 
         $post = $this->posts->create($request->all(), $signup->id);
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -108,7 +108,10 @@ class PostsController extends ApiController
     {
         $northstarId = getNorthstarId($request);
 
-        $signup = $this->signups->get($northstarId, $request['campaign_id']);
+        // Get the campaign id from the request by campaign_id or action_id.
+        $campaignId = $request['campaign_id'] ? $request['campaign_id'] : get_campaign_by_action_id($request['action_id'])->id;
+
+        $signup = $this->signups->get($northstarId, $campaignId);
 
         if (! $signup) {
             $signup = $this->signups->create($request->all(), $northstarId);
--- a/app/Http/Requests/PostRequest.php
+++ b/app/Http/Requests/PostRequest.php
@@ -28,7 +28,7 @@ class PostRequest extends Request
         return [
             'campaign_id' => 'required_without:action_id|integer',
             'northstar_id' => 'nullable|objectid',
-            'type' => 'required_without:action_id|string|in:photo,voter-reg,text,share-social',
+            'type' => 'required|string|in:photo,voter-reg,text,share-social',
             // @TODO: eventually, deprecate action in the payload and make action_id required when all systems have been updated.
             'action' => 'required_without:action_id|string',
             'action_id' => 'required_without:action,campaign_id|integer|exists:actions,id',
--- a/app/Http/Requests/PostRequest.php
+++ b/app/Http/Requests/PostRequest.php
@@ -30,7 +30,7 @@ class PostRequest extends Request
             'northstar_id' => 'nullable|objectid',
             'type' => 'required_without:action_id|string|in:photo,voter-reg,text,share-social',
             // @TODO: eventually, deprecate action in the payload and make action_id required when all systems have been updated.
-            'action' => 'required_without:action_id|string',
+            'action' => 'required|string',
             'action_id' => 'required_without:action,campaign_id|integer|exists:actions,id',
             'why_participated' => 'nullable|string',
             'text' => 'nullable|string|max:256',
--- a/app/Http/Requests/PostRequest.php
+++ b/app/Http/Requests/PostRequest.php
@@ -26,7 +26,7 @@ class PostRequest extends Request
         }
 
         return [
-            'campaign_id' => 'required|integer',
+            'campaign_id' => 'required_without:action_id|integer',
             'northstar_id' => 'nullable|objectid',
             'type' => 'required|string|in:photo,voter-reg,text,share-social',
             // @TODO: eventually, deprecate action in the payload and make action_id required when all systems have been updated.
--- a/app/Http/Requests/PostRequest.php
+++ b/app/Http/Requests/PostRequest.php
@@ -30,7 +30,7 @@ class PostRequest extends Request
             'northstar_id' => 'nullable|objectid',
             'type' => 'required_without:action_id|string|in:photo,voter-reg,text,share-social',
             // @TODO: eventually, deprecate action in the payload and make action_id required when all systems have been updated.
-            'action' => 'required|string',
+            'action' => 'required_without:action_id|string',
             'action_id' => 'required_without:action,campaign_id|integer|exists:actions,id',
             'why_participated' => 'nullable|string',
             'text' => 'nullable|string|max:256',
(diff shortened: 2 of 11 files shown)


app/Http/Requests/PostRequest.php
```diff
@@ -26,7 +26,7 @@ public function rules()
         }
 
         return [
-            'campaign_id' => 'required|integer',
+            'campaign_id' => 'required_without:action_id|integer',
             'northstar_id' => 'nullable|objectid',
             'type' => 'required|string|in:photo,voter-reg,text,share-social',
             // @TODO: eventually, deprecate action in the payload and make action_id required when all systems have been updated.
```
katiecrane: just below this should `action_id` also be `required_without:campaign_id`?

app/Repositories/SignupRepository.php
```diff
@@ -19,7 +19,7 @@ public function create($data, $northstarId)
         $signup = new Signup;
 
         $signup->northstar_id = $northstarId;
-        $signup->campaign_id = $data['campaign_id'];
+        $signup->campaign_id = isset($data['campaign_id']) ? $data['campaign_id'] : get_campaign_by_action_id($data['action_id'])->id;
```
katiecrane: Can the `campaign_id` kind of trickle down to here once it's grabbed from the action the first time? If not, is something like this needed for posts as well?

app/Http/Requests/PostRequest.php
chloealee: great catch!

app/Repositories/SignupRepository.php
chloealee: ah great call, updated that @katiecrane ! 

> is something like this needed for posts as well?

We don't need this because when a post is created, it [grabs the `campaign_id` from the signup it belongs to](https://github.com/DoSomething/rogue/blob/master/app/Repositories/PostRepository.php#L99)!

app/Http/Controllers/PostsController.php
```diff
@@ -108,10 +108,13 @@ public function store(PostRequest $request)
     {
         $northstarId = getNorthstarId($request);
 
-        $signup = $this->signups->get($northstarId, $request['campaign_id']);
+        // Get the campaign id from the request by campaign_id or action_id.
+        $campaignId = $request['campaign_id'] ? $request['campaign_id'] : get_campaign_by_action_id($request['action_id'])->id;
```
DFurnes: Instead of a `get_campaign_by_action_id` helper, can we make this a static method on the `Action` model?

DFurnes: Or actually, this should probably be on the `Campaign` model since that's what we're fetching, e.g.:

```php
$campaignId = $request['campaign_id'] ?: Campaign::fromActionId($request['action_id'])->id;
```

docs/endpoints/posts.md
```diff
@@ -168,12 +168,14 @@ This will automatically create or update the corresponding signup.
 POST /api/v3/posts
 ```
 
-- **campaign_id**: (int) required.
+- **campaign_id**: (int) required without action_id.
   The Campaign ID of the campaign that the user's post is associated with.
 - **type**: (string) required.
   The type of post submitted. Must be one of the following types: `photo`, `voter-reg`, `text`, `share-social`. `share-social` posts will be auto-accepted unless an admin sets a custom `status`.
-- **action**: (string) required.
+- **action**: (string) required without action_id.
```
DFurnes: I think it might be helpful to "remove" these from the main argument list, and list them below (so we start by listing required and optional fields, and then show these deprecated options beneath). That way there's a bit less "parsing" necessary when skimming the docs.

app/Http/Controllers/PostsController.php
chloealee: oh yes, I like that a lot better!

docs/endpoints/posts.md
chloealee: @DFurnes yeah, I agree that it is a little hard to read! Do you mean separate into different lists of required and optional params?

DFurnes: Yeah, I think grouping these into "required", "optional", and "deprecated" buckets is a nice idea.

chloealee: great! Added the "required" and "optional" sections for now ("deprecated" to come later when there is anything relevant for this category!).

DFurnes: Oh sorry, I wasn't clear! I was suggesting moving `campaign_id` and `action` into the "deprecated" section, since right now it's somewhat hard to parse out that it's "either provide these two fields or this one" and that `action_id` is preferred.

chloealee: OHH I see what you mean! Totally a lot clearer that way - updated!

docs/endpoints/posts.md
```diff
@@ -168,31 +168,41 @@ This will automatically create or update the corresponding signup.
...
-  The type of post submitted. Must be one of the following types: `photo`, `voter-reg`, `text`, `share-social`. `share-social` posts will be auto-accepted unless an admin sets a custom `status`.
-- **action**: (string) required.
-  Describes the bucket the action is tied to. A campaign could ask for multiple types of actions throughout the life of the campaign.
-- **quantity**: (int|nullable) optional.
+Required params:
+
+- **action_id**: (int) required without action or campaign_id.
+  The action ID of the action that the user's post is associated with.
+- **file**: (multipart/form-data) required for `photo` posts.
+  File to save of post image.
```
chloealee: I actually don't think this is true. If we do want to make this validation, we might have to add in `type` as a requirement again in order to do this?

app/Http/Requests/PostRequest.php
```diff
@@ -26,12 +26,12 @@ public function rules()
         }
 
         return [
-            'campaign_id' => 'required|integer',
+            'campaign_id' => 'required_without:action_id|integer',
             'northstar_id' => 'nullable|objectid',
-            'type' => 'required|string|in:photo,voter-reg,text,share-social',
+            'type' => 'required_without:action_id|string|in:photo,voter-reg,text,share-social',
```
chloealee: hi @DFurnes ! Continuing on from [your comment](https://github.com/DoSomething/rogue/pull/841#discussion_r261268220), right now, I don't think we actually require a photo post to have a file (somewhere this got lost unless I wasn't looking hard enough for this code!). This might make it a little more difficult to have the validation in the post request but I was thinking we could throw an error later, after we make the db query anyway for the type of post. But that probably isn't ideal... 

Re: "
and we [allow different statuses](https://github.com/DoSomething/rogue/blob/4a12be9f680f7cec776be1e81573357440a0cd5a/app/Http/Requests/PostRequest.php#L66-L77) based on the type of post as well", this is just for `PATCH` requests, in which case we already have the post type from the post object so I don't think this would be an issue here!

DFurnes: > Continuing on from [your comment](https://github.com/DoSomething/rogue/pull/841#discussion_r261268220), right now, I don't think we actually require a photo post to have a file (somewhere this got lost unless I wasn't looking hard enough for this code!).

Yup, you're right. We definitely should, though.

> this is just for `PATCH` requests, in which case we already have the post type from the post object

Hmm, are you sure? It looks like we use this for [post creation as well](https://github.com/DoSomething/rogue/blob/4a12be9f680f7cec776be1e81573357440a0cd5a/app/Http/Requests/PostRequest.php#L40) to me.

chloealee: Oh shoot you're totally right, @DFurnes ! I missed that line! Ok so I think it is better to revert back to still requiring `type` then, cc @rapala61 @aaronschachter . 

@DFurnes to revert this PR back, can I just git rebase and drop all the new commits that adds this work? 

Also, [here](https://www.pivotaltracker.com/n/projects/2019429/stories/164308611) is a card to make sure we are actually requiring the `file` when a `photo` post is created!

docs/endpoints/posts.md
chloealee: Resolved [here](https://github.com/DoSomething/rogue/pull/837/files#r261276786).

app/Http/Requests/PostRequest.php
rapala61: @chloealee reading through the comments. Am I understanding correctly that type will continue to be expected regardless of sending `action_id`? Also, another question related to the changes I have to make in [Gateway-js](https://github.com/DoSomething/gateway-js/blob/master/server/src/endpoints/rogue/posts.js#L62). I'll move that line to Gambit so that in the event `action_id` is not included in the payload, `action: 'default'` will be sent instead. Is this inn line with the changes in this PR?

chloealee: > Am I understanding correctly that `type` will continue to be expected regardless of sending `action_id`

@rapala61 yes that is correct! We do some validation around the post type (so if it is a photo post, we want to validate that there is a `file` in the payload as well) which we want to have this type handy instead of doing a DB call. 

> I'll move that line to Gambit so that in the event `action_id: 1234` is not included in the payload, `action: 'default'` will be sent instead. Is this in line with the changes in this PR?

Yes! This is a little tricky though because if the action has a different name than `default`, it won't find the action in the query. I think the work around right now is that we're creating separate actions for SMS posts as with the name of `default`. I think this is how we have to go for now if there is no `action_id` because there is no way for a user to enter this info in gambit I think? cc @ngjo

ngjo: Yup - I'm working w/ Anthony tomorrow to make sure he's setting up action ids in Rogue for SMS for the time being to have `default` so things work!

chloealee: Thanks for confirming, @ngjo !

DFurnes: So to wrap up this conversation, we need to revert the `required_without:action_id` on this line to `required`!

chloealee: ah yes, thanks for catching that, @DFurnes ! I was rebasing and it must have changed back then! Updated :)

PR #840, opened 2019-02-27
adds migration to increase a post's text character limit from 255 to 500

chloealee:
#### What's this PR do?
adds migration to increase a post's text character limit from 255 to 500

👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/163867787) Pivotal Card.

#### How to deploy

### DIFF
--- a/database/migrations/2019_02_27_201348_increase_text_character_count_in_posts_table.php
+++ b/database/migrations/2019_02_27_201348_increase_text_character_count_in_posts_table.php
@@ -14,7 +14,7 @@ class IncreaseTextCharacterCountInPostsTable extends Migration
     public function up()
     {
         Schema::table('posts', function (Blueprint $table) {
-            $table->string('text')->default(NULL)->change();
+            $table->string('text')->default(null)->change();
             $table->string('text', 500)->nullable()->change();
         });
     }
--- a/database/migrations/2019_02_27_201348_increase_text_character_count_in_posts_table.php
+++ b/database/migrations/2019_02_27_201348_increase_text_character_count_in_posts_table.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class IncreaseTextCharacterCountInPostsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('posts', function (Blueprint $table) {
+            $table->string('text', 500)->change();
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('posts', function (Blueprint $table) {
+            $table->string('text', 255)->change();
+        });
+    }
+}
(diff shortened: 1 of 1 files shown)


chloealee:
when I run this on my local, I'm getting: 
```
  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; ch
  eck the manual that corresponds to your MariaDB server version for the right syntax to use near
  ''' COLLATE utf8_unicode_ci' at line 1 (SQL: ALTER TABLE posts CHANGE text text VARCHAR(500) DEF
  AULT ''NULL'' COLLATE utf8_unicode_ci)
```

not sure why it is saying `text text` but I think this is the correct way to do this?

DFurnes:
It looks like this is upset at the `DEFAULT ''NULL''`, although the error message definitely doesn't make that very clear! The `text text` is a funny one, but technically correct – it's because we have a field named "text" with a SQL type of `TEXT`.

Looking at the structure for this table, you can see that something's a little off:

```sql
CREATE TABLE `posts` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  # ...
  `url` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
  `text` varchar(255) COLLATE utf8_unicode_ci DEFAULT 'NULL', # <--- ?!?
```

Somewhere along the line we must've tried to set the default for this column to `NULL` but accidentally set it to the string `'NULL'`! Hopefully if we change that back to `NULL`, this query will clear up.

chloealee:
ah, @DFurnes thanks for explaining all of this! Where did you find that structure of the table?

DFurnes:
I snagged that from [TablePlus](https://tableplus.com)'s structure tab, but I assume the same thing would show up in SequelPro or elsewhere (although differentiating `'NULL'` from `NULL` is still a sneaky one regardless!).

chloealee:
ah yes it is hard to tell the difference just in the GUI in Sequel Pro - thank you! I tried to just set that to `nullable()` again to do the trick but it didn't seem to like that - is there another way that we can change that default value?

DFurnes:
@chloealee I'm not sure off the top of my head! If changing it with Laravel's schema builder doesn't work, you could try changing it with a raw SQL query using `DB::statement('ALTER TABLE...')`

chloealee:
thanks, @DFurnes ! That worked! This is now updated and ready for final 👀

database/migrations/2019_02_27_201348_increase_text_character_count_in_posts_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('posts', function (Blueprint $table) {
+            DB::statement('ALTER TABLE posts MODIFY COLUMN text VARCHAR(500)');
```
DFurnes: @chloealee Can we fix the `'NULL'` string default here using the `DB::statement()`, and then use the normal schema builder to change the column length? We don't want to be stuck making all our schema changes using manual SQL statements from now on!

chloealee: oh yes great call! Updated and when changing that, I could use the schema builder for both!

PR #841, opened 2019-02-27
Throws 500 error code and adds logging if no action is found

chloealee:
#### What's this PR do?
Throws 500 error code if no action is found and logs what is going on (before it was throwing a misleading `404` error code since we were using `findOrFail`). 

- Try to create a post via the `POST /api/v3/posts` endpoint with an `action` that doesn't exist. 
- You should receive a `500` error code with this message in response: `"No action found.",`. 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/164286560) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -87,7 +87,11 @@ class PostRepository
                 'campaign_id' => $signup->campaign_id,
                 'post_type' => $data['type'],
                 'name' => $data['action'],
-            ])->firstOrFail();
+            ])->first();
+
+            if (! $action) {
+                abort(500, 'No action found.');
+            }
 
             $actionId = $action->id;
         }
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -97,7 +97,7 @@ class PostRepository
                     'name' => $data['action'],
                 ]);
 
-                throw new ValidationException::withMessages(array_fill_keys(['campaign_id', 'post_type', 'name'], 'An action with the given fields does not exist.'));
+                throw ValidationException::withMessages(array_fill_keys(['campaign_id', 'post_type', 'name'], 'An action with the given fields does not exist.'));
             }
 
             $actionId = $action->id;
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -9,6 +9,7 @@ use Rogue\Models\Review;
 use Rogue\Models\Signup;
 use Rogue\Services\Registrar;
 use Intervention\Image\Facades\Image;
+use Illuminate\Validation\ValidationException;
 
 class PostRepository
 {
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -90,6 +90,12 @@ class PostRepository
             ])->first();
 
             if (! $action) {
+                info('action_not_found', [
+                    'campaign_id' => $signup->campaign_id,
+                    'post_type' => $data['type'],
+                    'name' => $data['action'],
+                ]);
+
                 abort(500, 'No action found.');
             }
 
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -96,7 +96,7 @@ class PostRepository
                     'name' => $data['action'],
                 ]);
 
-                abort(500, 'No action found.');
+                throw new ValidationException::withMessages(array_fill_keys(['campaign_id', 'post_type', 'name'], 'An action with the given fields does not exist.'));
             }
 
             $actionId = $action->id;
--- a/docs/endpoints/posts.md
+++ b/docs/endpoints/posts.md
@@ -188,6 +188,16 @@ Required params:
 Optional params:
 
 - **quantity**: (int|nullable).
+- **campaign_id**: (int) required without action_id.
+  The Campaign ID of the campaign that the user's post is associated with.
+- **type**: (string) required.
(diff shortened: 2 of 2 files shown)


app/Repositories/PostRepository.php
```diff
@@ -81,13 +81,17 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
...
             $action = Action::where([
                 'campaign_id' => $signup->campaign_id,
-                'post_type' => $type,
+                'post_type' => $data['type'],
                 'name' => $data['action'],
-            ])->firstOrFail();
+            ])->first();
+
+            if (! $action) {
+                abort(500, 'No action found.');
```
aaronschachter: This branch is named `422`... is there a reason to return a 500 instead of a 422?

Also - would it make sense to log this message so we can find specific instances of these errors in Papertrail?

DFurnes: We _could_ manually create a validation error here like so:

```php
throw new ValidationException::withMessages(array_fill_keys(['campaign_id', 'post_type', 'name'], 'An action with the given fields does not exist.'));
```

My concern with that is that we'd be breaking our contract of how validation works, since a request might be partially completed (e.g. a signup has been created, and now we're moving on to creating the post). Ideally, all of our validation happens before we make any "actual" changes to the database.

chloealee: > 
This branch is named `422`... is there a reason to return a 500 instead of a 422?

that was a mistake by my part and got lazy to rename!

> 
Also - would it make sense to log this message so we can find specific instances of these errors in Papertrail?

Yes - thank you for the reminder! Added!

app/Http/Requests/PostRequest.php
```diff
@@ -26,12 +26,12 @@ public function rules()
         }
 
         return [
-            'campaign_id' => 'required|integer',
+            'campaign_id' => 'required_without:action_id|integer',
             'northstar_id' => 'nullable|objectid',
-            'type' => 'required|string|in:photo,voter-reg,text,share-social',
+            'type' => 'required_without:action_id|string|in:photo,voter-reg,text,share-social',
```
DFurnes: Will this make our conditional validation rules more complicated (since we'd need to make a database query to determine what kind of a payload to expect for a given `action_id`)? Thinking about how `file` (should) be required only for `photo` type posts, and we [allow different statuses](https://github.com/DoSomething/rogue/blob/4a12be9f680f7cec776be1e81573357440a0cd5a/app/Http/Requests/PostRequest.php#L66-L77) based on the type of post as well.

chloealee: Sorry @DFurnes this PR got a little mixed up bc I accidentally branched from this - but these changes were intended for updates in #837 ! 

I'm going to move the discussion there!

app/Repositories/PostRepository.php
chloealee: @DFurnes ah I see what you're saying about how we'd want this before the signup gets created too! Would it be really messy to put the logic in the [Post Request](https://github.com/DoSomething/rogue/blob/master/app/Http/Requests/PostRequest.php#L27)? Or, I actually think maybe the [Post Controller](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/PostsController.php#L110) might be better to do that logic to to the DB query to the `actions` table with those variables?

DFurnes: Since this is temporary, I'm good with manually throwing a `ValidationException` here (see code block above). It's probably not worth complicating things too too much for something we'll likely tear out as soon as other systems are updated to send `action_id`.

(But yeah, it's tricky! Ideally, we want all our validation to happen in the `PostRequest`, for example using the [`exists`](https://laravel.com/docs/5.8/validation#rule-exists) operator. Unfortunately, I don't think Laravel has a good way for us to do as complicated of a validation rule as this would be!)

chloealee: Sounds good, thanks for clarifying, @DFurnes ! Updated :)

docs/endpoints/posts.md
```diff
@@ -210,6 +211,8 @@ Deprecated params:
   The Campaign ID of the campaign that the user's post is associated with.
 - **action**: (string) required without action_id.
   Describes the bucket the action is tied to. A campaign could ask for multiple types of actions throughout the life of the campaign.
+  - **type**: (string) required without action_id.
```
DFurnes: I think this snuck in here in a rebase! (It's correctly noted as "required" above.)

chloealee: ah yes, thanks for catching!

app/Repositories/PostRepository.php
```diff
@@ -87,7 +88,17 @@ public function create(array $data, $signupId, $authenticatedUserRole = null)
...
+
+            if (! $action) {
+                info('action_not_found', [
+                    'campaign_id' => $signup->campaign_id,
+                    'post_type' => $data['type'],
+                    'name' => $data['action'],
+                ]);
+
+                throw new ValidationException::withMessages(array_fill_keys(['campaign_id', 'post_type', 'name']), 'An action with the given fields does not exist.');
+            }
```
chloealee: @DFurnes I'm getting this error but I think everything looks ok to me? 
```
Unexpected 'withMessages' (T_STRING), expecting variable (T_VARIABLE) or '$' on line 100.
```

DFurnes: Oh oops, since we're calling a static method on that class (which returns an instance), we can just `throw ValidationException::withMessages(…)`.

chloealee: ah that worked! Thank you!

PR #842, opened 2019-02-28
Add 'location_name' computed field.

DFurnes:
#### What's this PR do?
This pull request adds a `location_name` computed field that we can use to fill in the "anonymous gallery" in the upcoming Prom campaign, where posts will be "attributed" to the user's state:

screen shot 2019-02-28 at 2 13 47 pm

This turns out to be way more time-consuming than I would've imagined (40-60ms each lookup):

screen shot 2019-02-28 at 2 03 01 pm

Since that ends up accounting for quite a bit of time on an "index" page with 20+ results, I decided to cache these lookups. Since a state's region code is never (rarely? haha) going to change & takes up very little memory, these are cached forever. Much better:

screen shot 2019-02-28 at 2 05 35 pm

The best I can tell, that first 80ms lookup is just from initializing the Redis connection. (Either that, or it just has a hard time remembering Minnesota. Dunno what Redis has against the Land of 10,000 Lakes!)

👑

#### Relevant tickets
References [#164214415](https://www.pivotaltracker.com/story/show/164214415).

#### How to deploy

### DIFF
--- a/app/Http/Transformers/PostTransformer.php
+++ b/app/Http/Transformers/PostTransformer.php
@@ -48,6 +48,7 @@ class PostTransformer extends TransformerAbstract
             ],
             'status' => $post->status,
             'location' => $post->location,
+            'location_name' => $post->location_name,
             'created_at' => $post->created_at->toIso8601String(),
             'updated_at' => $post->updated_at->toIso8601String(),
         ];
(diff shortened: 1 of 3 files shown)


DFurnes:
Going to merge this in now that we've fixed that up!

PR #847, opened 2019-03-05
Updates GET /api/v3/campaigns endpoint to filter by internal_title and include actions

chloealee:
#### What's this PR do?
Updates `GET /api/v3/campaigns` endpoint to filter by `internal_title` and adds `include=actions`.
  - This is so CallPower can find a `action_id` to send along when creating a Post in Rogue. 
    - It will find the campaign by the `internal_title` and match the `action` by `internal_name` that is the standardized for all CallPower actions. This way, we can get the `action_id`. 

Can you use `filter[internal_name]=` & `include=actions` to find the specific campaign you're looking for? 

#### Relevant tickets
The first implementation step in [this](https://docs.google.com/document/d/1aDNLKZ6BPto65Ce5RDJkXf8LpKGIFdftO13WgZgkO48/edit#) document. 

#### How to deploy

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -30,6 +30,11 @@ class CampaignsController extends ApiController
     {
         $query = $this->newQuery(Campaign::class);
 
+        if (has_include($request, 'actions')) {
+            // Eagerly load the `action` relationship.
+            $query->with('actions');
+        }
+
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Campaign::$indexes);
 
--- a/app/Http/Transformers/CampaignTransformer.php
+++ b/app/Http/Transformers/CampaignTransformer.php
@@ -8,6 +8,15 @@ use League\Fractal\TransformerAbstract;
 class CampaignTransformer extends TransformerAbstract
 {
     /**
+     * List of resources possible to include
+     *
+     * @var array
+     */
+    protected $availableIncludes = [
+        'actions',
+    ];
+
+    /**
      * Transform resource data.
      *
      * @param \Rogue\Models\Campaign $campaign
@@ -23,5 +32,16 @@ class CampaignTransformer extends TransformerAbstract
             'created_at' => $campaign->created_at->toIso8601String(),
             'updated_at' => $campaign->updated_at->toIso8601String(),
         ];
+    }
+
+    /**
+     * Include the actions.
+     *
+     * @param \Rogue\Models\Campaign $campaign
+     * @return \League\Fractal\Resource\Collection
+     */
+    public function includeActions(Campaign $campaign)
+    {
+        return $this->collection($campaign->actions, new ActionTransformer);
     }
 }
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -31,7 +31,7 @@ class Campaign extends Model
      *
      * @var array
      */
-    public static $indexes = ['id'];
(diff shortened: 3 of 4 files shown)


katiecrane:
I have a question about this - is `internal_title` stored in Call Power and not `campaign_id`?

chloealee:
Great q! This is mostly for go-forward records. `campaign_id` is not stored in CallPower and is a little tricky to map back to our `campaign_id`s. 

This is how I _think_ the logic will have to follow: 
- For historical records, I _think_ we'll have to get a list of campaigns and what they're called in CallPower to match up their `campaign_id`. We'll then hardcode this logic to match, if X name in CallPower, `campaign_id` = Y. 
- For go-forward records, I think we'll have to notify whoever sets up the campaign in CallPower to name it exactly what the campaign name is in Rogue. This way, we can search Rogue for the campaign name with these changes. 

For actions for both historical and go-forward, we'll want to also come up with an action `name` that is the same for ALL CallPower actions. This way, when CallPower sends a Post to Rogue, we can hardcode the `action` and `campaign_id` to find the `action_id`.

Now that I think about this though, we are eventually going to deprecate `action` for creating a post so we might want to find a way for CallPower to find the `action_id` and just send that along, instead of the `campaign_id` and `action`... 

This seems like a lot of calls to Rogue though 🤔

chloealee:
hm maybe I can add an `include=actions` to this endpoint as well so that we can match the name to get the `action_id`?

chloealee:
@katiecrane updated this! So now, CallPower should be able to search by `internal_title` and then will receive the campaign and all of it's actions. If we standardize the action name across CallPower actions, then we can get the `action_id` and just this when creating a Post instead of `campaign_id` and `action`, since `action` will be deprecated in the future. 

Not sure though if there will only be one CallPower action per campaign - let me double check in the Technical Spec to confirm.

katiecrane:
@chloealee thank you for explaining! If this is mainly for records moving forward, why are we going with `internal_title` instead of `campaign_id` in CallPower?

chloealee:
@katiecrane sure thing! I don't think there is any mapping in CallPower for our campaigns and there are multiple campaigns that people can use CallPower for.

DFurnes:
@chloealee I'm also confused by the use of campaign's `internal_title` here, rather than the actual `action_id`. What if a campaign has more than one CallPower action (or we change a campaign name)?

katiecrane:
@chloealee I thought that we were in control of whatever field in CallPower would be the `internal_title`, would we not be able to use `campaign_id` or `action_id` for that field in CallPower instead?

chloealee:
@DFurnes @katiecrane  yeah - it is super finicky and not fool proof for sure. I'm trying to think of a way that we can grab the `action_id` from Rogue. 

CallPower doesn't have any mapping to our `campaign_id`s created in Rogue so I was thinking that if we have the `internal_name`, we can get this info from this endpoint in Rogue. 

> would we not be able to use campaign_id or action_id for that field in CallPower instead?

We totally can input whatever we want here! However, for historical posts, there are calls across many different campaigns. So we wouldn't be able to hardcode this since they're not for the same campaign. 

Is there something I'm missing maybe/overthinking and overcomplicating this?

chloealee:
Closing this PR! After chatting with @DFurnes, we found a better solution than this. More details in the [technical spec](https://docs.google.com/document/d/1aDNLKZ6BPto65Ce5RDJkXf8LpKGIFdftO13WgZgkO48/edit).

PR #849, opened 2019-03-06
Adds CallPower campaign id to action creation

chloealee:
#### What's this PR do?
- Adds `phone-call` post `type` and post `type` validation. 
- Updates documentation.
- Adds CallPower campaign id to action creation form (front end and backend): 
screen shot 2019-03-06 at 12 51 21 pm
- Adds CallPower campaign id to campaign id show page: 
screen shot 2019-03-06 at 12 50 45 pm

Are you able to create a new action with a CallPower campaign id? Does this save in the db correctly? 

@ngjo @lkpttn can we add this to the campaign creation playbook? LMK if you have any questions! 
#### Relevant tickets
Fixes (this)[https://www.pivotaltracker.com/n/projects/2019429/stories/164436959] Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -87,6 +87,7 @@ class ActionsController extends Controller
         $this->validate($request, [
             'name' => 'string',
             'post_type' => 'string',
+            'callpower_campaign_id' => 'integer',
             'reportback' => 'boolean',
             'civic_action' => 'boolean',
             'scholarship_entry' => 'boolean',
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -47,7 +47,7 @@ class ActionsController extends Controller
             'name' => 'required|string',
             'campaign_id' => 'required|integer|exists:campaigns,id',
             'post_type' => 'required|string|in:photo,voter-reg,text,share-social,phone-call',
-            'callpower_campaign_id' => 'integer',
+            'callpower_campaign_id' => 'required_if:post_type,phone-call|integer',
             'noun' => 'required|string',
             'verb' => 'required|string',
         ]);
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -47,7 +47,7 @@ class ActionsController extends Controller
             'name' => 'required|string',
             'campaign_id' => 'required|integer|exists:campaigns,id',
             'post_type' => 'required|string|in:photo,voter-reg,text,share-social,phone-call',
-            'callpower_campaign_id' => 'required_if:post_type,phone-call|integer',
+            'callpower_campaign_id' => 'required_if:post_type,phone-call|integer|exists:actions,callpower_campaign_id',
             'noun' => 'required|string',
             'verb' => 'required|string',
         ]);
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -86,8 +86,8 @@ class ActionsController extends Controller
     {
         $this->validate($request, [
             'name' => 'string',
-            'post_type' => 'string',
-            'callpower_campaign_id' => 'integer',
+            'post_type' => 'string|in:photo,voter-reg,text,share-social,phone-call',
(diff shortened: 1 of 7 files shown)


app/Http/Controllers/Legacy/Web/ActionsController.php
```diff
@@ -45,7 +46,8 @@ public function store(Request $request)
         $this->validate($request, [
             'name' => 'required|string',
             'campaign_id' => 'required|integer|exists:campaigns,id',
-            'post_type' => 'required|string',
+            'post_type' => 'required|string|in:photo,voter-reg,text,share-social,phone-call',
+            'callpower_campaign_id' => 'integer',
```
DFurnes: Let's make sure this is [`required_if`](https://laravel.com/docs/5.5/validation#rule-required-if) the `post_type` is `phone-call`.

resources/assets/app.scss
```diff
@@ -40,8 +40,8 @@ button.delete {
   }
 }
 
-.-half {
+.-third {
```
DFurnes: Were we using this `.-half` class anywhere else?

We also should be careful with modifiers like this. Usually we pair a `.-x` modifier with another class (so like, `.form-item.-third`), so that it wouldn't conflict with another class using a modifier of the same name (like `.container__row.-third`).

app/Http/Controllers/Legacy/Web/ActionsController.php
DFurnes: We also probably want to make sure we don't attribute a single CallPower campaign to more than one action, so let's add an [`exists`](https://laravel.com/docs/5.5/validation#rule-exists) rule for that.

resources/views/actions/create.blade.php
```diff
@@ -29,11 +29,15 @@
                             @endforeach
                         </select>
                     </div>
-                    <div class="form-item -half">
+                    <div class="form-item -third">
+                        <label class="field-label">CallPower Campaign ID</label>
+                        <input type="text" name="callpower_campaign_id" class="text-field" placeholder="e.g. 4" value="{{ old('callpower_campaign_id') }}">
```
DFurnes: It'd be nice to conditionally hide this field if we're not making a `phone-call` action, but that's a little bit more complicated to do outside of React-land. If you feel up for a little vanilla JS though, it'd be a nice UX win!

app/Http/Controllers/Legacy/Web/ActionsController.php
chloealee: great call @DFurnes ! 

With `integer` also in the validation, it was throwing an error when I tried to create other post types, saying that `callpower_campaign_id` had to be an integer (and it was empty). Seems like this is a [known issue](https://github.com/laravel/framework/issues/17032) with some hacky fixes. To get around this, I just removed `integer` validation since we're using the `exists` validation and I think that should work!

resources/assets/app.scss
chloealee: ah great call! I just added the `.-half` back in and created a separate element for the thirds.

app/Http/Controllers/Legacy/Web/ActionsController.php
chloealee: > We also probably want to make sure we don't attribute a single CallPower campaign to more than one action, so let's add an exists rule for that.

@DFurnes oh shoot - testing this on QA I think I misunderstood what you were saying here! I need to update this validation in a new PR but not sure how to make sure a CallPower campaign only belongs to one action  with the `exists` rule?

DFurnes: > With `integer` also in the validation, it was throwing an error when I tried to create other post types, saying that `callpower_campaign_id` had to be an integer (and it was empty). Seems like this is a [known issue](https://github.com/laravel/framework/issues/17032) with some hacky fixes. To get around this, I just removed `integer` validation since we're using the `exists` validation and I think that should work! 

This seems to be working for me. Can you try again?

```php
'post_type' => 'required|string|in:photo,voter-reg,text,share-social,phone-call',
'callpower_campaign_id' => 'required_if:post_type,phone-call|integer'
```

> I need to update this validation in a new PR but not sure how to make sure a CallPower campaign only belongs to one action with the `exists` rule?

Oops, sorry. Meant to link to its opposite – the [`unique`](https://laravel.com/docs/5.5/validation#rule-unique) rule!

chloealee: ah got it! PR [here](https://github.com/DoSomething/rogue/pull/850)!

PR #850, opened 2019-03-06
Updates callpower_campaign_id validation

chloealee:
#### What's this PR do?
-Updates validation so `callpower_campaign_id` is an integer and has a unique action `id` so a CallPower campaign can only belong to one action.

- You shouldn't be able to add a `callpower_campaign_id` that isn't an integer.
- There should only be one action associated to one CallPower campaign - you should not be able to make two actions for the same CallPower campaign. 

**Question**: Is this the correct logic that we want? I think we'll need to put this rule in place in order to find the `action_id` but is it ok if we only have one action per CallPower campaign? Would we ever want multiple? cc @sohaibhasan @mjain-1 @ngjo 

#### Relevant tickets
Fixes validation in #849 

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -88,7 +88,10 @@ class ActionsController extends Controller
         $this->validate($request, [
             'name' => 'string',
             'post_type' => 'string|in:photo,voter-reg,text,share-social,phone-call',
-            'callpower_campaign_id' => ['required_if:post_type,phone-call', Rule::unique('actions')->ignore($action)],
+            'callpower_campaign_id' => [
+                'required_if:post_type,phone-call',
+                Rule::unique('actions')->ignore($action->id),
+            ],
             'reportback' => 'boolean',
             'civic_action' => 'boolean',
             'scholarship_entry' => 'boolean',
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -47,7 +47,7 @@ class ActionsController extends Controller
             'name' => 'required|string',
             'campaign_id' => 'required|integer|exists:campaigns,id',
             'post_type' => 'required|string|in:photo,voter-reg,text,share-social,phone-call',
-            'callpower_campaign_id' => 'nullable|required_if:post_type,phone-call|integer|unique:actions,id',
+            'callpower_campaign_id' => 'nullable|required_if:post_type,phone-call|integer|unique:actions',
             'noun' => 'required|string',
             'verb' => 'required|string',
         ]);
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/Legacy/Web/ActionsController.php
```diff
@@ -47,7 +47,7 @@ public function store(Request $request)
             'name' => 'required|string',
             'campaign_id' => 'required|integer|exists:campaigns,id',
             'post_type' => 'required|string|in:photo,voter-reg,text,share-social,phone-call',
-            'callpower_campaign_id' => 'required_if:post_type,phone-call|exists:actions,callpower_campaign_id',
+            'callpower_campaign_id' => 'nullable|required_if:post_type,phone-call|integer|unique:actions,id',
```
DFurnes: I think we want this to be to be `unique:actions,callpower_campaign_id` right (since that's the column on that table that we want to be unique, not `id`)? If so, we can actually shorten this to `unique:actions` as a shorthand!

DFurnes: Let's also make sure to update the rule in the `update` method down below, which is still using `exists`! Since we want to "ignore" the existing action when checking uniqueness, we'll need to follow the instructions under the ["Forcing A Unique Rule To Ignore A Given ID"](https://laravel.com/docs/5.8/validation#rule-unique) heading there.

DFurnes:
> I think we'll need to put this rule in place in order to find the action_id but is it ok if we only have one action per CallPower campaign? Would we ever want multiple?

I think we _have_ to have one action per CallPower campaign. We need some way to determine how to bucket these when they enter our ecosystem – if we had a single CallPower campaign tied to more than one action, we'd be unable to decide where to put an incoming call!

We _could_ decide to have multiple CallPower campaigns attributed to one action, but that would increase complexity on the Rogue end (we'd need a separate table for this many-to-one mapping) and seems like something we'd likely not want to do in practice (since we'd be "losing" some data when importing calls then).

chloealee:
That totally makes sense, thanks @DFurnes ! I think that is a good idea to keep things simple and you're right, probably won't be having multiple actions for one CallPower campaign anyway!

app/Http/Controllers/Legacy/Web/ActionsController.php
```diff
@@ -84,10 +85,16 @@ public function store(Request $request)
      */
     public function update(Request $request, Action $action)
     {
+        Validator::make($request, [
+            'callpower_campaign_id' => [
+                'required_if:post_type,phone-call',
+                Rule::unique('actions')->ignore($action),
+            ],
+        ]);
+
```
chloealee: @DFurnes is this the correct way to do this?

DFurnes: Does it work when you test it out? It looks like this is instantiating a validator, but not using it (which you'd do by calling `->validate()` on the returned "validator" object, or conditionally running code on `->fails()`).

You can group this with the rest of our validation rules in the `$this->validate(…)` call below:

```php
$this->validate($request, [
    // ...
    'callpower_campaign_id' => ['required_if:post_type,phone-call', Rule::unique('actions')->ignore($action)],
```

chloealee: @DFurnes got this to work! I had to include `$action->id` instead of just `$action` because for some reason it seemed to be grabbing the `campaign_id` here.

PR #852, opened 2019-03-08
Updates GET api/v3/actions to be able to filter by callpower_campaign_id 

chloealee:
#### What's this PR do?
- Updates `GET api/v3/actions` to be able to filter by `callpower_campaign_id`.
- Makes `callpower_campaign_id` a unique index in the `actions` table.
- Updates Action Transformer to return `callpower_campaign_id` for each action.
- Updates documentation. 

Can you do this? 

This is needed to find the action `id` when creating a post from CallPower records. 

#### Relevant tickets
Fixes the first task in [this](https://www.pivotaltracker.com/n/projects/2019429/stories/164476751) Pivotal Card.

#### How to deploy


app/Models/Action.php
```diff
@@ -36,7 +36,7 @@ class Action extends Model
      *
      * @var array
      */
-    public static $indexes = ['id'];
+    public static $indexes = ['id', 'callpower_campaign_id'];
```
DFurnes: I don't believe this column has an index in the database. Can we add one as part of this PR?

docs/endpoints/actions.md
```diff
@@ -24,6 +24,7 @@ Example Response:
             "name": "august-2018-turbovote",
             "campaign_id": 9005,
             "post_type": "voter-reg",
+            "callpower_campaign_id": null,
```
DFurnes: This is so nitpicky, but let's leave these as `null` since they're `post_type` of `voter-reg` and `share-social`. (Or, if we want to demonstrate how this field would look, let's change one of these actions' type to `phone-call`).

chloealee: Not nitpicky at all! This is great - we should have accurate documentation for less confusion! :)

database/migrations/2019_03_08_204743_adds_callpower_campaign_id_unique_index_in_actions_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('actions', function (Blueprint $table) {
+            $table->index('callpower_campaign_id');
```
DFurnes: Since we ensure that this field is unique in our validation rules, we can use a [unique index](https://laravel.com/docs/5.8/migrations#indexes) here.

chloealee: ah yes - updated!

PR #855, opened 2019-03-12
Uses date-fns to format dates so it shows up as valid dates in Chrome and Safari 

chloealee:
#### What's this PR do?
Uses [date-fns](https://date-fns.org/v1.30.1/docs/format) to format dates so it shows up as valid dates in Chrome and Safari.
 
👀 

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/164282296) Pivotal Card.

#### How to deploy

### DIFF
--- a/resources/assets/components/CampaignIdSingle/index.js
+++ b/resources/assets/components/CampaignIdSingle/index.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { map, isEmpty } from 'lodash';
-import { parse } from 'date-fns';
+import { format } from 'date-fns';
 
 import './campaignidsingle.scss';
 
@@ -82,13 +82,9 @@ class CampaignIdSingle extends React.Component {
             <p>–</p>
           )}
           <h4>Start Date</h4>
-          <p>{parse(campaign.start_date)}</p>
+          <p>{format(new Date(campaign.start_date), 'MM/dd/yyyy')}</p>
           <h4>End Date</h4>
-          <p>
-            {campaign.end_date
-              ? new Date(campaign.end_date).toDateString()
-              : '–'}
-          </p>
+          <p>{campaign.end_date ? format(new Date(campaign.end_date)) : '–'}</p>
         </div>
         <div className="container__block -narrow">
           <a className="button" href={`/campaign-ids/${campaign.id}/edit`}>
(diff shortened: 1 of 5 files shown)


resources/assets/components/CampaignIdSingle/index.js
```diff
@@ -81,8 +82,7 @@ class CampaignIdSingle extends React.Component {
             <p>–</p>
           )}
           <h4>Start Date</h4>
-          <p>{new Date(campaign.start_date).toDateString()}</p>
-
+          <p>{parse(campaign.start_date)}</p>
```
chloealee: @DFurnes I think this is the best way to do this since `campaign.start_date` is a date but when I do this, I get the error: 

```
index.js:319 Uncaught TypeError: 3 arguments required, but only 1 present
    at parse (index.js:319)
```

According to the [documentation](https://date-fns.org/v1.9.0/docs/parse), I think it only takes one argument?

chloealee: oh wait nvm I see now that I was on the wrong version of the documentation!

resources/assets/components/CampaignIdSingle/index.js
```diff
@@ -81,29 +82,19 @@ class CampaignIdSingle extends React.Component {
             <p>–</p>
           )}
           <h4>Start Date</h4>
-          <p>{new Date(campaign.start_date).toDateString()}</p>
-
+          <p>{format(new Date(campaign.start_date), 'MM/dd/yyyy')}</p>
```
DFurnes: I just want to double-check that this is working in Safari now.

I haven't pulled down the branch to my local, but just testing this snippet on QA now and it seems like `new Date(window.STATE.campaign.start_date)` still results in `Invalid Date`.

chloealee: @DFurnes ah thanks for pointing out! When I hit rogue.test in Safari, it sends me to the google search for that. Is there anything I have to update to be able to hit local in Safari?

chloealee: ack I got this to work now - sorry for all the pinging!

chloealee: @DFurnes you're right, this doesn't seem to work either: 
Screen Shot 2019-03-12 at 1 50 08 PM

Looks like because `start_date` and `end_date` are `DATETIME` in the database and probably should be a `TIMESTAMP`?

chloealee: @DFurnes updated with some regex and now it seems to be working both in Chrome and Safari!

resources/assets/components/CampaignIdSingle/index.js
```diff
@@ -46,7 +47,9 @@ class CampaignIdSingle extends React.Component {
   render() {
     const campaign = this.props.campaign;
     const actions = this.props.actions;
-
+    console.log(
+      format(new Date(campaign.start_date.replace(/-/g, '/')), 'MM/dd/yyyy'),
+    );
```
katiecrane: Do we still want this log?

chloealee: oops we do not! Thanks for catching, @katiecrane ! Removed :)

PR #856, opened 2019-03-13
Date fns update

chloealee:
#### What's this PR do?
Updates formatting - things were breaking on QA and I realized I was using the wrong formatting! More from the [docs](https://date-fns.org/v1.9.0/docs/format). 

👀 

#### Relevant tickets
Fixes mistake in #855 

#### How to deploy

### DIFF
--- a/resources/assets/components/CampaignIdSingle/index.js
+++ b/resources/assets/components/CampaignIdSingle/index.js
@@ -85,7 +85,7 @@ class CampaignIdSingle extends React.Component {
           <p>
             {format(
               new Date(campaign.start_date.replace(/-/g, '/')),
-              'MM/dd/yyyy',
+              'MM/DD/YYYY',
             )}
           </p>
           <h4>End Date</h4>
@@ -93,7 +93,7 @@ class CampaignIdSingle extends React.Component {
             {campaign.end_date
               ? format(
                   new Date(campaign.end_date.replace(/-/g, '/')),
-                  'MM/dd/yyyy',
+                  'MM/DD/YYYY',
                 )
               : '–'}
           </p>
@@ -106,13 +106,13 @@ class CampaignIdSingle extends React.Component {
             Last updated:{' '}
             {format(
               new Date(campaign.updated_at.replace(/-/g, '/')),
-              'MM/dd/yyyy h:m:s',
+              'MM/DD/YYYY h:m:s',
             )}
             <br />
             Created:{' '}
             {format(
               new Date(campaign.created_at.replace(/-/g, '/')),
-              'MM/dd/yyyy h:m:s',
+              'MM/DD/YYYY h:m:s',
             )}
           </p>
         </div>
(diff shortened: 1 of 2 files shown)


resources/assets/components/CampaignIdSingle/index.js
```diff
@@ -83,18 +83,12 @@ class CampaignIdSingle extends React.Component {
           )}
           <h4>Start Date</h4>
           <p>
-            {format(
-              new Date(campaign.start_date.replace(/-/g, '/')),
-              'MM/dd/yyyy',
-            )}
+            {format(new Date(campaign.start_date.replace(/-/g, '/')), 'M/d/yy')}
```
DFurnes: It might be cleaner to use `date-fn`'s [`parse`](https://date-fns.org/v1.30.1/docs/parse) method here (which I believe should support the given format), rather than applying a regex find-replace to "hack" things into the format that JavaScript is expecting.

chloealee: @DFurnes I thought about using that but think it would get a little messy too trying to format the dates since they are `DATETIME`s and `parse` expects a `TIMESTAMP`?

DFurnes: Where are you seeing that `parse` expects a `TIMESTAMP`?

chloealee: Oh, I totally read that wrong! I'm still having a hard time getting this to work though, still seeing "Invalid Date" when doing something like: `{parse(campaign.updated_at, 'd.MM.yy h:m:s', new Date()).toString()}`

DFurnes: I'm not sure about the usage of `parse()` with three parameters and `.toString()` there. Can you link to where in the docs you're seeing that?

This seems to work for me:

```js
format(parse(campaign.updated_at), 'M/d/YY');
```

chloealee: hm that still doesn't work for me when I test out in Safari :-/ 

I'm getting this error: `TypeError: 3 arguments required, but only 1 present` from `parse`. 

I found the example [here](https://date-fns.org/v2.0.0-alpha.27/docs/parse#v2.0.0-breaking-changes) to use three parameters after seeing this error (and even tried to drop the `.toString()` that I found [here](https://github.com/date-fns/date-fns/blob/master/docs/unicodeTokens.md) but doesn't seem to be working :-/ 

What is weird though is that I thought we were using `v1.30.1` but it seems to still be adopting the new rules...

DFurnes: Huh – yeah, it looks like we [ended up with the 2.0 beta](https://github.com/DoSomething/rogue/blob/ec3eb8a9629a87b2e18b5adf351dd16476296ce1/package-lock.json#L2918-L2922) somehow.

DFurnes: Not sure why that's happening, but looks like you can use [`toDate`](https://date-fns.org/v2.0.0-alpha.27/docs/toDate) for this in 2.x:

```js
format(toDate("2019-03-28"), 'M/d/YY');
```

chloealee: @DFurnes updated to the correct version and all is working now! Thanks for the help on that! PR [here](https://github.com/DoSomething/rogue/pull/857).

PR #857, opened 2019-03-13
Updates to correct version of date-fns and uses combination of format and parse to get correct date

chloealee:
#### What's this PR do?
Updates to correct version of date-fns and uses combination of format and parse to get correct date

#### How should this be reviewed?
Everything working correctly in Safari and Chrome?

#### Relevant tickets
Fixes mistake in #856 

#### How to deploy
https://github.com/DoSomething/rogue/blob/master/docs/development/deployments.md

PR #858, opened 2019-03-14
Adds Edit and Delete Functionality for Actions

chloealee:
#### What's this PR do?
- Adds Edit and Delete Functionality for Actions: 
  - Edit and Delete buttons only show up if a campaign is not yet live: 
Screen Shot 2019-03-14 at 5 14 29 PM
  - When you click on the "EDIT ACTION" button, it takes you to this form: 
Screen Shot 2019-03-14 at 5 15 16 PM
  - Pressing the "Delete Action" button deletes the action and removes from the page with React. 
- Updates all forms to keep old form values if there are validation errors. 
- Updates `GET /api/v3/actions` to be able to filter by `campaign_id` 
- Updates `DELETE /actions/:id` to return a response 
- Updates documentation. 
- Updates bug with dates to show what it is intended instead of `updated_at` dates. 

Can you update and a delete an action? 
If the campaign has already started, do these buttons still show up? 

While coding this, found a bug on https://activity.dosomething.org/campaign-ids page - the dates are not what they actually are in the database. Wrote a card for that [here](https://www.pivotaltracker.com/n/projects/2019429/stories/164637404)!

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/164022716) Pivotal Card.

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -150,9 +150,15 @@ class ActionsController extends Controller
      */
     public function destroy(Action $action)
     {
-        $action->delete();
+        $trashed = $action->delete();
 
-        // Log that an action was deleted.
-        info('action_deleted', ['id' => $action->id]);
+        if ($trashed) {
+            // Log that an action was deleted.
+            info('action_deleted', ['id' => $action->id]);
+
+            return $this->respond('Action deleted.', 200);
+        }
+
+        return response()->json(['code' => 500, 'message' => 'There was an error deleting the action']);
     }
 }
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Http\Controllers\Legacy\Web;
 
+use Rogue\PostType;
 use Rogue\Models\Action;
 use Illuminate\Http\Request;
 use Illuminate\Validation\Rule;
@@ -23,16 +24,8 @@ class ActionsController extends Controller
      */
     public function create($campaignId)
     {
-        $postTypes = [
-            'text',
-            'photo',
-            'voter-reg',
-            'share-social',
-            'phone-call',
-        ];
-
         return view('actions.create')->with([
-            'postTypes' => $postTypes,
+            'postTypes' => PostType::all(),
             'campaignId' => (int) $campaignId,
         ]);
     }
@@ -82,20 +75,10 @@ class ActionsController extends Controller
      */
     public function edit($campaignId, $actionId)
     {
-        $action = Action::find($actionId);
-
-        $postTypes = [
-            'text',
-            'photo',
-            'voter-reg',
-            'share-social',
-            'phone-call',
-        ];
-
         return view('actions.edit')->with([
             'campaignId' => $campaignId,
-            'action' => $action,
-            'postTypes' => $postTypes,
+            'action' => Action::find($actionId),
+            'postTypes' => PostType::all(),
         ]);
     }
 
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -82,9 +82,20 @@ class ActionsController extends Controller
      */
     public function edit($campaignId, $actionId)
     {
+        $action = Action::find($actionId);
+
+        $postTypes = [
+            'text',
+            'photo',
+            'voter-reg',
+            'share-social',
+            'phone-call',
+        ];
+
         return view('actions.edit')->with([
-            'campaignId' => (int) $campaignId,
-            'actionId' => (int) $actionId,
+            'campaignId' => $campaignId,
+            'action' => $action,
+            'postTypes' => $postTypes,
         ]);
     }
 
@@ -115,6 +126,8 @@ class ActionsController extends Controller
 
         // Log that an action was updated.
         info('action_updated', ['id' => $action->id]);
+
+        return redirect()->route('campaign-ids.show', $action->campaign_id);
     }
 
     /**
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -152,13 +152,13 @@ class ActionsController extends Controller
     {
         $trashed = $action->delete();
 
-        if ($trashed) {
-            // Log that an action was deleted.
-            info('action_deleted', ['id' => $action->id]);
-
-            return $this->respond('Action deleted.', 200);
+        if (! $trashed) {
+            return response()->json(['code' => 500, 'message' => 'There was an error deleting the action']);
         }
 
-        return response()->json(['code' => 500, 'message' => 'There was an error deleting the action']);
+        // Log that an action was deleted.
+        info('action_deleted', ['id' => $action->id]);
+
+        return $this->respond('Action deleted.', 200);
     }
 }
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -122,6 +122,19 @@ class ActionsController extends Controller
             'verb' => 'string',
         ]);
 
+        $checkboxes = [
+                        'reportback',
+                        'civic_action',
+                        'scholarship_entry',
+                        'anonymous',
+                      ];
+
+        foreach ($checkboxes as $checkbox) {
+            if (! isset($request[$checkbox])) {
+                $request[$checkbox] = 0;
+            }
+        }
+
         $action->update($request->all());
 
         // Log that an action was updated.
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -78,6 +78,17 @@ class ActionsController extends Controller
     }
 
     /**
+     * Edit an existing action.
+     */
+    public function edit($campaignId, $actionId)
+    {
+        return view('actions.edit')->with([
+            'campaignId' => (int) $campaignId,
+            'actionId' => (int) $actionId,
+        ]);
+    }
+
+    /**
      * Update the specified resource in storage.
      *
      * @param  \Illuminate\Http\Request  $request
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -112,7 +112,7 @@ class ActionsController extends Controller
             'post_type' => 'string|in:photo,voter-reg,text,share-social,phone-call',
             'callpower_campaign_id' => [
                 'required_if:post_type,phone-call',
-                Rule::unique('actions')->ignore($action->id),
+                Rule::unique('actions')->whereNotNull('callpower_campaign_id')->ignore($action->id),
             ],
             'reportback' => 'boolean',
             'civic_action' => 'boolean',
--- a/app/Http/Controllers/Legacy/Web/CampaignIdsController.php
+++ b/app/Http/Controllers/Legacy/Web/CampaignIdsController.php
@@ -85,7 +85,6 @@ class CampaignIdsController extends Controller
         return view('pages.campaign_id_single')
             ->with('state', [
                 'campaign' => $campaign,
-                'actions' => $campaign->actions,
             ]
         );
     }
--- a/app/Models/Action.php
+++ b/app/Models/Action.php
@@ -36,7 +36,7 @@ class Action extends Model
      *
      * @var array
      */
-    public static $indexes = ['id', 'callpower_campaign_id'];
+    public static $indexes = ['id', 'campaign_id', 'callpower_campaign_id'];
 
     /**
      * Each action belongs to a campaign.
--- a/resources/assets/components/Action/index.js
+++ b/resources/assets/components/Action/index.js
@@ -96,7 +96,7 @@ class Action extends React.Component {
 
             <button
               className="button delete -tertiary"
-              onClick={e => this.deleteSignup(signup.id, e)}
+              onClick={e => this.props.deleteAction(action.id, e)}
             >
               Delete Action
             </button>
--- a/resources/assets/components/Action/index.js
+++ b/resources/assets/components/Action/index.js
@@ -85,6 +85,16 @@ class Action extends React.Component {
             </li>
           </ul>
         </div>
+        {new Date(campaign.start_date.replace(/-/g, '/')) > Date.now() ? (
+          <div className="container__row">
+            <a
+              className="button -secondary"
+              href={`${campaign.id}/actions/${action.id}/edit`}
+            >
+              Edit Action
+            </a>
+          </div>
+        ) : null}
       </div>
     );
   }
--- a/resources/assets/components/Action/index.js
+++ b/resources/assets/components/Action/index.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { format, parse } from 'date-fns';
 
 import './action.scss';
 
@@ -85,7 +86,7 @@ class Action extends React.Component {
             </li>
           </ul>
         </div>
-        {new Date(campaign.start_date.replace(/-/g, '/')) > Date.now() ? (
+        {new Date(format(parse(campaign.start_date), 'M/d/YY')) > Date.now() ? (
           <div className="container__row">
             <a
               className="button -secondary"
--- a/resources/assets/components/Action/index.js
+++ b/resources/assets/components/Action/index.js
@@ -93,6 +93,13 @@ class Action extends React.Component {
             >
               Edit Action
             </a>
+
+            <button
+              className="button delete -tertiary"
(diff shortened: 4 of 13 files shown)


resources/views/actions/edit.blade.php
```diff
@@ -0,0 +1,100 @@
...
+                                value="{{ old('verb') }}"
+                            @else
+                                value="{{ $action->verb }}"
+                            @endif
+                        >
+                    </div>
+                    <div class="form-item">
+                        <label class="field-label">Post Details</label>
+                        <label class="option -checkbox">
+                          <input type="checkbox" id=reportback name=reportback value=1 {{ old('reportback') || $action->reportback ? 'checked="checked"' : '' }}>
```
chloealee: This doesn't work perfectly right now. For example, if the Action is a reportback, and I want to edit it to not be a reportback but there is a validation error, I would expect the form to keep it unchecked when the validation error pops up with `old('reportback')` here. However, this changes the checkbox back to being checked when the validation error pops up because it satisfies `$action->reportback` since it hasn't been updated yet. 

I'm not sure how to get this to work the way I want it to because I can't really depend on discerning at what stage this is at (if it the second time and we want to grab `old('reportback')` vs `$action->reportback`). Kind of hard to explain via text so lmk if anyone wants to talk through this! (Also tried to do a Giphy Capture but for some reason wouldn't let me press "UPDATE ACTION" 🤔 ).

app/Http/Controllers/Legacy/Web/ActionsController.php
```diff
@@ -77,6 +77,28 @@ public function store(Request $request)
...
+    {
+        $action = Action::find($actionId);
+
+        $postTypes = [
+            'text',
+            'photo',
+            'voter-reg',
+            'share-social',
+            'phone-call',
+        ];
```
katiecrane: I wonder if there's a way to save this list somewhere so it doesn't have to be listed out manually in more than once place?

app/Http/Controllers/Legacy/Web/ActionsController.php
```diff
@@ -90,7 +112,7 @@ public function update(Request $request, Action $action)
             'post_type' => 'string|in:photo,voter-reg,text,share-social,phone-call',
             'callpower_campaign_id' => [
                 'required_if:post_type,phone-call',
-                Rule::unique('actions')->ignore($action->id),
+                Rule::unique('actions')->whereNotNull('callpower_campaign_id')->ignore($action->id),
```
katiecrane: Could you explain why we need the `->ignore($action->id)` piece of this? Having trouble wrapping my head around it (sorry I know that piece isn't new here!)

app/Http/Controllers/Legacy/Web/ActionsController.php
```diff
@@ -113,9 +150,15 @@ public function update(Request $request, Action $action)
...
     {
-        $action->delete();
+        $trashed = $action->delete();
+
+        if ($trashed) {
+            // Log that an action was deleted.
+            info('action_deleted', ['id' => $action->id]);
+
+            return $this->respond('Action deleted.', 200);
+        }
```
katiecrane: Do you think this way would be more readable? Totally up to you!
```
if (! $trashed) {
    return response()->json(['code' => 500, 'message' => 'There was an error deleting the action']);
}

// Log that an action was deleted.
info('action_deleted', ['id' => $action->id]);

return $this->respond('Action deleted.', 200);
```

app/Models/Action.php
```diff
@@ -36,7 +36,7 @@ class Action extends Model
      *
      * @var array
      */
-    public static $indexes = ['id', 'callpower_campaign_id'];
+    public static $indexes = ['id', 'campaign_id', 'callpower_campaign_id'];
```
katiecrane: Is this already an index in the database?

app/Http/Controllers/Legacy/Web/ActionsController.php
chloealee: Yes this was confusing to me too! So we want the validation logic to only allow one CallPower Campaign ID per action. To do that, we use `->ignore($action->id)`.

[Here's](https://laravel.com/docs/5.8/validation#forcing-a-unique-rule-to-ignore-a-given-id) the documentation for it but lmk if it still doesn't make sense!

app/Http/Controllers/Legacy/Web/CampaignIdsController.php
```diff
@@ -85,7 +85,6 @@ public function show(Campaign $campaign)
         return view('pages.campaign_id_single')
             ->with('state', [
                 'campaign' => $campaign,
-                'actions' => $campaign->actions,
```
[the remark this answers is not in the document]
chloealee: Yes it does! But I was passing it as a prop before and we actually want it to be in the state (since when we delete the action, we want to update the state to remove it from the page).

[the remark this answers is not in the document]
katiecrane: Ooo I see we use the API to grab the actions now?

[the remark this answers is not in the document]
chloealee: Exactly!

resources/assets/components/CampaignIdSingle/index.js
```diff
@@ -15,13 +17,70 @@ class CampaignIdSingle extends React.Component {
...
+          actions: keyBy(json.data, 'id'),
+        }),
+      );
+  }
+
+  // Delete an action.
+  deleteAction(actionId, event) {
+    event.preventDefault();
+    const confirmed = confirm(
+      '🚨🔥🚨 Are you sure you want to delete this? 🚨🔥🚨',
```
katiecrane: Excellent choice of emojis!! Could be clearer to add the word "action" at the end maybe!

app/Models/Action.php
chloealee: It is!

app/Http/Controllers/Legacy/Web/ActionsController.php
chloealee: omg yes this is so much better! Updated!

resources/assets/components/CampaignIdSingle/index.js
chloealee: Ah I just copy and pasted that from what we had already when deleting a post - I think maybe @sbsmith86 or @DFurnes 's original doing! :) Yes, I'll update that to include action - thanks!

resources/assets/components/CampaignIdSingle/index.js
```diff
@@ -15,13 +17,70 @@ class CampaignIdSingle extends React.Component {
...
+    );
+
+    if (confirmed) {
+      // Make API request to Rogue to delete the action.
+      const response = this.api.delete(`actions/${actionId}`);
+
+      response.then(result => {
```
katiecrane: Does this only happen if it was a good response?

resources/assets/components/Signup/index.js
```diff
@@ -300,7 +300,7 @@ class Signup extends React.Component {
     );
 
     if (confirmed) {
-      // Make API request to Rogue to update the quantity on the backend
+      // Make API request to Rogue to delete the post.
```
katiecrane: Good catch!!

resources/assets/components/CampaignIdSingle/index.js
chloealee: Yes! That is why I had to update the endpoint to return a response. It wasn't doing anything until I returned something and then when it got a good response, the state updated.

resources/views/actions/create.blade.php
```diff
@@ -44,17 +44,17 @@
...
+                          <input type="checkbox" id=civic_action name=civic_action value=1 {{ old('civic_action') ? 'checked="checked"' : '' }}>
                           <span class="option__indicator"></span>
                           <span>Civic action</span>
                         </label>
                         <label class="option -checkbox">
-                          <input type="checkbox" id=scholarship_entry name=scholarship_entry value=1>
+                          <input type="checkbox" id=scholarship_entry name=scholarship_entry value=1 {{ old('scholarship_entry') ? 'checked="checked"' : '' }}>
                           <span class="option__indicator"></span>
                           <span>Scholarship entry</span>
                         </label>
```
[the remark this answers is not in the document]
chloealee: ah yes, thanks for catching that @katiecrane ! Added :) 

This needs the `{{ old('anonymous') ? 'checked="checked"' : '' }}` because if there is a form validation error, this keeps the values the user submitted instead of wiping the form clean and making the user input them again.

resources/views/actions/edit.blade.php
katiecrane: I get what you're saying @chloealee ! Hmmm and if you remove the `$action->reportback` piece it won't be checked when you first get to the page to edit if it should be?

app/Http/Controllers/Legacy/Web/ActionsController.php
katiecrane: Ah that documentation helped! So a CallPower Campaign ID can only live on one action, am I understanding that?

resources/views/actions/create.blade.php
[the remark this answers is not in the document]
katiecrane: That's so cool!

app/Http/Controllers/Legacy/Web/ActionsController.php
chloealee: Yes exactly!

resources/views/actions/edit.blade.php
chloealee: @katiecrane yes exactly! It's so funky!

katiecrane: @chloealee what is the value of `old('reportback')` if the user has not altered the form? Is it `false` or `null`? I'm wondering if there is a way to tell from that whether or not the user has touched that part of the form yet

chloealee: @katiecrane ooo that is a great call! From the [docs](https://laravel.com/docs/5.8/requests#retrieving-old-input), looks like it is `null`! However, I tried to do something like 
```
@if ($action->reportback)
  checked="checked"
@elseif (old('reportback')
  checked="checked"
@endif
```
but within an `@if` block, I don't have access to the `$action`'s reportback for some reason 🤔

chloealee: @katiecrane finally got this to work as expected! 
- On initial load, if the value is true for the action, the checkbox will be checked.
- If there are form validations, then the old values that the user wanted to uncheck/check are reflected, instead of what the database values.

app/Http/Controllers/Legacy/Web/ActionsController.php
chloealee: hm interesting! I agree this is a little weird to hardcode but not sure where would be a better way to access this list?

resources/views/actions/edit.blade.php
katiecrane: @chloealee YAYAYYYYY GREAT FIDDLING!!! 🎻

app/Http/Controllers/Legacy/Web/ActionsController.php
katiecrane: @chloealee yup I wasn't sure either, I was just thinking outloud!

resources/assets/components/CampaignIdSingle/index.js
katiecrane: strong emoji game all around!

app/Http/Controllers/Legacy/Web/ActionsController.php
aaronschachter: Could it make sense to add a `PostType` class similar to the [`ImportType` in Chompy](https://github.com/DoSomething/chompy/blob/master/app/Import.php#L33), or [`PasswordResetType` in Northstar](https://github.com/DoSomething/northstar/blob/509dcee48431b8283d9b436283d5a14a45139fec/app/PasswordResetType.php#L26)?

chloealee: ooo that is really cool @aaronschachter ! So would this work by creating a `PostType` class and then pass `$type` to the view similar to how you did this [here](https://github.com/DoSomething/chompy/pull/59/files#diff-e489ef4664b1c24151b9b01896103403R77)? Where does this `$type` variable come from?

resources/views/actions/edit.blade.php
```diff
@@ -70,22 +70,54 @@
...
-                          <input type="checkbox" id=reportback name=reportback value=1 {{ old('reportback') || $action->reportback ? 'checked="checked"' : '' }}>
+                          <input type="checkbox" id=reportback name=reportback value=1
+                            @if ($errors->any()))
+                                @if (old('reportback'))
+                                    checked="checked"
+                                @endif
+                            @elseif ($action->reportback)
+                                checked="checked"
+                            @endif
+                          >
```
katiecrane: yay! so glad this is working! would it make sense to pull this logic into a helper? can we really do that with blade files? 🤔

chloealee: I was looking online and [this could be a solution](https://stackoverflow.com/questions/47879154/laravel-checkbox-state-after-failed-validation)? Would this make sense to make into a separate PR since this one is a little hefty?

resources/assets/components/Action/index.js
```diff
@@ -84,6 +85,23 @@ class Action extends React.Component {
             </li>
           </ul>
         </div>
+        {new Date(campaign.start_date.replace(/-/g, '/')) > Date.now() ? (
```
DFurnes: Let's parse this safely like we did in #857, instead of relying on this regex replacement hack.

resources/views/actions/edit.blade.php
katiecrane: that seems like it could work! totally up to you!

resources/assets/components/Action/index.js
chloealee: oh great catch! I actually just deleted the regex since we don't even need that here!

DFurnes: Hmm, how come we can use `new Date(campaign.start_date)` directly here but not elsewhere? Are you sure this works properly across browsers?

DFurnes: Unrelated to this pull request, but noticed a typo that made it through [in #857](
https://github.com/DoSomething/rogue/pull/857/files#r267096683) when clicking through to compare with how we solved this there.

chloealee: > Hmm, how come we can use new Date(campaign.start_date) directly here but not elsewhere? Are you sure this works properly across browsers?

Shoot you're totally right, forgot to check in Safari. Updated!

chloealee: > Unrelated to this pull request, but noticed a typo that made it through in #857 when clicking through to compare with how we solved this there.

That is updated [here](https://github.com/DoSomething/rogue/pull/858/files#diff-0e7fbc8575f3965559b364378e6d6e54R144)!

app/Http/Controllers/Legacy/Web/ActionsController.php
aaronschachter: That `$type` variable is [set in the same file](https://github.com/DoSomething/chompy/blob/ae50b2f567e43458f0a35d46a4579260a22656da/app/Http/Controllers/ImportController.php#L60), saving the `type` query string parameter... but that's just specific to the `ImportController`. The intent for the `ImportType` and `PasswordResetType` classes I linked to is to DRY hardcoding strings like `rock-the-vote` all over the codebase, but instead calling `ImportType::rockTheVote()`. The `all()` function returns all valid import types. 

So for a PostType, we could define different readable functions like `voterRegistration()`, `text()`, `photo()`, `phoneCall()`, `socialShare()`, that return the string values we save in the database.. and an `all` function should return all of these string values, so this edit function could be refactored as:
```
    public function edit($campaignId, $actionId)
    {
        return view('actions.edit')->with([
            'campaignId' => $campaignId,
            'action' => Action::find($actionId),
            'postTypes' => PostTypes::all(),
        ]);
 }
```

aaronschachter: Not within scope of this PR, but we could eventually start replacing hardcoded post type strings with the newer `PostType` functions... which could help prevent typos for string values, or make it easy to change these values if we ever needed to

e.g.  [`getStatusRules`](https://github.com/DoSomething/rogue/blob/master/app/Http/Requests/PostRequest.php#L69) would look like:
```
    private function getStatusRules($type)
    {
        switch ($type) {
            PostType::photo():
                $rule = 'in:pending,accepted,rejected';
                break;
            case PostType::voterRegistration():
                $rule = 'in:pending,register-form,register-OVR,confirmed,ineligible,uncertain';
                break;
            case PostType::phoneCall():
                $rule = 'in:accepted,incomplete';
                break;
            default:
                $rule = 'in:pending,accepted,rejected';
        }
        return $rule;
```

Related - we could also add a helper for PostStatus

chloealee: Ah super smart - thanks for explaining more, @aaronschachter ! Updated this :)

resources/views/actions/edit.blade.php
chloealee: @katiecrane I think that I will pull that into another PR since this one is getting a little beefy! Made a card for that [here](https://www.pivotaltracker.com/n/projects/2019429/stories/164772481)!

PR #845, opened 2019-03-05
increases text maximum character limit to 500 in Post Request

chloealee:
#### What's this PR do?
Increases text maximum character limit to 500 in Post Request.

#### How should this be reviewed?
Can you submit a text reportback with up to 500 characters in the `text` response?

#### Relevant tickets
Finishes up work in #840 

#### How to deploy
https://github.com/DoSomething/rogue/blob/master/docs/development/deployments.md

PR #860, opened 2019-03-18
Update text validation length to be consistent.

DFurnes:
#### What's this PR do?
This pull request just includes a quick fix to make sure that the `PUT v3/posts/:id` route validates text length consistently with the rules we use for `POST v3/posts`. References #845.

I updated the behavior in 0b18958 & the associated test in fb499c9. 🚥

I also cleaned this test up to use Laravel's `assertJsonValidationErrors` assertion, so we don't end up with a fatal error if the validation error doesn't fire & avoid having to copy-paste validation messages into our tests. I made similar touchups to two other nearby tests in 5c91d18.

I noticed this while in this file earlier this morning for #859.

#### Relevant tickets
N/A

#### How to deploy

### DIFF
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -1268,14 +1268,10 @@ class PostTest extends TestCase
 
         $response = $this->withAdminAccessToken()->patchJson('api/v3/posts/' . $post->id, [
             'quantity' => 'this is words not a number!',
-            'text' => 'This must be longer than 140 characters to break the validation rules so here I will create a caption that is longer than 140 characters to test.',
+            'text' => 'a' . str_repeat('h', 512), // ahhh...hhhhh!
         ]);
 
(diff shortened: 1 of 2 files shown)


tests/Http/PostTest.php
```diff
@@ -1268,14 +1267,10 @@ public function testValidationUpdatingAPost()
 
         $response = $this->withAdminAccessToken()->patchJson('api/v3/posts/' . $post->id, [
             'quantity' => 'this is words not a number!',
-            'text' => 'This must be longer than 140 characters to break the validation rules so here I will create a caption that is longer than 140 characters to test.',
+            'text' => 'a' . str_repeat('h', 512), // ahhh...hhhhh!
```
katiecrane: Maybe my favorite line of code you've ever written

PR #861, opened 2019-03-20
Fixes Date Bug for Campaign Ids - incorrect dates were being displayed

chloealee:
#### What's this PR do?
Fixes Date Bug for Campaign Ids - incorrect dates were being displayed. This updates using the correct formatting to make sure the date in the DB is the same that is being displayed. 

👀 
Documentation for the formatting [here](https://date-fns.org/v1.30.1/docs/format)!
#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/164637404) Pivotal Card.

#### How to deploy

### DIFF
--- a/resources/assets/components/Action/index.js
+++ b/resources/assets/components/Action/index.js
@@ -86,7 +86,8 @@ class Action extends React.Component {
             </li>
           </ul>
         </div>
-        {new Date(format(parse(campaign.start_date), 'M/d/YY')) > Date.now() ? (
+        {new Date(format(parse(campaign.start_date), 'MM/D/YYYY')) >
+        Date.now() ? (
           <div className="container__row">
             <a
               className="button -secondary"
--- a/resources/assets/components/Action/index.js
+++ b/resources/assets/components/Action/index.js
@@ -86,7 +86,7 @@ class Action extends React.Component {
             </li>
           </ul>
         </div>
-        {new Date(parse(campaign.start_date)) > Date.now() ? (
+        {parse(campaign.start_date) > Date.now() ? (
           <div className="container__row">
             <a
               className="button -secondary"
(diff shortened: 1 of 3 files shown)


resources/assets/components/Action/index.js
```diff
@@ -86,7 +86,8 @@ class Action extends React.Component {
             </li>
           </ul>
         </div>
-        {new Date(format(parse(campaign.start_date), 'M/d/YY')) > Date.now() ? (
+        {new Date(format(parse(campaign.start_date), 'MM/D/YYYY')) >
```
DFurnes: This is parsing a string into a Date, formatting into a string, and then once again into a Date. Instead:

```js
parse(campaign.start_date) > Date.now()
```

chloealee: oh right - updated!

resources/assets/components/Action/index.js
```diff
@@ -86,7 +86,7 @@ class Action extends React.Component {
             </li>
           </ul>
         </div>
-        {new Date(format(parse(campaign.start_date), 'M/d/YY')) > Date.now() ? (
+        {new Date(parse(campaign.start_date)) > Date.now() ? (
```
DFurnes: @chloealee Can you try the snippet I'd posted in [my comment](
https://github.com/DoSomething/rogue/pull/861#discussion_r267789300)? The [`parse`](https://date-fns.org/v1.30.1/docs/parse) function returns a `Date`, so there shouldn't be a reason to re-instantiate it by passing the result into the `Date` constructor:

```js
parse(campaign.start_date) > Date.now()
```

chloealee: @DFurnes oh sorry I wasn't looking close enough! Updated that!

PR #862, opened 2019-03-22
Adds an index to the location column on posts table.

weerd:
#### What's this PR do?
This PR adds an index to the `location` column on the `posts` table.

👀 

This update is needed for work related to allowing the post gallery to be filtered by the location.

#### Relevant tickets
Refs [PivotalTracker ID #164737236](https://www.pivotaltracker.com/story/show/164737236)

#### How to deploy

### DIFF
--- a/database/migrations/2019_03_08_204743_adds_callpower_campaign_id_unique_index_in_actions_table.php
+++ b/database/migrations/2019_03_08_204743_adds_callpower_campaign_id_unique_index_in_actions_table.php
@@ -26,7 +26,7 @@ class AddsCallpowerCampaignIdUniqueIndexInActionsTable extends Migration
     public function down()
     {
         Schema::table('actions', function (Blueprint $table) {
-            $table->dropUnique('callpower_campaign_id');
+            $table->dropUnique(['callpower_campaign_id']);
         });
     }
 }
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -54,7 +54,7 @@ class Post extends Model
      *
      * @var array
      */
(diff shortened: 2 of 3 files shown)


weerd:
One quick note, I basically referenced this [migration](https://github.com/DoSomething/rogue/blob/master/database/migrations/2017_03_09_192255_add_signup_id_index_to_events_table.php#L28) as an example, but I didn't pass the `dropIndex()` as an array, but as a string. According to the method signature/docblock for that method, it allows a string or array, and since I'm just passing a single ID I figured string would be fine.

DFurnes:
Oh hmm, can you double-check that works by doing a quick `php artisan migrate:rollback` on your local? I believe when you pass a string Laravel [interprets it as the index name](https://laravel.com/docs/5.5/migrations#dropping-indexes), not the column name.

weerd:
Screen Shot 2019-03-22 at 5 06 26 PM

Ok! Glad I asked. Switching to using an array. That's a tricky item! 😬

database/migrations/2019_03_08_204743_adds_callpower_campaign_id_unique_index_in_actions_table.php
```diff
@@ -26,7 +26,7 @@ public function up()
     public function down()
     {
         Schema::table('actions', function (Blueprint $table) {
-            $table->dropUnique('callpower_campaign_id');
+            $table->dropUnique(['callpower_campaign_id']);
```
weerd: @chloealee I did the same for this migration for `callpower_campaign_id` since it was giving a similar error when rolling back! All is good now for this one too!

chloealee: ah thank you, @weerd !

PR #864, opened 2019-04-01
force deletes action instead of soft deletes

chloealee:
#### What's this PR do?
Force deletes action instead of soft deletes.

Make sure that the action is no longer in the database when deleted. 

@ngjo wanted to create CallPower actions in the database and there was a mistake with one of them. We deleted it but the functionality only soft deletes actions when deleted via the web route. When she tried to create a new action with that soft deleted action's same `callpower_camapign_id, she wasn't able to because of the unique validation rule. 

We only allow users to delete actions if the campaign has not yet started. Because of this, the main functionality of deleted actions is if it really shouldn't be there. Thus, I think it actually makes sense to force delete these actions instead of soft deleting. 

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Legacy/Web/ActionsController.php
+++ b/app/Http/Controllers/Legacy/Web/ActionsController.php
@@ -133,11 +133,7 @@ class ActionsController extends Controller
      */
     public function destroy(Action $action)
     {
-        $trashed = $action->delete();
-
-        if (! $trashed) {
-            return response()->json(['code' => 500, 'message' => 'There was an error deleting the action']);
-        }
+        $action->forceDelete();
 
         // Log that an action was deleted.
         info('action_deleted', ['id' => $action->id]);


app/Http/Controllers/Legacy/Web/ActionsController.php
```diff
@@ -133,11 +133,7 @@ public function update(Request $request, Action $action)
      */
     public function destroy(Action $action)
     {
-        $trashed = $action->delete();
-
-        if (! $trashed) {
-            return response()->json(['code' => 500, 'message' => 'There was an error deleting the action']);
-        }
+        $action->forceDelete();
```
DFurnes: Not a big deal, but it might make sense to remove [the `SoftDeletes` trait](https://github.com/DoSomething/rogue/blob/b3219d7047dc69b389cdddb794c938ee0e7fe6a3/app/Models/Action.php#L10) and the `deleted_at` column from this model if we're not relying on soft deletion anymore.

chloealee: @DFurnes ah thanks for catching that! Made a PR for that [here](https://github.com/DoSomething/rogue/pull/865).

PR #867, opened 2019-04-02
upgrades just-extend and runs composer update

chloealee:
#### What's this PR do?
upgrades `just-extend` and runs `composer update` per [this](https://github.com/dosomething/rogue/network/alert/package-lock.json/just-extend/open) Github security alert. 

👀 

#### How to deploy

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -60,16 +60,16 @@
         },
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.90.1",
+            "version": "3.91.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/aws/aws-sdk-php.git",
-                "reference": "3262a84de9428766646223963c944b2be2cd09c0"
+                "reference": "808a8cde13d5ba458d6365828385fead955c933e"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/aws/aws-sdk-php/zipball/3262a84de9428766646223963c944b2be2cd09c0",
-                "reference": "3262a84de9428766646223963c944b2be2cd09c0",
+                "url": "https://api.github.com/repos/aws/aws-sdk-php/zipball/808a8cde13d5ba458d6365828385fead955c933e",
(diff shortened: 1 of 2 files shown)


chloealee:
I updated `just-extend` to be `"just-extend": ">=4.0.0",` but should that be `"just-extend": "^4.0.0",` instead?

DFurnes:
> I updated `just-extend` to be `"just-extend": ">=4.0.0",` but should that be `"just-extend": "^4.0.0",` instead?

Where did you change this?

chloealee:
In the `package-lock.json` file per the suggestion in the alert!

DFurnes:
Ah! It's probably not a wise idea to manually edit our `package-lock.json`, since those changes will be reverted next time someone installs a package using the CLI (and it's not a super friendly file to work with).

Since this is a sub-dependency, and [it's been updated upstream](https://github.com/sinonjs/nise/blob/242c9504b4a058ef10d849d14147f10ad303f474/package.json#L63), you should be able to just re-generate the lockfile to grab this and any other updated dependencies:

```
rm -rf node_modules package-lock.json
npm install
```

DFurnes:
> I updated `just-extend` to be `"just-extend": ">=4.0.0",` but should that be `"just-extend": "^4.0.0",` instead?

For your original question, yes! We'd definitely prefer `^4.0.0`, since `>=4.0.0` would pull in _anything_ that's a greater version number, including potentially breaking changes in `5.x.x` or `6.x.x`, etc.

chloealee:
> Ah! It's probably not a wise idea to manually edit our package-lock.json, since those changes will be reverted next time someone installs a package using the CLI (and it's not a super friendly file to work with).
Since this is a sub-dependency, and it's been updated upstream, you should be able to just re-generate the lockfile to grab this and any other updated dependencies:
rm -rf node_modules package-lock.json
npm install

Super helpful to know for the future, thank you @DFurnes ! Updated with what you suggested. I ran it outside of my vagrant - just want to make sure that is correct!

PR #868, opened 2019-04-03
Moves review gate to Posts Controller to allow script to bulk approve

chloealee:
#### What's this PR do?
Moves review gate to Posts Controller to allow script to bulk approve. Also, per @DFurnes 's comment below, this logic should actually live in the Posts Controller anyway since the repository should really only be responsible for storing or retrieving data.

👀 

#### Relevant tickets
References change made [here](https://github.com/DoSomething/rogue/pull/859/commits/7533ec2513919d7ecebb5670397a5c66cb1206c6). 

Needed for [this](https://www.pivotaltracker.com/n/projects/2019429/stories/165090581) Pivotal Card.
#### How to deploy

### DIFF
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -8,6 +8,7 @@ use Illuminate\Http\Request;
 use Rogue\Managers\PostManager;
 use Rogue\Managers\SignupManager;
 use Rogue\Http\Requests\PostRequest;
+use Illuminate\Support\Facades\Gate;
 use Rogue\Http\Transformers\PostTransformer;
 use Rogue\Http\Controllers\Traits\FiltersRequests;
 
@@ -150,6 +151,11 @@ class PostsController extends ApiController
     {
         // Only allow an admin/staff or the user who owns the post to update.
         $this->authorize('update', $post);
+
+        // But don't allow user's to review their own posts.
+        if (! Gate::allows('review', $post)) {
+            unset($request['status']);
+        }
 
         $this->posts->update($post, $request->validated());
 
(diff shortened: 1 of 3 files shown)


chloealee:
hmm the `testNonStaffUpdatePost` test fails with this new line since phpunit is run from the console. I'm not sure how to get around this with the new code. Looking for potentially another way to determine that this is run in a job instead of using `app()->runningInConsole()` to check this?

chloealee:
@DFurnes I think it is important to keep this unit test here but having a hard time deciding on a different way then using `app()->runningInConsole()` here! Two other ideas that aren't great but leaning towards the second option as the way to go:

1. Have the Bulk Review Script hit an API endpoint with machine token instead of using the Post Repositories [update](https://github.com/DoSomething/rogue/blob/master/app/Console/Commands/BulkReviewPosts.php#L78) method. 
2. Creating a new method in the Post Repository (maybe named `bulkReviewPosts`) that skips over the [gate check logic](https://github.com/DoSomething/rogue/blob/master/app/Repositories/PostRepository.php#L174-L176) so it can update posts? 

I like the second option better because it seems a little weird to have a job run inside of Rogue to hit it's own endpoint? Thoughts on this?

chloealee:
OR - is it ok to mark this test as incomplete for now, just while I run the bulk approval job? Campaign leads need to have these reviewed by EOD tomorrow!

DFurnes:
@chloealee Riffing off your second suggestion, can we move the gate logic out into the controller, since it's only relevant to web requests? The repository should really only be responsible for storing or retrieving data.

chloealee:
@DFurnes Love that suggestion - so smart! Updated :)

PR #869, opened 2019-04-10
Clean up post component & add campaign/action links.

DFurnes:
#### What's this PR do?
This pull request adds links to the campaign & action on each post in the "inbox". I've also separated signup & post properties for readability, and linked things that make sense (so you can easily hop over to a particular campaign, signup, or user).

Finally, I fixed the "broken image" fallback to actually show us Jen's face. Phew. 🐛 

##### Before:
Screen Shot 2019-04-10 at 12 41 13 PM

##### After:
Screen Shot 2019-04-10 at 12 44 13 PM

I dunno, look at it! 👀 I also updated the test & snapshot for this component.

We should really do a cleanup pass on all these components & re-enable linting at some point...

#### Relevant tickets
[#165188977](https://www.pivotaltracker.com/story/show/165188977)

#### How to deploy


DFurnes:
Another thought: as much as I truly love it, it _might_ be a good idea to find a more "neutral" icon for posts that don't have images, since we show 🐛 for any "sibling" text posts in the review queue right now.

app/Http/Transformers/PostTransformer.php
```diff
@@ -37,6 +37,7 @@ public function transform(Post $post)
             'type' => $post->type,
             'action' => $post->getActionName(),
             'action_id' => $post->action_id,
+            'campaign_id' => $post->campaign_id,
```
chloealee: is this removed since we have the `action_id` returned?

DFurnes: Sorry, could you clarify what you mean by "removed" here?

DFurnes: I added the `campaign_id` field to the post transformer because we store it on the post in the database, but do not currently expose it here (and it was handy to be able to easily snag for displaying in this component).

chloealee: oops my goof! I thought this was added instead of removed 🤦‍♀️

DFurnes:
Following up on my comment above, I updated the `PostTile` component to have a better fallback when displaying a text post. This should make the inbox look a little less broken on campaigns with text posts & hopefully reduce confusion from campaign leads:

Screen Shot 2019-04-10 at 2 14 29 PM

PR #872, opened 2019-04-24
Update dependencies & allow admins to edit actions. 

DFurnes:
#### What's this PR do?
This pull request updates dependencies (notably, Webpack 4 for less buggy re-builds) and adds the ability for admins to edit in-progress actions (which we'll want for [#165160940](https://www.pivotaltracker.com/story/show/165160940)).

As part of that, I updated this app's `window.AUTH` to be an object (filled in by a [view composer](https://laravel.com/docs/5.6/views#view-composers)) so that we'd have access to the user's role. This follows the same approach we've used in Phoenix.

I think everything's good, but can't hurt to click around on QA a bit before we deploy this!

nah.

#### Relevant tickets

#### How to deploy

### DIFF
--- a/resources/assets/components/Action/index.js
+++ b/resources/assets/components/Action/index.js
@@ -12,7 +12,7 @@ class Action extends React.Component {
   render() {
     const action = this.props.action;
     const campaign = this.props.campaign;
-    const campaignInProgress = parse(campaign.start_date) > Date.now();
+    const campaignInProgress = Date.now() > parse(campaign.start_date);
     const isAdmin = window.AUTH.role == 'admin';
 
     return (
--- a/resources/assets/components/Action/index.js
+++ b/resources/assets/components/Action/index.js
@@ -12,6 +12,8 @@ class Action extends React.Component {
   render() {
     const action = this.props.action;
     const campaign = this.props.campaign;
+    const campaignInProgress = parse(campaign.start_date) > Date.now();
+    const isAdmin = window.AUTH.role == 'admin';
 
     return (
       <div className="container__action">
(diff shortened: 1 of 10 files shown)


resources/assets/components/Action/index.js
```diff
@@ -12,6 +12,8 @@ class Action extends React.Component {
   render() {
     const action = this.props.action;
     const campaign = this.props.campaign;
+    const campaignInProgress = parse(campaign.start_date) > Date.now();
```
katiecrane: I've been confused about dates recently, but wouldn't a campaign be "in progress" if it's start date is `<` now? (meaning it has already passed?)

DFurnes: Oh yeah this is very wrong. 🙃

ngjo: This probably doesn't have anything to do with what it's like in Rogue, but I know that Gambit relies on this date and I think there is something funny with how they "read" the date too and treat it as closed?

DFurnes: Updated this in 47b1a2d! Thanks for catching @katiecrane!

DFurnes: Maybe we should just add a computed "open" boolean on the campaign response that does this for us so we don't have to risk silly mistakes like this in multiple places. 😂

DFurnes: I tested this a little more thoroughly this time and can confirm that it works as expected now: admins can edit actions for open & closed campaigns, but staffers only see the "edit" and "delete" buttons for campaigns that haven't started yet.

katiecrane: Oooo I love the computed boolean idea!!

PR #874, opened 2019-05-02
Fix two flaky tests.

DFurnes:
#### What's this PR do?
This pull request fixes occasional test failures in `testPostsIndexWithAnonymousPosts` and `testPostsIndexAsOwnerHiddenPosts`, both caused when the user IDs generated by Faker would unexpectedly collide. This will make our CI tests for this app more trustworthy.

✅ 👀

If you ever want to "battle test" a particular test, you can use the `--repeat` argument! For example, this runs the `testPostsIndexAsOwnerHiddenPosts` test 100 times to catch those flaky failures:

```
$ phpunit tests/Http/PostTest.php --filter PostsIndexAsOwnerHiddenPosts --repeat 100

.............F...............F...................F.............  63 / 100 ( 63%)
.....................................                           100 / 100 (100%)

Time: 46.03 seconds, Memory: 32.00MB

There were 3 failures:
...
```

#### Relevant tickets
N/A

#### How to deploy

### DIFF
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -1026,7 +1026,7 @@ class PostTest extends TestCase
     public function testPostsIndexAsOwnerHiddenPosts()
     {
         // Owners should only see accepted posts and their own hidden posts.
-        $ownerId = $this->faker->northstar_id;
+        $ownerId = $this->faker->unique()->northstar_id;
 
         // Create posts and associate to this $ownerId.
         $posts = factory(Post::class, 'accepted', 2)->create(['northstar_id' => $ownerId]);
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -803,11 +803,15 @@ class PostTest extends TestCase
     {
         // Create an accepted post.
         $this->mockTime('8/3/2017 14:00:00');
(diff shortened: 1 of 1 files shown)


katiecrane:
> If you ever want to "battle test" a particular test, you can use the --repeat argument! For example, this runs the testPostsIndexAsOwnerHiddenPosts test 100 times to catch those flaky failures:
THIS IS GOING TO COME IN SO HANDY!!!

tests/Http/PostTest.php
```diff
@@ -1022,7 +1026,7 @@ public function testPostsIndexAsOwner()
     public function testPostsIndexAsOwnerHiddenPosts()
     {
         // Owners should only see accepted posts and their own hidden posts.
-        $ownerId = $this->faker->northstar_id;
+        $ownerId = $this->faker->unique()->northstar_id;
```
katiecrane: I would have thought that since this one came first, the other one would never be the same, but the evidence shows otherwise!

DFurnes: Yeah, totes! I don't think it starts tracking uniqueness until you use the `unique()` helper though, so when it was choosing a "unique" value later on it didn't take this into account.

PR #875, opened 2019-05-02
Add new list of causes for campaigns.

DFurnes:
#### What's this PR do?
This pull request adds the [new campaign causes](https://docs.google.com/spreadsheets/d/1NYLfIjHGpEnd8m0C-xVJvHciCrMuqWYD6mTFisW263U/edit#gid=0) to Rogue's campaign edit form, and includes the ability to select multiple causes for a single campaign. Some other highlights and notes:

📜 Removes the `rogue:ashescampaignsimport` because we're awfully unlikely to use it again, and it would've required refactoring for the new Campaign `cause` schema. 7c66998

🔒 Simplifies the `role:staff,admin` middleware to apply to all "Campaign ID" routes, both for simplicity and because we'd (accidentally?) missed the `index` route in this list, and there's no reason that a "normal" user should need to visit Rogue's Campaign ID index. 4f191fc

☑️ Enables Forge's "pretty" [option field](http://forge.dosomething.org/#forms-option-field) appearance by adding the Modernizr classes to the page markup. We don't support browsers that'd fail any Modernizr tests on Rogue, so might as well make things look pretty! 7cbdd6f

🔀 Shares common validation rules between the "store" and "update" methods in the Campaign ID controller. This reduces some unnecessary duplication, and fixes an issue where you could "unset" a required field when updating a campaign because of inconsistent rules. 5955801

🐳 The big ones! Finally, 6a476b3 and be6808d add the [new list of cause spaces](https://docs.google.com/spreadsheets/d/1NYLfIjHGpEnd8m0C-xVJvHciCrMuqWYD6mTFisW263U/edit#gid=0), the ability to select multiple causes for a campaign via checkboxes & transformations between human and machine-friendly representations.

Screen Shot 2019-05-02 at 1 27 36 PM

I'd recommend reviewing this one commit-by-commit. 

We've got a meeting on the books to run through existing campaigns one-by-one and set their causes based on the new taxonomy, which I think is easier than trying to automate this somehow (considering many of the new causes don't cleanly map to old ones).

I've tested these changes with campaigns that have the "old" causes stored in the database and everything works fine – they'll just appear as if they don't have any cause attached.

#### Relevant tickets

#### How to deploy


app/Http/Controllers/Legacy/Web/CampaignIdsController.php
```diff
@@ -60,15 +55,11 @@ public function create()
...
     {
-        $this->validate($request, [
-            'internal_title' => 'required|string|unique:campaigns',
-            'cause' => 'required|string',
-            'impact_doc' => 'required|url',
-            'start_date' => 'required|date',
-            'end_date' => 'nullable|date|after:start_date',
-        ]);
+        $values = $this->validate($request, array_merge_recursive($this->rules, [
+            'internal_title' => [Rule::unique('campaigns')],
```
katiecrane: Why can't this be included in `$this->rules`?

katiecrane: Oh I see now! It is different per method

app/Http/Controllers/Legacy/Web/CampaignIdsController.php
```diff
@@ -16,15 +16,26 @@ class CampaignIdsController extends Controller
     public function __construct()
     {
         $this->middleware('auth');
-        $this->middleware('role:admin,staff', ['only' => ['create', 'store', 'edit', 'update', 'destroy']]);
+        $this->middleware('role:admin,staff');
+
+        $this->rules = [
+            'internal_title' => ['required', 'string'],
+            'cause' => ['required', 'array', 'between:1,5'],
```
katiecrane: Does this mean we limit to 5 causes per campaign?

katiecrane: and now I see in the screengrab that yes, it does mean that!

PR #876, opened 2019-05-03
Update campaign overview for our brave new world of multi-cause campaigns.

DFurnes:
#### What's this PR do?
This pull request updates the Campaign Overview page to group & sort campaigns by status & pending post count, rather than by cause (since in a multi-cause world, that's a confusing thing to do). I've also added a little search box so users can filter for a campaign based on ID, name, or cause.

Screen Shot 2019-05-03 at 1 50 05 PM

👀

↕️

#### Relevant tickets

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Legacy/Web/CampaignsController.php
+++ b/app/Http/Controllers/Legacy/Web/CampaignsController.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Http\Controllers\Legacy\Web;
 
+use Rogue\Models\Campaign;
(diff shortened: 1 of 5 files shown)


DFurnes:
Snuck in one more change in 86391c8 to add more descriptive headings to the two categories, since I could see folks getting confused why inactive campaigns don't show up under the "pending review" category otherwise. Also here's an example filtering by cause!

Screen Shot 2019-05-03 at 2 06 27 PM

PR #882, opened 2019-05-13
Handle Rendering Campaigns Overview with empty/null Categories

mendelB:
#### What's this PR do?
This PR updates the `CampaignOverview` component to handle the scenario where no campaigns are available per category, rendering a simple message instead of attempting to still filter the (`null`) campaigns value.

https://github.com/DoSomething/rogue/commit/0841bf3ecf4f2be7ee83e35390d5565454dcb6df then tentatively refactors some of the logic into a method. This felt DRYer to me, but if this seems too egregious I can just roll this back! (Also, should such a thing be in a method as opposed to its own functional component?)

And finally, https://github.com/DoSomething/rogue/pull/882/commits/e05bcdb755c15ea8a26800338ccedfcbff2ec6ab adds a step to our installation instructions to cover seeding the DB!

Delete all the campaigns from Rogue or review all pending posts and see what happens 😁 

While setting up Rogue, I attempted to visit the campaigns page before having seeded the database, and the page errored out due to an [attempted filtering](https://github.com/DoSomething/rogue/blob/2dc842f7b1a888b19884ab9c4b2d71e7f884f121/resources/assets/components/CampaignOverview/index.js#L31) of the [empty campaign data](https://github.com/DoSomething/rogue/blob/2dc842f7b1a888b19884ab9c4b2d71e7f884f121/resources/assets/components/CampaignOverview/index.js#L56-L57)

#### Relevant tickets
https://dosomething.slack.com/archives/C2BLZGBPH/p1557773627015300

#### How to deploy

### DIFF
--- a/docs/development/installation.md
+++ b/docs/development/installation.md
@@ -11,4 +11,7 @@ $ php artisan rogue:setup
 
 # And finally, build the frontend assets:
 $ npm run build
+
+# Also, feel free to seed the database with dummy data:
+$ php artisan db:seed
 ```
--- a/resources/assets/components/CampaignOverview/index.js
+++ b/resources/assets/components/CampaignOverview/index.js
@@ -53,8 +53,9 @@ class CampaignOverview extends React.Component {
    * @return {React.Element}
    */
   render() {
-    const pending = this.filterCampaigns(this.props.pending);
-    const etc = this.filterCampaigns(this.props.etc);
+    const { pending, etc } = this.props;
+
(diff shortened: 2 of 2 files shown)


mendelB:
@katiecrane I added a quick patch to the docblock param definitions in https://github.com/DoSomething/rogue/pull/882/commits/6e33a1b760dac53b98333427e225a724f1e3ffcc which should be completely inconsequential - but just hoping to get another 👍 before I merge. Thanks!

PR #884, opened 2019-05-17
Tag Information

mendelB:
#### What's this PR do?
This PR adds some helper text and FAQ links for Tags to the Campaign page and campaign Inbox.

A couple of updates accompanying the copy changes:
- https://github.com/DoSomething/rogue/pull/884/commits/d4a60ea5f515d091cd449a0d049bbfb1cd77c7fb rendering `props.children` from the `MultiValueFilter` to allow custom heading content. (This ended up deviating from the initial approach which didn't end up panning out, unfortunately)
- https://github.com/DoSomething/rogue/pull/884/commits/0bec859d7893001c37a52d03fa3bd833a7f1f975, https://github.com/DoSomething/rogue/pull/884/commits/1201dfdc70fea63a3e22d0c79bd287b57e1e8c80 hiding the tags completely from the campaign inbox posts until the post is reviewed (marked 'accepted' or 'rejected')
- https://github.com/DoSomething/rogue/pull/884/commits/625831bcf13515da0b0f872a9c1c51ccfde92826 add an anchor link to the tag info in the FAQ page to enable linking directly there.

Are the tags hidden from the campaign inbox until you review the post?

~Was hoping to get eyes on this controversial finessing of existing pattern to agnostically allow a subheader but in this case, it needs to be markup. Is this too crazy? Should we just add the header content as a static part of this component since it's only used in this case so far?~ I was getting some errors related to the output of these `subheading` nodes, and reevaluated, and after a bit of googling around, it seems like the best option here is to just utilize `props.children` which is admittedly more straightforward!

#### Relevant tickets
[Pivotal ID #165938663](https://www.pivotaltracker.com/story/show/165938663)
[Mocks](https://projects.invisionapp.com/share/V9RZLC9F8GQ#/screens/363026075)

#### How to deploy

📸 
Campaign Page:
image

Campaign Inbox:
image

image


chloealee:
omg I love this @mendelB !!!

mendelB:
☺️  so glad to hear that! I wasn't sure if this was like toooo crazy

katiecrane:
I think this is looking awesome so far!

resources/assets/components/Tags/index.js
```diff
@@ -76,11 +75,14 @@ class Tags extends React.Component {
...
     };
 
-    const showTags = this.props.disabled ? 'disabled' : 'enabled';
-
     return (
       <div>
-        <h4 className={showTags}>Tags</h4>
+        <h4>
+          Tags (see{' '}
+          <a href="/faq#tags" target="_blank">
```
mendelB: I opted to open this one in a new tab since otherwise, clicking the link would cause you to lose the post you'd just reviewed before being able to add tags.

resources/assets/components/CampaignSingle/index.js
```diff
@@ -181,7 +181,12 @@ class CampaignSingle extends React.Component {
         </div>
         <FilterBar onSubmit={this.filterPosts}>
           <DropdownFilter options={statusFilters} header={'Post Status'} />
-          <MultiValueFilter options={tagFilters} header={'Tags'} />
+          <MultiValueFilter options={tagFilters}>
+            <b>Tags</b> Not sure which tag to filter by?{' '}
+            <a className="font-normal" href="/faq#tags">
+              See tag definitions.
+            </a>
```
katiecrane: Is this text inside the `MultiValueFilter` what ends up as `this.props.children` in `MultiValueFilter/index.js`?

DFurnes: I think this is a perfect way to handle this! ⚛️

PR #888, opened 2019-06-06
Add GraphQL Service

mendelB:
#### What's this PR do?
This PR adds a new GraphQL service, which wraps the [Softonic GraphQL Client package](https://github.com/softonic/graphql-client)

_Does this make sense at all_? It's a pretty bare-bones service, but this affords us the ability to `use` and `query` with this service across the app with one central instantiation procedure with the proper config variables, etc.

I didn't love needing to add our own `query` method, but I wasn't sure how else to expose that functionality. 

_Does this need any tests_? I couldn't think of any test cases myself.

#### Relevant tickets
[Pivotal ID#166360211](https://www.pivotaltracker.com/story/show/166360211)

#### How to deploy


app/Services/GraphQL.php
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Run a GraphQL query using the client.
+     *
+     * @param  $query     String
+     * @param  $variables Array
+     * @return \Softonic\GraphQL\Response
+     */
+    public function query($query, $variables)
+    {
+        return $this->client->query($query, $variables);
```
katiecrane: I know you don't love this, but I for sure think it's worth it to have the config all set up properly in one place. I can't think of a more straightforward way to do it either

katiecrane: Maybe down the line we'll have like `getCampaignById()` that takes an `id`, sets up the `$query` and `$variables` and calls `$this->client->query()` all for us! Something to consider once it starts getting used!

mendelB: Gotcha, and that's a sweet thought -- will def keep that in mind when setting up the queries!

aaronschachter:
I could see us moving this GraphQL client into [Gateway](https://github.com/dosomething/gateway). It'd be handy for sending over a school name to Customer.io per a Northstar user's `school_id` over in https://www.pivotaltracker.com/story/show/169745644.

PR #889, opened 2019-06-25
GraphQL Service Updates (& new CampaignWebsite query)

mendelB:
#### What's this PR do?
This PR Updates the GraphQL service's `query` method to:
- safely catch & log any errors thrown when querying GraphQL via the client.
- parse out the data result fields into a flat Array of said fields (or `null` result).

Also adds a `getCampaignWebsiteByCampaignId` method (As @katiecrane [predicted](https://github.com/DoSomething/rogue/pull/888#discussion_r291398099)) which utilizes this new `query` format to (and you probably didn't see this coming) get a Campaign Website (Contentful campaign) from GraphQL queried by the `campaignId` (`legacyCampaignId` on Contentful) field.

I made some architectural decisions here which I'd love to be validated/torn apart!
- ~`try/catch`ing the client errors - This felt like a good move so that a single query error wouldn't completely break the site, but is that wrong and breaking the site would be the expected behavior anyway? Or is this ideal and we can be catching faulty queries in our monitoring systems instead of having everything explode?~ (https://github.com/DoSomething/rogue/pull/889#discussion_r297692642)
  - ~logging said errors~
- parsing out the data result - ~Though this adds a new `$queryKey` param to the `query` method, this still felt like a win to me to help encapsulate the parsing logic avoid the repetition in any new queries we add. (We know we always want the actual Array of fields).~ (https://github.com/DoSomething/rogue/pull/889#discussion_r297298568)

I strangely got a styleci complaint about an untouched file, which I fixed in https://github.com/DoSomething/rogue/commit/09b8f3cab08c45c2c1f3af0bde2c14cd88a6eb42 

#### Relevant tickets
https://www.pivotaltracker.com/story/show/166360256

#### How to deploy

### DIFF
--- a/app/Console/Commands/UpdatePostsWithActionIds.php
+++ b/app/Console/Commands/UpdatePostsWithActionIds.php
@@ -3,7 +3,6 @@
 namespace Rogue\Console\Commands;
 
 use DB;
-use Rogue\Models\Post;
 use Rogue\Models\Action;
 use Illuminate\Console\Command;
 
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -60,6 +60,6 @@ class GraphQL
             'campaignId' => $campaignId,
         ];
 
-        return $this->query($query, $variables, 'campaignWebsiteByCampaignId');
+        return $this->query($query, $variables);
     }
 }
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -2,7 +2,6 @@
 
 namespace Rogue\Services;
 
-use Illuminate\Support\Facades\Log;
 use Softonic\GraphQL\ClientBuilder;
 
 class GraphQL
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -22,7 +22,7 @@ class GraphQL
      * @param  $query     String
      * @param  $variables Array
      * @param  $queryKey  String
-     * @return Array|null
+     * @return array|null
      */
     public function query($query, $variables, $queryKey)
     {
@@ -44,7 +44,7 @@ class GraphQL
      * Query for a CampaignWebsite by campaignId field.
      *
      * @param  $campaignId String
-     * @return Array|null
+     * @return array|null
      */
     public function getCampaignWebsiteByCampaignId($campaignId)
     {
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -29,7 +29,8 @@ class GraphQL
             $response = $this->client->query($query, $variables);
         } catch (\Exception $exception) {
             Log::error(
-                'GraphQL request failed. Variables: '.json_encode($variables).' Exception: '.$exception->getMessage()
+                'GraphQL request failed. Query: '.$query
+                .' Variables: '.json_encode($variables).' Exception: '.$exception->getMessage()
             );
 
             return null;
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -31,6 +31,10 @@ class GraphQL
     /**
      * Query for a CampaignWebsite by campaignId field.
      *
+     * NOTE: We prefer to bundle queries with their associated usage logic.
+     * We've opted to utilize this helper method as an exception to avoid duplication.
+     * (See thread https://git.io/fjoqd).
+     *
      * @param  $campaignId String
      * @return array
      */
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -20,7 +20,7 @@ class GraphQL
      *
      * @param  $query     String
      * @param  $variables Array
-     * @return array|null
+     * @return array
      */
     public function query($query, $variables)
     {
@@ -34,17 +34,17 @@ class GraphQL
                 'exception' => $exception->getMessage(),
             ]);
 
-            return null;
+            return [];
         }
 
-        return $response ? $response->getData() : null;
+        return $response ? $response->getData() : [];
     }
 
     /**
      * Query for a CampaignWebsite by campaignId field.
      *
      * @param  $campaignId String
-     * @return array|null
+     * @return array
      */
     public function getCampaignWebsiteByCampaignId($campaignId)
     {
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Services;
 
+use Illuminate\Support\Facades\Log;
 use Softonic\GraphQL\ClientBuilder;
 
 class GraphQL
@@ -15,14 +16,28 @@ class GraphQL
     }
 
     /**
-     * Run a GraphQL query using the client.
+     * Run a GraphQL query using the client
+     * and parse the data result into a convenient format.
      *
      * @param  $query     String
      * @param  $variables Array
-     * @return \Softonic\GraphQL\Response
+     * @param  $queryKey  String
+     * @return Array|null
      */
-    public function query($query, $variables)
+    public function query($query, $variables, $queryKey)
     {
-        return $this->client->query($query, $variables);
+        // Use try/catch to avoid any GraphQL related errors breaking the application.
+        try {
+            $response = $this->client->query($query, $variables);
+        } catch (\Exception $exception) {
+            Log::error(
+                'GraphQL request failed. Variables: '.json_encode($variables).' Exception: '.$exception->getMessage()
+            );
+
+            return null;
+        }
+
+        return $response ? array_get($response->getData(), $queryKey) : null;
+    }
     }
 }
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -28,10 +28,11 @@ class GraphQL
         try {
             $response = $this->client->query($query, $variables);
         } catch (\Exception $exception) {
-            Log::error(
-                'GraphQL request failed. Query: '.$query
-                .' Variables: '.json_encode($variables).' Exception: '.$exception->getMessage()
-            );
+            Log::error('GraphQL request failed.', [
+                'query' => $query,
+                'variables' => $variables,
+                'exception' => $exception->getMessage(),
+            ]);
 
             return null;
         }
--- a/app/Services/GraphQL.php
+++ b/app/Services/GraphQL.php
@@ -39,5 +39,27 @@ class GraphQL
 
         return $response ? array_get($response->getData(), $queryKey) : null;
     }
+
+    /**
+     * Query for a CampaignWebsite by campaignId field.
+     *
+     * @param  $campaignId String
+     * @return Array|null
+     */
+    public function getCampaignWebsiteByCampaignId($campaignId)
+    {
+        $query = '
+        query GetCampaignWebsiteByCampaignId($campaignId: String!) {
+          campaignWebsiteByCampaignId(campaignId: $campaignId) {
+            title
+            slug
+          }
+        }';
+
+        $variables = [
+            'campaignId' => $campaignId,
+        ];
+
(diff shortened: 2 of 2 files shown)


app/Services/GraphQL.php
```diff
@@ -15,14 +16,50 @@ public function __construct()
...
-    public function query($query, $variables)
+    public function query($query, $variables, $queryKey)
     {
-        return $this->client->query($query, $variables);
+        // Use try/catch to avoid any GraphQL related errors breaking the application.
+        try {
+            $response = $this->client->query($query, $variables);
+        } catch (\Exception $exception) {
+            Log::error(
+                'GraphQL request failed. Variables: '.json_encode($variables).' Exception: '.$exception->getMessage()
```
mendelB: How's this log? Should we perhaps include the `$query` itself as well?

app/Services/GraphQL.php
```diff
@@ -15,14 +16,50 @@ public function __construct()
...
+
+    /**
+     * Query for a CampaignWebsite by campaignId field.
+     *
+     * @param  $campaignId String
+     * @return array|null
+     */
+    public function getCampaignWebsiteByCampaignId($campaignId)
+    {
+        $query = '
```
mendelB: @DFurnes I know you'd mentioned a while back potentially making use of dedicated .gql files to contain these queries -- do you think that's worth implementing now? I felt like this format wasn't quite as obnoxious as the HEREDOC

app/Console/Commands/UpdatePostsWithActionIds.php
```diff
@@ -3,7 +3,6 @@
 namespace Rogue\Console\Commands;
 
 use DB;
-use Rogue\Models\Post;
```
mendelB: 🤷‍♂ styleci cried about this -- It didn't seem to be used in this file so felt safe to remove!

katiecrane: agreed that this is safe to remove! I wonder why styleci didn't catch it before 🤔

app/Services/GraphQL.php
```diff
@@ -15,14 +16,50 @@ public function __construct()
...
+     * and parse the data result into a convenient format.
      *
      * @param  $query     String
      * @param  $variables Array
-     * @return \Softonic\GraphQL\Response
+     * @param  $queryKey  String
+     * @return array|null
      */
-    public function query($query, $variables)
+    public function query($query, $variables, $queryKey)
```
DFurnes: I think I'd recommend against having a `$queryKey` parameter here, since we might very well make queries that include multiple keys (e.g. querying `posts(…) { … } and `user(id: …) { … }` in one shot in the future.

app/Services/GraphQL.php
DFurnes: Yeah, I think including the query (or if it's parsed anywhere convenient, the query name) makes sense! 👍

app/Services/GraphQL.php
DFurnes: Yeah this looks pretty clean! I think the only thing `.gql` files would really get us is syntax highlighting & editor autocomplete, which probably isn't as necessary here as in other applications!

app/Services/GraphQL.php
katiecrane: @DFurnes do we currently make queries like that anywhere? I'd love to familiarize myself! I would have thought those would have been two separate queries making `$queryKey` okay to use here

DFurnes: One that comes to mind is [the petition block](https://github.com/DoSomething/phoenix-next/blob/abecd18eb47ea9bfddd492efb71de275514cec48/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js#L33-L42), where we check both for any existing submission & the user's first name for the "signature" field. This is definitely something that is possible (and encouraged) in GraphQL!

mendelB: Ah makes sense! Updated the parsing and removed the `$queryKey` param in https://github.com/DoSomething/rogue/pull/889/commits/45c6850df0efb7d0e867c97002072594b2dbfa24

app/Services/GraphQL.php
mendelB: Added the `$query` to the log output in https://github.com/DoSomething/rogue/pull/889/commits/1d5fc99d6ea1cd09e68c41ff59b8133e087dcfa0

app/Services/GraphQL.php
```diff
@@ -15,14 +16,49 @@ public function __construct()
...
      */
     public function query($query, $variables)
     {
-        return $this->client->query($query, $variables);
+        // Use try/catch to avoid any GraphQL related errors breaking the application.
+        try {
+            $response = $this->client->query($query, $variables);
+        } catch (\Exception $exception) {
+            Log::error(
+                'GraphQL request failed. Query: '.$query
```
DFurnes: Oh, I didn't think of this before but let's put these variables into the "meta" block rather than stringifying:

```php
Log::error('GraphQL request failed.', ['query' => $query, 'variables' => $variables, 'exception' => $exception->getMessage()]);
```

mendelB: Oh sweet, way cleaner!! Updated in https://github.com/DoSomething/rogue/pull/889/commits/53c090d550046cf0b04d9a7becdb427405da48b8

app/Services/GraphQL.php
```diff
@@ -15,14 +16,50 @@ public function __construct()
...
+            title
+            slug
+          }
+        }';
+
+        $variables = [
+            'campaignId' => $campaignId,
+        ];
+
+        return $this->query($query, $variables, 'campaignWebsiteByCampaignId');
```
DFurnes: I think this needs to be updated after removing the `$queryKey` parameter in 45c6850:

```php
return $this->query($query, $variables)['campaignWebsiteByCampaignId'];
```

app/Services/GraphQL.php
```diff
@@ -15,14 +16,50 @@ public function __construct()
...
+            Log::error('GraphQL request failed.', [
+                'query' => $query,
+                'variables' => $variables,
+                'exception' => $exception->getMessage(),
+            ]);
+
+            return null;
+        }
+
+        return $response ? $response->getData() : null;
```
DFurnes: Instead of returning `null` here, would `[]` make sense to keep the return type consistent? That'd also help us avoid scenarios where we'd read a "key" from null (or have to use the `array_get` helper anywhere we call this function).

app/Services/GraphQL.php
mendelB: 🤦‍♂

app/Services/GraphQL.php
mendelB: I think that makes sense for consistency! (Though I feel like we'd still need to use `array_get` since otherwise, we might end up with an `undefined index` error?)

app/Services/GraphQL.php
mendelB: Would it make sense for the individual query methods to be parsing out the top level query key themselves, or should that responsibility rest with the invoking file? I'm thinking the reason the former might make sense is in order to be consistent with queries which do use multiple queries in one, where we wouldn't be able to just return the top level query key?

app/Services/GraphQL.php
DFurnes: If we always return an array from this method, then grabbing a nonexistent index should just be a warning (which sorta makes sense to me since that'd be unexpected behavior).

app/Services/GraphQL.php
```diff
@@ -15,14 +16,50 @@ public function __construct()
...
-     * @return \Softonic\GraphQL\Response
+     * @return array|null
      */
     public function query($query, $variables)
     {
-        return $this->client->query($query, $variables);
+        // Use try/catch to avoid any GraphQL related errors breaking the application.
+        try {
+            $response = $this->client->query($query, $variables);
+        } catch (\Exception $exception) {
```
DFurnes: Should we swap this to a more specific type of exception? Would we want a network failure, for example, to continue as a real `RuntimeException` (so if this is happening in a queue runner, we could then retry that job later)?

mendelB: Hmm so the graphql-client's `query` method specifies two potential Exceptions in its docblock:
The [`\RuntimeException`](https://github.com/softonic/graphql-client/blob/f60c0ac3945b7413a4cea2cd6cf4894ad3ba788c/src/Client.php#L21) which it sounds like we *don't* want to catch (which makes sense! I didn't realize we could just retry these failed jobs, and I think that sounds like the better option than just logging the error (and sending c.io `null` campaign `title` & `slug` values in the case of the current overarching work)).

Then there's the `\UnexpectedValueException` for an invalid JSON response body. 
https://github.com/softonic/graphql-client/blob/f60c0ac3945b7413a4cea2cd6cf4894ad3ba788c/src/Client.php#L20

I'm thinking now that maybe we don't need to `catch` even the latter for the same reason we won't `catch` the former? I guess I'm not sure of the scenario where we'd even get this `UnexpectedValueException` as opposed to just the `TransferException` (my futzing about with graphql to prompt errors in the response only managed to yield the `TransferException` due to `400`s), but either way, would it make sense to just rely on the retry in both cases?

DFurnes: Yeah, I think that's the right call not to catch these. Thanks for looking into it! 🙇

mendelB: And right back atcha -- thanks for all the helpful feedback! (I just had one more outstanding question about the response parsing https://github.com/DoSomething/rogue/pull/889#discussion_r297408575 if you're not tired of me yet 🕺)

app/Services/GraphQL.php
DFurnes: Hmmm, yeah that's a fair point... 🤔 

Now that you mention it, I wonder if it would make more sense to actually move this whole query _out_ of the `GraphQL` class and into the specific places we use it (so in the job itself, like how we colocate queries with their respective components in Phoenix). That way you could easily reference or change the query in the same place it's being used...

mendelB: Would it be an issue tho that we'd have to repeat that query across the [Post](https://github.com/DoSomething/rogue/blob/bc9561140a8ed98f39a7c257feaa725b8e5e7d73/app/Models/Post.php#L210-L243) and [Signup](https://github.com/DoSomething/rogue/blob/bc9561140a8ed98f39a7c257feaa725b8e5e7d73/app/Models/Signup.php#L139-L160) `toBlinkPayload` methods (I think those are the locations you mean), since we'll be needing these campaign fields in both places?

Ohh or maybe we can add the logic for the query as a method on the Campaign model itself so it's more specific? Or for that price maybe it's just worth having it defined in the service...

I dunno, maybe based on the Phoenix component usage example this indeed _would_ be worth repeating across the various places as we'd do with two different components using the same query (though in that case would we define it once as a fragment)? (lol sorry I'm making this so confusing 😖 🙃I guess my mental model for gracefully implementing graphQL in an app like this needs some serious therapy.)

DFurnes: I don't think there's a right answer, to be honest!

Philosophically, I think it does make sense to keep the queries as close to the place they're used since it makes it easier to understand & edit in the future… but I can also see how that'd make our (already pretty hefty) models even heftier!

I'm good with trying whichever approach you think works best for now, and we can always adjust and try other things if we end up making more GraphQL queries in this app.

mendelB: Fair! How's this for a compromise for the time being:
I'll leave the helper method since we're -as of now- using it identically across those models, but I'll also leave a comment specifying our preferred bundling approach so that this doesn't set a philosophically deviant precedent?

DFurnes: Smart! 👍

mendelB: Added with https://github.com/DoSomething/rogue/pull/889/commits/8f95dfc20e107c471db42fd11aa5d629e84ea404!

PR #890, opened 2019-07-02
Add Campaign Website Fields to Post/Signup Blink Payload

mendelB:
#### What's this PR do?
This PR adds the Campaign Website (Phoenix Contentful Campaign) `title` and `slug` fields to the `Signup` and `Post` Blink Payloads. We are finally utilizing the `GraphQL` service!

Should be super straightforward! I added some very basic test mocking, to ensure any test calls don't error out or result in extraneous requests to our GraphQL application.

For the field names - I utilized the existing pattern of `campaign_` prefixed values from the ID and run ID.

This payload is forwarded to Customer.io and triggered emails, thus, more campaign detail allows for more customized messaging! 

#### Relevant tickets
https://www.pivotaltracker.com/story/show/166360256

#### How to deploy


app/Models/Post.php
```diff
@@ -217,13 +218,18 @@ public function toBlinkPayload()
...
+
         return [
             'id' => $this->id,
             'signup_id' => $this->signup_id,
             'quantity' => $quantity,
             'why_participated' => $this->signup->why_participated,
             'campaign_id' => (string) $this->campaign_id,
             'campaign_run_id' => (string) $this->signup->campaign_run_id,
+            'campaign_title' => array_get($campaignWebsite, 'title'),
+            'campaign_slug' => array_get($campaignWebsite, 'slug'),
```
katiecrane: Since these are selected in the query we'll always have at least `null` for these values right?

mendelB: Based on my testing, if there's no matching entry for the provided `campaignId`, it'll just send back `null` as a top-level value without the fields themselves - even if they're specified in the query.

Buuut it seems we could get away with just grabbing the values without the `array_get` and it still wouldn't fail which is surprising to me! I had no idea that `null['index']` just returns `null`, or am I tripping?

image

So we could just have 
```php
...
'campaign_title' => $campaignWebsite['title'],
'campaign_slug' => $campaignWebsite['slug'],
...
```
And would we prefer that to the `array_get`s?

katiecrane: Yeah I think that style is better, nothing is nested here and you're not using any dot notation which I believe is the upside of [`array_get`](https://laravel.com/docs/5.2/helpers#method-array-get)! Most important though is that it won't error out no matter what it gets back from GraphQL which it sounds like it won't, my tests showed the same results as yours!

mendelB: Makes sense! updated in https://github.com/DoSomething/rogue/pull/890/commits/7849807c00e1e4a4330e541cd6862378b7ba1a3b

PR #894, opened 2019-07-11
Update Campaign->cause attribute accessor method

mendelB:
#### What's this PR do?
It seems that accessing the `Campaign->cause` attribute for a Campaign with `null` or empty string `cause` value returns an array with a single empty string value. 

image

This is because our `getCauseAttribute` accessor method always `explode`s the value, expecting a comma separated string.

This PR adds some logic to just output an empty array for an [`blank`](https://laravel.com/docs/5.5/helpers#method-blank) value.

Does this make sense at all or am I just completely off the mark about this observation / solution? Is the idea perhaps that since valid Campaigns require a Cause, an 'empty' cause is indeed what we desire?

I bumped into this while testing out adding the `cause` attribute to `Post`/`Signup` blink payloads. While testing for cases where the attribute was unset or null or an empty string, I noticed this array with an empty string value being returned, which I figured could cause some trouble or at least be a misrepresentation of the true value. 

#### Relevant tickets
Related to this effort [Pivotal ID#166360256](https://www.pivotaltracker.com/story/show/166360256)

#### How to deploy

### DIFF
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -88,7 +88,9 @@ class Campaign extends Model
      */
     public function getCauseAttribute()
     {
-        return explode(',', $this->attributes['cause']);
+        $cause = $this->attributes['cause'];
+
+        return blank($cause) ? [] : explode(',', $cause);
     }
 
     /**


mendelB:
> Look at that blank() helper already coming in for the win!

😄 

Shoot, I had meant, but forgotten to hat-tip @mshmsh5000 for this!

PR #896, opened 2019-07-15
Update Signup & Post Blink Payload Fields

mendelB:
#### What's this PR do?
This PR updates the Blink (-> Customer.io) payloads for the `Signup` and `Post` models:
- ~https://github.com/DoSomething/rogue/commit/aee5595c59e95321ac7d37c4066fc668a237b7ad casts the campaign website obtained values as `string`s to ensure we don't end up with any `null` values which would be dropped by Customer.io (https://customer.io/docs/attributes)~ (https://github.com/DoSomething/rogue/pull/896#discussion_r303962122)
- https://github.com/DoSomething/rogue/commit/dc304169f4e479c9471d282fa95c312863aa5c7d adds the Campaign `cause` attribute, fetched as the original attribute - a comma separated String, since Array's aren't supported by Customer.io
- https://github.com/DoSomething/rogue/commit/c41a30c559282c653fd5aa5fe9efb4a713c89691 adds a bunch of Action attributes to the payload, ~while also casting those as `string`s/`boolean`s to prevent dropped fields.~

I'm not really sure any of these values need to be `cast` at all? We are doing it for the `campaign_id`/`campaign_run_id` fields (this was the context I found https://github.com/DoSomething/rogue/pull/417). They seem to be dropped from the c.io payload if they're `null` but I don't think that'd be an issue either way?

#### Relevant tickets
[Pivotal ID#166360256](https://www.pivotaltracker.com/story/show/166360256)

#### How to deploy


app/Models/Post.php
```diff
@@ -218,22 +218,35 @@ public function toBlinkPayload()
         // Blink expects quantity to be a number.
         $quantity = $this->quantity === null ? 0 : $this->quantity;
 
+        // Array values aren't supported in Customer.io event attributes.
+        $campaign_cause = implode(',', (array) $this->signup->campaign['cause']);
```
mendelB: An alternative method would be to just obtain the original DB value without the [accessor formatting](https://github.com/DoSomething/rogue/blob/c79e6c145f6f3a923246b64c78bf6772cebd6d8e/app/Models/Campaign.php#L84-L94). (I thought it was a drop clunkier, but the advantage I guess is that we don't need to do explicit formatting on the value and just grab it as-is from the campaign?)
```suggestion
        $campaign_cause = (string) optional($this->signup->campaign)->getAttributes()['cause'];
```

katiecrane: That `optional()` helper is amazing who knew???? 

The 2nd option (straight from DB) seems like there is slightly less room for error but I agree that the way you have it now is cleaner, so your call!

mendelB: Hmm, ok, in that case, I kinda feel like going with the 2nd option so it's less error-prone rather than cleaner!

app/Models/Signup.php
```diff
@@ -146,6 +146,9 @@ public function toBlinkPayload()
         // Blink expects quantity to be a number.
         $quantity = $this->quantity === null ? 0 : $this->quantity;
 
+        // Array values aren't supported in Customer.io event attributes.
+        $campaign_cause = implode(',', (array) $this->campaign['cause']);
```
mendelB:
```suggestion
        $campaign_cause = (string) optional($this->campaign)->getAttributes()['cause'];
```

See https://github.com/DoSomething/rogue/pull/896#discussion_r303555913

app/Models/Post.php
```diff
@@ -218,22 +218,36 @@ public function toBlinkPayload()
...
             'signup_id' => $this->signup_id,
             'quantity' => $quantity,
             'why_participated' => $this->signup->why_participated,
             'campaign_id' => (string) $this->campaign_id,
             'campaign_run_id' => (string) $this->signup->campaign_run_id,
-            'campaign_title' => $campaignWebsite['title'],
-            'campaign_slug' => $campaignWebsite['slug'],
+            'campaign_title' => (string) $campaignWebsite['title'],
+            'campaign_slug' => (string) $campaignWebsite['slug'],
+            'campaign_cause' => $campaign_cause,
```
katiecrane: Does this need to be cast as a string as well? If it's not set `optional()` returns `null` I think? Or is it okay if this one is `null` and gets dropped?

app/Models/Post.php
```diff
@@ -218,22 +218,36 @@ public function toBlinkPayload()
...
             'northstar_id' => $this->northstar_id,
             'type' => $this->type,
             'action' => $this->getActionName(),
             'action_id' => $this->action_id,
+            'action_type' => (string) $action['action_type'],
+            'scholarship_entry' => (bool) $action['scholarship_entry'],
+            'civic_action' => (bool) $action['civic_action'],
+            'quiz' => (bool) $action['quiz'],
+            'online' => (bool) $action['online'],
+            'time_commitment' => (string) $action['time_commitment'],
```
katiecrane: I wanted to flag that for customers we are limited to sending 30 attributes at a time. This PR puts us at 30. BUT I believe this is sent at an event, so it is instead limited at 10000 bytes rather than a specific number of attributes? Wanted to double check that we are okay here! Docs are [here](https://customer.io/docs/api/) under "Limits."

app/Models/Post.php
mendelB: I cast it as a string up top when I set the variable, do you think it'd read clearer if it were done here? 

(In general, I also wasn't completely sure that any of this even _needed_ to be cast since we shouldn't have trouble if customer.io just drops them. (Maybe just to be safe? In which case it might make sense to cast _all_ the other `null`-able fields in this payload as well?))

katiecrane: OH I see it now! Jeez sorry! I think it's fine as is!

app/Models/Post.php
mendelB: 👍 That was my understanding as well! I also did a bit of testing in a byte counter online and we were well below the `10000` limit even with some contrived long af dummy data.

app/Models/Post.php
katiecrane: Hmm I think that's a @ngjo question - I'm not sure if it's confusing in c.io to see fields "missing"

mendelB: As [confirmed in Slack](https://dosomething.slack.com/archives/C0YGXUE01/p1563225297003600), we've removed the extra `cast`ing for these new attributes https://github.com/DoSomething/rogue/pull/896/commits/dae3df95b11c4680d68e76d42ff08264cb9174c9

PR #898, opened 2019-07-29
Update Rogue's "delete user" command.

DFurnes:
#### What's this PR do?
This pull request makes some updates to our "delete user" command (following the example of DoSomething/northstar#902), now that we'll be starting to run it periodically:
- I've renamed it to `rogue:delete` since it's not GDPR-specific anymore.
- I added the ability to pipe in a CSV, so we don't need temporary storage.
- Rather than just hiding posts, the command now soft-deletes them (see below).

I also added a test for this command! 🚥

For clarity, I think it makes sense to soft-delete these posts and signup records after anonymizing them (especially since I believe Quasar is planning to rely on the `deleted_at` field to handle making these changes on their end). Any concerns?

#### Relevant Tickets
[#167327865](https://www.pivotaltracker.com/story/show/167327865)

#### How to deploy


app/Console/Commands/DeleteUsersCommand.php
```diff
@@ -0,0 +1,77 @@
...
+                    'details' => null,
+                    'location' => null,
+                    'url' => null,
+                ]);
+
+                $post->delete();
+            }
+
+            $this->info('Deleted: ' . $id . '(' . $posts->count() . ' posts and ' . $signups->count() . ' signups)');
+        }
```
katiecrane: Same question as in Northstar, would it be helpful to have a final "Done anonymizing all signups and posts" at the end here so it's easy to tell when the script has finished?

tests/Console/DeleteUsersCommandTest.php
```diff
@@ -0,0 +1,40 @@
...
+     * Assert that the given model has been anonymized.
+     *
+     * @param User $user
+     */
+    protected function assertAnonymized($post)
+    {
+        // The post and its signup should be soft deleted, and fields that may
+        // contain personally-identifiable information should be erased:
+        $this->assertSoftDeleted('posts', ['id' => $post->id, 'text' => null, 'location' => null]);
+        $this->assertSoftDeleted('signups', ['id' => $post->signup_id, 'why_participated' => null]);
```
katiecrane: I know that if some of the fields got anonymized correctly then odds are they all did, but for peace of mind it might be nice to check for all the fields being wiped here. So adding `details` and `url` for `posts` and `details` for `signups`?

DFurnes: Mmm, sure!

app/Console/Commands/DeleteUsersCommand.php
DFurnes: Added this in 889dbad.

tests/Console/DeleteUsersCommandTest.php
DFurnes: Added assertions for these in 96bdd6e!

app/Console/Commands/DeleteUsersCommand.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Execute the console command.
+     *
+     * @param AWS $aws
+     * @param Fastly $fastly
+     *
+     * @return mixed
+     */
+    public function handle(AWS $aws, Fastly $fastly)
```
katiecrane: Same comment here as in Northstar, could be worth adding a `confirm` to this!

PR #900, opened 2019-07-31
Add min/max to dimensions error message

aaronschachter:
#### What's this PR do?
* Adds a `config/posts.php` to define the min and max height/width for a Post image
* Updates Post image validation to reference the config
* Fixes a bug - typo in checking for max height

I've been testing against my local Phoenix to verify:

Screen Shot 2019-07-31 at 10 49 03 AM

Doesn't seem ideal that this validation message is used for dimensions across the entire app, but AFAIK this shouldn't make a big difference as we only store images on Posts.

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/166767808


resources/lang/en/validation.php
```diff
@@ -34,7 +37,7 @@
     'different'            => 'The :attribute and :other must be different.',
     'digits'               => 'The :attribute must be :digits digits.',
     'digits_between'       => 'The :attribute must be between :min and :max digits.',
-    'dimensions'           => 'The :attribute has invalid image dimensions.',
+    'dimensions'           => 'The :attribute size is invalid: must be at least '.$minImageSize['width'].'x'.$minImageSize['height'].' and no larger than '.$maxImageSize['width'].'x'.$maxImageSize['height'].'.',
```
DFurnes: We can customize this with Laravel's built-in [custom error messages](https://laravel.com/docs/5.8/validation#custom-error-messages) functionality, like `digits_between` above! This will read the provided arguments from anywhere we use this `dimensions` validator: 

```suggestion
    'dimensions'           => 'The :attribute size is invalid: must be at least :min_widthx:min_height and no larger than :max_widthx:max_height.',
```

aaronschachter: Great find, thanks! Updated.

PR #902, opened 2019-08-01
Add campaigns.impact_doc to API response

aaronschachter:
#### What's this PR do?

Fixes missing Proof Of Impact link from a `/campaigns-id/:id` page by returning it from the API response. Is this URL ok to make public?

Example response of a **GET** `/api/v3/campaigns/9003` request:
```
{
  "data": {
    "id": 9003,
    "internal_title": "Assimilated Zerodefect Projection",
    "start_date": "2019-04-25T03:12:15+00:00",
    "end_date": "2020-01-24T21:10:45+00:00",
    "is_open": true,
    "impact_doc": "https://www.google.com/",
    ...
  }
}
```

👀 

#### Relevant tickets
https://www.pivotaltracker.com/n/projects/2019429/stories/166356076

### DIFF
--- a/app/Http/Controllers/Legacy/Web/CampaignsController.php
+++ b/app/Http/Controllers/Legacy/Web/CampaignsController.php
@@ -49,7 +49,7 @@ class CampaignsController extends Controller
     }
 
     /**
-     * Show particular campaign and it's posts.
+     * Show particular campaign and its posts.
      *
      * @param  int $id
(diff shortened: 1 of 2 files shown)


DFurnes:
> Is this URL ok to make public?

Should be! AFAIK, these docs would only be accessible to folks with `@dosomething.org` emails.

PR #903, opened 2019-08-01
Removes duplicate README

aaronschachter:
#### What's this PR do?
Follow up to #891 - didn't realize there were 2 places to change things. Why have the duplicate README at all?

👀 

🔥


DFurnes:
> Why have the duplicate README at all?

I think we added the `docs/README.md` so that it'd be picked up by this app's [GitBook](https://dosomething.gitbook.io/rogue-documentation/). We could definitely delete the one in the root directory, though!

PR #908, opened 2019-09-24
Adds Reject Post Function to Hide Display of Runscope Posts

Vmack:
#### What's this PR do?
To effectively prevent runscope tests from appearing in a campaign lead's inbox, this PR adds a RejectPost function to delay the appearance of a 'test' post for 15 minutes after it's created and then marks the status of the post as 'rejected'.

In rogue.test, click on upload a photo. 
Add a photo and for the caption add 'test runscope upload' and click submit. 
In terminal and inside of rogue in vagrant  `cd` into the /storage/logs folder and `tail -f` the most current log file. 
In a separate tab in terminal run `php artisan queue:listen`. This will watch the files for any changes made 
You should see an update to the log file 15 minutes after creating the post

what else?
We want to delay the test posts so that the test experience would mimic the normal user experience 
#### Relevant tickets
Fixes #[161383417](https://www.pivotaltracker.com/story/show/161383417)

#### How to deploy

### DIFF
--- a/app/Jobs/RejectPost.php
+++ b/app/Jobs/RejectPost.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Rogue\Jobs;
+
+use Rogue\Models\Post;
+use Illuminate\Bus\Queueable;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+
+class RejectPost implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
+     * The post to send to Customer.io.
+     *
+     * @var Post
+     */
+    protected $post;
+
+    /**
+     * Create a new job instance.
+     *
+     * @return void
+     */
+    public function __construct(Post $post)
+    {
+        $this->post = $post;
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        // Set the status of post to rejected 
+        $this->post->status = 'rejected'; 
+        $this->post->save();
+
+        // Log
+        info('Automatically rejected post ' . $this->post->id);
+    }
+}
+
--- a/app/Jobs/RejectPost.php
+++ b/app/Jobs/RejectPost.php
@@ -38,11 +38,13 @@ class RejectPost implements ShouldQueue
     public function handle()
     {
 
+
         $this->post->status = 'rejected'; 
         $this->post->save();
 
         // Log
         info('Automatically rejected post ' . $this->post->id);
     }
+
     
 }
--- a/app/Jobs/RejectPost.php
+++ b/app/Jobs/RejectPost.php
@@ -9,13 +9,12 @@ use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Bus\Dispatchable;
 
-
 class RejectPost implements ShouldQueue
 {
     use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
 
     /**
-     * The post to send to Customer.io.
+     * The post to reject.
      *
      * @var Post
      */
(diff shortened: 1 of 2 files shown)


app/Providers/ModelServiceProvider.php
```diff
@@ -30,13 +34,23 @@ public function boot()
             }
         });
 
+         // When Posts are created reject test events.
+         Post::created(function ($post) {
+            // Check if caption matches 'Test runscope upload' and 'caption_ghost_test'                
+             if (in_array($post->text, ['Test runscope upload', 'caption_ghost_test'])) {
+                 // If it does, run the RejectPost function and delay for 15 minutes 
+                RejectPost::dispatch($post)->delay(now()->addMinutes(15));
```
DFurnes: Do you think we could delay for a shorter amount of time (say, 2 minutes)? As long as Runscope or Ghost Inspector would see the same "normal" behavior for the duration of their test run, I think we'd be good (and the less time these sit "unreviewed", the less likely a campaign lead is weirded out by them).

app/Providers/ModelServiceProvider.php
```diff
@@ -30,13 +34,23 @@ public function boot()
             }
         });
 
+         // When Posts are created reject test events.
+         Post::created(function ($post) {
+            // Check if caption matches 'Test runscope upload' and 'caption_ghost_test'
```
DFurnes: This comment feels unnecessary since it's so similar to the actual line of code it's describing. Generally, I think comments are best used to describe the "why" for a line (so, here, a good comment might describe why why we're waiting 15 minutes after the post is submitted to delete it, or why we're rejecting test posts in the first place).

app/Providers/ModelServiceProvider.php
```diff
@@ -6,8 +6,12 @@
 use Rogue\Models\Event;
 use Rogue\Models\Review;
 use Rogue\Models\Signup;
+use Rogue\Jobs\RejectPost;
 use Illuminate\Support\ServiceProvider;
 
+
+use function GuzzleHttp\Promise\is_rejected;
```
DFurnes: I think this might've been included by mistake! 👾

app/Jobs/RejectPost.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * Execute the job.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        // Set the status of post to rejected
```
DFurnes: Same idea here: `$this->post->status = 'rejected'` reads very closely to "set the status of post to rejected", so this comment can probably be omitted!

app/Providers/ModelServiceProvider.php
Vmack: That's actually a good idea! I can change that, thanks Dave!

app/Providers/ModelServiceProvider.php
Vmack: ok, gotcha!

app/Jobs/RejectPost.php
```diff
@@ -0,0 +1,49 @@
...
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+
+
+class RejectPost implements ShouldQueue
+{
+    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
+     * The post to send to Customer.io.
```
katiecrane: This comment can be updated to say "The post to reject"

app/Providers/ModelServiceProvider.php
katiecrane: If we see any weirdness we can always up the delay!

app/Providers/ModelServiceProvider.php
```diff
@@ -30,13 +34,23 @@ public function boot()
...
+        });
+
         // When Posts are saved create an event for them.
         Post::saved(function ($post) {
             $post->events()->create([
                 // @TODO: this should include the tags with the post
                 'content' => $post->toJson(),
                 // Use the authenticated user if coming from the web,
-                // otherwise use the id of the user in the request.
+                // Otherwise use the id of the user in the request.
```
katiecrane: This "otherwise" can remain lowercase ✍️

PR #909, opened 2019-09-30
Update to use Gateway 2.x & make API requests with user's token.

DFurnes:
#### What's this PR do?
This pull request makes two updates to unblock [Access Control & Auditing](https://docs.google.com/document/d/1OclKWYEtjo0LTI9DzKaVG1dEBY6kfAEEgYZVRzSlD7s) work in this app:

🆙 First, it just [updates to Gateway 2.x](https://github.com/DoSomething/gateway/releases)! (~Pending https://github.com/DoSomething/gateway/pull/126.~ Released as [v2.2.0](https://github.com/DoSomething/gateway/releases/tag/v2.2.0) and included here in fa9f4d6b).

🌵 It updates Rogue's registrar to fetch users _as the user who is making the request_.

This is super-important for a few reasons:
1. We shouldn't be handling access control in multiple places (e.g. in Northstar & then again here in Rogue)! That just makes it easy to forget a check & accidentally expose sensitive data.
2. Similarly, we don't want to be caching authenticated requests here since we'd be using one person's authorization to then show "privileged" data to someone else! Danger!
3. We need the real "identity" for the person making requests in Northstar so that we can log access to sensitive fields. When Rogue makes all requests as an admin, we have no way of knowing who's actually using each request on the other end.

I've got a follow-up PR with user-facing changes based on this. 🙈

I've tested this out on my local (reading from real Northstar on big campaign inboxes) and there's a bit of a performance hit, but no request timeouts. If we have time at some point, I'd like to update Rogue to use GraphQL for better batching of requests, but that's non-trivial.

#### Relevant tickets
[#167596656](https://www.pivotaltracker.com/story/show/167596656)

#### How to deploy


tests/Http/SignupTest.php
```diff
@@ -561,59 +561,20 @@ public function testSignupShowWithIncludedPostsAndParamsAsOwner()
...
-    {
-        $post = factory(Post::class)->create();
-        $signup = $post->signup;
-
-        // Test with annoymous user that only a user's first name is returned.
-        $response = $this->getJson('api/v3/signups?include=user');
-        $response->assertStatus(200);
-        $decodedResponse = $response->decodeResponseJson();
-
-        $this->assertEquals(true, empty($decodedResponse['data'][0]['user']['data']['birthdate']));
```
DFurnes: Note: These tests are no longer necessary since Rogue itself doesn't have any authorization logic for users. (It won't ever get anything from Northstar that the user making the request isn't allowed to see).

app/Repositories/CacheRepository.php
```diff
@@ -1,142 +0,0 @@
...
-                    $id = $this->unsetPrefix($key);
-                }
-
-                $items[$key] = $this->find($id);
-            }
-        }
-
-        return $items;
-    }
-}
```
katiecrane: Was this only used for caching users?

DFurnes: Yep, this was the last place this was being used. (It used to cache campaigns too when we used to load them from Drupal, but that's no longer the way of things!)

tests/Http/SignupTest.php
katiecrane: So this test is essentially redundant to a similar test in Northstar?

weerd:
> Similarly, we don't want to be caching authenticated requests here since we'd be using one person's authorization to then show "privileged" data to someone else! Danger!

Ya, this is no bueno 😱

PR #911, opened 2019-10-01
Update Rogue admin interface to use preview fields.

DFurnes:
#### What's this PR do?
This pull request updates Rogue's admin interface to remove sensitive fields (like birthdate or mobile number) in favor of the new "preview" fields. It also updates dependencies & fixes a strange bug due to not returning a Promise and trying to `then` it later (dunno why it worked in the first place, or why we have the same method in two different files).

Screen Shot 2019-10-01 at 11 05 30 AM

Screen Shot 2019-10-01 at 11 01 58 AM

I'd recommend peeking commit-by-commit.

I moved two helper functions into the `UserInformation` component since they're only used in that one place (and so extracting them into `helpers.js` made for unnecessary jumping around & contributed to the helper file just becoming a catch-all).

#### Relevant tickets
[#167596656](https://www.pivotaltracker.com/story/show/167596656)

#### How to deploy

### DIFF
--- a/resources/assets/components/Users/UserInformation.js
+++ b/resources/assets/components/Users/UserInformation.js
@@ -1,69 +1,74 @@
 import React from 'react';
 import { isEmpty } from 'lodash';
 import PropTypes from 'prop-types';
-import { displayUserInfo, displayCityState } from '../../helpers';
+
+/* eslint-disable react/prop-types */
+
+/**
+ * Returns a readable City and State string.
+ *
+ * @param {Object} user
+ * @return {ReactElement|null}
+ */
+const UserLocation = ({ user }) => {
+  if (!user || !user.addr_city || !user.addr_state) {
+    return null;
+  }
+
+  return (
+    <span>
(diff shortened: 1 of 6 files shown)


resources/assets/components/Users/UserInformation.js
```diff
@@ -1,69 +1,74 @@
...
-        )}
+        <h2 className="heading">
+          <UserName
+            user={props.user}
+            link={props.linkSignup ? `/signups/${props.linkSignup}` : null}
+          />
+        </h2>
         <p>
-          {props.user.email ? (
+          {props.user.email_preview ? (
```
katiecrane: Is the idea that admins will no longer be able to see full emails in Rogue at all?

DFurnes: Yep! If they need to for some reason, they can hop over to Aurora.

katiecrane: LOCK. IT. DOWN. 👏 🔒

PR #915, opened 2019-10-03
Link to Aurora on user overview page.

DFurnes:
#### What's this PR do?
This pull request adds a link to a user's Aurora profile on the "user overview" page, which is useful if you [need to lookup their email or mobile](https://dosomething.slack.com/archives/C09ANFQLA/p1570026635041100?thread_ts=1570026476.040100&cid=C09ANFQLA). However.... rather than add logic to Rogue to know about different Aurora URLs, I decided to build off the work I did yesterday in DoSomething/graphql#136 by reworking this to use GraphQL! 🙈

This pull request adds a "modern" application shell in c9fb6df which uses [Reach Router](https://reach.tech/router) & [Apollo Client](https://www.apollographql.com/docs/react/) to fetch data for this interface. It also sets us up for using client-side routing (rather than having to hard-code a container for each page). (c9fb6df)

I then refactored `UserOverview`, `SignupCard`, `PostTile`, and `UserInformation` to query their data dependencies via GraphQL. For `PostTile` and `UserInformation`, they're configured to accept both response formats (either GraphQL or REST APIs) for now. (998d10c)

Finally, in 3cfbff3 I added the permalink. 🔗

I'd recommend reviewing commit-by-commit & "ignoring" white-space in the diff.

This introduces one little "snag" for local development – this screen will load content from whichever GraphQL environment is set in `GRAPHQL_URL`, rather than your local instance. You can still develop against local data by running a local GraphQL server with `QUERY_ENV = local`.

#### Relevant tickets
Follow-up to [#167596656](https://www.pivotaltracker.com/story/show/167596656) & some pre-work for Q4. 😇

#### How to deploy

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "@babel/runtime": "^7.5.4",
     "@dosomething/forge": "^6.9.1",
     "@dosomething/gateway": "^2.0.1",
-    "@reach/router": "^1.2.1",
     "apollo-cache-inmemory": "^1.6.3",
     "apollo-client": "^2.6.4",
     "apollo-link": "^1.2.13",
@@ -62,7 +61,9 @@
     "lodash": "^4.17.14",
     "prop-types": "^15.6.2",
     "react": "^16.10.1",
-    "react-dom": "^16.10.1"
+    "react-dom": "^16.10.1",
+    "react-router": "^5.1.2",
+    "react-router-dom": "^5.1.2"
   },
   "devDependencies": {
     "@babel/core": "^7.5.4",
--- a/package.json
+++ b/package.json
@@ -40,9 +40,18 @@
     }
   },
   "dependencies": {
+    "@apollo/react-hooks": "^3.1.2",
     "@babel/runtime": "^7.5.4",
     "@dosomething/forge": "^6.9.1",
     "@dosomething/gateway": "^2.0.1",
+    "@reach/router": "^1.2.1",
+    "apollo-cache-inmemory": "^1.6.3",
+    "apollo-client": "^2.6.4",
+    "apollo-link": "^1.2.13",
+    "apollo-link-batch-http": "^1.2.13",
+    "apollo-link-context": "^1.0.19",
+    "apollo-link-error": "^1.1.12",
+    "apollo-link-persisted-queries": "^0.2.2",
     "classnames": "^2.2.6",
     "core-js": "^3.1.4",
     "date-fns": "^1.30.1",
--- a/resources/assets/components/PostTile/index.js
+++ b/resources/assets/components/PostTile/index.js
@@ -5,7 +5,7 @@ import { getImageUrlFromProp } from '../../helpers';
 import './post-tile.scss';
 
 const PostTile = ({ post }) => {
-  const url = getImageUrlFromProp(post, 'edited');
+  const url = post.media ? getImageUrlFromProp(post, 'edited') : post.url;
   const altText = `Post #${post.id}`;
 
   if (!url) {
--- a/resources/assets/components/SignupCard/index.js
+++ b/resources/assets/components/SignupCard/index.js
@@ -23,9 +23,9 @@ const SignupCard = ({ signup }) => {
             <h2 className="heading">
               {campaign.internalTitle || 'Unknown Campaign'}
             </h2>
-            <h4 className="heading">Campaign ID: {campaign.id}</h4>
+            <h4 className="heading">Campaign ID: {campaign.id || 'N/A'}</h4>
             <h4 className="heading">
-              Campaign Start Date: {campaign ? campaign.startDate : '???'}
+              Campaign Start Date: {campaign.startDate || '???'}
             </h4>
           </div>
           <div className="container__row">
--- a/resources/assets/components/SignupCard/index.js
+++ b/resources/assets/components/SignupCard/index.js
@@ -1,92 +1,66 @@
 import React from 'react';
-import { map } from 'lodash';
-import PropTypes from 'prop-types';
-import './signup-card.scss';
 import PostTile from '../PostTile';
-import { RestApiClient } from '@dosomething/gateway';
-import { extractPostsFromSignups } from '../../helpers';
 
-class SignupCard extends React.Component {
-  render() {
-    const signup = this.props.signup;
-    const campaign = this.props.campaign;
-    const gallerySize = 4;
+import './signup-card.scss';
 
-    const extraPostCount = signup.posts.data.length - gallerySize;
+const POST_GALLERY_SIZE = 4;
 
-    const posts = signup.posts.data
-      .slice(0, gallerySize)
-      .map(post => <PostTile key={post.id} post={post} />);
+const SignupCard = ({ signup }) => {
+  const campaign = signup.campaign || {};
 
-    const signupUrl = `/signups/${signup.id}`;
+  const extraPostCount = signup.posts.length - POST_GALLERY_SIZE;
+  const posts = signup.posts.slice(0, POST_GALLERY_SIZE).map(post => (
+    <li key={post.id}>
+      <PostTile post={post} />
+    </li>
+  ));
 
-    return (
-      <article className="container__row signup-card">
-        <a href={signupUrl}>
-          <div className="container__block -half">
-            <div className="container__row">
-              <h2 className="heading">
-                {campaign ? campaign.internal_title : signup.campaign_id}
-              </h2>
-              <h4 className="heading">Campaign ID: {signup.campaign_id}</h4>
-              {campaign ? (
-                <h4 className="heading">
-                  Campaign Start Date: {campaign.start_date}
-                </h4>
-              ) : null}
-            </div>
-            <div className="container__row">
-              <h4 className="heading">Why Statement</h4>
-              <p>{signup.why_participated}</p>
-            </div>
+  return (
+    <article className="container__row signup-card">
+      <a href={`/signups/${signup.id}`}>
+        <div className="container__block -half">
+          <div className="container__row">
+            <h2 className="heading">
+              {campaign.internalTitle || 'Unknown Campaign'}
+            </h2>
+            <h4 className="heading">Campaign ID: {campaign.id}</h4>
+            <h4 className="heading">
+              Campaign Start Date: {campaign ? campaign.startDate : '???'}
+            </h4>
           </div>
-          <div className="container__block -half">
-            {signup.quantity ? (
-              <div className="container__row figure -left -center">
-                <div className="figure__media">
-                  <div className="quantity">{signup.quantity}</div>
-                </div>
-                <div className="figure__body">
-                  <h4 className="reportback-noun-verb">things done</h4>
-                </div>
-              </div>
-            ) : null}
-            {posts.length ? (
-              <div className="container__row">
-                <h4>Items</h4>
-                <ul className="gallery">
-                  {posts}
-                  {extraPostCount > 0 ? (
-                    <li className="figure__media">
-                      <div className="quantity">+{extraPostCount}</div>
-                    </li>
-                  ) : null}
-                </ul>
-              </div>
-            ) : null}
+          <div className="container__row">
+            <h4 className="heading">Why Statement:</h4>
+            <p>{signup.whyParticipated}</p>
           </div>
-        </a>
-      </article>
-    );
-  }
-}
-
-SignupCard.propTypes = {
-  signup: PropTypes.shape({
-    posts: PropTypes.object,
-    why_participated: PropTypes.string,
-    signup_id: PropTypes.number,
-    campaign_id: PropTypes.string,
-    quantity: PropTypes.number,
-  }).isRequired,
-  campaign: PropTypes.shape({
-    internal_title: PropTypes.string,
-    start_date: PropTypes.string,
-  }),
-};
-
-SignupCard.defaultProps = {
-  campaign: null,
+        </div>
+        <div className="container__block -half">
+          {signup.quantity ? (
+            <div className="container__row figure -left -center">
+              <div className="figure__media">
+                <div className="quantity">{signup.quantity}</div>
+              </div>
+              <div className="figure__body">
+                <h4 className="reportback-noun-verb">things done</h4>
+              </div>
+            </div>
+          ) : null}
+          {posts.length ? (
+            <div className="container__row">
+              <h4>Items</h4>
+              <ul className="gallery -quartet">
+                {posts}
+                {extraPostCount > 0 ? (
+                  <li className="figure__media">
+                    <div className="quantity">+{extraPostCount}</div>
+                  </li>
+                ) : null}
+              </ul>
+            </div>
+          ) : null}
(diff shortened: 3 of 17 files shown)


resources/assets/components/UserOverview/index.js
```diff
@@ -1,149 +1,85 @@
...
 
-    return (
-      <div>
-        <div className="container__block">
-          <h2 className="heading -emphasized -padded">
-            <span>User Info</span>
-          </h2>
-        </div>
+  if (loading) return <p>Loading...</p>;
+  if (error) return <p>Error :(</p>;
```
DFurnes: These could probably both use some love, but seemed out of scope for this pull request! Now that we're on React 16 here, I'd also love to start adding some more error boundaries so smaller issues don't crash the whole app.

resources/assets/components/UserOverview/index.js
```diff
@@ -1,149 +1,85 @@
...
-        this.setState({
-          campaigns: keyBy(json.data, 'id'),
-        }),
-      );
-  }
-
-  render() {
-    const user = this.props.user;
+const UserOverview = ({ id }) => {
+  const { loading, error, data } = useQuery(USER_OVERVIEW_QUERY, {
```
DFurnes: Note that this is following [the "modern" Apollo Client hooks API](https://www.apollographql.com/docs/react/data/queries/#executing-a-query), so looks a little different than usual!

resources/assets/components/UserOverview/index.js
```diff
@@ -1,149 +1,85 @@
...
+    variables: { id },
+  });
```
DFurnes: Note that this is following [the "modern" Apollo Client hooks API](https://www.apollographql.com/docs/react/data/queries/#executing-a-query), so looks a little different than usual!

resources/assets/components/Users/UserInformation.js
```diff
@@ -1,8 +1,19 @@
...
 import { isEmpty } from 'lodash';
-import PropTypes from 'prop-types';
+import { useQuery } from '@apollo/react-hooks';
 
-/* eslint-disable react/prop-types */
+export const UserInformationFragment = gql`
+  fragment UserInformation on User {
+    id
+    displayName
+    # age
```
DFurnes: I forgot to add this to our GraphQL schema, hence why it's commented out for now! 🙈

resources/assets/helpers.js
```diff
@@ -260,3 +260,24 @@ export function getImageUrlFromPost(post, type) {
...
+ * @return {String}
+ */
+export function env(key) {
+  return (window.ENV || {})[key];
+}
+
+/*
+ * Change the keys in the given object to camelCase. This
+ * is useful as a "compatibility layer" for components
+ * that recieve data from GraphQL & REST APIs.
```
mendelB: So smart! 💡

resources/assets/graphql.js
```diff
@@ -0,0 +1,48 @@
...
+export default uri => {
+  // Handle making persisted queries so we send fewer bytes per request.
+  const persistedLink = createPersistedQueryLink();
+
+  // Create the HTTP link! This is our terminating link that batches up
+  // GraphQL queries and makes the actual HTTP request to our server.
+  const httpLink = new BatchHttpLink({ uri });
+
+  return new ApolloClient({
+    link: ApolloLink.from([errorLink, authLink, httpLink]),
```
mendelB: Are we missing the `persistedLink` here?

package.json
```diff
@@ -40,9 +40,18 @@
     }
   },
   "dependencies": {
+    "@apollo/react-hooks": "^3.1.2",
     "@babel/runtime": "^7.5.4",
     "@dosomething/forge": "^6.9.1",
     "@dosomething/gateway": "^2.0.1",
+    "@reach/router": "^1.2.1",
```
mendelB: Out of curiosity -- why `@reach/router` over `react-router`?

resources/assets/components/Users/UserInformation.js
```diff
@@ -43,53 +53,36 @@ const UserName = ({ user, link }) => {
   return <span>{displayName}</span>;
 };
 
-const UserInformation = props => (
+export default ({ user, linkSignup, children }) => (
```
mendelB: Is this `default` preferred over our usual approach of exporting the component names explicitly?

resources/assets/components/Users/UserInformation.js
```diff
@@ -43,53 +53,36 @@ const UserName = ({ user, link }) => {
...
         </p>
       </div>
     ) : null}
 
-    {props.children}
+    {children}
   </div>
 );
-
-UserInformation.propTypes = {
```
mendelB: Are we cool without the prop type checking?

resources/assets/components/SignupCard/index.js
```diff
@@ -1,92 +1,66 @@
...
+    <article className="container__row signup-card">
+      <a href={`/signups/${signup.id}`}>
+        <div className="container__block -half">
+          <div className="container__row">
+            <h2 className="heading">
+              {campaign.internalTitle || 'Unknown Campaign'}
+            </h2>
+            <h4 className="heading">Campaign ID: {campaign.id}</h4>
+            <h4 className="heading">
+              Campaign Start Date: {campaign ? campaign.startDate : '???'}
```
mendelB: Why do we condition the `campaign`s existence here as opposed to [above](https://github.com/DoSomething/rogue/pull/915/files#diff-ff380d564bdc83de86dff5ffd7dcd30bR24) re the `campaign.internalTitle`?

resources/assets/components/SignupCard/signup-card.scss
```diff
@@ -7,31 +7,6 @@
...
-          padding-left: 0px;
-        }
-
-        &:nth-of-type(5n + 1) {
-          clear: both;
-        }
-      }
-    }
-  }
-
```
mendelB: Just confirming  -- is this just not used any longer?

resources/assets/components/UserOverview/index.js
mendelB: Oh neato!! Is this related to removing the Prop type checking on the components, or totally unrelated?

resources/assets/graphql.js
DFurnes: Oh gee! Yep, I was moving too quick! 🏃

resources/assets/components/Users/UserInformation.js
DFurnes: I don't have much of a preference, but I think you're right that we tend to name and export! I'll keep that convention going here. 👍

resources/assets/components/Users/UserInformation.js
DFurnes: I haven't found a lot of value from prop-type checks since they don't run as static analysis on the pull request (and so I've often missed them if I'm not careful). I also think that as we move towards GraphQL queries, we get the benefits of a typed response out of the box (so re-typing with prop-types seems a little redundant).

resources/assets/components/SignupCard/index.js
DFurnes: Oh, no particular reason. I'll swap both of these to use the `|| 'fallback'` pattern.

resources/assets/components/Users/UserInformation.js
mendelB: Ohhh right, that's interesting re GraphQL! I wonder if there's a way to like actually run static analysis on the PR so it's more useful for non-gql props, and then we could have a healthy balance of type checking where appropriate (with maybe more lenient lint rules or some concession for gql props 🤔)?

resources/assets/components/SignupCard/signup-card.scss
DFurnes: From what I could tell, this is currently broken on `master` (showing zero posts for the first campaign, where I have 8) and all tiles are full-size:

Screen Shot 2019-10-03 at 3 29 50 PM

 I just replaced with `.gallery.-quartet` which works out of the box:

Screen Shot 2019-10-03 at 3 29 58 PM

resources/assets/components/UserOverview/index.js
DFurnes: Only sorta (re: their redundancy with typed GraphQL responses described above).

package.json
katiecrane: Same question!

resources/views/layouts/master.blade.php
```diff
@@ -46,4 +46,5 @@
     </body>
 
     {{ scriptify($auth, 'AUTH') }}
+    {{ scriptify(['GRAPHQL_URL' => config('services.graphql.url')], 'ENV') }}
```
katiecrane: I'm having trouble finding what this `scriptify` does, could you link me to docs? I'm not sure I'm finding the right thing

resources/assets/components/PostTile/index.js
```diff
@@ -5,7 +5,7 @@ import { getImageUrlFromProp } from '../../helpers';
 import './post-tile.scss';
 
 const PostTile = ({ post }) => {
-  const url = getImageUrlFromProp(post, 'edited');
+  const url = post.media ? getImageUrlFromProp(post, 'edited') : post.url;
```
katiecrane: 👏

resources/assets/graphql.js
DFurnes: Fixed in 250ad970.

resources/assets/components/Users/UserInformation.js
DFurnes: Changed in 0b63d480.

resources/assets/components/Users/UserInformation.js
DFurnes: I don't think we can with prop-types (since JavaScript is fairly lenient and so there's often no way for something like ESLint to know what types would be passed to a component until the app runs). But I have [some ideas](https://www.typescriptlang.org) on how we might do more static analysis in the future…

resources/assets/components/SignupCard/index.js
DFurnes: Updated in 1ab4eb09.

resources/views/layouts/master.blade.php
DFurnes: It's a [helper function](https://github.com/DoSomething/rogue/blob/c555b2e5f64dc76d71c62f9c80c6e3b6fcf08bca/app/helpers.php#L6-L16) that we made (originally in Phoenix, and then copied over here), which takes the given variable and JSON-encodes it in a JavaScript variable.

So, in this case, it ends up writing `window.ENV = { GRAPHQL_URL: '…' };`

package.json
DFurnes: I'd wanted to play around with it since the API looks a bit cleaner than the one that we're used to, and it was [effectively positioned as a "preview" of upcoming changes to React Router](https://reacttraining.com/blog/reach-react-router-future/).

Buuuuuut after double-checking right now, it seems like the [newer  React Router 5.1 release](https://github.com/ReactTraining/react-router/releases/tag/v5.1.0) with the simplified hooks-based API came out 10 days ago, so I'll just swap to use that!

resources/assets/components/SignupCard/index.js
```diff
@@ -1,92 +1,66 @@
...
-              <h4 className="heading">Why Statement</h4>
-              <p>{signup.why_participated}</p>
-            </div>
+  return (
+    <article className="container__row signup-card">
+      <a href={`/signups/${signup.id}`}>
+        <div className="container__block -half">
+          <div className="container__row">
+            <h2 className="heading">
+              {campaign.internalTitle || 'Unknown Campaign'}
```
katiecrane: I would have thought this would just return a boolean, is this a shortcut to get to something like `campaign.internalTitle ? campaign.internalTitle : 'Unknown Campaign`?

resources/assets/components/SignupCard/signup-card.scss
katiecrane: YESSSSSS

resources/assets/components/SignupCard/index.js
DFurnes: Yep! It's a trick that takes advantage of the fact that `||` will short-circuit once it gets something "truthy", and so if `campaign.internalTitle` exists it'll do that, and only evaluate the second part of the expression if not.

katiecrane: Once it gets something truthy won't it return `true` though? Sorry this is blowing my mind a bit 🐡

DFurnes: Nope! It just returns the "truthy" value. This used to be [a common, if somewhat flawed, way to set default values](https://www.codereadability.com/javascript-default-parameters-with-or-operator/) before that became part of the language (with default parameters for functions).

resources/assets/components/UserOverview/index.js
```diff
@@ -1,149 +1,85 @@
...
-  getUserActivity(id) {
-    const request = this.api.get('api/v3/signups', {
-      filter: {
-        northstar_id: id,
-      },
-      include: 'posts',
-      orderBy: 'created_at,desc',
-      limit: 80,
-    });
+  ${UserInformationFragment}
```
katiecrane: Is this what allows it to use `...UserInformation` above?

resources/assets/components/UserOverview/index.js
katiecrane: This is fair for now!

resources/assets/components/SignupCard/index.js
katiecrane: > First of all, it isn’t obvious why the OR operator would be used to assign default values. It takes time for this hack to become intuitive.

This sums up my feelings 😂 but it does work! Very cool!

resources/assets/components/UserOverview/index.js
DFurnes: Yep, exactly! This follows a similar pattern we use in Phoenix to query data [via a central "loader" component](https://github.com/DoSomething/phoenix-next/blob/155df29f183c74c44e9710ab68fc7e137e7eb135/resources/assets/components/utilities/ContentfulEntryLoader/ContentfulEntryLoader.js#L21-L46) but colocate the actual "data requirements" with the components themselves (like [here](https://github.com/DoSomething/phoenix-next/blob/155df29f183c74c44e9710ab68fc7e137e7eb135/resources/assets/components/actions/PetitionSubmissioncAction/PetitionSubmissionAction.js#L22-L33)).

resources/views/layouts/master.blade.php
katiecrane: Well that explains why I couldn't find it via Google! Thanks Dave!

package.json
DFurnes: Updated to use `react-router` with the `useParams` hook in 1b2b3eee.

DFurnes:
Finally, just uncommented `age` in 38a0291 after adding that field in DoSomething/graphql#138.

resources/assets/components/Users/UserInformation.js
DFurnes: And uncommented in 38a0291.

PR #916, opened 2019-10-10
Remove Glide feature flag & unused code paths.

DFurnes:
#### What's this PR do?
This pull request removes the `DS_ENABLE_GLIDE` feature flag in 7300122, since we've been using Glide to format images for quite a long time now without issues! I also did some cleanup & refactoring while I was in some of these files. Notably:

📝 No longer create `edited_XXXX.jpg` images when creating posts, since they're only used when Glide is disabled. This should significantly speed up the `POST v3/posts` endpoint. adba0e8

🆙 Increases the rate limit (from 30/min. to 60/min.) for `GET /images/{id}`. This should reduce the number of alarms we get when folks in the office (or elsewhere) are digging deep into older campaign galleries, while still shielding us from abusive behavior. 5f1a68e

🔗 Removes code that handled Data-URL uploads, since that's not a thing that we allow in the v3 API. This lets us remove a lot of relatively confusing code! 185311f

I'd recommend reviewing this commit-by-commit.

Our locals & CI environments aren't using S3, so I'll do an extra round of testing on QA.

#### Relevant tickets
This is some pre-work to [#165980718](https://www.pivotaltracker.com/story/show/165980718).

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Legacy/Web/ImagesController.php
+++ b/app/Http/Controllers/Legacy/Web/ImagesController.php
@@ -34,7 +34,7 @@ class ImagesController extends Controller
         $this->aws = $aws;
         $this->fastly = $fastly;
 
-        $this->middleware('throttle:30');
+        $this->middleware('throttle:60');
     }
 
     /**
--- a/app/Http/Controllers/Legacy/Web/ImagesController.php
+++ b/app/Http/Controllers/Legacy/Web/ImagesController.php
@@ -2,7 +2,6 @@
 
 namespace Rogue\Http\Controllers\Legacy\Web;
 
-use Carbon\Carbon;
 use Rogue\Models\Post;
 use Rogue\Services\AWS;
 use Rogue\Services\Fastly;
@@ -47,11 +46,6 @@ class ImagesController extends Controller
      */
     public function show(Post $post, Request $request)
     {
-        if (! config('features.glide')) {
-            abort(501, 'Glide image URLs are not enabled in this environment.');
-        }
-
-        // Create the Glide server.
         $server = ServerFactory::create([
             'response' => new LaravelResponseFactory($request),
             'cache' => new Flysystem(new MemoryAdapter()),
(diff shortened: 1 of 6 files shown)


app/Http/Controllers/Legacy/Web/ImagesController.php
```diff
@@ -35,7 +34,7 @@ public function __construct(Filesystem $filesystem, AWS $aws, Fastly $fastly)
         $this->aws = $aws;
         $this->fastly = $fastly;
 
-        $this->middleware('throttle:30');
+        $this->middleware('throttle:60');
```
katiecrane: YESSSSSSSS

DFurnes:
I added a rather hefty commit to rename `AWS` to `ImageStorage` in d50f5f3, since we use that class to store images on the local disk as well (for development) so it's kinda confusing to call it AWS. Plus, there are plenty of other AWS services we use that _don't_ live in that class! 🙃 

I've also noticed that we _do_ in fact [rely on `storeImageData` in `ImagesController`](https://github.com/DoSomething/rogue/blob/5a03b0d1d1712c550066f1d05416d2d985ed053a/app/Http/Controllers/Legacy/Web/ImagesController.php#L107), so I'll need to do a little more work tomorrow to make sure I don't accidentally break that "rotate image" endpoint as part of this cleanup.

app/Http/Controllers/Legacy/Web/ImagesController.php
```diff
@@ -74,75 +69,33 @@ public function show(Post $post, Request $request)
     /**
      * Edits and overwrites an image based on given request parameters.
      *
-     * @TODO - Currently rotates and overwrites both the original image and
-     * processed, edited image. We will not need to work with the edited image
-     * when Glide processing is turned on, so we should remove this logic when
-     * that is live on prod.
-     *
```
katiecrane: `@TODONE`

app/Repositories/PostRepository.php
```diff
@@ -3,47 +3,30 @@
 namespace Rogue\Repositories;
 
 use Rogue\Models\Post;
-use Rogue\Services\AWS;
 use Rogue\Models\Action;
 use Rogue\Models\Review;
 use Rogue\Models\Signup;
-use Rogue\Services\Registrar;
```
katiecrane: Wow this wasn't being used like at all, good catch!

PR #917, opened 2019-10-15
More cleanup!

DFurnes:
#### What's this PR do?
This pull request includes some more cleanup. 🍃

Commit-by-commit! 👁

This is just some more follow-up from #916. My goal is to remove more "old" unused functionality to simplify things & continue to unify places that we use different approaches.

#### Relevant tickets
N/A

### DIFF
--- a/app/Http/Controllers/TagsController.php
+++ b/app/Http/Controllers/TagsController.php
@@ -41,17 +41,15 @@ class TagsController extends ApiController
     /**
      * Updates a post's tags when added or deleted.
      *
-     * @param  \Illuminate\Http\Request $request
      * @param  Post $post
+     * @param  \Illuminate\Http\Request $request
      * @return \Illuminate\Http\Response
      */
-    public function store(Request $request, Post $post)
+    public function store(Post $post, Request $request)
     {
         $request->validate([
             'tag_name' => 'required|string',
         ]);
-
-        $post = $this->post->find($post->id);
 
         // If the post already has the tag, remove it. Otherwise, add the tag to the post.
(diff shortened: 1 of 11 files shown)


app/Services/Fastly.php
```diff
@@ -2,52 +2,51 @@
...
         }
 
-        $service = config('services.fastly.service_id');
-
-        $purgeResponse = $this->post('service/'.$service.'/purge/'.$cacheKey);
-
+        $purgeResponse = $this->post('service/'.$this->service.'/purge/'.$cacheKey);
         info('image_cache_purge_successful', ['response' => $purgeResponse]);
-
-        return $purgeResponse;
```
katiecrane: Do we not want to return anything here? The result here is also what's being returned by `purge()` right?

app/Http/Controllers/TagsController.php
```diff
@@ -41,18 +41,16 @@ public function __construct(PostRepository $post)
...
      * @return \Illuminate\Http\Response
      */
-    public function store(Request $request, Post $post)
+    public function store(Post $post, Request $request)
     {
         $request->validate([
             'tag_name' => 'required|string',
         ]);
 
-        $post = $this->post->find($post->id);
```
katiecrane: Love this edit!!

app/Services/Fastly.php
DFurnes: Since we never use this API response in our app logic, I decided not to return it. I updated these to be `void` functions in 2d22da98 to make this a little bit more straightforward.

PR #918, opened 2019-10-16
Add 'originals/{post}' route.

DFurnes:
#### What's this PR do?
This pull request adds an `originals/{post}` endpoint which allows staffers (or the post's owner) to download the original file that they uploaded. This prevents rando's from downloading a user's high-resolution uploads, and also lets us lock down the S3 bucket since it won't be receiving any public traffic anymore.

I did some refactoring of authorization logic in 01b4122 to make it easier to read. Then 3efd318 adds this new endpoint (which just proxies the unmodified S3 file if the user is authorized to see it).

🍃

#### Relevant tickets
[#165980718](https://www.pivotaltracker.com/story/show/165980718)

#### How to deploy


app/helpers.php
```diff
@@ -93,17 +93,28 @@ function getNorthstarId($request)
...
  */
-function is_staff_user()
+function is_staff_user() : bool
 {
-    $isStaffUser = token()->exists() && in_array(token()->role, ['admin', 'staff']);
-    $isStaffClient = token()->exists() && in_array('admin', token()->scopes());
-
-    if ($isStaffUser || $isStaffClient) {
+    // If this is a machine client, then it's de-facto an admin:
+    if (token()->exists() && ! token()->id()) {
```
katiecrane: Is there any way this could be abused?

DFurnes: I don't believe so. The only way you should be able to get a token without an ID attached is via the client credentials grant, which is only used by internal "machine" applications.

routes/web.php
```diff
@@ -42,3 +42,4 @@
 
 // Images
 $router->post('images/{postId}', 'Legacy\Web\ImagesController@update');
+$router->get('originals/{post}', 'Legacy\Web\OriginalsController@show');
```
katiecrane: Is there a reason we want this new route to be in `Legacy`?

DFurnes: We really don't! 😩 I need to move these into just a plain ol' `Web/` namespace...

app/helpers.php
katiecrane: Someone couldn't delete the ID part from their token then? It all comes bundled together?

routes/web.php
katiecrane: Gotcha, sounds out of scope for this PR!

app/helpers.php
DFurnes: Ah nope! If the user were to modify [their token](https://github.com/DoSomething/northstar/blob/master/documentation/authentication.md#access-tokens), the signature would become invalid & refused by the server.

katiecrane: 🔒

PR #920, opened 2019-10-16
Use short hashes for image URLs.

DFurnes:
#### What's this PR do?
This pull request updates our `images/{post}` endpoint to use a deterministic (but hard to guess) hash based on the post's ID, using [`hashids`](https://hashids.org/php/). This prevents a curious user from looking at unapproved post's images (while still letting us cache those in Fastly).

👀

We can't simply apply authorization logic to the `images/{post}` endpoint (like we do for originals in #918) because we want to cache these generated Glide responses in Fastly (and so most requests won't ever hit our servers). That means we want the image to be public but not easily enumerable.

Users are only able to view `posts/{id}` once a post has been approved (or if it's theirs), and so they would not be able to guess the URL for `images/{hash}` until then. This allows admins and users themselves to see unapproved posts without letting anyone else find them until they're in the gallery.

This should be a safe change since we only reference these images based on the post's `url` (rather than trying to create the URLs by hand). However, I'll be sure to give a heads up in #deploys before pushing this up to production. 💬

#### Relevant tickets
[#169188892](https://www.pivotaltracker.com/story/show/169188892)

#### How to deploy

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -181,7 +181,8 @@ class Post extends Model
      *
      * @return string
      */
-    public function getHashAttribute() {
+    public function getHashAttribute()
+    {
         return app(Hashids::class)->encode($this->id);
     }
 
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Models;
 
+use Hashids\Hashids;
 use Rogue\Services\GraphQL;
 use Rogue\Events\PostTagged;
 use Illuminate\Support\Facades\Cache;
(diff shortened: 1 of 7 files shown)


app/Models/Post.php
```diff
@@ -167,7 +195,7 @@ public function getMediaUrl()
             return null;
         }
 
-        return url('images/' . $this->id);
+        return url('images/' . $this->hash);
```
katiecrane: I would have expected an update for when we first store the image to use this as well?

DFurnes: Whatcha mean?

katiecrane: Great question let me try to rephrase - when do we store the image at `'images/' . $this->hash`?

DFurnes: Ah, we don't! This is our [Glide](https://glide.thephpleague.com) route that renders images for use in galleries (or wherever). It used to use the post's ID to load the database record (and from that, the actual S3 file path), but this PR swaps it to use this predictable hash instead so that someone can't just iterate over every image and find unapproved ones.

DFurnes: Correct!

katiecrane: ohhhhhhhh 💡

PR #923, opened 2019-10-18
Update "user index" page to use React and GraphQL.

DFurnes:
#### What's this PR do?
This pull request updates Rogue's "user index" page to render via our new "modern" application shell & run searches via GraphQL (using the new `users` query introduced in DoSomething/graphql#140).

It also adds a new `app.blade.php` view that renders the `#app` container and has a little nicer "loading" placeholder than a blank page. 😇

👀

I'm knocking these out first because they're low-hanging fruit & should make it easier to tackle the more complicated elements (like the campaign inbox). 🍒

#### Relevant tickets

#### How to deploy


resources/assets/components/UserIndex.js
```diff
@@ -0,0 +1,109 @@
...
+};
+
+/**
+ * The user index!
+ *
+ * @param {string} term
+ */
+const UserIndex = () => {
+  const [term, setTerm] = useState(null);
+  const inputElement = useRef(null);
```
DFurnes: Here's [documentation for the `useRef` hook](https://reactjs.org/docs/hooks-reference.html#useref) if you're interested!

app/Http/Controllers/Web/UsersController.php
```diff
@@ -1,74 +0,0 @@
...
-        // Attempt to fetch all users.
-        $users = $this->registrar->search($query);
-
-        if ($users->count() === 0) {
-            return redirect()->route('users.index')->with('status', 'No user found!');
-        }
-
-        return redirect()->route('users.show', [$users->first()->id]);
-    }
-}
```
katiecrane: All the logic that was in here got moved to the frontend?

resources/assets/components/PagePlaceholder.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+const PagePlaceholder = () => (
+  <div className="placeholder flex-center-xy">
+    <div className="spinner" />
```
katiecrane: For a future PR I'd love to see Tongue Cat incorporated here somehow 😝

resources/assets/components/UserIndex.js
katiecrane: > Essentially, useRef is like a “box” that can hold a mutable value in its .current property.

a-ha! thanks for including the link 😸

app/Http/Controllers/Web/UsersController.php
DFurnes: Yep!

resources/assets/components/PagePlaceholder.js
DFurnes: yessss 👅

PR #924, opened 2019-10-21
Add `filter[is_open]` to Campaigns API.

DFurnes:
#### What's this PR do?
This pull request allows us to filter the `v3/campaigns` API via the computed `is_open` field, using new `whereOpen` and `whereClosed` query scopes. This will allow us to group open vs. closed campaigns on the "campaign overview" page without loading all 700+ campaigns in-memory.

Here's how this would look via the REST API:

```
https://activity.dosomething.org/api/v3/campaigns?filter[is_open]=true
```

🚥 👀

📂

#### Relevant tickets

#### How to deploy

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -33,6 +33,18 @@ class CampaignsController extends ApiController
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Campaign::$indexes);
 
+        // Apply scope for the "computed" is_open field:
+        if (isset($filters['is_open'])) {
+            // We can use this to find only open campaigns with a truthy filter
+            // applied, e.g. `?filter[is_open]=true`, or closed campaigns with
+            // a falsy filter, e.g. `?filter[is_open]=false`.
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/CampaignsController.php
```diff
@@ -33,6 +33,18 @@ public function index(Request $request)
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Campaign::$indexes);
 
+        // Apply scope for the "computed" is_open field:
+        if (isset($filters['is_open'])) {
+            // We can use this to find only open campaigns with a truthy filter
+            // applied, e.g. `?filter[is_open]=true`, or closed campaigns with
+            // a falsy filter, e.g. `?filter[is_open]=false`.
```
weerd: Curious if helpful at all to just have a corresponding `is_closed` filter, just so it's available if preferred to read it as such? Defs not a blocker and not needed for now, but could be a nice to have 🤷‍♂

DFurnes: Yeah, that's an interesting idea. 📕

PR #927, opened 2019-10-24
Use GraphQL for data fetching on campaign index.

DFurnes:
#### What's this PR do?
This pull request updates the "Campaign Overview" index to use GraphQL for data fetching & hooks for state management. I've also consolidated the `/campaign-ids` index into this view, since they're rather redundant otherwise.

Screen Shot 2019-11-01 at 12 35 05 PM

It's probably worth taking a peek at DoSomething/graphql#142 and DoSomething/rfcs#6 if you haven't already for background on the new pagination approach. Then I'd recommend viewing the whole diff, since this is effectively a full re-write.

🍃

#### Relevant tickets

#### How to deploy


DFurnes:
Oooof, meant to make this a draft PR! Please ignore for now. 😇

DFurnes:
This is ready for review!! 🎉 (Just waiting on DoSomething/graphql#147 before it's mergeable).

resources/assets/components/CampaignsTable.js
```diff
@@ -0,0 +1,163 @@
...
+  const search = filter.toLowerCase();
+
+  if (!data) {
+    return [];
+  }
+
+  return data.campaigns.edges.filter(campaign => {
+    if (search === '') {
+      return true;
+    }
```
mendelB: Not sure if it makes enough of a difference -- but do you think if we just returned the `edges` array point-blank on top of this filter method (or only call `filterCampaigns` if there's a a `filter`), we'd avoid having to filter through the campaigns by default even when there's no `search`? (or is that such a cheap process that it doesn't matter 🤷‍♂ )

resources/assets/components/CampaignsTable.js
```diff
@@ -0,0 +1,163 @@
...
+      cause.name.toLowerCase().includes(search),
+    );
+
+    return matchesId || matchesTitle || matchesCause;
+  });
+};
+
+/**
+ * This component handles fetching & paginating a list campaign
+ * of campaigns matching the given filter(s).
```
mendelB: 👾 *A list of campaigns

resources/assets/components/CampaignsTable.js
```diff
@@ -0,0 +1,163 @@
...
+      updateQuery,
+    });
+  };
+
+  // If we've filtered all results & can load more, do so automatically:
+  useEffect(() => {
+    if (noFilteredResults && hasNextPage) {
+      handleViewMore();
+    }
+  }, [filter, endCursor]);
```
mendelB: So this will keep loading campaigns until we find a match for our search? What if the first found match is not the intended result, but we stop auto-loading campaigns? (I guess you could still manually press 'view more' until you find it?)
Related -- are we worried at all about this loading _all_ campaigns for completely unmatched searches?

resources/assets/helpers.js
```diff
@@ -281,3 +289,19 @@ export function env(key) {
...
+ */
+export const updateQuery = (previous, { fetchMoreResult }) => {
+  return mergeWith({}, previous, fetchMoreResult, (dest, src) => {
+    // By default, Lodash's `merge` would try to merge *each* array
+    // item (e.g. `edges[0]` with then next page's `edges[0]`).
+    if (isArray(dest)) {
+      return [...dest, ...src];
+    }
+  });
+};
```
mendelB: @DFurnes Would you mind articulating how this works? 😖 sorry. I'm having trouble grokking this offhand

resources/assets/components/CampaignsTable.js
DFurnes: Oh good call! It definitely can't hurt to avoid doing we don't have to. 😄

DFurnes: Updated in bc832f4.

resources/assets/components/CampaignsTable.js
DFurnes: Oof, good eye! Fixed in 1d1ef8f.

resources/assets/components/CampaignsTable.js
DFurnes: >So this will keep loading campaigns until we find a match for our search?

Yep, this will only automatically load more if the given filter has no results on what we've loaded so far (to save awkwardly pressing "view more" a bunch of times if the target campaign is near the bottom of the list).

> What if the first found match is not the intended result, but we stop auto-loading campaigns? (I guess you could still manually press 'view more' until you find it?)

Yep! We could also just keep on loading pages regardless, but my hunch is that usually folks are looking for one of the campaigns near the top of the list (with a lot of pending posts).

> Related -- are we worried at all about this loading all campaigns for completely unmatched searches?

I'll try to check in on how this is performing for folks after a while in production, but I don't think that's too big a deal if so since this query performs reasonably. And if we ever get around to setting up something like Algolia for search, that'd be a great way to make this a lot smoother.

resources/assets/helpers.js
DFurnes: Yeah it's kinda wonky! By default, Lodash's `merge` has a kinda weird way of merging arrays (it'll merge each element in the array with the corresponding one in the new array). So, by default, we'd do something like this:

```js
merge([item1, item2, item3], [item4, item5, item6]);
// => [merge(item1, item4), merge(item2, item5), merge(item3, item6)];
```

That's obviously not what we want here though! 😂 This "customizer" function just overrides Lodash's "merge strategy" for arrays, telling it to instead concatenate them:

```js
merge([item1, item2, item3], [item4, item5, item6]);
// => [item1, item2, item3, item4, item5, item6];
```

mendelB: Aha lol, fair, thanks for clarifying! (And I should have taken a peak at [the documentation](https://lodash.com/docs/4.17.15#mergeWith) for `mergeWith` where this is literally their provided example 😅(though they're using `Arrya.concat`, any specific reason you diverged, or just cuz?))

resources/assets/components/CampaignsTable.js
mendelB: Makes sense! Yeah I can totally hear @weerd's voice ringing out with the call for Algolia which would be super neat here!

resources/assets/helpers.js
DFurnes: I just thought that spreading the arrays read a little clearer (since with `Array.concat` it's kinda unclear which is being concatenated to which at first glance).

PR #930, opened 2019-10-28
Add `cursor` and experimental `?after=` query param to Campaigns API.

DFurnes:
#### What's this PR do?
This pull request adds some API support for DoSomething/graphql#142:

👇 Adds a `cursor` field to the campaign transformer, which is just the Base-64 encoded ID of the campaign. This is suggested in GraphQL's documentation, and makes sense to me:

> Using cursors for pagination gives additional flexibility if the pagination model changes in the future. As a reminder that the cursors are opaque and that their format should not be relied upon, we suggest base64 encoding them.

↪️ Adds an experimental `?after=` query parameter to `/api/v3/campaigns`, which accepts the cursor that we generate in the transformer. This allows us to say "get me all the items after this one" and sidestep some bugs we could encounter with traditional `?page=#` pagination.

tell us, how can we review, to test that this works

what else?

#### Relevant tickets
Fixes #

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Traits/FiltersRequests.php
+++ b/app/Http/Controllers/Traits/FiltersRequests.php
@@ -83,6 +83,11 @@ trait FiltersRequests
             if (in_array($column, $indexes)) {
                 $query = $query->orderBy($column, $direction);
             }
+        } else {
+            // If we don't specify an ordering in the query, we should default
+            // to order by ID. (Fun fact: MySQL makes no guarantees of ordering
+            // if we don't include this in the query!)
+            $query->orderBy('id', 'asc');
         }
 
         return $query;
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/Traits/FiltersRequests.php
```diff
@@ -83,6 +83,11 @@ public function orderBy($query, $orderBy, $indexes)
             if (in_array($column, $indexes)) {
                 $query = $query->orderBy($column, $direction);
             }
+        } else {
+            // If we don't specify an ordering in the query, we should default
+            // to order by ID. (Fun fact: MySQL makes no guarantees of ordering
+            // if we don't include this in the query!)
+            $query->orderBy('id', 'asc');
```
katiecrane: A smart addition!

PR #931, opened 2019-10-31
Fixes an issue with cursor pagination & sorted results.

DFurnes:
#### What's this PR do?
This pull request fixes an issue with cursor pagination (introduced in #930). That implementation didn't take into account results that have `orderBy` (and so the ID alone wouldn't get us the proper position in the sorted results).

I did some refactoring in 9313d2f, 01059be, and 9af53c4. I added a passing test for `?after=` with a standard non-sorted response in 582b69e and a failing test in 04231d5. Finally, I fixed the issue with using `after` & `orderBy` together in c8749e7.

Finally, I renamed `after` to `cursor[after]` for clarity in 5792139.

This [Shopify Engineering blog post](https://engineering.shopify.com/blogs/engineering/pagination-relative-cursors) includes a good explanation & visualization of this issue under the "Sorting and Skipping Records" heading. (Cursor pagination is also used by Facebook, [GitHub](https://developer.github.com/v4/object/repository/), and [Slack](https://slack.engineering/evolving-api-pagination-at-slack-1c1f644f8e12)).

#### Relevant tickets
[#169352891](https://www.pivotaltracker.com/story/show/169352891)

#### How to deploy

### DIFF
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -217,7 +217,20 @@ class Campaign extends Model
      */
     public function getCursor()
     {
-        return base64_encode($this->id);
+        $sortCursor = '';
+
+        // If we're ordering results, we need to include that column's
+        // value in the cursor (so we can say "get me things after this"):
+        if ($orderBy = request()->query('orderBy')) {
+            [ $column, $direction ] = explode(',', $orderBy, 2);
+
+            // Check that it's okay to expose this column in the cursor:
+            if (in_array($column, self::$sortable)) {
+                $sortCursor = '.' . $this->{$column};
+            }
+        }
+
+        return base64_encode($this->id . $sortCursor);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


app/Models/Campaign.php
```diff
@@ -199,4 +209,56 @@ public function setArbitraryDateString($attribute, $value)
...
+        $sortCursor = isset($cursor[1]) ? $cursor[1] : null;
+
+        $orderBy = request()->query('orderBy');
+        if ($orderBy && $sortCursor) {
+            // If we're sorting by a column, things get a lil' tricky:
+            [ $column, $direction ] = explode(',', $orderBy, 2);
+
+            // Check that we're allowed to sort by this column:
+            if (in_array($column, self::$sortable)) {
+                $query->where($column, '<', $sortCursor)
```
mendelB: Am I misunderstanding or does the `'<'` assume that the sort is _descending_?

DFurnes: Whoops! It sure does. 🤦‍♂ Fixed this up to work either way in 2699887c.

DFurnes:
I've overridden the `laravel` preset to continue to use `length_ordered_imports` in 16af869, which clears up those style warnings by keeeping the same rule we'd inherited from the preset before.

PR #933, opened 2019-11-05
Fix issues with rotation API.

DFurnes:
#### What's this PR do?
This pull request updates our image rotation endpoint to live under the API namespace and require the `staff` or `admin` role (previously anyone could rotate posts). 🔒 I've tested this with the local filesystem driver & S3 driver and both seem to work swimmingly now.

👀

I have not updated Rogue's admin interface to use this new `/api/v3/posts/{id}/rotate` endpoint, since I'm going to be swapping these out to use the new GraphQL powered component very soon.

#### Relevant tickets
[#169346181](https://www.pivotaltracker.com/story/show/169346181)

#### How to deploy

### DIFF
--- a/app/Services/ImageStorage.php
+++ b/app/Services/ImageStorage.php
@@ -6,6 +6,7 @@ use Rogue\Models\Post;
 use Intervention\Image\Image;
 use Illuminate\Support\Facades\Storage;
 use Symfony\Component\HttpFoundation\File\File;
+use Intervention\Image\Facades\Image as ImageManager;
 use Symfony\Component\HttpKernel\Exception\HttpException;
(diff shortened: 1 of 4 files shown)


app/Services/ImageStorage.php
```diff
@@ -17,6 +18,19 @@ class ImageStorage
...
+     * Get image for the given post.
+     *
+     * @param Post $post
+     * @return Image
+     */
+    public function get(Post $post)
+    {
+        $contents = Storage::get($post->getMediaPath());
+
+        return ImageManager::make($contents);
```
weerd: 👌

PR #935, opened 2019-11-05
Add "individual post" permalink (powered by GraphQL).

DFurnes:
#### What's this PR do?
This pull request adds a `post/{id}` view that shows an individual post. This interface fetches (and mutates!) data via GraphQL, allowing us to sidestep some of the complexity that's built up around doing this with REST & a custom in-memory store.

Kapture 2019-11-05 at 17 47 53

This relies on the mutations and field added in DoSomething/graphql#153. I've split this into individual commits for each feature, which may be helpful while reviewing.

This adds a feature I've wanted for a long time (since otherwise the best we can do is link to a signup and say "look at the Nth one") and also makes it super-simple to build the campaign inbox!

#### Relevant tickets
[#169512736](https://www.pivotaltracker.com/story/show/169512736)

### DIFF
--- a/app/Http/Controllers/TagsController.php
+++ b/app/Http/Controllers/TagsController.php
@@ -51,10 +51,10 @@ class TagsController extends ApiController
             'tag_name' => 'required|string',
         ]);
 
-        // If a tag slug is sent in, change to the tag name.
+        // If a tag slug is sent in (dashed or lowercase), change to the tag name.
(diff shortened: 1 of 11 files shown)


DFurnes:
This is ready for review! I'll add support for showing post "sibling" thumbnails and displaying the modal to edit a post's quantity in a follow-up PR.

PR #936, opened 2019-11-06
Add "edit quantity" modal to GraphQL-backed post interface.

DFurnes:
#### What's this PR do?
This pull request adds the "edit quantity" modal to the GraphQL-powered post interface, which can currently be found at `/posts/:id`. This uses [react-useportal](https://github.com/alex-cory/react-useportal) to handle rendering the modal, and the new `updatePostQuantity` GraphQL mutation to update this field on the backend. (f4c8d2a)

It also includes two tiny fixes to the "rotate" functionality:

💤 Adds a (hook-powered) `LazyImage` component which lets us lazy-load images! f968ad7

💣 Removes the `Expires` header from Glide images so browsers don't cache them! I missed this in #934 but hopefully this should sort everything out for us! 5b77545

👀

This gets us pretty much to feature-parity with the old implementation, nice!

#### Relevant tickets
[#169512736](https://www.pivotaltracker.com/story/show/169512736), [#169346181](https://www.pivotaltracker.com/story/show/169346181)

#### How to deploy

### DIFF
--- a/resources/assets/components/utilities/QuantityForm.js
+++ b/resources/assets/components/utilities/QuantityForm.js
@@ -1,0 +1,74 @@
+import gql from 'graphql-tag';
+import classNames from 'classnames';
+import React, { useState } from 'react';
+import { useMutation } from '@apollo/react-hooks';
+
+export const QuantityFormFragment = gql`
+  fragment QuantityForm on Post {
+    id
+    quantity
+    impact
+  }
+`;
+
+const UPDATE_QUANTITY_MUTATION = gql`
+  mutation UpdateQuantityMutation($id: Int!, $quantity: Int!) {
+    updatePostQuantity(id: $id, quantity: $quantity) {
+      ...QuantityForm
+    }
+
+    ${QuantityFormFragment}
+  }
+`;
+
+const isInteger = string => /^\d+$/.test(string);
+
+const QuantityForm = ({ post }) => {
+  const [newQuantity, setNewQuantity] = useState(post.quantity);
+  const [updateQuantity, { loading }] = useMutation(UPDATE_QUANTITY_MUTATION, {
+    variables: { id: post.id },
+  });
+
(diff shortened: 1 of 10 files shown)


resources/assets/components/utilities/QuantityForm.js
```diff
@@ -0,0 +1,74 @@
...
+  const [newQuantity, setNewQuantity] = useState(post.quantity);
+  const [updateQuantity, { loading }] = useMutation(UPDATE_QUANTITY_MUTATION, {
+    variables: { id: post.id },
+  });
+
+  const handleSubmit = () =>
+    updateQuantity({ variables: { quantity: Number(newQuantity) } });
+
+  return (
+    <>
```
katiecrane: I don't think I've seen these empty tags used before, can you do this as a wrapper so you're still returning something valid?

katiecrane:
I am curious what this looks like! Is it exactly the same?

resources/assets/components/utilities/QuantityForm.js
DFurnes: Yep! These are just a "short syntax" for [React fragments](https://reactjs.org/docs/fragments.html).

DFurnes:
Pretty much! I just didn't re-create the "it'll be back soon" for the history log, since… 😅 

#### Before:

Screen Shot 2019-11-07 at 10 52 59 AM


#### After:

Screen Shot 2019-11-07 at 10 53 35 AM

resources/assets/components/utilities/QuantityForm.js
katiecrane: Nice! That was my guess but I only remember seeing the full `<React.Fragment>` used before. I like this style better!

katiecrane:
Oh rad! I missed that it's like a number input now, that's really cool!

PR #937, opened 2019-11-07
Add 'cursor[after]' support to 'api/v3/posts'.

DFurnes:
#### What's this PR do?
This pull request extracts the experimental cursor methods from #930 into a `HasCursor` trait so that we can try it out for posts as well. I also ran into a bug with the "sorted" cursor logic and fixed it up in b262123 (see below).

I'd recommend reviewing this one commit-by-commit (mainly because extracting the trait adds a lot of noise that's not "really" there).

The bug in c6ff2ba occurs when the "sorted" column (e.g. `created_at` or `pending_posts`) has items with the same value. In that case, we want either items after the primary sort column (so `created_at < cursor.created_at` for example) or, if the primary sort has duplicates, after the secondary sort (so `created_at == cursor.created_at && id > cursor.id`).

Previously, I'd been checking for `created_at <= cursor.created_at && id > cursor.id` which applies the secondary sort in cases where we don't want it.

(I believe the reason this bug didn't appear with campaigns in #930 because we only ran this "secondary sort" logic at the "end" of the list when we were sorting by `pending_count,desc`, and so there was no opportunity for the `id > cursor.id` to mess things up).

Anyways, [yikes](https://media.giphy.com/media/5QMPpCzH6yxEqhev5A/giphy.gif).

#### Relevant tickets

#### How to deploy

### DIFF
--- a/app/Http/Controllers/PostsController.php
+++ b/app/Http/Controllers/PostsController.php
@@ -70,8 +70,7 @@ class PostsController extends ApiController
      */
     public function index(Request $request)
     {
-        $query = $this->newQuery(Post::class)
-            ->orderBy('created_at', 'desc');
+        $query = $this->newQuery(Post::class);
 
         if (has_include($request, 'signup')) {
             // Eagerly load the `signup` relationship.
@@ -95,6 +94,22 @@ class PostsController extends ApiController
         // If the northstar_id param is passed, only allow admins, staff, or owner to see anonymous posts.
         if (array_has($filters, 'northstar_id')) {
             $query = $query->withoutAnonymousPosts();
+        }
+
+        // This endpoint always returns posts in reverse chronological order. We'll
+        // therefore "force" the query string so that we can use it in `getCursor`.
+        // @TODO: There must be a more elegant way of doing this...
+        $query->orderBy('created_at', 'desc')
+            ->orderBy('id', 'asc');
+
+        $request->query->set('orderBy', 'created_at,desc');
+
+        // Experimental: Allow paginating by cursor (e.g. `?cursor[after]=OTAxNg==`):
+        if ($cursor = array_get($request->query('cursor'), 'after')) {
+            $query->whereAfterCursor($cursor);
+
+            // Using 'cursor' implies cursor pagination:
+            $this->useCursorPagination = true;
         }
 
         return $this->paginatedCollection($query, $request);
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/PostsController.php
```diff
@@ -97,6 +96,22 @@ public function index(Request $request)
             $query = $query->withoutAnonymousPosts();
         }
 
+        // This endpoint always returns posts in reverse chronological order. We'll
+        // therefore "force" the query string so that we can use it in `getCursor`.
+        // @TODO: There must be a more elegant way of doing this...
+        $query->orderBy('created_at', 'desc')
+            ->orderBy('id', 'asc');
```
katiecrane: Why are there two `orderBy`s? In case there are identical timestamps on `created_at`?

DFurnes: Yep!

app/Models/Post.php
```diff
@@ -57,6 +59,16 @@ class Post extends Model
...
 
+    /**
+     * Attributes that can be sorted by.
+     *
+     * This array is manually maintained. It does not necessarily mean that
+     * any of these are actual indexes on the database... but they should be!
+     *
+     * @var array
+     */
+    public static $sortable = ['created_at'];
```
katiecrane: Aren't you sorting by `id` as well?

DFurnes: This one's a little awkward, because we don't actually choose `orderBy` for this endpoint! But on other endpoints, like campaigns, we'd use this array to decide what `?orderBy=` values are allowed (and then, to decide whether or not we're safe to encode a given columns value into the `cursor` value in a transformer).

DFurnes: I bet there's a clearer way of specifying this (especially with the awkward forced query-string on this endpoint), but I haven't stumbled upon it yet. Let me know if anything comes to mind!

DFurnes: So we'd include `id` if it was a primary sort option (available to `?orderBy=`), but since it's just our hard-coded secondary sort that we use alongside whatever is chosen as the primary column, it doesn't hit this logic.

katiecrane: AH I think for me this would make more sense to me if the comment said something like `Attributes that can be sorted by using the ?orderBy query parameter`

DFurnes: Yeah, that's way clearer! ✏️

PR #938, opened 2019-11-08
Render campaign inbox with GraphQL.

DFurnes:
#### What's this PR do?
This pull request updates the "inbox" view to use GraphQL for data fetching & mutations. You can see the meat of the work in 66b9a2a, and luckily it largely gets to ride on the coat-tails of the work done for #935! The only additional work is showing a [paginated list of posts](https://github.com/DoSomething/graphql/pull/157), rather than just one.

Some additional clean-up that's also included:

⚙️ Adds `error` and `loading` states to the __Chrome__ component, since we tend to repeat the same thing in all these different views. 631630d

⌛️ The review buttons will now show a spinner while waiting for the mutation to finish. c36afe6

🌱 Fixes an issue where the ModelFactory's `cause` seeder would set `cause => [1, 2]` rather than the actual causes, e.g. `cause => ['bullying', 'education']`. a3c6330

I'd recommend reviewing commit-by-commit this time!

> The only additional work is showing a [paginated list of posts](https://github.com/DoSomething/graphql/pull/157), rather than just one.

Woah... pagination!? It's true! 🌟 

This is a feature we weren't able to implement without [cursor pagination](https://github.com/DoSomething/rfcs/pull/6), since fetching a second page of pending posts would give inconsistent results (since if you'd reviewed some or all of the first page, a "one page" offset on the "pending posts" query would either skip posts or show duplicates).

#### Relevant tickets


resources/assets/components/UserIndex.js
```diff
@@ -23,12 +22,20 @@ const USER_SEARCH_QUERY = gql`
...
+  const subtitle = 'User profiles &amp; signups...';
+
   const { loading, error, data } = useQuery(USER_SEARCH_QUERY, {
     variables: { term },
   });
 
-  if (loading) return <div className="spinner" />;
-  if (error) return <span>Error :(</span>;
+  if (loading) {
+    return <Shell title={title} subtitle={subtitle} loading />;
```
katiecrane: I don't see `Shell` imported or a component by that name in Rogue, is this something built in to React?

resources/assets/components/utilities/Chrome.js
```diff
@@ -14,4 +14,30 @@ const Chrome = ({ title, subtitle, children }) => (
   </div>
 );
 
+const Chrome = ({ title, subtitle, children, error, loading }) => {
```
katiecrane: I've never understood why stuff is called Chrome and why there is a browser called Chrome too and it's ungoogleable because I just get results about the browser! Can you explain this to me?

DFurnes: Oh haha, yeah that's fair! 😆 It's a [user interface term](https://stackoverflow.com/a/5072092) (and I think at least [partially the namesake for the browser](https://www.quora.com/Why-is-Google-Chrome-browser-named-as-Chrome), but it is kinda confusing now! We could call this `Shell` or `Layout` perhaps to be clearer.

resources/assets/components/UserIndex.js
DFurnes: Whoops, forgot to update that to be `Chrome`!

resources/assets/components/utilities/Chrome.js
katiecrane: OOOOO finally, the mystery is solved. Thanks! I dig `Shell` because it seems more descriptive to me, but a potential change for another day!

DFurnes: I renamed to `Shell` in e44e5b53! 🐚

resources/assets/components/UserIndex.js
DFurnes: Renamed this component to `Shell` in e44e5b53 so this now works! 😁

DFurnes:
I found one more place where we'd been duplicating "shell" markup in the component and updated it to use the `Shell` utility component in 55af3131048cb7094e09bed875ae4d2f7e1ea24e!

PR #941, opened 2019-11-11
Fix Fastly caching issues.

DFurnes:
#### What's this PR do?
This pull request… hopefully… fixes the caching issues I've been wrestling with today.

I think the easiest thing will just be testing this on QA! 🙃

Frustratingly, Symfony (which Laravel's HTTP stack is built on) [forces certain `Cache-Control` headers](https://github.com/symfony/http-foundation/blob/d8f1bf9f099852ddd0c009589634ecadb7e4a9d8/ResponseHeaderBag.php#L255-L285), so we can't follow [Fastly's instructions](https://docs.fastly.com/en/guides/cache-control-tutorial#applying-different-cache-rules-for-fastly-and-browsers) without it being overridden with a `private` at the end (which disables Fastly caching).

I'm still not certain why #940 didn't solve our issues, but I've tested these headers in a [Fastly Fiddle](https://fiddle.fastlydemo.net/fiddle/bd6771fc) and confirmed that given the headers that are being output by my local app… this _should_ correctly cache in Fastly & the user's browser.

#### Relevant tickets
[#169690043](https://www.pivotaltracker.com/story/show/169690043)

#### How to deploy

### DIFF
--- a/app/Http/Controllers/Web/ImagesController.php
+++ b/app/Http/Controllers/Web/ImagesController.php
@@ -63,9 +63,9 @@ class ImagesController extends Controller
         $response->headers->set('Surrogate-Control', 'max-age=31536000');
         $response->headers->set('Surrogate-Key', 'post-'.$post->id);
 
-        // But we want browsers to always re-validate before serving this image from local
-        // cache, so that image rotations/deletions are respected immediately:
-        $response->headers->set('Cache-Control', 'no-cache, public');
+        // And we want browsers to cache but re-validate with us before returning cached content
+        // on a subsequent load, so that image rotations/deletions are respected immediately:
+        $response->headers->set('Cache-Control', 'max-age=31536000, must-revalidate, no-cache, public');
 
         return $response;
     }


DFurnes:
I deployed this to QA to double-check that this works & can confirm we're seeing the right behavior:

```
$ curl -svo /dev/null https://activity-qa.dosomething.org/images/bj0OVZWz7p -H "Fastly-Debug: 1"

< Cache-Control: no-cache, public
< Expires: Wed, 11 Nov 2020 19:13:32 GMT
< Surrogate-Control: max-age=31536000
< Surrogate-Key: post-1331620
< Fastly-Debug-TTL: (H cache-ewr18122-EWR 31535989.345 0.000 11) (M cache-iad2140-IAD - - 0)
< X-Cache: MISS, HIT
```

Reloading the browser loads from local cache, until an image is rotated at which point it will load from Fastly (and if this is the first load in Fastly, then the request goes through to the actual backend). You can also see reasonable TTL's in the `Fastly-Debug-TTL` header this time.

app/Http/Controllers/Web/ImagesController.php
```diff
@@ -63,9 +63,9 @@ public function show(string $hash, Request $request)
         $response->headers->set('Surrogate-Control', 'max-age=31536000');
         $response->headers->set('Surrogate-Key', 'post-'.$post->id);
 
-        // But we want browsers to always re-validate before serving this image from local
-        // cache, so that image rotations/deletions are respected immediately:
-        $response->headers->set('Cache-Control', 'no-cache, public');
+        // And we want browsers to cache but re-validate with us before returning cached content
+        // on a subsequent load, so that image rotations/deletions are respected immediately:
+        $response->headers->set('Cache-Control', 'max-age=31536000, must-revalidate, no-cache, public');
```
katiecrane: With this does it no longer add the `private` to the end?

DFurnes: That part was fixed in #940, although I didn't dig into why at the time (which is what it seems like that snippet does). Symfony seems to require Responses to always have either `private` or `public` on them (and if neither is set, it'll generally mark it as `private` by default).

This is also accepted by Symfony without changes, and seems to be accepted by Fastly as well (whereas before, even though the `Cache-Control` and `Surrogate-Control` looked correct to me, they were being discarded by Fastly when run in production).

PR #942, opened 2019-11-12
Update "show campaign" interface to use GraphQL.

DFurnes:
#### What's this PR do?
This pull request updates the "show campaign" view (which lets users view a campaign's accepted and rejected posts & filter by tags) to use GraphQL for data fetching. I was also able to combine the "inbox" into this view (since it now supports pagination) to simplify things further.

Screen Shot 2019-11-12 at 5 04 20 PM

A cool feature which is worth calling out is that the status is actually stored in the URL now, so we can link directly to `campaigns/:id/pending` or any other status (and when you change the filter, the URL is automatically updated).

I'd recommend reviewing commit-by-commit!

It's probably worth noting that this switches the "tag" filters from multi-select to single-select. This slightly reduces functionality, but @lkpttn and I agreed that this is a very uncommon use case… aaaand it's been broken on production for an unclear amount of time anyway. 😅  

#### Relevant tickets

#### How to deploy


DFurnes:
Something that doesn't show up in the screenshot or code diff but seems worth highlighting... building these interfaces via GraphQL is _screamingly_ faster than querying them from our PHP backend (mainly because we can efficiently batch up includes):

| Page            | Before | After | Comparison  |
|-----------------|--------|-------|-------------|
| `7950/pending`  | 4.22s  | 0.95s | 4.5x faster |
| `1144/accepted` | 3.79s  | 1.28s | 3.0x faster |
| `9013/pending`  | 2.71s  | 0.87s | 3.1x faster |
| `8026/accepted` | 2.53s  | 0.51s | 4.9x faster |
| `7/pending`     | 5.27s  | 0.74s | 7.1x faster |
| `7/accepted`    | 9.22s  | 1.23s | 7.5x faster |

resources/assets/components/utilities/Select.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { map } from 'lodash';
+
+const Select = ({ values, value, onChange }) => (
+  <div className="select">
+    <select value={value} onChange={event => onChange(event.target.value)}>
```
mendelB: Does the `value` attribute set the default value on `select`? (I thought that was only through the `options`s `selected` attribute?)

DFurnes: It [does](https://reactjs.org/docs/forms.html#the-select-tag)! It actually yells at you if you use `selected`, which is what I'd originally done.

mendelB: Aha! Ya learn something new... 👨‍🏫

PR #945, opened 2019-11-14
More cleanup & minor bug fixes.

DFurnes:
#### What's this PR do?
This pull request does another round of cleanup after #944. In particular: it removes some more old code (unused components & helper classes), fixes some awkward table styling, and updates more links to use `Link` so we get lightning-fast client-side routing.

Just one visual change: I've added a `HelpLink` component for places that we want to link to documentation, since it's visually a bit less distracting than a big "see definitions" link:

Screen Shot 2019-11-14 at 11 03 08 AM

You can go commit-by-commit for context, but it might be easiest to just skim the whole diff.

Once this is merged, we should be ready to test before deploying to production!

#### Relevant tickets

#### How to deploy


resources/assets/components/ReviewablePost.js
```diff
@@ -161,10 +164,7 @@ const ReviewablePost = ({ post }) => {
         {post.status !== 'PENDING' ? (
           <div className="container__row">
             <h4>
-              Tags
-              <a className="footnote" href="/faq#tags" target="_blank">
-                (see definitions)
-              </a>
+              Tags <HelpLink to="/faq#tags" title="Tag definitions" />
```
mendelB: 👏

DFurnes:
> I'm unfamiliar with the context for those image url and file parsing helpers, but I'm assuming that these have been retired due to GraphQL queried post media url's etc.?

Yep, the image URL ones are no longer necessary since we're querying via GraphQL!

Rogue's photo uploader was temporarily removed in #943 because it had never been updated to support multiple actions or other post types. I think most of the file helpers are unnecessary even when we re-build this tool (since, for example, we shouldn't be stripping EXIF data anymore) but I'll grab any that we do need then, if so!

PR #947, opened 2019-11-14
Use "counter caches" for pending count instead of heavy query.

DFurnes:
#### What's this PR do?
~This pull request (hopefully) fixes an issue introduced in #946. That pull request fixed the problem where campaigns with zero pending posts wouldn't be shown in the results, but introduced a new issue (only on QA! 🤔) where the list would "loop" around to the beginning like this:~

This pull request instead fixes the issue introduced in #946 by side-stepping it entirely!

This was all caused because `pending_count` would be null if there weren't yet any posts for a campaign (because we're applying this by doing a big old `SELECT COUNT(*)` on the posts table and left-joining it to campaigns). I tried a couple different ways to get `NULL` column cursors to work ([harder than you think!](https://github.com/DoSomething/rogue/pull/946)), and just couldn't get it to work. So instead I took a step back, and thought back to [this gem from Ruby on Rails](https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache)...

So, this pull request adds a new `pending_count` (and `accepted_count`!) integer columns on each campaign, which are either updated by the `campaign:refresh-counts` scheduled job or immediately upon reviewing a post for campaign.

Ignore the first few commits! The fun stuff starts here:

🥞 Adds the `pending_count` and `accepted_count` columns to the database. d41a51b

📬 Adds model events & a scheduled job to actually write to them. 34a217c

I snuck in an unrelated fix in eb2a61786ceb9feea0d7d157990b2f23816f7b33 to stop referencing an image on `twooter.biz` that was 1) 404ing and 2) prohibited since it's `http://` and we're `https://`.

#### Relevant tickets
[#169683587](https://www.pivotaltracker.com/story/show/169683587)

#### How to deploy

### DIFF
--- a/app/Console/Commands/RefreshCampaignCounts.php
+++ b/app/Console/Commands/RefreshCampaignCounts.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Rogue\Console\Commands;
+
+use Rogue\Models\Campaign;
+use Illuminate\Console\Command;
+
+class RefreshCampaignCounts extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'campaign:refresh-counts';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Refresh post counts for all campaigns.';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        foreach (Campaign::cursor() as $campaign) {
+            $this->line('Refreshing cached post counts for ' . $campaign->id . '.');
+            $campaign->refreshCounts();
+        }
+    }
+}
--- a/app/Console/Commands/RefreshCampaignCounts.php
+++ b/app/Console/Commands/RefreshCampaignCounts.php
@@ -12,7 +12,7 @@ class RefreshCampaignCounts extends Command
      *
      * @var string
      */
-    protected $signature = 'campaign:refresh-counts';
+    protected $signature = 'rogue:refresh-counts';
 
     /**
      * The console command description.
--- a/app/Http/Transformers/CampaignTransformer.php
+++ b/app/Http/Transformers/CampaignTransformer.php
@@ -23,6 +23,7 @@ class CampaignTransformer extends TransformerAbstract
             'is_open' => $campaign->isOpen(),
             'impact_doc' => $campaign->impact_doc,
             'cause' => $campaign->cause,
+            'accepted_count' => $campaign->accepted_count,
             'pending_count' => is_staff_user() ? $campaign->pending_count : null,
             'cause_names' => $campaign->getCauseNames(),
             'created_at' => $campaign->created_at->toIso8601String(),
--- a/app/Http/Transformers/CampaignTransformer.php
+++ b/app/Http/Transformers/CampaignTransformer.php
@@ -23,7 +23,7 @@ class CampaignTransformer extends TransformerAbstract
             'is_open' => $campaign->isOpen(),
             'impact_doc' => $campaign->impact_doc,
             'cause' => $campaign->cause,
-            'pending_count' => $campaign->pending_count,
+            'pending_count' => is_staff_user() ? $campaign->pending_count : null,
             'cause_names' => $campaign->getCauseNames(),
             'created_at' => $campaign->created_at->toIso8601String(),
             'updated_at' => $campaign->updated_at->toIso8601String(),
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -536,7 +536,7 @@ class Post extends Model
      */
     public static function getPostCount(Campaign $campaign, string $status)
     {
-        return (new Post)->newModelQuery()
+        return (new self)->newModelQuery()
             ->where('campaign_id', $campaign->id)
             ->where('status', $status)
             ->whereReviewable()
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -275,8 +275,7 @@ class Post extends Model
         $campaign_cause = optional($this->signup->campaign)->getAttributes()['cause'];
 
         // Fetch Campaign Website information via GraphQL.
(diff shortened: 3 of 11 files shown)


resources/views/layouts/master.blade.php
```diff
@@ -8,7 +8,7 @@
 
         <title>Rogue</title>
 
-        <link rel="icon" type="image/png" href="http://twooter.biz/Gifs/tonguecat.png">
+        <link rel="apple-touch-icon-precomposed" href="/assets/images/apple-touch-icon-precomposed.png">
```
katiecrane: You know I gotta ask that if this isn't TongueCat, then what is it?

DFurnes: Tongue Cat lives on in the "empty" component!!

Screen Shot 2019-11-14 at 2 32 38 PM

DFurnes: But now the touch icon is simply our logo. 🙃

katiecrane: Well I don't agree but I'll let it happen 😂

app/Models/Traits/HasCursor.php
```diff
@@ -55,10 +55,11 @@ public function scopeWhereAfterCursor($query, $cursor, $defaultSort = 'id,desc')
...
                         // query, since MySQL's `WHERE =` won't compare with `NULL`.
-                        $query->where(function ($query) use ($column, $sortCursor) {
-                            $query->where($column, '=', $sortCursor)
-                                ->orWhereNull($column);
-                        });
+                        if (is_null($sortCursor)) {
+                            $query->whereNull($column);
+                        } else {
+                            $query->where($column, '=', $sortCursor);
+                        }
```
DFurnes: Oh ok right –– the reason I had changed this was because this doesn't handle cases where a page would overlap between sorted values (e.g. a page with 3, 2, 1, 1, 1, `null`, `null`, `null`). Without the `orWhereNull`, this page would not include those three nulls that it's supposed to.

DFurnes:
This is ready for re-review now! I think this is all-in-all a lot simpler, and the campaign index loads way faster now to boot (from 1.2 seconds down to ~150ms).

app/Console/Commands/RefreshCampaignCounts.php
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        foreach (Campaign::cursor() as $campaign) {
+            $this->line('Refreshing cached post counts for ' . $campaign->id . '.');
+            $campaign->refreshCounts();
```
katiecrane: Aren't we gonna have to run this pretty often for it to not be confusing for admins? Like if you review all the posts, then come back to the campaign index, won't it be confusing if the inbox number hasn't changed?

app/Models/Post.php
```diff
@@ -275,7 +275,8 @@ public function toBlinkPayload()
         $campaign_cause = optional($this->signup->campaign)->getAttributes()['cause'];
 
         // Fetch Campaign Website information via GraphQL.
-        $campaignWebsite = app(GraphQL::class)->getCampaignWebsiteByCampaignId($this->campaign_id);
+        // $campaignWebsite = app(GraphQL::class)->getCampaignWebsiteByCampaignId($this->campaign_id);
```
katiecrane: Should this line be removed?

app/Providers/ModelServiceProvider.php
```diff
@@ -52,6 +52,10 @@ public function boot()
 
         // When Reviews are saved create an event for them.
         Review::saved(function ($review) {
+            info('review', ['campaign' => $review->post->campaign_id]);
+            // Update the "counter cache" on this campaign:
+            $review->post->campaign->refreshCounts();
```
katiecrane: Oh wait I see that this count is getting updated whenever there is a review, and the job is just to make sure things don't get wildly out of sync. Is that right?

database/migrations/2019_11_14_203229_add_pending_count_to_campaigns.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('campaigns', function (Blueprint $table) {
+            $table->integer('pending_count')->default(0);
+            $table->integer('accepted_count')->default(0);
```
katiecrane: Can we put these columns after `end_date` so that the timestamps are still the last columns on this table?

app/Console/Commands/RefreshCampaignCounts.php
DFurnes: We [run this on-demand with a model event](https://github.com/DoSomething/rogue/pull/947/files#diff-01de341882d0c02d6377ad12e035b315R57) so it'll always be up-to-date after reviewing a post! I'm thinking we'll run the scheduled job every 5-10 minutes or so, so that won't be too out-of-date either.

app/Models/Post.php
DFurnes: Gah, yes!! 😖 I have to remove that to fix an issue communicating with the local GraphQL server when developing (need a better fix!) and forgot to uncomment here.

app/Providers/ModelServiceProvider.php
DFurnes: Yes! The job is to "re-sync" if things somehow got wonky, and to take into account newly submitted posts (which don't trigger the on-demand count).

database/migrations/2019_11_14_203229_add_pending_count_to_campaigns.php
DFurnes: Oh sure!

app/Http/Transformers/CampaignTransformer.php
```diff
@@ -23,7 +23,8 @@ public function transform(Campaign $campaign)
             'is_open' => $campaign->isOpen(),
             'impact_doc' => $campaign->impact_doc,
             'cause' => $campaign->cause,
-            'pending_count' => $campaign->pending_count,
+            'accepted_count' => $campaign->accepted_count,
```
katiecrane: Does this need `is_staff_user()` too?

DFurnes: I don't think so, since this is the same as paging through the gallery & counting by hand.

PR #948, opened 2019-11-15
Fix some issues with counter cache.

DFurnes:
#### What's this PR do?
This pull request fixes some issues with the new counter caches added in #947:

🔃 Fixes an issue where paginated lists would continually loop back to the beginning. This turned out to be a frustratingly simple bug – we were checking if the cursor was "truthy" (rather than just whether it existed), and so a cursor value of `0` was the same as saying "no cursor". 68c682e

🐎 The "recount" query for each campaign takes about a second to run, which is not too bad when re-calculating this for a single campaign when someone reviews a post… but [quite slow](https://my.papertrailapp.com/events?focus=1131997651941478401&q=program%3Aapp%2Fscheduler.7334&selected=1131997534979104772) when trying to do it for over 700 campaigns on a schedule!

🍃 It also removes the heavy query to include a on-demand `pending_count` on the `campaigns/:id` API endpoint, since we store this using the counter cache. 7e966a8

Finally, I included three quick quality-of-life improvements:

✅ Adds the count of "accepted" posts for each campaign to the index! We had to remove this a while back when performance was suffering, but we can now add it without a penalty. 8db90c0

🔢 Loads 80 campaigns per page instead of 15. This should make it quicker for campaign leads to find campaigns with less round-trip "search" queries to the backend. 02c245f

🚀 Optimistically render any changed "review" status (so a post will appear to be accepted or rejected immediately). This makes the interface feel a lot more snappy, especially since the real backend request now takes an extra 1-2 seconds to refresh the counter cache. 7b9eb73

I'd recommend reviewing these commit-by-commit (linked above).

Hopefully this is the last of the Rogue fixes! 🤞

#### Relevant tickets
[#169683587](https://www.pivotaltracker.com/story/show/169683587)

### DIFF
--- a/resources/assets/components/ReviewButton.js
+++ b/resources/assets/components/ReviewButton.js
@@ -26,6 +26,16 @@ const ReviewButton = ({ post, status, children }) => {
       id: post.id,
       status: status,
     },
+    // We'll optimistically update the interface with the given status
+    // before waiting for the full network round-trip. Snappy!
+    optimisticResponse: {
+      __typename: 'Mutation',
+      reviewPost: {
+        __typename: 'Post',
+        id: post.id,
+        status: status,
+      },
+    },
   });
 
   const statusClass = `-${status.toLowerCase()}`;
--- a/app/Console/Commands/RecountPostsCommand.php
+++ b/app/Console/Commands/RecountPostsCommand.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Console\Commands;
 
+use Rogue\Models\Post;
 use Rogue\Models\Campaign;
 use Illuminate\Console\Command;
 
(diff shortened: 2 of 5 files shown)


resources/assets/components/ReviewButton.js
```diff
@@ -26,6 +26,16 @@ const ReviewButton = ({ post, status, children }) => {
       id: post.id,
       status: status,
     },
+    // We'll optimistically update the interface with the given status
+    // before waiting for the full network round-trip. Snappy!
```
katiecrane: Snappy!! If something goes wrong will it switch back to the status from the db?

DFurnes: Yep! This "pretends" the given response was returned, and then replaces it with the real response once we get it. (So for things like this where they're almost certain to be the same, it's a pretty cool way to make it seem a ton faster.)

katiecrane: THIS IS SO COOL!!!!!!

PR #949, opened 2019-11-15
Allow users to choose how to sort campaigns.

DFurnes:
#### What's this PR do?
This pull request adds the ability to choose the order that campaigns are sorted on the index (either by ID, pending count, or accepted count). This makes it easy to find, say, the newest campaigns (by sorting `id,desc`) or most popular campaigns (by sorting `accepted_count,desc`).

sort-columns

👀

This was quick to add & should help address some of the issues campaign leads have raised with finding campaigns that aren't near the top of the list.

#### Relevant tickets
N/A

#### How to deploy

### DIFF
--- a/resources/assets/components/CampaignsTable.js
+++ b/resources/assets/components/CampaignsTable.js
@@ -1,17 +1,22 @@
 import { get } from 'lodash';
 import gql from 'graphql-tag';
 import { Link } from 'react-router-dom';
-import React, { useEffect } from 'react';
 import { useQuery } from '@apollo/react-hooks';
+import React, { useState, useEffect } from 'react';
 
 import Empty from './Empty';
 import { updateQuery } from '../helpers';
+import OrderableHeading from './utilities/OrderableHeading';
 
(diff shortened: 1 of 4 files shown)


resources/assets/components/CampaignsTable.js
```diff
@@ -110,9 +117,24 @@ const CampaignsTable = ({ isOpen, filter }) => {
       <table className="table">
         <thead>
           <tr>
-            <td>Campaign</td>
-            <td>Pending</td>
-            <td>Accepted</td>
+            <OrderableHeading
```
katiecrane: "Orderable" is a hard word to say

DFurnes: And to spell! 😂 Any ideas on a better term for this? Maybe `SortableHeading`?

katiecrane: SORTABLE I like that better!

DFurnes: Yeah me too! Swapped in 49f6f51.

PR #352, opened 2017-08-10
Add custom Faker providers.

DFurnes:
#### What's this PR do?
Add custom faker providers for Northstar IDs and Campaign IDs, so we can easily create test data that works well when querying against our other test environments. For example, here's some fake reportbacks I just seeded:

screen shot 2017-08-10 at 12 23 21 pm

#### How should this be reviewed?
👀

#### Any background context you want to provide?
For now, this talks to Northstar QA & Phoenix Thor, since that's what I have configured on my local. What are other people using?

#### Relevant tickets
[#149662237](https://www.pivotaltracker.com/story/show/149662237)

#### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.
- [ ] Pinged a PM if this is a larger PR that would benefit from some additional testing love.

PR #951, opened 2019-11-18
Add school_id to posts

aaronschachter:
#### What's this PR do?

Adds a new `school_id` column, to the Posts table as well as API support to write, update, and view.

Create a post with `school_id` to verify it works. What was really tripping me up is that I can't get `PATCH /v3/posts` to work within Postman at all, but the tests are passing.

I've been thinking about adding a field to the Actions table to indicate whether the post should collect `school_id`. It seems like the better place to save this information vs anywhere in Phoenix. This gets me thinking that perhaps the Post create/update requests would lookup the Action and only save the `school_id` value if `should_collect_school_id` is set to true ... and that way Phoenix or any other client could send `school_id` without needing to make the lookup to see if the action accepts a `school_id`. 

Another tangent is that I was trying to get a [`FakerSchoolId` class to work](https://github.com/DoSomething/rogue/commit/e51e762327b444d2875d3f6a73c93bbabd86f266), but kept getting an `InvalidArgumentException: Unknown formatter "school_id"` when I'd try to call the `$this->faker->school_id` . I left it out for now to work on figuring out how/when to send over the user school from Phoenix, but it was a bummer as it looked like this should've worked (refs  https://github.com/DoSomething/rogue/pull/352)

#### Relevant tickets
https://www.pivotaltracker.com/story/show/169549658

#### How to deploy

### DIFF
--- a/app/Http/Requests/PostRequest.php
+++ b/app/Http/Requests/PostRequest.php
@@ -42,6 +42,7 @@ class PostRequest extends Request
             'text' => 'nullable|string|max:500',
             'location' => 'nullable|iso3166',
             'postal_code' => 'nullable|max:10',
+            'school_id' => 'nullable|string|max:255',
             'quantity' => 'nullable|integer',
             'file' => 'image|dimensions:min_width='.$minImageSize['width'].',min_height='.$minImageSize['height'].',max_width='.$maxImageSize['width'].',max_height='.$maxImageSize['height'],
             'status' => $this->getStatusRules($this->type),
@@ -62,6 +63,7 @@ class PostRequest extends Request
             'location' => 'nullable|iso3166',
             'quantity' => 'nullable|integer',
             'status' => $this->getStatusRules($this->post->type),
+            'school_id' => 'nullable|string|max:255',
         ];
     }
 
--- a/app/Http/Transformers/PostTransformer.php
+++ b/app/Http/Transformers/PostTransformer.php
@@ -66,6 +66,7 @@ class PostTransformer extends TransformerAbstract
             $response['source_details'] = $post->source_details;
             $response['remote_addr'] = '0.0.0.0';
             $response['details'] = $post->details;
+            $response['school_id'] = $post->school_id;
         }
 
         return $response;
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -1219,15 +1219,14 @@ class PostTest extends TestCase
     }
 
     /**
-     * Test for updating a post successfully.
+     * Test for updating a post with invalid status.
      *
-     * PATCH /api/v3/posts/186
+     * PATCH /api/v3/posts/:id
      * @return void
      */
-    public function testUpdatingAPhotoWithBadStatus()
+    public function testUpdatingAPhotoWithInvalidStatus()
     {
         $post = factory(Post::class)->create();
-        $signup = $post->signup;
 
         $this->mock(Blink::class)->shouldReceive('userSignupPost');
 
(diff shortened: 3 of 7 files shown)


app/Http/Requests/PostRequest.php
```diff
@@ -42,6 +42,7 @@ public function rules()
             'text' => 'nullable|string|max:500',
             'location' => 'nullable|iso3166',
             'postal_code' => 'nullable|max:10',
+            'school_id' => 'nullable|string|max:255',
```
katiecrane: I thought these were integers?

katiecrane:
> I've been thinking about adding a field to the Actions table to indicate whether the post should collect school_id. It seems like the better place to save this information vs anywhere in Phoenix. This gets me thinking that perhaps the Post create/update requests would lookup the Action and only save the school_id value if should_collect_school_id is set to true ... and that way Phoenix or any other client could send school_id without needing to make the lookup to see if the action accepts a school_id.

I thought the way we were gonna do this was to send over `school_id` when a post is created if a `school_id` exists on the Northstar user, since it's a column on the Posts table and any post _can_ have a `school_id` (we just might not be planning to use it for anything specifically).

app/Http/Requests/PostRequest.php
aaronschachter: Although we are using the numeric `universal-id` column as our School ID, we're saving as a [string](https://github.com/DoSomething/graphql/blob/99a98fd34e55c44b50ba97d3a644c9d253a1d84e/src/schema/schools.js#L15) for two reasons:
- support a `school-not-available` value to save if user selects that they can't find their school
- in case we ever want to use our own MongoDB ID's as identifiers (there's been talk about potentially querying colleges from a different dataset)

aaronschachter:
Initially we did chat about always sending including `school_id` for every post if it exists on the user.  Upon a data hangout with @ngjo and @mjain-1  (and a separate @DFurnes  hangout about https://github.com/DoSomething/phoenix-next/pull/1694) -- it seemed a bit extreme to save `school_id` on all posts when the user's post for any given campaign may not be affiliated with their current school.

It definitely would make things easier for Phoenix development to always send over the `school_id` if it exists on the user. In theory we could go back and null out post values for non TFJ posts should we regret the decision, but the more we add School Finders to campaigns (which is no longer a hardcoded thing per https://github.com/DoSomething/phoenix-next/pull/1721) the trickier that would be. 

It seemed like a good first step and compromise would be to only save `school_id` via hardcoded check for the TFJ action ID. We could go this route, but it got me thinking about the ideal way to configure this without hard-coding (which is beyond the scope of this PR.. but felt like relevant "background context to provide").

app/Http/Requests/PostRequest.php
katiecrane: Those are two very good reasons, good thinking!!

katiecrane:
@aaronschachter thanks for adding more context! So the current plan is to only send `school_id` for TFJ but you're thinking about potential better ways to do that?

app/Http/Transformers/PostTransformer.php
```diff
@@ -66,6 +66,7 @@ public function transform(Post $post)
             $response['source_details'] = $post->source_details;
             $response['remote_addr'] = '0.0.0.0';
             $response['details'] = $post->details;
+            $response['school_id'] = $post->school_id;
```
katiecrane: Does this need to be added to the c.io payload as well?

aaronschachter: Doesn't seem to hurt sending over while we're here.. @ngjo and I were chatting over whether we need this (and the perhaps-more-helpful school name) over in https://www.pivotaltracker.com/story/show/169745644/comments/208852773

katiecrane: Ah yeah I guess school name is probably more helpful for c.io, but like you said, id can't hurt

tests/Http/PostTest.php
```diff
@@ -1279,7 +1304,7 @@ public function testValidationUpdatingAPost()
 
     /**
      * Test that a user can update their own post, but can't
-     * change it's review status.
+     * change its review status.
```
katiecrane: Thanks for cleaning up these doc blocks!

aaronschachter:
> So the current plan is to only send school_id for TFJ but you're thinking about potential better ways to do that?

Yep! I suppose the easiest way to hardcode for TFJ would be to check the action ID on the Rogue side, and have Phoenix always send over a `school_id` if set. (I also have been mulling over whether Rogue could query Northstar to find `school_id` per the `northstar_id` parameter, but that's probably just wild 🐅)

The idea of always sending over the `school_id` and determining when to save on the Rogue side was what got me thinking about something like a `collect_school_id` boolean flag on an Action. But I could see confusion in a client sending over `school_id` parameter and then it not being populated because the action doesn't collect school id's (unless we had an additional `message` type of API response property that informed "This action does not save school ID's")... brainstorming here, more than writing a spec..

PR #959, opened 2019-11-26
Add action permalink

aaronschachter:
#### What's this PR do?

Adds a web `/actions/:id` route and a `ShowAction` page component to render a permalink view of an action. I kept navigating to this path to spot check which action I was selecting while creating different submission content types over in Phoenix for testing `school_id` collection.

Screen Shot 2019-11-26 at 1 46 56 PM

👀 

It'd be useful to add an `Action.campaign` nested property over in GraphQL to display the campaign's internal title as well as ID

#### Relevant tickets
https://www.pivotaltracker.com/story/show/169549621/comments/209139906

#### How to deploy

### DIFF
--- a/resources/assets/pages/ShowAction.js
+++ b/resources/assets/pages/ShowAction.js
@@ -41,7 +41,7 @@ const ShowAction = () => {
         <a href={`/campaign-ids/${campaignId}`}>{`Campaign ${campaignId}`}</a>
       }
     >
-      <Action action={data.action} campaign={{ id: campaignId }} />
+      <Action action={data.action} campaign={{ id: campaignId }} isPermalink />
     </Shell>
   );
 };
--- a/resources/assets/pages/ShowAction.js
+++ b/resources/assets/pages/ShowAction.js
@@ -3,6 +3,7 @@ import React, { useState } from 'react';
 import { useParams } from 'react-router-dom';
 import { useQuery } from '@apollo/react-hooks';
 
+import NotFound from './NotFound';
 import Action from '../components/Action';
 import Shell from '../components/utilities/Shell';
 import MetaInformation from '../components/utilities/MetaInformation';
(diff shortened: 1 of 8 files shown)


resources/assets/pages/ShowAction.js
```diff
@@ -0,0 +1,49 @@
...
+import Action from '../components/Action';
+import Shell from '../components/utilities/Shell';
+import MetaInformation from '../components/utilities/MetaInformation';
+
+const SHOW_ACTION_QUERY = gql`
+  query ShowActionQuery($id: Int!) {
+    action(id: $id) {
+      id
+      name
+      campaignId
```
DFurnes: Should we include other fields (like `noun`, `verb`, `reportback`, etc.) in this query? Although I guess we'd have to add some conditionals for camel-cased vs. snake-cased formatting, since the other place this `Action` component is used is still using the plain REST API...

resources/assets/pages/ShowAction.js
```diff
@@ -0,0 +1,49 @@
...
+
+  if (error) {
+    return <Shell error={error} />;
+  }
+
+  if (loading) {
+    return <Shell title="Action" loading />;
+  }
+
+  const { campaignId, name } = data.action;
```
DFurnes: Should we add a "not found" error handler as well (just something simple [like this](https://github.com/DoSomething/rogue/blob/e795e177698df11811d66b8f8673c15a20e99f40/resources/assets/pages/ShowPost.js#L46-L53))?

resources/assets/pages/ShowAction.js
```diff
@@ -0,0 +1,49 @@
...
+
+  return (
+    <Shell
+      title={name}
+      subtitle={
+        <a href={`/campaign-ids/${campaignId}`}>{`Campaign ${campaignId}`}</a>
```
DFurnes: This is really subjective, but it looks a little odd to have a link here to me (the blue on purple).

What if we added this link below the action, like [on the post permalink](https://github.com/DoSomething/rogue/blob/e795e177698df11811d66b8f8673c15a20e99f40/resources/assets/pages/ShowPost.js#L61-L70)?

resources/assets/pages/ShowAction.js
aaronschachter: Yes! This has been resolved in #960 by rendering Actions via GraphQL from both this permalink and the campaign permalink.

PR #960, opened 2019-12-02
Cleanup admin Actions + Campaigns

aaronschachter:
#### What's this PR do?

Ever try to fix onnne little thing (https://github.com/DoSomething/rogue/pull/959#discussion_r351001867) before the holiday break and then end up fixing a TODO for cleaning up the `CampaignsIdController`? This PR started out by refactoring the `Action` component to query GraphQL for all data, but soon made its way to refactoring the `CampaignIdSingle` component to query GraphQL as well, since it included the `Action` component. 

I was getting pretty confused about the `CampaignIdSingle` vs `Campaign` naming, and figured it'd be a good time to refactor the `campaign-ids` URL as just `campaigns`, taking a page from how the Actions Controller permalink works in #959.

Here's a breakdown of all changes:

* Rename `CampaignIdSingle` component dir as `Campaign`, query Campaign and Action information from GraphQL (blocked by https://github.com/DoSomething/graphql/pull/177)

   * Uses the Campaign.action GraphQL field to display all actions, which doesn't have pagination. As far as I can tell from some spot checking I did in our QA dataset, only Grab the mic (campaign 8017) has enough actions that need pagination. I'm assuming this is a non-blocker because it's an old campaign.

* Rename `CampaignIdsController` as `CampaignsController` - display the React app for index and show methods

* Change `/campaign-ids/:id` and `/campaign-ids/create` to `/campaigns` route 

* Direct user to action permalink after creating action, instead of the campaign permalink (the action permalink contains a footer link to navigate back to the campaign permalink)

👀 

The `Campaign.action` property was extremely convenient to build out the full Campaign permalink, and had me thinking about the question around deprecating the simple queries over in https://github.com/DoSomething/rfcs/pull/6. It does feel like they are worth keeping when the data set isn't usually something we need to paginate, but then again -- we should do due diligence and add pagination for the Grab the Mic actions in a future PR.

#### How to deploy


DFurnes:
> The `Campaign.action` property was extremely convenient to build out the full Campaign permalink, and had me thinking about the question around deprecating the simple queries over in [DoSomething/rfcs#6](https://github.com/DoSomething/rfcs/pull/6). It does feel like they are worth keeping when the data set isn't usually something we need to paginate, but then again -- we should do due diligence and add pagination for the Grab the Mic actions in a future PR.

Yeah, I think I'm leaning in this direction as well! Definitely seems handy for simple queries.

resources/assets/bootstrap.js
```diff
@@ -4,10 +4,9 @@ import React from 'react';
 import ReactDom from 'react-dom';
 
 import { ready } from './helpers';
-import mountContainer from './utilities/MountContainer';
+// import mountContainer from './utilities/MountContainer';
```
aaronschachter: @DFurnes -- Looks like we're ok to delete this file as well?

aaronschachter: Removed this in #962

aaronschachter: Removed this in #962

PR #964, opened 2019-12-03
Adds ActionStats resource

aaronschachter:
Updated description:

### What's this PR do?

This pull request adds a `GET /action-stats` endpoint and `ActionStat` model to store the sum quantity of approved posts for a given `action_id` and `school_id`.

 

Example request:

`GET http://rogue.test/api/v3/action-stats`

Example response:

```

{
  "data": [
    {
      "id": 1,
      "action_id": 3,
      "school_id": "3401460",
      "accepted_quantity": 1428
    },
    {
      "id": 2,
      "action_id": 1,
      "school_id": "3401460",
      "accepted_quantity": 1485
    },
    {
      "id": 3,
      "action_id": 6,
      "school_id": "3401460",
      "accepted_quantity": 148
    }
  ],
  "meta": {
    "pagination": {
      "total": 3,
      "count": 3,
      "per_page": 20,
      "current_page": 1,
      "total_pages": 1,
      "links": [
        
      ]
    }
  }
```
 ----

Original description:

### What's this PR do?

This pull request adds a `/actions/:action/schools/:schoolId` route to return the aggregate sum of accepted `posts.quantity` for the given Action and School ID's. See background context for WIP

Example request (with random seed data):

`GET http://rogue.test/api/v3/actions/4/schools/quo`

Example response:

```
{
  "quantity": 135
}
```

It's making the calculation each time it receives a request for now, even for anonymous users. At minimum we can require authentication - but still leaves us susceptible to malicious authenticated user polling the endpoint.

We could cache each result, but that doesn't help if an attacker cycled through numeric action/school ID's. I'm beginning to think it'd be ideal to have something like an `ActionStats` Model we could use, and update the calculation upon a post review.  This way we'd easily be able to create leaderboard stats on the fly, and expose in something like an `actions/:id/schools` endpoint.

### Relevant tickets

References https://www.pivotaltracker.com/story/show/169659230

### Checklist

### DIFF
--- a/app/Http/Controllers/ActionStatsController.php
+++ b/app/Http/Controllers/ActionStatsController.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Rogue\Http\Controllers;
+
+use Rogue\Models\ActionStat;
+use Illuminate\Http\Request;
+use Rogue\Http\Transformers\ActionStatTransformer;
+
+class ActionStatsController extends ApiController
+{
+    /**
+     * @var Rogue\Http\Transformers\ActionTransformer;
+     */
+    protected $transformer;
+
+    /**
+     * Create a controller instance.
+     */
+    public function __construct()
+    {
+        $this->transformer = new ActionStatTransformer;
+    }
+
+    /**
+     * Display a listing of the resource.
+     *
+     * @return \Illuminate\Http\Response
+     */
+    public function index(Request $request)
+    {
+        $query = $this->newQuery(ActionStat::class);
+
+        $filters = $request->query('filter');
+        $query = $this->filter($query, $filters, ActionStat::$indexes);
+
+        return $this->paginatedCollection($query, $request);
+    }
+}
--- a/app/Http/Controllers/ActionStatsController.php
+++ b/app/Http/Controllers/ActionStatsController.php
@@ -2,8 +2,8 @@
 
 namespace Rogue\Http\Controllers;
 
-use Rogue\Models\ActionStat;
 use Illuminate\Http\Request;
+use Rogue\Models\ActionStat;
 use Rogue\Http\Transformers\ActionStatTransformer;
 
 class ActionStatsController extends ApiController
--- a/app/Http/Transformers/ActionStatTransformer.php
+++ b/app/Http/Transformers/ActionStatTransformer.php
@@ -10,16 +10,14 @@ class ActionStatTransformer extends TransformerAbstract
     /**
      * Transform resource data.
      *
-     * @param \Rogue\Models\Action $action
+     * @param \Rogue\Models\ActionStat $actionStat
      * @return array
      */
-    public function transform(ActionStat $stat)
+    public function transform(ActionStat $actionStat)
     {
-        return [
-            'id' => $stat->id,
-            'action_id' => $stat->action_id,
-            'school_id' => $stat->school_id,
-            'accepted_quantity' => $stat->accepted_quantity,
-        ];
+        $result = $actionStat;
+        unset($result->deleted_at);
+
+        return $result;
     }
 }
--- a/app/Http/Transformers/ActionStatTransformer.php
+++ b/app/Http/Transformers/ActionStatTransformer.php
@@ -19,8 +19,8 @@ class ActionStatTransformer extends TransformerAbstract
             'action_id' => $actionStat->action_id,
             'school_id' => $actionStat->school_id,
             'accepted_quantity' => $actionStat->accepted_quantity,
-            'created_at' => $campaign->created_at->toIso8601String(),
-            'updated_at' => $campaign->updated_at->toIso8601String(),
+            'created_at' => $actionStat->created_at->toIso8601String(),
+            'updated_at' => $actionStat->updated_at->toIso8601String(),
         ];
     }
 }
--- a/app/Http/Transformers/ActionStatTransformer.php
+++ b/app/Http/Transformers/ActionStatTransformer.php
@@ -15,9 +15,12 @@ class ActionStatTransformer extends TransformerAbstract
      */
     public function transform(ActionStat $actionStat)
     {
-        $result = $actionStat;
-        unset($result->deleted_at);
-
-        return $result;
+        return [
+            'action_id' => $actionStat->action_id,
+            'school_id' => $actionStat->school_id,
+            'accepted_quantity' => $actionStat->accepted_quantity,
+            'created_at' => $campaign->created_at->toIso8601String(),
+            'updated_at' => $campaign->updated_at->toIso8601String(),
+        ];
     }
 }
--- a/app/Http/Transformers/ActionStatTransformer.php
+++ b/app/Http/Transformers/ActionStatTransformer.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Rogue\Http\Transformers;
+
+use Rogue\Models\ActionStat;
+use League\Fractal\TransformerAbstract;
+
+class ActionStatTransformer extends TransformerAbstract
+{
+    /**
+     * Transform resource data.
+     *
+     * @param \Rogue\Models\Action $action
+     * @return array
+     */
+    public function transform(ActionStat $stat)
+    {
+        return [
+            'id' => $stat->id,
+            'action_id' => $stat->action_id,
+            'school_id' => $stat->school_id,
+            'accepted_quantity' => $stat->accepted_quantity,
+        ];
+    }
+}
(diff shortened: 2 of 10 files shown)


aaronschachter:
Digging into this more and liking the idea of a new model (started on an `ActionStat` model in a new-or-maybe-this branch). A `GET /action-stats` endpoint with filters on `school_id` and `action_id` would give us the leaderboard data for any given action, and also the list of all actions any given school has taken.

aaronschachter:
Updated this branch per the `ActionStat` model discussed in https://github.com/DoSomething/rogue/pull/964#issuecomment-561339240. Will add documentation and tests in a following PR if this looks good to go.

app/Http/Transformers/ActionStatTransformer.php
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Transform resource data.
+     *
+     * @param \Rogue\Models\ActionStat $actionStat
+     * @return array
+     */
+    public function transform(ActionStat $actionStat)
+    {
+        $result = $actionStat;
+        unset($result->deleted_at);
```
DFurnes: This should [return an array](https://fractal.thephpleague.com/transformers/) with the format we want for this API endpoint. This ensures that if we add/remove/rename properties down the line, we can keep the same HTTP contract (and we don't need to worry about an expanding list of properties to `unset` here either!

database/migrations/2019_12_03_105400_create_action_stats_table.php
```diff
@@ -0,0 +1,36 @@
...
+    public function up()
+    {
+        Schema::create('action_stats', function (Blueprint $table) {
+            $table->increments('id');
+            $table->unsignedInteger('action_id')->index()->comment('The action id the stat is for.');
+            $table->foreign('action_id')->references('id')->on('actions');
+            $table->string('school_id', 255)->index()->comment('The school id the stat is for.');
+            $table->unsignedInteger('accepted_quantity')->comment('The sum of accepted post quantity for the school and action.');
+            $table->timestamps();
+            $table->softDeletes();
```
DFurnes: Does this need to be soft-deletable? I'm having a hard time thinking of how we'd use this.

routes/api.php
```diff
@@ -51,4 +51,7 @@
     // actions
     $this->get('actions', 'ActionsController@index');
     $this->get('actions/{action}', 'ActionsController@show');
+
+    // action stats
+    $this->get('action-stats', 'ActionStatsController@index');
```
DFurnes: This is such an odd one to think about, since it's really stats about the action + school together! I wonder if we'd want to explicitly name it as `action-school-stats` (since it's not really applicable for non-school actions)? Or nest it under a placeholder "schools" resource, like `/actions/:action_id/schools/:school_id/stats`?

app/Http/Controllers/ActionStatsController.php
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public function index(Request $request)
+    {
+        $query = $this->newQuery(ActionStat::class);
+
+        $filters = $request->query('filter');
+        $query = $this->filter($query, $filters, ActionStat::$indexes);
+
+        return $this->paginatedCollection($query, $request);
+    }
```
DFurnes: Would adding a "show" endpoint here make sense (rather than filtering the index down to an array of one)?

routes/api.php
aaronschachter: My thought was to name it exactly per model for simplicity sake (once this is in place, I'll be starting on adding this into GraphQL).

I went with `ActionStat` vs using the word `School` somewhere in the name -- because in design meetings there had been talk of eventually renaming `school_id` as something like `affiliation_id`, e.g. to support youth groups, after school groups, etc. But because we're not there yet, it seemed a lot clearer to store these as `school_id`.

I do like the idea of defining the `school_id` and `action_id` via URL params instead of query params... but it potentially adds a lot overhead that I thought might not be necessary. But if we went this route (and still assuming `ActionStat` as the model name), it could look like this?

`GET /actions/:action_id/stats` - This is our leaderboard for an action. GraphQL query could be named `Action.getStats(options)`

`GET /actions/:action_id/schools/:school_id/stats` - This is the school's stat for the action, which we'll display to user upon selecting a school in a School Finder (we'll need to add a new `actionId` parameter to the `additionalContent` in a `SectionBlock` to know which `ActionStat` to display). GraphQL query could be `School.getStatsForActionId()`

`GET /schools/:school_id/stats` - All accepted actions a school has taken, GraphQL `School.getStats(options)`

aaronschachter: Thinking more about the GraphQL methods makes me think we could get away with just a single index route, because we're likely never going to use this endpoint anywhere else. It's also got me leaning more towards just calling the model `ActionSchoolStat` or maybe potentially better or worse `SchoolAction` for have a single index endpoint (e.g. `action-school-stats` like you suggested)

DFurnes: Ahh, that makes sense re: the naming & potential renames in the pipeline. I can also see the argument that `?filter[…]=` is easier to deal with that URL-params in our resolvers. I'm good to leave it as is & we can always see if something starts to make more sense as it solidifies!

> `GET /actions/:action_id/stats` - This is our leaderboard for an action. GraphQL query could be named `Action.getStats(options)`
> 
> `GET /schools/:school_id/stats` - All accepted actions a school has taken, GraphQL `School.getStats(options)`

Nitpick, but I think I'd recommend naming these as `stats` or `statistics` rather than `getStats` (so it's a "thing" in the schema, rather than an imperative request).

> `GET /actions/:action_id/schools/:school_id/stats` - This is the school's stat for the action, which we'll display to user upon selecting a school in a School Finder (we'll need to add a new `actionId` parameter to the `additionalContent` in a `SectionBlock` to know which `ActionStat` to display). GraphQL query could be `School.getStatsForActionId()`

Again, sorta nitpicky but this could be an optional argument on `School.statistics`, so we could do something like this:

```gql
{
  school(…) {
    name
    stats { # <-- no argument, so we leave `?filter[action_id]=` out!
      actionId
      acceptedQuantity
    }
    thisAction: stats(actionId: 934) { # <-- ?filter[action_id]=934
      acceptedQuantity
    }
}
```

database/migrations/2019_12_03_105400_create_action_stats_table.php
```diff
@@ -0,0 +1,36 @@
...
+     * @return void
+     */
+    public function up()
+    {
+        Schema::create('action_stats', function (Blueprint $table) {
+            $table->increments('id');
+            $table->unsignedInteger('action_id')->index()->comment('The action id the stat is for.');
+            $table->foreign('action_id')->references('id')->on('actions');
+            $table->string('school_id', 255)->index()->comment('The school id the stat is for.');
+            $table->unsignedInteger('accepted_quantity')->comment('The sum of accepted post quantity for the school and action.');
```
DFurnes: Ah sorry, one more nitpick! Can we name this `accepted_count` so it matches up with [the Campaign equivalent](https://github.com/DoSomething/rogue/blob/dfa30cb5e750826478ba3054a35c9a85fa187573/app/Http/Transformers/CampaignTransformer.php#L26)?

aaronschachter: This isn't the count of accepted posts though... it's their sum quantity (or "aggregate impact", in biz-cat speak)

DFurnes: Oh you're totally right! I got myself mixed up there.

app/Providers/ModelServiceProvider.php
```diff
@@ -53,9 +54,21 @@ public function boot()
...
             $review->post->campaign->refreshCounts();
 
+            // If this post is associated with a school, update the school action stats.
+            $schoolId = $review->post->school_id;
+
+            if ($schoolId) {
+                $actionId = $review->post->action_id;
+                ActionStat::updateOrCreate(
+                    ['action_id' => $actionId, 'school_id' => $schoolId],
+                    ['accepted_quantity' => Post::getAcceptedQuantitySum($actionId, $schoolId)]
```
aaronschachter: Testing this a little more throughly (having worked fixing inbox locally over in #966) - when I debug the `$review->post` here, its status is still `pending`, so first review results in a quantity of 0.  Will update this branch with the fix, but I think the campaign refreshCounts should get moved to the same place? Seems it: https://www.pivotaltracker.com/n/projects/2328687/stories/170111820

app/Http/Controllers/ActionStatsController.php
aaronschachter: Going to leave this as just an `index` endpoint per https://github.com/DoSomething/rogue/pull/964/files#discussion_r353974591

PR #965, opened 2019-12-03
Add redirects for old `/campaign-ids` routes.

DFurnes:
### What's this PR do?

This pull request includes some cleanup & adds redirects for the old `/campaign-ids` and `/campaign-ids/:id` routes that were refactored in #960. (They're still linked from Contentful & folks may have bookmarks saved on their machines).

### How should this be reviewed?

I did some cleanup in 0d2d5d0 (moving exclusively client-side routes so they're solely in `routes/web.php` and grouping the few remaining server-side routes). I also updated this file to use Laravel's standard `Route` facade to match the framework documentation.

Then the bug fix is in c0f333b! 🐞

### Any background context you want to provide?

I wanted to use Laravel's [`Route::redirect`](https://laravel.com/docs/5.5/routing#redirect-routes) here, but it doesn't support route arguments (for the `/campaign-ids/:id` to `/campaigns/:id` redirect). Bummer!

### Relevant tickets

References [Pivotal #170082975](https://www.pivotaltracker.com/story/show/170082975).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.

PR #966, opened 2019-12-04
Rogue Admin cleanup

aaronschachter:
### What's this PR do?

This pull request fixes some few bugs and adds a few enhancements:

* Adds missing `/actions/:id` route - removed in #965 

* Removes deprecated `Route::model` for `campaign_id`, fixes Post links to campaigns - removed in #960 

* Adds link to the Action permalink to a Post view 

* Removes `school` field from our Post gallery GraphQL and checks for `post.schoolId` instead

* Display `NotFound` component for Campaign, School pages when invalid ID

* Adds a `FakerSchoolId` to use valid school ID's from GraphQL when creating fake posts

* Adds a `userId` prop to the `UserInformation` to use if the `user` prop is empty. The posts we generate are using fake user ID's from our dev Northstar database, but we may not have those ID's locally when working in GraphQL.

👀 

The `/signup/:id` permalink also breaks when viewing activity for users that don't exist in your local Northstar DB (which they won't, once you seed your local DB with test data) -- but holding off on fixing that to keep this easier to review (and manually test).

### Relevant tickets

References https://www.pivotaltracker.com/story/show/169549761

### Checklist

### DIFF
--- a/resources/assets/components/ReviewablePost.js
+++ b/resources/assets/components/ReviewablePost.js
@@ -40,6 +40,7 @@ export const ReviewablePostFragment = gql`
     deleted
 
     actionDetails {
+      id
       name
       noun
       verb
@@ -51,10 +52,6 @@ export const ReviewablePostFragment = gql`
     }
 
     schoolId
-    school {
-      id
-      name
-    }
 
     signupId
     signup {
@@ -183,12 +180,12 @@ const ReviewablePost = ({ post }) => {
             details={{
               ID: <Link to={`/posts/${post.id}`}>{post.id}</Link>,
               Campaign: (
-                <a href={`/campaign-ids/${post.campaign.id}`}>
+                <a href={`/campaigns/${post.campaign.id}`}>
                   {post.campaign.internalTitle}
                 </a>
               ),
               Action: (
-                <a href={`/campaign-ids/${post.campaign.id}#actions`}>
+                <a href={`/actions/${post.actionDetails.id}`}>
                   {post.actionDetails.name}
                 </a>
               ),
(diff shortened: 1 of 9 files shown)


katiecrane:
> Adds a userId prop to the UserInformation to use if the user prop is empty. The posts we generate are using fake user ID's from our dev Northstar database, but we may not have those ID's locally when working in GraphQL.

I'm confused about this - if we don't have those IDs locally when working in GraphQL, wouldn't we not have `user` AND not have `userId`?

resources/assets/components/ReviewablePost.js
```diff
@@ -209,7 +211,7 @@ const ReviewablePost = ({ post }) => {
             title="Signup Information"
             details={{
               ID: <a href={`/signups/${post.signupId}`}>{post.signupId}</a>,
-              User: <Link to={`/users/${post.user.id}`}>{post.user.id}</Link>,
+              User: <Link to={`/users/${post.userId}`}>{post.userId}</Link>,
```
katiecrane: Is the switch here because we are guaranteed to have `userId`?

katiecrane:
oh wait @aaronschachter - is it because `userId` is stored on the post so we always have it, but `user.id` we'd be pulling from Northstar via GraphQL?

aaronschachter:
>  is it because userId is stored on the post so we always have it, but user.id we'd be pulling from Northstar via GraphQL?

Yup - that's why. When using Rogue seed data, GraphQL won't find any corresponding Northstar users for the Rogue `post.user_id` values created by our `FakerNorthstarId` provider. The ideal fix would be to get user ID's from the local Northstar (which will always change based on seed data as well) instead of the hardcoded dev ID's from `FakerNorthstarId`.

PR #967, opened 2019-12-04
Fix performance issue on signups with many posts.

DFurnes:
### What's this PR do?

This pull request fixes a performance issue with the `Signup.quantity` field on posts with many signups. Because we updated this to store quantity on individual posts, the `$signup->posts` relationship was loading _all_ the posts on a signup into memory to sum them up:

Screen Shot 2019-12-04 at 4 00 13 PM

Yikes! I adjusted this to create an "aggregate" query with the query builder rather than loading these all and using [Eloquent's Collection](https://laravel.com/docs/6.x/eloquent-collections) class to tally them up in-memory. Much better:

Screen Shot 2019-12-04 at 5 43 54 PM

This also removes the `features.v3QuantitySupport` feature flag since it shipped a long time ago.

👀 

This was causing intermittent failures for our Ghost Inspector user, which has 5,937 posts on their "Merit Over Money" signup. I'll make a backlog ticket to see if we can remove this `Signup.quantity` field altogether on the base response (and make it opt-in `?include=quantity` so we don't have to run this extra query unnecessarily).

### Relevant tickets

References [Pivotal #170107177](https://www.pivotaltracker.com/story/show/170107177).

### Checklist


DFurnes:
It may not be worth doing right now, but we should also think about whether we can delete the `quantity` and `quantity_pending` columns on the `signups` table (since they're no longer used).

@katiecrane Since you were around when we made the switch, any reasons you can think not to?

app/Models/Signup.php
```diff
@@ -197,29 +197,13 @@ public function toQuasarPayload()
...
-        // logic to store everything in the quanity column and not use the quanity_pending
-        // column at all. We only want to return what is in the quanity_pending column
-        // if is the only place quanity is stored.
-        if (! config('features.v3QuantitySupport') || $this->posts->isEmpty()) {
-            if (! is_null($this->quantity_pending) && is_null($this->quantity)) {
-                return $this->quantity_pending;
-            }
-
-            return $this->quantity;
-        }
```
aaronschachter: I was just [asking about trashing all of this](https://dosomething.slack.com/archives/C0YGXUE01/p1575329708161500) as I was looking into aggregating quantity over in #964.

DFurnes: Ah, a good coincidence! 💥

aaronschachter:
Re: saving `quantity` / `quantity_pending` on the Signup -- in the pre-Phoenix-Next days, Gambit used to have default messaging that could work for any campaign and we'd use the fields for a user's photo submission confirmation: e.g. `Great, we've got you down for {{quantity_pending}} {{rb_noun}} {{rb_verb}}. If you've got more {{rb_noun}} {{rb_verb}}, text START to submit another photo`. I believe we removed support for those [tags](https://github.com/DoSomething/gambit-admin/wiki/Tags) long ago -- so I'd imagine we're okay to remove them.

DFurnes:
Oh interesting – it turns out we're updating the `quantity` on the post [when we create it](https://github.com/DoSomething/rogue/blob/7862db45ba1e8770e8736dee071e85a11426a17e/app/Repositories/PostRepository.php#L134) or [update it](https://github.com/DoSomething/rogue/blob/c91c291076697c262b40c9fca4666286e79dd383/app/Repositories/PostRepository.php#L155), so we might be able to simply return it from the database rather than re-calculate it on each "read".

I'd be a teeny bit worried that there are ways we modifying posts _outside_ of the `POST v3/posts` endpoint though (such as running scripts or the `DELETE v3/posts`) that could cause this database field to get out of sync though... if we do need this, we should consider making it a model event.

DFurnes:
> Re: saving `quantity` / `quantity_pending` on the Signup -- in the pre-Phoenix-Next days, Gambit used to have default messaging that could work for any campaign and we'd use the fields for a user's photo submission confirmation: e.g. `Great, we've got you down for {{quantity_pending}} {{rb_noun}} {{rb_verb}}. If you've got more {{rb_noun}} {{rb_verb}}, text START to submit another photo`. I believe we removed support for those [tags](https://github.com/DoSomething/gambit-admin/wiki/Tags) long ago -- so I'd imagine we're okay to remove them.

Iiiinteresting! Yeah, I also vaguely remember the idea of admins "approving" changes in quantity which I think got scrapped along the way.

katiecrane:
@DFurnes in terms of admins approving quantity, I think that kind of implicitly happens now that posts have distinct quantities. If we needed to know the "approved" quantity, we could sum up the quantity of all the approved posts.

As you found, the `quantity` on the Signup does get updated when a new post comes in! I'm VERY in favor of making that trigger as a model event (which I kind of thought it already did...) - so we'll want to keep that. As far as I can tell, `quantity_pending` is no longer used at all, let's trash it!

DFurnes:
Thanks @aaronschachter and @katiecrane! I've updated this PR to read quantity directly from the appropriate database columns in d2fbe377 & update `signups.quantity` via a model event 
in ac939bb. Finally, I dropped `quantity_pending` in c89b806e.

app/Models/Signup.php
```diff
@@ -31,7 +31,7 @@ class Signup extends Model
      *
      * @var array
      */
-    protected $fillable = ['id', 'northstar_id', 'campaign_id', 'quantity', 'quantity_pending', 'why_participated', 'source', 'source_details', 'details', 'created_at', 'updated_at'];
+    protected $fillable = ['id', 'northstar_id', 'campaign_id', 'why_participated', 'source', 'source_details', 'details', 'created_at', 'updated_at'];
```
katiecrane: Should `quantity` be fillable still?

tests/Http/SignupTest.php
```diff
@@ -642,46 +642,6 @@ public function testSignupsIndexAsAdminWithFilters()
...
-        $response = $this->withAdminAccessToken()->getJson('api/v3/signups?orderBy=quantity,asc');
-        $decodedResponse = $response->decodeResponseJson();
-
-        $response->assertStatus(200);
-
-        // Assert results are returned in ascending order.
-        $this->assertEquals(1, $decodedResponse['data'][0]['quantity']);
-        $this->assertEquals(2, $decodedResponse['data'][1]['quantity']);
-    }
-
```
katiecrane: Is this functionality not something we do anymore?

app/Models/Signup.php
DFurnes: Nope – we don't want anyone to set quantity on the post by hand (only via the model event).

katiecrane: Did you mean on the signup?

DFurnes: Oh yep - sorry! 🤦‍♂ We do want folks to set quantity on the posts, but the field on the signup should not be mass-assignable & only updated by this model event.

tests/Http/SignupTest.php
DFurnes: I don't think so (I believe it was added for Domino), and it's a little awkward to write out the test since this is calculated based on the posts. We could update this to sort by a different field if we want coverage for this behavior.

katiecrane: Oh DOMINO blast from the past!

katiecrane: So does this `orderBy` parameter not work anymore? I don't remember seeing anything about that in this PR so I'm having trouble wrapping my head around this

DFurnes: It does! I had removed this test when I originally thought we weren't storing `quantity` on the post anymore, but should probably re-add it to keep the test coverage.

DFurnes: Re-added this test in f0a485c.

PR #970, opened 2019-12-06
Fix order of operations for updating Campaign post counts

aaronschachter:
### What's this PR do?

This pull request makes changes per https://github.com/DoSomething/rogue/pull/964#discussion_r354011950 to update aggregate campaign information in the same place we update aggregate school impact -- which is after the reviewed post's status has been updated.

It also adds unit tests for both aggregate calculations.

👀 

🌵 

### Relevant tickets

References [Pivotal #170111820](https://www.pivotaltracker.com/n/projects/2328687/stories/170111820).

### Checklist

### DIFF
--- a/app/Repositories/PostRepository.php
+++ b/app/Repositories/PostRepository.php
@@ -175,7 +175,7 @@ class PostRepository
     }
 
     /**
-     * Updates a post's status after being reviewed and updates aggregate information.
+     * Creates a Review for the Post, then updates Post status and re-aggregates.
      *
      * @param array Post $post
      * @param string $status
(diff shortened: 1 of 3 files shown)


app/Repositories/PostRepository.php
```diff
@@ -200,7 +200,10 @@ public function reviews(Post $post, $status, $comment = null, $admin = null)
         $post->status = $status;
         $post->save();
 
-        // If the Post has a school ID, update School's aggregate impact for the Action.
+        // Update the "counter cache" on the Post Campaign:
+        $post->campaign->refreshCounts();
```
katiecrane: I'm trying to think of if there is anyway a post could get reviewed without hitting this function - but I guess vast majority would come through here so we probably would never be too far off?

PR #974, opened 2020-01-03
Updating End Date Field on Campaign Page

lindsayrmaher:
### What's this PR do?

This pull request updates the endDate field in the Campaign page so that it is looking for the correct property name `endDate` vs `end_date`

👀 

All of the other file changes are coming from styleCi spacing failures.

### Relevant tickets

References [Pivotal #170090422](https://www.pivotaltracker.com/story/show/170090422).

### Checklist

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -10,5 +10,4 @@ disabled:
   - phpdoc_no_package
   - phpdoc_summary
   - phpdoc_var_without_name
-  - phpdoc_to_comment
   - alpha_ordered_imports
--- a/package-lock.json
+++ b/package-lock.json
@@ -3,20 +3,20 @@
   "lockfileVersion": 1,
   "dependencies": {
     "@apollo/react-common": {
-      "version": "3.1.2",
-      "resolved": "https://registry.npmjs.org/@apollo/react-common/-/react-common-3.1.2.tgz",
-      "integrity": "sha512-6+gTeBZoIyCE6VnHD2EI9Wz+Dm05MBxplUTmVgswzvTe05lUO78SxGgB3XJc9GM/TmNexgCc0eS84vUpG/f6Tg==",
+      "version": "3.1.3",
+      "resolved": "https://registry.npmjs.org/@apollo/react-common/-/react-common-3.1.3.tgz",
+      "integrity": "sha512-Q7ZjDOeqjJf/AOGxUMdGxKF+JVClRXrYBGVq+SuVFqANRpd68MxtVV2OjCWavsFAN0eqYnRqRUrl7vtUCiJqeg==",
       "requires": {
         "ts-invariant": "^0.4.4",
         "tslib": "^1.10.0"
       }
     },
     "@apollo/react-hooks": {
-      "version": "3.1.2",
-      "resolved": "https://registry.npmjs.org/@apollo/react-hooks/-/react-hooks-3.1.2.tgz",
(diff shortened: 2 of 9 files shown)


.styleci.yml
```diff
@@ -10,5 +10,4 @@ disabled:
   - phpdoc_no_package
   - phpdoc_summary
   - phpdoc_var_without_name
-  - phpdoc_to_comment
```
katiecrane: What does this rule do and why don't we want it?

resources/assets/components/Campaign/index.js
```diff
@@ -133,7 +133,7 @@ const Campaign = ({ id }) => {
         <p>{format(parse(campaign.startDate), 'MM/D/YYYY')}</p>
         <h4>End Date</h4>
         <p>
-          {campaign.end_date
+          {campaign.endDate
```
katiecrane: Oh snap, so was this always showing as `-` even if there really was an end date?

lindsayrmaher: Apparently!

.styleci.yml
lindsayrmaher: In a previous version of Laravel this was enabled by default, so this was overriding it to disable. But it looks like it's now disabled by default, so we don't need to override!

katiecrane: Ah gotcha! And I messed up the double negative there, so hard!

package-lock.json
```diff
@@ -5359,14 +5359,14 @@
...
-          "resolved": "https://registry.npmjs.org/wrappy/-/wrappy-1.0.2.tgz",
+          "resolved": false,
           "integrity": "sha1-tSQ9jz7BqjXxNkYFvA0QNuMKtp8=",
           "dev": true,
           "optional": true
         },
         "yallist": {
           "version": "3.0.3",
-          "resolved": "https://registry.npmjs.org/yallist/-/yallist-3.0.3.tgz",
+          "resolved": false,
```
katiecrane: Did you run `npm update` on this branch?

katiecrane: I'm unsure if all these `resolved` fields should have switched to `false`, seems unexpected?

lindsayrmaher: I ran npm install.. but i thought that seemed a lil odd as well.

katiecrane: I'm not sure if having all of these as `false` could cause issues, but it does seem unexpected and seems folks have fixed with [`rm -rf node_modules && npm install`](https://github.com/RocketChat/Rocket.Chat/issues/15408). Seems worth a try? cc: @DFurnes if you have any insights (maybe this is not a problem?)

DFurnes: Yeah, it's frustrating how many little things like this seem to be wonky with npm. It can't hurt to try a lil' `rm -rf node_modules && npm install` and see if that sorts things out!

lindsayrmaher: Ok! i deleted node modules and reinstalled and I'm not seeing anything on my `git diff`

lindsayrmaher: @DFurnes & @katiecrane what do you think about leaving? or maybe i should rebase with master?

DFurnes: Might be worth trying that again, but also deleting `package-lock.json` (to force it to regenerate from the `package.json` constraints). If that doesn't sort things out, though, I'm good with merging it in as-is. 👻

lindsayrmaher: ok, that seems to have updated even more of the packages lol 🙈

DFurnes: That makes sense! For better of for worse, npm is really conservative with updating sub-dependencies when you have a lockfile (even when running `npm update`) so they can fall behind.

And it looks like this fixed the `resolved` weirdness, so that's good too! 😸

katiecrane: Yeah I think it looks less funky now!

PR #976, opened 2020-01-03
Add cursor support to signups index.

DFurnes:
### What's this PR do?

This pull request adds support for using `?cursor[after]=` to paginate through signups.

This is a pre-requisite for DoSomething/graphql#185, which in turn is a pre-requisite for adding pagination to a user's list of signups here. 🐢

This follows the same pattern established in #930 and #937.

The [GraphQL Pagination RFC](https://github.com/DoSomething/rfcs/pull/6) has more context on why we're approaching pagination in this way.

### Relevant tickets

References [Pivotal #170503307](https://www.pivotaltracker.com/story/show/170503307).

### Checklist

### DIFF
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -3,12 +3,13 @@
 namespace Rogue\Models;
 
 use Rogue\Services\GraphQL;
+use Rogue\Models\Traits\HasCursor;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\SoftDeletes;
 
 class Signup extends Model
 {
-    use SoftDeletes;
+    use SoftDeletes, HasCursor;
 
     /**
      * The attributes that should be cast to native types.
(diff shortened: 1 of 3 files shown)


app/Models/Signup.php
```diff
@@ -45,6 +46,20 @@ class Signup extends Model
...
+     * should not contain any sensitive or private data (as they'll be encoded
+     * in cursors).
+     *
+     * This array is manually maintained. It does not necessarily mean that
+     * any of these are actual indexes on the database... but they should be!
+     *
+     * @var array
+     */
+    public static $sortable = [
+        'campaign_id', 'updated_at', 'northstar_id', 'id', 'quantity', 'source',
```
Vmack: What does the 'id' track? Is it the page number?

DFurnes: This is the signup's ID, which is used to uniquely identify it (such as referring to which signup a given post was created under).

PR #977, opened 2020-01-06
Extract signup "gallery" with pagination support.

DFurnes:
### What's this PR do?

This pull request adds a "load more" button to request additional signups on a user's profile. This builds on the work done in #976 and DoSomething/graphql#185.

I extracted a `SignupGallery` to handle the display of this interface element (following the patterns used in `ReviewablePostGallery`).

This is particularly helpful when viewing staff or power-user accounts, since otherwise folks were having to look up signup IDs in Looker in order to get to them from Rogue.

### Relevant tickets

References [Pivotal #170503307](https://www.pivotaltracker.com/story/show/170503307).

### Checklist

### DIFF
--- a/resources/assets/components/SignupGallery.js
+++ b/resources/assets/components/SignupGallery.js
@@ -1,0 +1,90 @@
+import React from 'react';
+import gql from 'graphql-tag';
+import { get, map } from 'lodash';
+import { useQuery } from '@apollo/react-hooks';
+
+import Empty from './Empty';
+import Shell from './utilities/Shell';
+import SignupCard, { SignupCardFragment } from './SignupCard';
+import { updateQuery } from '../helpers';
+
+const SIGNUP_GALLERY_QUERY = gql`
+  query SignupGalleryQuery($userId: String, $cursor: String) {
+    signups: paginatedSignups(userId: $userId, after: $cursor, first: 10) {
+      edges {
+        cursor
+        node {
+          ...SignupCardFragment
+        }
+      }
+
+      pageInfo {
+        hasNextPage
+        endCursor
+      }
+    }
+  }
+
+  ${SignupCardFragment}
+`;
+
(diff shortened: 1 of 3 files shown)


resources/assets/pages/ShowUser.js
```diff
@@ -84,9 +66,7 @@ const ShowUser = () => {
         <h2 className="heading -emphasized -padded">
           <span>Campaigns</span>
         </h2>
-        {map(signups, signup => {
-          return <SignupCard key={signup.id} signup={signup} />;
-        })}
+        <SignupGallery userId={id} />
```
mendelB: This may be irrelevant to this work, but I notice we use `user.id` [above](https://github.com/DoSomething/rogue/pull/977/files#diff-34c9ce668130a04c6c47524a5004c9bcR56) and `id` here -- these should be the same ID, right? Would it make sense to make these consistent?

resources/assets/components/SignupGallery.js
```diff
@@ -0,0 +1,90 @@
...
+  };
+
+  if (signups.length == 0) {
+    return loading ? (
+      <div className="h-content flex-center-xy">
+        <div className="spinner" />
+      </div>
+    ) : (
+      <div className="h-content flex-center-xy">
+        <Empty copy="This user has no signups yet." />
```
mendelB: I love the optimism in this copy 👏

resources/assets/pages/ShowUser.js
DFurnes: Ah yep, they totally should. I updated these to both read from `user.id` in 85fc28d51f34bc6fc5927d8facfe86c6e12c5df9.

DFurnes: I also added a "not found" handler in 6ac1565 just to be safe!

PR #978, opened 2020-01-07
Updates timezone definition in campaign getter

lindsayrmaher:
### What's this PR do?

This pull request updates the seeding for our faker factory to set the time to midnight when creating new campaign dates, as well as add getters for the start and end dates. These getters get our formatting set up and in the correct time zone.

👀 

Originally we approached this by updating our data type query in Graphql, but making the update here helps us to be more consistent and accurate in the dates and time zones we are using for both campaigns and scholarships.

### Relevant tickets

References [Pivotal # 170090422](https://www.pivotaltracker.com/story/show/170090422).

### Checklist

### DIFF
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -2,6 +2,7 @@
 
 namespace Rogue\Models;
 
+use Carbon\Carbon;
 use Rogue\Types\Cause;
 use Rogue\Models\Traits\HasCursor;
 use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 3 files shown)


app/Models/Campaign.php
```diff
@@ -154,6 +155,20 @@ public function setCauseAttribute($value)
...
+     */
+    public function getStartDateAttribute()
+    {
+        $value = $this->attributes['start_date'];
+        if (! $value) {
+            return null;
+        }
+        $date = (new Carbon($value))->format('Y-m-d');
+
+        return new Carbon($date, 'America/New_York');
```
DFurnes: It might be nice to add a comment here explaining why we're setting this to EST.

lindsayrmaher:
oooo @aaronschachter I like that idea!

PR #979, opened 2020-01-07
Quick clean up for adding a helper function

lindsayrmaher:
### What's this PR do?

This pull request adds a helper function to help DRY #978 a bit

👀 

Aaron [pointed this out](https://github.com/DoSomething/rogue/pull/978#pullrequestreview-339507973) on my last PR.

### Relevant tickets

References [Pivotal # 170090422](https://www.pivotaltracker.com/story/show/170090422).

### Checklist

### DIFF
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -138,6 +138,12 @@ function has_include($request, $include)
     return false;
 }
 
+/**
+ * Converts a date to a specific format and timezone
+ * 
+ * @param $value
+ * @param $timezone str
+ */
 function convert_to_date($value, $timezone = 'UTC')
 {
     $date = (new Carbon($value))->format('Y-m-d');
--- a/app/helpers.php
+++ b/app/helpers.php
@@ -138,7 +138,7 @@ function has_include($request, $include)
     return false;
 }
 
-function formatDateAttribute($value, $timezone)
+function formatDateAttribute($value, $timezone = 'UTC')
 {
     $date = (new Carbon($value))->format('Y-m-d');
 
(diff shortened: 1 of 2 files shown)


app/helpers.php
```diff
@@ -137,3 +137,10 @@ function has_include($request, $include)
 
     return false;
 }
+
+function formatDateAttribute($value, $timezone)
```
DFurnes: Nitpicky, but Laravel's convention is to use `snake_case` for helper functions. I'm also wondering if this would be a bit clearer named as `to_date` (since it's more making a "conversion" rather than formatting).

DFurnes: It might also be nice to set a default `$timezone` value here – perhaps `$timezone = 'UTC'`?

lindsayrmaher: would it be too much to say `convert_to_date`? i'm all for making it more clear!

DFurnes: That sounds great!

app/helpers.php
```diff
@@ -137,3 +137,10 @@ function has_include($request, $include)
 
     return false;
 }
+
+function convert_to_date($value, $timezone = 'UTC')
```
DFurnes: Oh sorry, one last thing –– let's add a docblock to this function!

aaronschachter: Would it make sense to default to 'America/New_York' instead of `UTC`?

DFurnes: I think `UTC` makes the most sense as a default, since that's the timezone our dates are stored in. Converting this date to be the beginning of the day in a different timezone is a specific case.

PR #981, opened 2020-01-14
Cleanup post factory

aaronschachter:
### What's this PR do?

This pull request modifies the Model Factory to cleanup a few TODO's in the codebase:

* Creates a base Post factory class, and different factory states for post type + post status per https://github.com/DoSomething/rogue/pull/836#discussion_r365419349

* Modifies the Action factory to collect School ID by default, making it applicable to include a School ID by default in our Post factory

👀 

There's still an outstanding cleanup issue to create `action_stat` records for all approved posts with school id's, but that will be addressed in a future PR if there's time.

### Relevant tickets

References [Pivotal #170693976](https://www.pivotaltracker.com/story/show/170693976).

### Checklist

### DIFF
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -68,7 +68,7 @@ $factory->define(Post::class, function (Generator $faker) {
          * on behalf of their school (because they don't have a school_id set on their
          * Northstar profile).
          */
-        'school_id' => rand(0, 1) ? $this->faker->school_id : null,
+        'school_id' => $this->faker->optional()->school_id,
         'source' => 'phpunit',
     ];
 });
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -10,8 +10,6 @@ use Rogue\Models\Campaign;
 use Rogue\Models\Reaction;
 use Rogue\Types\ActionType;
 use Rogue\Types\TimeCommitment;
-use Rogue\Services\ImageStorage;
-use Symfony\Component\HttpFoundation\File\UploadedFile;
 
 /**
  * Here you may define all of your model factories. Model factories give
@@ -25,7 +23,7 @@ use Symfony\Component\HttpFoundation\File\UploadedFile;
 // Action Factory
 $factory->define(Action::class, function (Generator $faker) {
     return [
-        'name' => $this->faker->slug(),
+        'name' => title_case($this->faker->unique()->words(3, true)), 
         'campaign_id' => factory(Campaign::class)->create()->id,
         'post_type' => 'photo',
         'action_type' => $this->faker->randomElement(ActionType::all()),
@@ -43,11 +41,8 @@ $factory->define(Action::class, function (Generator $faker) {
 // Post Factory
 $factory->define(Post::class, function (Generator $faker) {
     $faker->addProvider(new FakerNorthstarId($faker));
+    $faker->addProvider(new FakerPostUrl($faker));
     $faker->addProvider(new FakerSchoolId($faker));
-
-    $uploadPath = $faker->file(storage_path('fixtures'));
-    $upload = new UploadedFile($uploadPath, basename($uploadPath), 'image/jpeg');
-    $url = app(ImageStorage::class)->put($faker->unique()->randomNumber(5), $upload);
 
     return [
         'campaign_id' => function () {
@@ -65,10 +60,8 @@ $factory->define(Post::class, function (Generator $faker) {
                 'campaign_id' => $attributes['campaign_id'],
             ])->id;
         },
-        'quantity' => $faker->randomNumber(2),
         'northstar_id' => $this->faker->northstar_id,
         'text' => $faker->sentence(),
-        'type' => 'photo',
         'location' => 'US-'.$faker->stateAbbr(),
         /**
          * Although our action models are set to collect school, not all users will create posts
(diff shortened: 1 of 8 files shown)


aaronschachter:
My 9 month old daughter opened that PR too early 👶 💻

database/seeds/DatabaseSeeder.php
```diff
@@ -17,28 +17,33 @@ public function run()
     {
         // Create 10 campaigns with signups & posts.
         factory(Campaign::class, 10)->create()->each(function (Campaign $campaign) {
-            // Add a "default" action so this functions as expected in the "dev" environment.
```
aaronschachter: This doesn't seem applicable anymore.. removing the `default` title and using the factory title seems to work as I'd expect it to.

database/factories/ModelFactory.php
```diff
@@ -22,15 +20,30 @@
  * @see https://laravel.com/docs/5.5/database-testing#writing-factories
  */
 
+// Action Factory
```
aaronschachter: I moved this to the top of the file instead of the bottom, to give more context to setting the `school_id` in the Post factory  below.

database/factories/ModelFactory.php
```diff
@@ -48,55 +61,67 @@
...
 });
 
-// @TODO: These should all extend a photo-less "base" post, instead.
-$factory->defineAs(Post::class, 'text', function (Generator $faker) {
-    $faker->addProvider(new FakerNorthstarId($faker));
-    $faker->addProvider(new FakerSchoolId($faker));
+/**
+ * Photo post factory states.
+ */
+$factory->defineAs(Post::class, 'photo-accepted', function (Generator $faker) use ($factory) {
```
aaronschachter: As I'm thinking about a factory for an `ActionStat` and states for Actions -- I'm considering changing the naming convention for these Post factory states to `[status]-[type]-post`, e.g. `accepted-photo-post`. That way our Actions factory states could be having a convention like `[type]-action`, e.g.`photo-action`, `text-action`.

database/factories/ModelFactory.php
```diff
@@ -48,55 +61,67 @@
             ])->id;
         },
         'northstar_id' => $this->faker->northstar_id,
-        'url' => $url,
```
katiecrane: So a plain `post` will default to a `text` post since we don't include a `url`?

database/factories/ModelFactory.php
```diff
@@ -48,55 +61,67 @@
...
 
-// @TODO: These should all extend a photo-less "base" post, instead.
-$factory->defineAs(Post::class, 'text', function (Generator $faker) {
-    $faker->addProvider(new FakerNorthstarId($faker));
-    $faker->addProvider(new FakerSchoolId($faker));
+/**
+ * Photo post factory states.
+ */
+$factory->defineAs(Post::class, 'photo-accepted', function (Generator $faker) use ($factory) {
+    return array_merge($factory->raw(Post::class), [
```
katiecrane: OMG look at this use of `array_merge`, this makes this SOOOOO much more readable!!! 🎉

database/factories/ModelFactory.php
katiecrane: Oh wait I see now that this is like a base `post` and won't be used on its own will it?

aaronschachter: That's the idea -- this came from a chat @DFurnes and I were having over in https://github.com/DoSomething/rogue/pull/836#discussion_r365403495.

tests/Unit/Models/PostModelTest.php
```diff
@@ -60,7 +60,9 @@ public function testSiblingRelationship()
      */
     public function testBlinkPayload()
     {
-        $post = factory(Post::class)->create();
+        $post = factory(Post::class)->create([
+            'school_id' => 'Example School ID',
```
katiecrane: I'm confused about how this `school_id` gets linked up to `San Dimas High School`?

aaronschachter: It's because we're always returning that school name via [GraphQL mock](https://github.com/DoSomething/rogue/blob/d77c9b879e168deaf5118962c2da4fabcb2c4d75/tests/WithMocks.php#L60) (in honor of Bill and Ted 3 release this summer).

database/factories/ModelFactory.php
```diff
@@ -48,55 +61,67 @@
...
         'text' => $faker->sentence(),
         'location' => 'US-'.$faker->stateAbbr(),
-        // @TODO: Only set school if the action is set to collect school ID.
-        'school_id' => $this->faker->school_id,
+        /**
+         * Although our action models are set to collect school, not all users will create posts
+         * on behalf of their school (because they don't have a school_id set on their
+         * Northstar profile).
+         */
+        'school_id' => rand(0, 1) ? $this->faker->school_id : null,
```
DFurnes: You could use Faker's [`optional` modifier](https://github.com/fzaninotto/Faker#modifiers) to make this a little more readable:

```
'school_id' => $this->faker->optional()->school_id,
```

DFurnes:
I didn't realize this until poking into the Factory API just now, but I think we could use [Factory States](https://laravel.com/docs/5.5/database-testing#factory-states) to simplify this a bit.

They seem to automatically inherit from the base factory, and would let us combine independent `pending`, `accepted`, `photo`, and `text` states rather than making all possible permutations (like `pending-post`, `accepted-post`, `pending-text`, etc.)

tests/Unit/Models/PostModelTest.php
katiecrane: Ah gotcha! What a great lil' easter egg 😂

aaronschachter:
I've implemented the fun suggestions, and they definitely clean this PR up substantially! Thanks for the 👀 @DFurnes and @katiecrane

PR #983, opened 2020-01-17
Clean up environment variables & remove old middleware.

DFurnes:
### What's this PR do?

This pull request cleans up environment variables to use standard values set by our Terraform config, allowing us to [remove these old duplicates](https://github.com/DoSomething/infrastructure/blob/0e39adb7f1c3a90c729a80724a07dcdbba958727/applications/rogue/main.tf#L56-L62). I've cleaned up this application's `.env.example` to remove variables that we shouldn't be using from local machines (like S3 or SQS).

It also removes the unused legacy authentication middleware & Dusk config.

👀

📖

### Relevant tickets

References [Pivotal #170779819](https://www.pivotaltracker.com/story/show/170779819).

### Checklist


config/filesystems.php
```diff
@@ -57,10 +57,10 @@
 
         's3' => [
             'driver' => 's3',
-            'key' => env('S3_KEY'),
-            'secret' => env('S3_SECRET'),
-            'region' => 'us-east-1',
-            'bucket' => env('S3_BUCKET', 'ds-rogue-qa'),
+            'key' => env('AWS_ACCESS_KEY'),
+            'secret' => env('AWS_SECRET_KEY'),
+            'region' => env('AWS_REGION', 'us-east-1'),
```
DFurnes: Note that we don't currently set this from Terraform, so it'll fall back to the `us-east-1` (same as before).

config/filesystems.php
```diff
@@ -57,10 +57,10 @@
 
         's3' => [
             'driver' => 's3',
-            'key' => env('S3_KEY'),
-            'secret' => env('S3_SECRET'),
-            'region' => 'us-east-1',
-            'bucket' => env('S3_BUCKET', 'ds-rogue-qa'),
+            'key' => env('AWS_ACCESS_KEY'),
+            'secret' => env('AWS_SECRET_KEY'),
```
DFurnes: These are set [here](https://github.com/DoSomething/infrastructure/blob/0e39adb7f1c3a90c729a80724a07dcdbba958727/components/iam_app_user/main.tf#L27-L28) in our Terraform config.

app/Http/Kernel.php
```diff
@@ -57,7 +57,6 @@ class Kernel extends HttpKernel
         'auth.basic' => \Illuminate\Auth\Middleware\AuthenticateWithBasicAuth::class,
         'guard' => \DoSomething\Gateway\Server\Middleware\SetGuard::class,
         'scopes' => \DoSomething\Gateway\Server\Middleware\RequireScope::class,
-        'legacy-auth' => \Rogue\Http\Middleware\AuthenticateWithApiKey::class,
```
DFurnes: This middleware (and the test helper below) were unused.

.env.example
```diff
@@ -4,12 +4,9 @@ APP_DEBUG=true
 APP_LOG=daily
 APP_LOG_LEVEL=debug
 APP_URL=http://rogue.test
-ROGUE_API_KEY=abc123
 
-# Feature Flags
+# Feature Flags:
 DS_ENABLE_BLINK=false
-DS_ENABLE_GLIDE=true
-DS_ENABLE_V3_QUANTITY_SUPPORT=false
```
DFurnes: These feature flags no longer exist! 🎏

aaronschachter: I removed this over in #980 -- happy to close that PR out instead, although it seems like maybe we'd want to also remove `app.json` here too?

.env.example
```diff
@@ -31,43 +24,15 @@ REDIS_HOST=127.0.0.1
...
-
 NORTHSTAR_URL=https://identity-dev.dosomething.org
 NORTHSTAR_AUTH_ID=dev-oauth
 NORTHSTAR_AUTH_SECRET=
 NORTHSTAR_CLIENT_ID=dev-machine
 NORTHSTAR_CLIENT_SECRET=
 
-BLINK_URL=
-BLINK_USERNAME=
-BLINK_PASSWORD=
```
aaronschachter: Don't we need this for sending activity events to Customer.io?

DFurnes: We do, but it's not something we want to be doing from local machines (since publishing an event to Blink might have side-effects, like triggering an email). Same argument as why we wouldn't want to encourage using S3 or SQS in the `.env.example`!

.env.example
DFurnes: Oh nice! I'll rebase to fix that conflict and knock these others out of `app.json`.

.env.example
aaronschachter: Ahh, ok.

PR #984, opened 2020-01-30
Adds migration for the Contentful campaign id column in campaigns table

lindsayrmaher:
### What's this PR do?

This pull request creates the migration for adding the contentfulCampaignId to the Campaigns table!

Does making this column nullable make sense? Or should it be required? 

Is the campaignID stored as a string or a number? How could one check this in the future? 

Doing some prep work for getting filters set up in Phoenix on the explore campaigns page.

### Relevant tickets

References [Pivotal #170987264](https://www.pivotaltracker.com/story/show/170987264).

### Checklist

### DIFF
--- a/database/migrations/2020_01_30_155905_add_contentful_campaign_i_d_to__campaign_table.php
+++ b/database/migrations/2020_01_30_155905_add_contentful_campaign_i_d_to__campaign_table.php
@@ -1,0 +1,32 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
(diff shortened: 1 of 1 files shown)


database/migrations/2020_01_30_155905_add_contentful_campaign_i_d_to__campaign_table.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('campaigns', function (Blueprint $table) {
+            $table->string('contentful_campaign_id')->nullable->comment('Add the related contentful campaign id when applicable.');
```
katiecrane: Let's add this column after the `id` column so it doesn't show up all the way at the end after the timestamps and everything

PR #985, opened 2020-01-30
Campaign Cause Filter

mendelB:
### What's this PR do?

This pull request adds a filter to the `/campaigns` index API endpoint & Campaign model to allow filtering by included cause.

**Updated** as per our conversation [in Slack](https://dosomething.slack.com/archives/C0YGXUE01/p1580861249011600) to allow filtering by multiple causes with some more reliable sanitization.

How's the filtering Regex? 
How's the filter name? (I thought `cause` implied a singular cause, as opposed to _included_ cause. I wasn't quite sure between `has_cause` and `includes_cause` though)
Does the test logic make sense?

I went with a `regexp` operator instead of a `LIKE` so that we obtain more precise matches. I couldn't think of an elegant way to allow fuzzy matching without being -in my view- way too generous about the filter. (Some of our cause words overlap, like `lgbtq-rights` & `gender-rights` and we want our cause pages to yield precise matches.)

You can mess around with the regex [here](https://regex101.com/r/D1dRxd/6).

I opted to test for a case where we have three comma-separated causes to ensure we're testing first, last, and regular ole middle words as far as commas are concerned. I hope this made sense!

### Relevant tickets

References [Pivotal #170734672](https://www.pivotaltracker.com/story/show/170734672).

### Checklist

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -60,7 +60,7 @@ class CampaignsController extends ApiController
 
         // Apply scope for campaigns including specified causes:
         if (isset($filters['causes'])) {
-            $query->whereCauses($filters['causes']);
+            $query->withCauses($filters['causes']);
         }
 
         // Experimental: Allow paginating by cursor (e.g. `?cursor[after]=OTAxNg==`):
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -59,8 +59,8 @@ class CampaignsController extends ApiController
         }
 
         // Apply scope for campaigns including specified causes:
-        if (isset($filters['has_cause'])) {
-            $query->whereHasCauses($filters['has_cause']);
+        if (isset($filters['causes'])) {
+            $query->whereCauses($filters['causes']);
         }
 
         // Experimental: Allow paginating by cursor (e.g. `?cursor[after]=OTAxNg==`):
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -59,8 +59,8 @@ class CampaignsController extends ApiController
         }
 
         // Apply scope for campaigns including specified causes:
-        if (isset($filters['causes'])) {
-            $query->withCauses($filters['causes']);
+        if (isset($filters['cause'])) {
+            $query->withCauses($filters['cause']);
         }
 
         // Experimental: Allow paginating by cursor (e.g. `?cursor[after]=OTAxNg==`):
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -58,9 +58,9 @@ class CampaignsController extends ApiController
             }
         }
 
-        // Apply scope for campaigns including specified cause:
+        // Apply scope for campaigns including specified causes:
         if (isset($filters['has_cause'])) {
-            $query->whereHasCause($filters['has_cause']);
+            $query->whereHasCauses($filters['has_cause']);
         }
 
         // Experimental: Allow paginating by cursor (e.g. `?cursor[after]=OTAxNg==`):
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -58,6 +58,11 @@ class CampaignsController extends ApiController
             }
         }
 
+        // Apply scope for campaigns including specified cause:
+        if (isset($filters['has_cause'])) {
+            $query->whereHasCause($filters['has_cause']);
+        }
+
         // Experimental: Allow paginating by cursor (e.g. `?cursor[after]=OTAxNg==`):
         if ($cursor = array_get($request->query('cursor'), 'after')) {
             $query->whereAfterCursor($cursor);
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -136,8 +136,8 @@ class Campaign extends Model
         // Sanitize the input to prevent melicious regex pattern injection:
         $sanitizedCause = preg_replace("/[^\w-]/", '', $cause);
 
-        // Regex matches on comma separated words to ensure precise match excepting first
-        // and last words (matching on only a following or preceding comma).
+        // Regex tests for complete match against one of the comma separated causes.
+        // (https://regex101.com/r/RaDXos/5).
         return $query->where('cause', 'regexp', '(^|,)'.$sanitizedCause.'(,|$)');
     }
 
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -112,7 +112,6 @@ class Campaign extends Model
     }
 
     /**
-<<<<<<< HEAD
      * Scope a query to only include campaigns with an associated Contentful 'Website' entry.
      */
     public function scopeWhereHasWebsite($query)
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -136,7 +136,8 @@ class Campaign extends Model
         // Sanitize the input to prevent melicious regex pattern injection:
         $sanitizedCause = preg_replace("/[^\w-]/", '', $cause);
 
-        // Regex tests for complete match against one of the comma separated causes.
+        // Regex matches on comma separated words to ensure precise match.
+        // Accounts for first and last words only having a comma on one side.
         // (https://regex101.com/r/RaDXos/5).
         return $query->where('cause', 'regexp', '(^|,)'.$sanitizedCause.'(,|$)');
     }
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -4,6 +4,7 @@ namespace Rogue\Models;
 
 use Carbon\Carbon;
 use Rogue\Types\Cause;
+use Illuminate\Support\Str;
 use Rogue\Models\Traits\HasCursor;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Builder;
@@ -134,7 +135,7 @@ class Campaign extends Model
     public function scopeWhereHasCause($query, $cause)
     {
         // Sanitize the input to prevent melicious regex pattern injection:
-        $sanitizedCause = preg_replace("/[^\w-]/", '', $cause);
+        $sanitizedCause = Str::slug($cause);
 
         // Regex matches on comma separated words to ensure precise match.
         // Accounts for first and last words only having a comma on one side.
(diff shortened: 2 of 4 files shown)


mendelB:
irrelevant to this PR but I noticed this in the Campaigns endpoint documentation:
> These endpoints require the role `admin` or `staff` to use.

But this doesn't seem to be the case? I seem to be able to access this endpoint in incognito, and don't see any auth on this controller.

DFurnes:
> But this doesn't seem to be the case? I seem to be able to access this endpoint in incognito, and don't see any auth on this controller.

Oh that definitely sounds like a lie. Let's remove that!

app/Models/Campaign.php
```diff
@@ -110,6 +110,17 @@ public function scopeWhereClosed($query)
...
     }
 
+    /**
+     * Scope a query to only include campaigns containing specified cause within
+     * "cause" field.
+     */
+    public function scopeWhereHasCause($query, $cause)
+    {
+        // Regex matches on comma separated words to ensure precise match excepting first
+        // and last words (matching on only a following or preceding comma).
```
katiecrane: I might be reading this wrong but I'm confused - can you explain to me what exactly we're doing here?

mendelB: I basically used the comma separation as 'word boundaries', so that we'd get precise matches (so the `filter[has_cause]=rights` wouldn't yield `lgbtq-rights` & `gender-rights`). The exceptions are the first and last causes, which do not have a preceding comma (first) or following comma (last), so in those case the regex only looks for the one comma they _do_ have.

Does this make sense? You can play with the regex matching [here](https://regex101.com/r/D1dRxd/6) to see what I mean.

D'you think this comment could use a little re-writing for clarity?

katiecrane: AH that sandbox thing really helped - is this summary correct? We look at each comma-separated cause, and signal a match if we have a complete match on one of the comma-separated causes?

DFurnes: I'm not sure how long that link sticks around, but if it's a forever-thing it'd be nice to include in the comment! I think seeing it actually visualized like that (and being able to type in different example inputs) is super helpful.

app/Models/Campaign.php
```diff
@@ -110,6 +110,17 @@ public function scopeWhereClosed($query)
...
+        return $query->where('cause', 'regexp', '(^|,)'.$cause.'(,|$)');
```
DFurnes: Is this possible to abuse? (Like, could someone enter a super-nasty regex as the "cause" and lock things up?)

mendelB: Oof, y'know I made a mental note to double-check regarding injection, and then completely. forgot to! I'll look into it!

app/Models/Campaign.php
mendelB: >  is this summary correct? We look at each comma-separated cause, and signal a match if we have a complete match on one of the comma-separated causes?

Precisely so!

> I'm not sure how long that link sticks around, but if it's a forever-thing it'd be nice to include in the comment! I think seeing it actually visualized like that (and being able to type in different example inputs) is super helpful.

That's a really good idea! I'll look into that

katiecrane: Ahhh okay the comma stuff had me confused - I took `matches on comma separated words` to include the fact that the first word would not have a comma before it and same idea for last word. I think this would have been more clear for me:
```
Regex matches on comma separated words to ensure precise match.  Accounts for first and last words only having a comma on one side.
```
And I love the idea of including that link if it's forever!

app/Models/Campaign.php
mendelB: Added a line to sanitize the input (by stripping out all non-alphanumeric (aside from hyphens) characters) in https://github.com/DoSomething/rogue/pull/985/commits/75de1925df325650338bcccec570d42a3d66ea51

app/Models/Campaign.php
mendelB: Simplified the comment and added a link to the regex sandbox in https://github.com/DoSomething/rogue/pull/985/commits/3c1db33a7ffa7710d08138a52dcdacde12d4f420

mendelB: Oh whoops I missed your suggestion @katiecrane I like that clarification better!

katiecrane: love it!

katiecrane: oh lol either way!

mendelB: Updated in https://github.com/DoSomething/rogue/pull/985/commits/393e1f7d9ffe106bd6d16885d39cc5281da6d3cf

app/Models/Campaign.php
```diff
@@ -110,6 +110,21 @@ public function scopeWhereClosed($query)
...
     }
 
+    /**
+     * Scope a query to only include campaigns containing specified cause within
+     * "cause" field.
+     */
+    public function scopeWhereHasCause($query, $cause)
+    {
+        // Sanitize the input to prevent melicious regex pattern injection:
+        $sanitizedCause = preg_replace("/[^\w-]/", '', $cause);
```
weerd: Doesn't Laravel have some helper for this? Quick check and I found this [`prepareForValidation()`](https://laravel.com/docs/6.x/validation#prepare-input-for-validation) method. But I also think there's some quick helper, I just can't remember well and my google-foo is failing me, haha 😅

mendelB: Hmm interesting! I'm not sure where we'd put the `prepareForValidation` method here? I googled around for input sanitization because I honestly did expect there to be an easy built-in method, but couldn't really find anything beyond folks saying to use `pre_replace`.
But maybe an interesting spin on this might be to validate the request if the `has_cause` filter is set using something like [not_regex](https://laravel.com/docs/6.x/validation#rule-not-regex)?

weerd: Sorry, I edited my question after since I realized it wasn't super clear... more-so meant possibility of just using `Str::slug()` method like it does in `prepareForValidation()` to just convert it to a slug-ified string, which I suspect would get rid of any regexy-ness 🤔

weerd: My suggestion may not be super robust though, and `preg_replace()` may just be better!

weerd: I tried this in Tinker and seemed to work well:

```php
>>> Illuminate\Support\Str::slug('(^|,)addiction(,|$)');
=> "addiction"
```

mendelB: Ohhh gotcha, that's an interesting idea! I think it'll indeed essentially be running the same sanitization to filter out any special characters which is convenient. I guess my one reservation is that it may be less clear as to what we're doing? I dunno, I'm kinda into it -- I assume our cause list will always be the same slug-like format!

weerd: Maybe to keep it clear, don't inline it and keep the variable named `$sanitizedCause`, plus this way if the sanitized cause ends up being empty after the sanitization, you could do a check on this variable and eschew the query if it turns out it was bad (and now empty) input? 🤔

docs/endpoints/campaigns.md
```diff
@@ -11,8 +9,9 @@ GET /api/v3/campaigns
 ### Optional Query Parameters
 
 - **filter[column]** _(string)_
-  - Filter results by the given column: `id`, `is_open`
+  - Filter results by the given column: `id`, `is_open`, `has_cause`
   - You can filter by more than one value for the ID column, e.g. `/campaigns?filter[id]=121,122`
+  - The `has_cause` filter (`/campaigns?filter[has_cause]=education`) will return campaigns _including_ the specified cause. (Can filter by one value only).
```
aaronschachter: Out of scope for this PR, but I'm curious if we've talked at all about converting the `cause` single value text field into a table like `campaign_causes`?  This is how the Ashes taxonomy worked -- and would allow us to filter by multiple causes, and also avoid using regex to find results...

app/Models/Campaign.php
mendelB: I gave it a go in https://github.com/DoSomething/rogue/pull/985/commits/4411570e342c27a42ca63509f5fa3a260b622e59! If the string is empty, the filter just won't yield any matches since we'll be searching for a comma-separated empty string! 😌

docs/endpoints/campaigns.md
mendelB: We didn't really have that discussion pertaining to this work, but I agree that that would be a helpful update, and actually formalize cause association! (Definitely feels more relevant now that we're starting to actually _use_ this taxonomy)

mendelB:
Rebased to resolve merge conflicts. 
Would love one more 👀 👍 on this if anyone has a chance since we've updated a fair bit per feedback updates!

app/Http/Controllers/CampaignsController.php
```diff
@@ -58,6 +58,11 @@ public function index(Request $request)
             }
         }
 
+        // Apply scope for campaigns including specified cause:
+        if (isset($filters['has_cause'])) {
```
DFurnes: Sorry for not thinking about this earlier, but would it be clearer to have this as just `cause`?

app/Models/Campaign.php
```diff
@@ -126,6 +127,21 @@ public function scopeWhereDoesNotHaveWebsite($query)
...
     }
 
+    /**
+     * Scope a query to only include campaigns containing specified cause within
+     * "cause" field.
+     */
+    public function scopeWhereHasCause($query, $cause)
+    {
+        // Sanitize the input to prevent melicious regex pattern injection:
+        $sanitizedCause = Str::slug($cause);
```
DFurnes: Also just thought of this – would it be clearer (and perhaps safer) to match against the list of valid causes?

```php
$sanitizedCause = array_intersect([strtolower($cause)], Cause::all());
```

app/Http/Controllers/CampaignsController.php
```diff
@@ -58,6 +58,11 @@ public function index(Request $request)
             }
         }
 
+        // Apply scope for campaigns including specified causes:
+        if (isset($filters['causes'])) {
```
DFurnes: This is super nitpicky, but should this be `cause` (so, that it's singular like `?filter[id]=1,2,3`)?

mendelB: Ya know what, I may have misread your comment in the Slack thread. Will update! This would also be consistent with the [`tag` filter](https://github.com/DoSomething/rogue/blob/cb4da53aba1f74c7f77c722b3085b7f05c186bdf/app/Http/Controllers/PostsController.php#L89-L92)

mendelB: Updated in https://github.com/DoSomething/rogue/pull/985/commits/2efce6c03f0c82eb946017e85d6001836ddfdc8f

PR #988, opened 2020-02-03
Campaign Has-Website Filter

mendelB:
### What's this PR do?

This pull request adds a filter to the Campaigns index API endpoint to scope out campaigns with/without an assigned `contentful_campaign_id`

How's the naming?
Does it make sense to have both scopes for with/without? (It seemed like a logical way to apply this filter).

Now that we have this new column 'linking' to associated 'Campaign Websites' in Contentful (#450), this filter will allow us to properly scope out these campaigns via the API

### Relevant tickets

References [Pivotal #171027117](https://www.pivotaltracker.com/n/projects/2401401/stories/171027117).

### Checklist

### DIFF
--- a/app/Http/Controllers/CampaignsController.php
+++ b/app/Http/Controllers/CampaignsController.php
@@ -54,7 +54,7 @@ class CampaignsController extends ApiController
             if (filter_var($filters['has_website'], FILTER_VALIDATE_BOOLEAN)) {
                 $query->whereHasWebsite();
             } else {
-                $query->whereDoesntHaveWebsite();
+                $query->whereDoesNotHaveWebsite();
             }
         }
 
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/CampaignsController.php
```diff
@@ -49,6 +49,15 @@ public function index(Request $request)
             }
         }
 
+        // Apply scope for the has_website filter:
+        if (isset($filters['has_website'])) {
+            if (filter_var($filters['has_website'], FILTER_VALIDATE_BOOLEAN)) {
+                $query->whereHasWebsite();
+            } else {
+                $query->whereDoesntHaveWebsite();
```
aaronschachter: Minor, but what do you think about renaming as `whereDoesNotHaveWebsite` ? I think it's a pinch more readable but not a biggie :neckbeard:

mendelB: That's a good call! Updated in https://github.com/DoSomething/rogue/pull/988/commits/c9672bf202399761a4e955099741f3f9bc37a483

PR #989, opened 2020-02-11
Add 'DELETE api/v3/users/:id' endpoint.

DFurnes:
### What's this PR do?

This pull request adds a `DELETE api/v3/users/:id`, which we can use to automate this portion of the deletion process from Northstar (rather than relying on a developer running an Artisan command in a one-off dyno).

Does it do the thing it says it does? Did I properly lock it down? 👀

This is part of the [Self-Service Account Deletion](https://docs.google.com/document/d/1vj_AcfZ_I92dtBH6y83RhSFYJCshQy2wAXhpecAWlWo/edit#) work.

### Relevant tickets

References [Pivotal #170953002](https://www.pivotaltracker.com/story/show/170953002).

### Checklist


app/Http/Controllers/UsersController.php
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public function destroy($id)
+    {
+        // Find the user's signups, anonymize, and soft-delete.
+        $signups = Signup::withTrashed()->where('northstar_id', $id);
+        foreach ($signups->cursor() as $signup) {
+            $signup->update([
+                'why_participated' => null,
+                'details' => null,
+            ]);
```
mendelB: Sorry, I know this is not new functionality, but out of curiosity -- is the idea that we purge the user's signups & posts of data & _soft_ delete so that we still retain their count for our total impact numbers?

DFurnes: Yes! The (totally undocumented) intent here is that we want to keep as much non-identifiable information about deleted activity as we can so that we could infer trends down the line. (This came out of an audit that Megha & I did when initially building this out to see which fields could possibly include PII & what had anonymized analytic value).

DFurnes: I imagine that soft-deleted records would _not_ count for impact in Quasar (but I'm not 100% sure on that).

mendelB: Ah ok, that makes more sense, thanks for clarifying!

PR #992, opened 2020-02-12
Script to clean up causes

katiecrane:
### What's this PR do?

This pull request creates an artisan command to clean up causes! It grabs all campaigns with `cause` set, then it takes that cause and makes it lowercase and also replaces spaces with hyphens. This operation is performed even if the cause is already in that format, so messed up causes will be fixed and ones that are already formatted correctly will remain as they are.

Will this clean up all the funky causes?

We don't want to have to worry about transforming these causes in other places, let's store them all the same!

### Relevant tickets

References [Pivotal #171159859](https://www.pivotaltracker.com/story/show/171159859).

### Checklist

### DIFF
--- a/app/Console/Commands/StandardizeCauses.php
+++ b/app/Console/Commands/StandardizeCauses.php
@@ -43,7 +43,7 @@ class StandardizeCauses extends Command
         // Grab every campaign that has a cause set
         $campaignsWithCause = Campaign::whereNotNull('cause')->get();
 
-        foreach($campaignsWithCause as $campaign) {
+        foreach ($campaignsWithCause as $campaign) {
             $this->info('Updating campaign '.$campaign->id);
 
             $oldCause = $campaign->getOriginal('cause');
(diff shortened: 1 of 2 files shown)


app/Console/Commands/StandardizeCauses.php
```diff
@@ -0,0 +1,67 @@
...
+        $campaignsWithCause = Campaign::whereNotNull('cause')->get();
+
+        foreach ($campaignsWithCause as $campaign) {
+            $this->info('Updating campaign '.$campaign->id);
+
+            $oldCause = $campaign->getOriginal('cause');
+            $this->info('--From:'.$oldCause);
+
+            // Make the causes lowercase and replaces spaces with hyphens
+            $newCause = str_replace(' ', '-', strtolower($oldCause));
```
aaronschachter: Just checking -- do any causes have trailing spaces we need to look for?

katiecrane: OOO that is a good call - I'm not sure if they do but I'm going to add a `trim()` to this so we don't have to worry if they do

katiecrane:
@aaronschachter I did a non-trivial refactor to get the whitespace trimming working - would you mind taking another 👀 when you get a chance?

PR #1001, opened 2020-03-17
Add <5 minute time commitment

katiecrane:
### What's this PR do?

This pull request adds another option for `time_commitment` on actions, <5 minutes. For the numerical value that is stored in the database, we store those in terms of hours. So we have like `3.0+` for "more than 3 hours" and we have `<0.5` for "less than 30 minutes. `5/60` is `0.083333333` so we store `<0.083`. Is that too specific? Should we do `<0.08` instead? None of them look particularly "nice" to me so I chose going with `0.083` since it's closer to `5/60` 🤷‍♀ 

Does this add <5 minutes as an option for time commitment for actions? Any tests needed?

We want to add this for some quicker actions. We have <30 but that makes people think it might take 30 minutes, saying <5 shows them that it really will be quick!

### Relevant tickets

References [Pivotal #171753530](https://www.pivotaltracker.com/story/show/171753530).

### Checklist

### DIFF
--- a/app/Types/TimeCommitment.php
+++ b/app/Types/TimeCommitment.php
@@ -4,6 +4,7 @@ namespace Rogue\Types;
 
 class TimeCommitment extends Type
 {
+    private const LESS_THAN_FIVE_MINUTES = '<0.083';
     private const LESS_THAN_THIRTY_MINUTES = '<0.5';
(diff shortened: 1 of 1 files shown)


lindsayrmaher:
`<.083` doesn't feel too specific to me. especially since all the other time frames are measured the same way. This makes sense to me!

PR #1009, opened 2020-04-01
Attach client identifier to requests.

DFurnes:
### What's this PR do?

This pull request simply attaches this application's name to our GraphQL requests, so that we can segment these out in Apollo Engine when looking at metrics:

Screen Shot 2020-04-01 at 3 59 29 PM

👀

Support for this was added in DoSomething/graphql#213.

I made a similar change to Phoenix in DoSomething/phoenix-next#2011.

### Relevant tickets

References [Pivotal #171001230](https://www.pivotaltracker.com/story/show/171001230).

### Checklist

### DIFF
--- a/resources/assets/graphql.js
+++ b/resources/assets/graphql.js
@@ -44,5 +44,6 @@ export default uri => {
   return new ApolloClient({
     link: ApolloLink.from([errorLink, authLink, persistedLink, httpLink]),
     cache: new InMemoryCache(),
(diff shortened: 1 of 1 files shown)


DFurnes:
Oh, good point! It doesn't look like that client has built-in support like Apollo, but it's definitely worth seeing if we can sort out in the future (and same thing for Gambit/Gambit Admin)!

PR #1013, opened 2020-05-01
Public Images & CORS

mendelB:
### What's this PR do?

This pull request (experimentally) adds a new `asset` middleware group for public `/images/{hash}` routes to enable assigning the https://github.com/fruitcake/laravel-cors handler to this route specifically since it doesn't fall into our `api` or `web` groups.

To be completely honest, I've been floundering, trying to figure this one out. I haven't been able to validate that this will work locally beyond assuming this should work logically! (For some reason, I'm unable to replicate the CORS issues I'm seeing on the dev environment related to the issue below locally).

Does this make sense? Is the middleware name ok?

See https://dosomething.slack.com/archives/CUQMU4Q6B/p1588343622106600

### Relevant tickets

References [Pivotal #172624016](https://www.pivotaltracker.com/story/show/172624016).

### Checklist

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -45,7 +45,7 @@ class Kernel extends HttpKernel
         ],
 
         'asset' => [
-          \Barryvdh\Cors\HandleCors::class,
+            \Barryvdh\Cors\HandleCors::class,
         ],
     ];
 
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -43,6 +43,10 @@ class Kernel extends HttpKernel
             // 'throttle:60,1',
             \Barryvdh\Cors\HandleCors::class,
         ],
+
+        'asset' => [
+          \Barryvdh\Cors\HandleCors::class,
+        ],
     ];
 
     /**
--- a/routes/api.php
+++ b/routes/api.php
@@ -10,7 +10,7 @@
  */
 
 // Assets
-$router->get('images/{hash}', 'Web\ImagesController@show');
+$router->get('images/{hash}', 'Web\ImagesController@show')->middleware('asset');
 
 // v3 routes
 $router->group(['prefix' => 'api/v3', 'middleware' => ['guard:api']], function () {


mendelB:
Awesome, thanks! I'll give this a go 🙏

DFurnes:
> To be completely honest, I've been floundering, trying to figure this one out. I haven't been able to validate that this will work locally beyond assuming this should work logically! (For some reason, I'm unable to replicate the CORS issues I'm seeing on the dev environment related to the issue below locally).

This is tricky to test out IIRC, since it doesn't actually set the `Access-Control-Allow-Origin` header unless you provide an `Origin` on the request, so I think you'd need something like…

```
$ curl -I http://rogue.test/images/LkanRbjnNp -H "Origin: https://example.com"
```

mendelB:
> > To be completely honest, I've been floundering, trying to figure this one out. I haven't been able to validate that this will work locally beyond assuming this should work logically! (For some reason, I'm unable to replicate the CORS issues I'm seeing on the dev environment related to the issue below locally).
> 
> This is tricky to test out IIRC, since it doesn't actually set the `Access-Control-Allow-Origin` header unless you provide an `Origin` on the request, so I think you'd need something like…
> 
> ```
> $ curl -I http://rogue.test/images/LkanRbjnNp -H "Origin: https://example.com"
> ```

😖 I did actually try that, imitating your call in https://github.com/DoSomething/northstar/pull/797, but I couldn't seem to activate a failure! (Maybe because this wasn't an `XMLHttpRequest`?)

DFurnes:
Since cURL is a command-line tool and doesn't have to worry about cross-origin attacks, it's not going to mark the request as a failure like a browser will. It'll just include or exclude the `Access-Control-Allow-Origin` header (that the browser then interprets to allow or disallow the request).

One thing that it turns out is _extra_ tricky here is that these images are getting cached in Fastly (and so we may have a cached version that didn't set this header). This means we may have to set this rule on that end, since Rogue isn't actually getting hit for these requests...

mendelB:
> Since cURL is a command-line tool and doesn't have to worry about cross-origin attacks, it's not going to mark the request as a failure like a browser will. It'll just include or exclude the `Access-Control-Allow-Origin` header (that the browser then interprets to allow or disallow the request).
> 
> One thing that it turns out is _extra_ tricky here is that these images are getting cached in Fastly (and so we may have a cached version that didn't set this header). This means we may have to set this rule on that end, since Rogue isn't actually getting hit for these requests...

Aha! You know what, that would probably explain why this is _still_ not working on dev/QA:
image


I'm indeed seeing the `Access-Control-Allow-Origin` locally, but not on QA!

DFurnes:
Yeah, bummer. Ideally, I think we'd just save `Access-Control-Allow-Origin: *` from the app & call it a day, but that unfortunately doesn't seem to be something this package lets us do and so there's no good caching strategy there. 😕 So to Fastly it is!

We can follow the example [from this old issue](https://github.com/DoSomething/internal/issues/416), but this time we'll want to apply it using Terraform. That should be pretty straightforward, I think. Let me know if you want to pair on it!

mendelB:
> but that unfortunately doesn't seem to be something this package lets us do and so there's no good caching strategy there

Ah do you mean because it sets the header per request origin specifically, instead of the `*`?

> We can follow the example from this old issue, but this time we'll want to apply it using Terraform. That should be pretty straightforward, I think. Let me know if you want to pair on it!

I'd love to!! What would the process be for this work, would we need a platforms ticket for this specifically? (Also, we're not publically launching Volunteer Credits yet, so this wouldn't be a _time pressing_ blocker.)

DFurnes:
> Ah do you mean because it sets the header value per request origin specifically, instead of the *?

Yep! Although hmm –– now that I look at it we do [`Vary`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary) based on the `Origin` header so I'd actually assume this would work... we'd just be hitting the backend a bit more than we perhaps want to.

DFurnes:
So all that to say, I'm actually back to being surprised this isn't already working! I'm totally down to spend some time looking into this sometime next week (just snag a slot on my calendar), and if it does end up needing Fastly work we can pair on that implementation.

mendelB:
The plot thickens 🥁 

Ok, much appreciated! I'll throw some time on your cal next week for some 🕵️

PR #1015, opened 2020-05-05
Make voter-reg referrals visible to the referrer

aaronschachter:
### What's this PR do?

This pull request modifies the `scopeWhereVisible` function in our `Post` model to allow a referrer to see completed voter registrations they have referred. This fixes a [bug found while testing the new Alpha OVRD "Get 3 friends to register" section](https://www.pivotaltracker.com/story/show/172417306/comments/213979491) -- staff users can see their referrals but non-staff users cannot, without these changes in place. I've tested my local Phoenix against this branch locally and confirmed a non-staff can now see their referrals.

👀 

We may eventually want to display more statuses than our completed statuses if we get around to building out displaying a [list of non-completed voter registration referrals to the alpha](https://www.pivotaltracker.com/story/show/171043026) - -but we can change this code up then should we move forward with that feature.

### Relevant tickets

References [Pivotal #172417306](https://www.pivotaltracker.com/story/show/172417306).

### Checklist

### DIFF
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -1140,6 +1140,38 @@ class PostTest extends TestCase
     }
 
     /**
+     * Test for retrieving voter registration posts as a referrer.
+     * A referrer should see all completed voter registrations that they have referred.
+     *
+     * GET /api/v3/posts
+     * @return void
+     */
+    public function testPostsIndexAsVoterRegistrationReferrer()
+    {
+        $referrerUserId = $this->faker->unique()->northstar_id;
+
+        // Add two completed voter reg referrals for our referrer, which should be visible to them.
+        factory(Post::class)->states('voter-reg', 'register-form')->create(['referrer_user_id' => $referrerUserId]);
+        factory(Post::class)->states('voter-reg', 'register-OVR')->create(['referrer_user_id' => $referrerUserId]);
+
+        // Add a non-completed voter reg referral for our referrer, which shouldn't be visible.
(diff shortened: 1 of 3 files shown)


tests/Http/PostTest.php
```diff
@@ -1139,6 +1139,38 @@ public function testPostsIndexAsOwnerHiddenPosts()
...
+
+        // Add a completed voter reg referral for a different referrer, which shouldn't be visible.
+        factory(Post::class)->states('voter-reg', 'register-form')->create(['referrer_user_id' => $this->faker->unique()->northstar_id]);
+
+        // Add a pending photo post, which shouldn't be visible.
+        factory(Post::class)->states('photo', 'pending')->create();
+
+        $response = $this->withAccessToken($referrerUserId)->getJson('api/v3/posts');
+        $response->assertStatus(200);
+        $response->assertJsonCount(2, 'data');
```
katiecrane, line 1171: Do you think we need to make sure the 2 posts visible here are the correct 2? I feel like odds that something breaks so the wrong 2 posts are showing up are pretty slim but in terms of covering all our bases...

aaronschachter: I was following by example from the other tests, but agree that it couldn't hurt! Updated this branch with [`assertJsonFragment`](https://laravel.com/docs/5.7/http-tests#assert-json-fragment) checks for the expected post ID's.

katiecrane: @aaronschachter looks awesome!!

PR #1017, opened 2020-05-11
Only allow admins to edit Contentful Campaign ID field

katiecrane:
### What's this PR do?

This pull request hides the Contentful Campaign ID field on the edit campaign view to everyone except admins! It does NOT hide this field on the view campaign page.

As admin:
Screen Shot 2020-05-11 at 3 19 58 PM

As staff:
Screen Shot 2020-05-11 at 3 21 18 PM

Will this prevent anyone who is not an admin from editing this field? Can admins still edit this field?

If we link to a Contentful page in Rogue that's not published, it'll take down `/campaigns`! So we want to make sure this field is only being added when it really should be.

### Relevant tickets

References [Pivotal #172383488](https://www.pivotaltracker.com/story/show/172383488).

### Checklist

### DIFF
--- a/resources/views/campaigns/edit.blade.php
+++ b/resources/views/campaigns/edit.blade.php
@@ -24,6 +24,7 @@
                         >
                     </div>
 
+                @if (Auth::user()->role === 'admin')
                     <div class="form-item">
                         <label class="field-label">Contentful Campaign ID <em>(optional)</em></label>
                         <input type="text" name="contentful_campaign_id" class="text-field"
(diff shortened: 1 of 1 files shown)


resources/views/campaigns/edit.blade.php
```diff
@@ -24,6 +24,7 @@
                         >
                     </div>
 
+                @if (Auth::user()->role === 'admin')
```
mendelB, line 27: Would it be worth using the `isAdmin` helper for this? (I'm biased because I updated those helpers recently 🙃)
https://github.com/DoSomething/rogue/blob/261c3a67d16be6e9e982352cc3aa70c0e262e2a8/app/helpers.php#L94-L107

mendelB: LOL, totally I think the advantage here is that the helper keeps the logic to determine an admin encapsulated to one central location.

katiecrane: oh I forgot we had that!!!!! will update to use it

PR #1018, opened 2020-05-15
Adds referral_post_created and referral_post_updated Customer.io events

aaronschachter:
### What's this PR do?

This pull request sends two new events, `referral_post_created` and `referral_post_updated` to Customer.io when a post with a `referral_user_id` is created or updated. Each event is created on behalf of the referrer -- so we can send email or SMS notifications to an alpha if a beta completes a voter registration from their OVRD page.

It's passing along a minimal set of attributes for now to keep this easier to maintain and review. The majority of the changes in this PR are for new classes for a Customer.io API client and the new `CreateCustomerIoEvent` job that will be used to create these events.

👀 
Once this is merged,it should be straight forward enough to manually test this out with QA Test Import form in Chompy - verifying the events are created when a referral voter reg post is created or updated, and that the events are not created when the post does not have a referrer.

I wanted the `CreateCustomerIoEvent` job to be very general, so it can eventually be used to deprecate the various jobs that post to Blink in order to create Customer.io events. The only tricky piece there would be sending along a `POST /v2/messages?origin=signup` request to the Gambit API on signup creation (Blink handles that currently), but this work would all be for an epic cleanup sprint anyway. 

### Relevant tickets

*  [Pivotal #172318228](https://www.pivotaltracker.com/story/show/172318228).
* Discussion in [#dev-applications](https://dosomething.slack.com/archives/CUQMU4Q6B/p1588960979193300?thread_ts=1588365799.150600&cid=CUQMU4Q6B)

### Checklist

### DIFF
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -83,7 +83,6 @@ $factory->define(Post::class, function (Generator $faker) {
          */
         'school_id' => $this->faker->optional()->school_id,
         'source' => 'phpunit',
-        'referrer_user_id' => $this->faker->optional()->northstar_id,
     ];
 });
 
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -1360,6 +1360,29 @@ class PostTest extends TestCase
     }
 
     /**
+     * Test that updating a referral post sends an event to Customer.io.
+     *
+     * PATCH /api/v3/posts/:id
+     * @return void
+     */
+    public function testUpdatingAReferralPost()
+    {
+        $post = factory(Post::class)->create([
+            'referrer_user_id' => $this->faker->northstar_id,
+            'type' => 'voter-reg',
+            'status' => 'step-1',
+        ]);
+
+        $this->mock(CustomerIo::class)
+          ->shouldReceive('trackEvent')
+          ->with($post->referrer_user_id, 'referral_post_updated', $post->getReferralPostEventPayload());
+
+        $response = $this->withAdminAccessToken()->patchJson('api/v3/posts/' . $post->id, [
+            'status' => 'register-form',
+        ]);
+    }
+
+    /**
      * Test for updating a post with invalid status.
      *
      * PATCH /api/v3/posts/:id
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -86,8 +86,6 @@ class PostTest extends TestCase
         $this->mock(CustomerIo::class)
             ->shouldReceive('trackEvent');
 
-
-
         // Create the post!
         $response = $this->withAccessToken($northstarId)->json('POST', 'api/v3/posts', [
             'campaign_id'      => $campaignId,
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -10,7 +10,9 @@ use Rogue\Models\Signup;
 use Rogue\Models\Campaign;
 use Rogue\Models\Reaction;
 use Rogue\Services\Fastly;
+use Rogue\Services\GraphQL;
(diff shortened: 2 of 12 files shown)


tests/Http/PostTest.php
```diff
@@ -1349,6 +1359,29 @@ public function testUpdatingAPhotoPost()
...
+    {
+        $post = factory(Post::class)->create([
+            'referrer_user_id' => $this->faker->northstar_id,
+            'type' => 'voter-reg',
+            'status' => 'step-1',
+        ]);
+
+        $this->mock(CustomerIo::class)
+          ->shouldReceive('trackEvent')
+          ->with($post->referrer_user_id, 'referral_post_updated', $post->getReferralPostEventPayload());
```
mendelB, line 1378: any reason we don't assert this above in the post created test?

tests/Http/SignupTest.php
```diff
@@ -590,7 +590,7 @@ public function testSignupIndexWithIncludedUser()
      */
     public function testSignupsIndexAsAdminWithFilters()
     {
-        $northstarId = $this->faker->northstar_id;
+        $northstarId = $this->faker->unique()->northstar_id;
```
mendelB, line 593: Is this because the referral ID would be set to the same ID as the post's user?

tests/Http/PostTest.php
```diff
@@ -76,6 +78,14 @@ public function testCreatingAPostAndSignup()
             ->shouldReceive('userSignup')
             ->shouldReceive('userSignupPost');
 
+        // Mock the GraphQL API calls.
+        $this->mock(GraphQL::class)
+            ->shouldReceive('getUserById');
```
mendelB, line 83: Why do we need to mock this here specifically and not in the updated or general event payload tests?

database/factories/ModelFactory.php
```diff
@@ -83,7 +83,6 @@
          */
         'school_id' => $this->faker->optional()->school_id,
         'source' => 'phpunit',
-        'referrer_user_id' => $this->faker->optional()->northstar_id,
```
mendelB, line 86: Is this to avoid triggering the c.io event sporadically unless we explicitly want to test?

tests/Http/PostTest.php
aaronschachter: Because this post has a `referrer_user_id` set - that's the only time we need to query GraphQL upon post create or update -- to show the post user's display name to the referrer. In Customer.io, we'll have all of the referrer user's info, so the message template would look something like:
```
Hey {{customer.first_name}}!

Guess what? Your friend {{event.user_display_name)) just registered to vote through your online voter registration drive.
```

tests/Http/PostTest.php
aaronschachter: Yeah, this one gets to me - I wish we could. It's because we don't have a `post->id` set -- so the test would fail. The reason I thought it'd be wise to include the post id in the Customer.io event was more for debugging, although I guess we'd need it anyway should we ever link a referrer to a referral post permalink (although seems really rare that we'd do so on post creation, cos it'd be pending in most cases, but you never know)

tests/Http/SignupTest.php
aaronschachter: There's the slight chance it could be, because our northstar_id faker selects a random element from a list of valid dev Northstar ID's. This is one of those tests that will randomly fail and then pass on a re-run, this should hopefully fix at least this one (there's a few).

database/factories/ModelFactory.php
aaronschachter: Exactly! We'd have had to add the mocks into a lot more tests -- I think it's better to explicitly test that we only call GraphQL when it's a referral post.

PR #1019, opened 2020-05-19
Adding Laravel Scout

weerd:
### What's this PR do?

This pull request adds the initial setup for Laravel Scout, so we can begin indexing our campaign data in Algolia.

👀 

### Relevant tickets

### Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
     "sokil/php-isocodes": "^3.0",
     "myclabs/php-enum": "^1.7",
     "softonic/graphql-client": "^1.2",
-    "hashids/hashids": "^4.0"
+    "hashids/hashids": "^4.0",
+    "laravel/scout": "^7.2.1"
   },
   "require-dev": {
     "filp/whoops": "~2.0",
--- a/config/scout.php
+++ b/config/scout.php
@@ -1,0 +1,91 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | Default Search Engine
+    |--------------------------------------------------------------------------
+    |
+    | This option controls the default search connection that gets used while
+    | using Laravel Scout. This connection is used when syncing all models
+    | to the search service. You should adjust this based on your needs.
+    |
+    | Supported: "algolia", "null"
+    |
+    */
+
+    'driver' => env('SCOUT_DRIVER', 'algolia'),
+
+    /*
+    |--------------------------------------------------------------------------
+    | Index Prefix
+    |--------------------------------------------------------------------------
+    |
+    | Here you may specify a prefix that will be applied to all search index
+    | names used by Scout. This prefix may be useful if you have multiple
+    | "tenants" or applications sharing the same search infrastructure.
+    |
+    */
+
+    'prefix' => env('SCOUT_PREFIX', ''),
+
+    /*
+    |--------------------------------------------------------------------------
+    | Queue Data Syncing
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -27,7 +27,8 @@
     "sokil/php-isocodes": "^3.0",
     "myclabs/php-enum": "^1.7",
     "softonic/graphql-client": "^1.2",
-    "hashids/hashids": "^4.0"
+    "hashids/hashids": "^4.0",
+    "laravel/scout": "^7.2.1"
```
weerd, line 31: This was the latest version that works with the current version of Laravel (v5.5) used in Rogue.

config/scout.php
```diff
@@ -0,0 +1,91 @@
...
+    |
+    | This option controls the default search connection that gets used while
+    | using Laravel Scout. This connection is used when syncing all models
+    | to the search service. You should adjust this based on your needs.
+    |
+    | Supported: "algolia", "null"
+    |
+    */
+
+    'driver' => env('SCOUT_DRIVER', 'algolia'),
```
DFurnes, line 18: Should we default to `null` so this can be deployed without `ALGOLIA_APP_ID` and `ALGOLIA_SECRET` being set? We could then set `SCOUT_DRIVER=algolia` on each environment as a sort of feature flag?

config/scout.php
```diff
@@ -0,0 +1,91 @@
...
+    | Index Prefix
+    |--------------------------------------------------------------------------
+    |
+    | Here you may specify a prefix that will be applied to all search index
+    | names used by Scout. This prefix may be useful if you have multiple
+    | "tenants" or applications sharing the same search infrastructure.
+    |
+    */
+
+    'prefix' => env('SCOUT_PREFIX', ''),
```
DFurnes, line 31: Since we'll have multiple apps in here eventually, we'll probably want to set this right? I wonder if it'd make sense to have our [Terraform `heroku_app` module](https://github.com/DoSomething/infrastructure/blob/master/components/heroku_app/main.tf) set an `APP_NAME` environment variable that we can just read in here...

config/scout.php
```diff
@@ -0,0 +1,91 @@
...
+    | Queue Data Syncing
+    |--------------------------------------------------------------------------
+    |
+    | This option allows you to control if the operations that sync your data
+    | with your search engines are queued. When this is set to "true" then
+    | all automatic data syncing will get queued for better performance.
+    |
+    */
+
+    'queue' => env('SCOUT_QUEUE', false),
```
DFurnes, line 44: We could probably just hard-code this to `true`!

weerd:
Oh ya. Good notes on the config settings. I hadn't actually tweaked anything just yet, I was just pushing up the default config to just get the file in the repo; I probs should have clarified that.

I don't think it's an issue since I'm not actually using the `Searchable` trait in any models just yet, so nothing is technically being used.

config/scout.php
weerd: Yep yep. I was gonna set this in an upcoming PR 👯

config/scout.php
weerd: I didn't set anything yet because I needed to research this a bit more. I'm not sure if we ever want Contentful to add data to indexed campaigns if it means we need to then specify the rogue specific prefix or whatnot. I just wasn't sure and wanted to review this option a bit more.

config/scout.php
weerd: I think it's technically fine since I'm not actually using the `Searchable` trait in a model yet, so none of this config is technically being used, but that being said, I do think this suggestion makes sesne!

PR #1021, opened 2020-05-20
Change post minimum dimensions to 50x50

katiecrane:
### What's this PR do?

This pull request lowers the minimum image size required for posts from 400 by 400 to 50 by 50.

Image I can submit now but couldn't before:
Screen Shot 2020-05-20 at 11 52 26 AM

Image I still can't submit:
Screen Shot 2020-05-20 at 11 52 56 AM

Any flags?

There's a whole incident channel linked from the card but something happened and Gambit images were coming in real small so we want to lower this limit so we can push them through so we don't lose them!

### Relevant tickets

References [Pivotal #172939835](https://www.pivotaltracker.com/story/show/172939835).

### Checklist

### DIFF
--- a/config/posts.php
+++ b/config/posts.php
@@ -12,8 +12,8 @@ return [
     */
     'image' => [
         'min' => [
-            'height' => 400,
-            'width' => 400,
+            'height' => 50,
+            'width' => 50,
         ],
         'max' => [
             'height' => 4000,


weerd:
Wait. Is this not bad for our posts if so tiny? I guess they could just be rejected?

katiecrane:
@weerd yes, folks reviewing posts will need to hide these from the gallery rather than outright rejecting all of them (if they were all going to be rejected I don't think we'd really care about saving them) - this is the easiest way to save the posts from the incident, other options involved more significant work in Blink/Gambit!

PR #1022, opened 2020-05-21
Enabling campaigns to be searchable on Algolia.

weerd:
### What's this PR do?

This pull request enables `Campaign` model records to be indexed on Algolia when updated (and also for batch importing via an `artisan` command) and thus searchable!

I also tested this locally to make sure that if the `SCOUT_DRIVER` env variable isn't set and campaign entries are updated, Rogue doesn't crap out. Everything seems to work as it should, but doesn't send any data to Algolia as expected!

This means we can merge this code in without worry that it'll break anything on different Rogue environments, and the `SCOUT_DRIVER` env variable does essentially act as a feature flag to turn Scout (and Algolia indexing) "on".

👀 

Laravel makes things so easy ✨ 

### Relevant tickets

### Checklist


app/Models/Campaign.php
```diff
@@ -267,4 +269,14 @@ public function setArbitraryDateString($attribute, $value)
...
     }
+
+    /**
+     * Get the indexable data array for the model.
+     *
+     * @return array
+     */
+    public function toSearchableArray()
+    {
+        return Arr::except($this->toArray(), ['impact_doc']);
```
weerd, line 280: This is pretty nice. We can easily _remove_ items to be indexed by passing them into this array. `impact_doc` for example is attribute we don't need indexed by Algolia so we remove it before sending off the data to the Search API.

config/scout.php
```diff
@@ -41,7 +41,7 @@
     |
     */
 
-    'queue' => env('SCOUT_QUEUE', false),
+    'queue' => env('SCOUT_QUEUE', true),
```
weerd, line 44: Since Rogue already has queue's set up, I didn't need to set them up and could just turn this on, which is recommended!

app/Models/Campaign.php
mendelB: Super neat!

app/Models/Campaign.php
```diff
@@ -267,4 +269,14 @@ public function setArbitraryDateString($attribute, $value)
 
         $this->attributes[$attribute] = $this->fromDateTime($value);
     }
+
+    /**
+     * Get the indexable data array for the model.
+     *
+     * @return array
+     */
+    public function toSearchableArray()
```
mendelB, line 278: How does this work? Is Scout just auto watching things etc. since we implement the `Searchable ` module, and we're just extending our own customized array of data which they pull?

weerd: It's adds a model observer that triggers whenever any activity happens on the model utilizing it. So when creating/updating/deleting etc, the observer is triggered :)

phpunit.xml
```diff
@@ -32,11 +32,12 @@
     <php>
         <env name="APP_ENV" value="testing"/>
         <env name="CACHE_DRIVER" value="array"/>
+        <env name="DB_DATABASE" value="rogue_test"/>
+        <env name="DS_ENABLE_BLINK" value="false"/>
+        <env name="DS_ENABLE_GLIDE" value="true"/>
+        <env name="SCOUT_DRIVER" value="null"/>
```
weerd, line 38: When running tests locally and then checking the Algolia `campaigns` index, I was initially surprised to see that the number of records jumped significantly! Realized that running the tests was causing creating and updates to the index. So setting the value in PHPUnit to "null" keeps the tests from sending data to Algolia!

katiecrane:
THIS IS SO COOL!!!

.env.example
```diff
@@ -32,7 +32,7 @@ GRAPHQL_URL=https://graphql-dev.dosomething.org
 # Search Engine
 ALGOLIA_APP_ID=
 ALGOLIA_SECRET=
-SCOUT_DRIVER=algolia
+SCOUT_DRIVER=
```
weerd, line 35: Removed pre-setting this to "algolia" since Circle CI uses the .env.example file for setting its initial env variables and by having it set to "algolia" it was trying to connect and failing. Keeping it empty allows this env variable to default to "null" and eliminate the errors we were seeing with Circle CI trying to connect and send data to Algolia.

PR #1023, opened 2020-05-21
Updating to use app env as prefix.

weerd:
### What's this PR do?

This pull request simply adds a Scout Prefix for the indexes related to Rogue.

I'm still figuring out what the ideal setup for all this is, but for the time being I've decided on setting up platform specific API Keys and for each platform having 3 keys per environment. So for Rogue we have:

- Rogue[Development]: handles indexes prefixed by either `local_*` or `development_*`
- Rogue[QA]: handles indexes prefixed by`qa_*`
- Rogue[Production]: handles indexes prefixed by`production_*`

GraphQL will have its own keys but will handle the same prefixed indexes. What this means is that, for example, there will be a "GraphQL[Production]" API Key, that will allow requests to create/update records in any indexes that are prefixed by `production_*`. So Rogue and GraphQL can both work with production data, but API keys are specific to the apps in case we accidentally expose one or need to change it, we don't have to remember to change in multiple places, just the specific app affected.

👀 

I felt grabbing the `config('app.env')` variable made sense for the `SCOUT_PREFIX` setting since all our apps (even our `local`) should have these set and we won't have to worry about typos resulting in unique indexes potentially created on algolia's end.

_Long explanations for simple one-liner changes are my specialty!_ 🕺 

### Relevant tickets

### Checklist

### DIFF
--- a/config/scout.php
+++ b/config/scout.php
@@ -28,7 +28,7 @@ return [
     |
     */
 
-    'prefix' => env('SCOUT_PREFIX', ''),
+    'prefix' => env('SCOUT_PREFIX', config('app.env').'_'),
 
     /*
     |--------------------------------------------------------------------------


weerd:
Having the separate `env_*` prefixed indices also helps make it easier for development purposes if we need to wipe any records or clear the indexes out. It's quite easy and we can ensure we don't affect indexes in other envs!

config/scout.php
```diff
@@ -28,7 +28,7 @@
     |
     */
 
-    'prefix' => env('SCOUT_PREFIX', ''),
+    'prefix' => env('SCOUT_PREFIX', config('app.env').'_'),
```
DFurnes, line 31: Would prefixing with the application name be helpful? Thinking about once we get Contentful stuff in here too...

weerd: I'm not 100% sure on the correct answer here, and we'll likely find the best approach as we proceed further. I suspect we'd want different platforms sharing indexes for example if we end up deciding we _do_ want Contentful campaign information integrated with Rogue campaign info. Otherwise when searching for a campaign we'd have to hit 2 different indexes which could get messy. 🤷‍♂️

DFurnes: Ohhhhhh! You're totally right! Never mind me. 🙈

PR #1024, opened 2020-05-22
Add Acion relationship to Campaign data.

weerd:
### What's this PR do?

This pull request helps include campaign Action data to each campaign when it is indexed. It also enables updating the index when a change is made to an action entry attached to a campaign.

👀 

I tested this a bit locally and seems to work fine!

### Relevant tickets

### Checklist

### DIFF
--- a/app/Models/Action.php
+++ b/app/Models/Action.php
@@ -2,10 +2,13 @@
 
 namespace Rogue\Models;
 
+use Laravel\Scout\Searchable;
 use Illuminate\Database\Eloquent\Model;
 
 class Action extends Model
 {
+    use Searchable;
+
     /**
      * The attributes that should be cast to native types.
      *
@@ -31,6 +34,13 @@ class Action extends Model
         'name', 'campaign_id', 'post_type', 'action_type', 'time_commitment', 'callpower_campaign_id',
         'reportback', 'civic_action', 'scholarship_entry', 'anonymous', 'online', 'quiz', 'noun', 'verb', 'collect_school_id', 'volunteer_credit',
     ];
+
+    /**
+     * The relationships that should be touched on save.
+     *
+     * @var array
+     */
+    protected $touches = ['campaign'];
 
     /**
      * Attributes that can be queried when filtering.
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -277,6 +277,39 @@ class Campaign extends Model
      */
     public function toSearchableArray()
     {
-        return Arr::except($this->toArray(), ['impact_doc']);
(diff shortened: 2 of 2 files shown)


app/Models/Campaign.php
```diff
@@ -277,6 +277,39 @@ public function setArbitraryDateString($attribute, $value)
...
+            'actions',
+            'contentful_campaign_id',
+            'cause',
+            'end_date',
+            'id',
+            'internal_title',
+            'pending_count',
+            'secondary_causes',
+            'start_date',
+        ]);
```
weerd, line 313: With the addition of the action relationship data, I switched to using the `only()` method instead of `except()` so it's easier at a glance to see exactly what we are sending to Algolia. Since Algolia has a cap for the size of each record, it's best to really only send exactly the data you want to be able to search on. While not all of these items seem relevant for search in Phoenix, we think it could help with Admins searching in Rogue in a future update!

app/Models/Action.php
```diff
@@ -32,6 +35,13 @@ class Action extends Model
         'reportback', 'civic_action', 'scholarship_entry', 'anonymous', 'online', 'quiz', 'noun', 'verb', 'collect_school_id', 'volunteer_credit',
     ];
 
+    /**
+     * The relationships that should be touched on save.
+     *
+     * @var array
+     */
+    protected $touches = ['campaign'];
+
```
katiecrane, line 44: This is how Algolia is able to know to pull it in yeah?

app/Models/Campaign.php
katiecrane: Love that switch, makes it way more obvious to see what is getting pulled in!

app/Models/Action.php
weerd: Not Algolia but Laravel. This tells Laravel that when an Action record is updated, to let the parent Campaign model know and trigger observed events, and the Scout `toSearchableArray()`is one of them and that sends stuff to Algolia :D

katiecrane: Right right, I skipped a few steps there - just meant this is what will start the chain of events, so cool!

weerd: oh! Yep, yep! How the dance commences on the Action record side of things! 💃 🕺

katiecrane: Dance: COMMENCE!!!

katiecrane:
@weerd I just noticed a missing letter in your title! "acion"

weerd:
gah! Thanks @katiecrane I hate typos! 🙌 Also helps when needing to search for this PR in the future 🔮

PR #1026, opened 2020-05-28
Adds GroupType model and GET /api/v3/group-types endpoint

aaronschachter:
### What's this PR do?

This pull request starts on a `GroupType` resource, which will be used as a lookup table for the upcoming `Groups` resource needed for [Group OVRD campaigns](https://www.pivotaltracker.com/epic/show/4570244).

Going commit by commit, we can see this PR adds:

* `app/models/GroupType` -- our `fillable` property determines which fields are editable

* `database/migrations/[datetime]_create_group_types_table` - this migration will:
    * create the new `group_types` SQL table - Laravel will know this corresponds our `GroupType` model based on the [table/class names](https://laravel.com/docs/5.5/eloquent#eloquent-model-conventions)
    * add a unique index on `name` field (two group types should not have the same name)
    * add `created_at` and `updated_at` fields via calling the `timestamps()` method

At this point, we can run `php artisan migrate` to create the `group_types` table. I used [Sequel Pro](https://www.sequelpro.com/) to connect to my local Rogue DB, insert a record into `group_types` to verify it saved correctly.

Next we want to return data via API, so we'll need to add new files:

* `app/Http/Transformers/GroupTypeTransformer` - used to transform a `GroupType` instance into API response object

* `app/Http/Controllers/GroupTypeController` - to be used by the `group-types` API endpoint, sets our `GroupTypeTransformer` and adds an index method to return a paginated collection of `GroupType` data

And then modify `routes/web` to add a `GET /api/v3/group-types` index, using our newly added `GroupTypeController`.

We can test locally by visiting `http://rogue.test/api/v3/group-types` in browser (or by making a GET request Postman) to verify our database record we manually inserted earlier shows up in the API response.

Finally, we're now in good shape to start on some tests, adding a new [factory](https://laravel.com/docs/5.5/database-testing#writing-factories) for our new `GroupType` model, and adding a new test file -- `tests/Http/GroupTypeTest`

To run the tests, ssh into vagrant from your homestead directory. Once you'er in SSH, navigate to your Rogue directory and run:

```
phpunit tests/Http/GroupTypeTest.php 
```
cc @lindsayrmaher and @Dinnall as they'll be working on adding a new Groups resource next sprint

* Pull this branch down
* Run `php artisan migrate`
* Add new record to `group_types`, e.g.
```
INSERT INTO `group_types` (`name`, `created_at`, `updated_at`)
VALUES
	('Test Group Type', '2020-03-24 18:33:56', '2020-03-24 18:33:56');
```
* Verify data appears in `GET /api/v3/group-types` request

We've still got a ways to go -- next up: 

* Add GraphQL `groupType(id)` and `groupTypes` queries to read data via GraphQL
* Adding a web controller to add and update records via Laravel blade templates.

### Relevant tickets

### Checklist

### DIFF
--- a/tests/Http/GroupTypeTest.php
+++ b/tests/Http/GroupTypeTest.php
@@ -16,12 +16,18 @@ class GroupTypeTest extends TestCase
     public function testGroupTypeIndex()
     {
         // Create five group types.
-        $first = factory(GroupType::class, 5)->create();
+        $groupTypes = factory(GroupType::class, 5)->create();
 
         $response = $this->getJson('api/v3/group-types');
         $decodedResponse = $response->decodeResponseJson();
 
         $response->assertStatus(200);
         $this->assertEquals(5, $decodedResponse['meta']['pagination']['count']);
+
+        foreach ($groupTypes as $groupType) {
+            $this->assertDatabaseHas('group_types', [
+                'name' => $groupType->name,
+            ]);
+        }
     }
 }
--- a/tests/Http/GroupTypeTest.php
+++ b/tests/Http/GroupTypeTest.php
@@ -22,27 +22,5 @@ class GroupTypeTest extends TestCase
 
         $response->assertStatus(200);
         $this->assertEquals(5, $decodedResponse['meta']['pagination']['count']);
-
-        foreach ($groupTypes as $groupType) {
-            $this->assertDatabaseHas('group_types', [
-                'name' => $groupType->name,
-            ]);
-        }
-    }
-
-    /**
-     * Test that the group_type name field is unique.
-     *
-     * @return void
-     */
-    public function testUniqueGroupTypeNameIndex()
-    {
-        $this->expectException(QueryException::class);
-
-        $groupType = factory(GroupType::class)->create();
-
-        factory(GroupType::class)->create([
-            'name' => $groupType->name,
-        ]);
     }
 }
--- a/tests/Http/GroupTypeTest.php
+++ b/tests/Http/GroupTypeTest.php
@@ -4,7 +4,6 @@ namespace Tests\Http;
 
 use Tests\TestCase;
 use Rogue\Models\GroupType;
-use Illuminate\Database\QueryException;
 
 class GroupTypeTest extends TestCase
 {
--- a/tests/Http/GroupTypeTest.php
+++ b/tests/Http/GroupTypeTest.php
@@ -4,18 +4,17 @@ namespace Tests\Http;
 
 use Tests\TestCase;
 use Rogue\Models\GroupType;
+use Illuminate\Database\QueryException;
 
 class GroupTypeTest extends TestCase
 {
     /**
      * Test that a GET request to /api/v3/group-types returns an index of all group types.
      *
-     * GET /api/v3/group-types
      * @return void
      */
     public function testGroupTypeIndex()
     {
-        // Create five group types.
         $groupTypes = factory(GroupType::class, 5)->create();
 
         $response = $this->getJson('api/v3/group-types');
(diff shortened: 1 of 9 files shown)


tests/Http/GroupTypeTest.php
```diff
@@ -0,0 +1,48 @@
...
+        $decodedResponse = $response->decodeResponseJson();
+
+        $response->assertStatus(200);
+        $this->assertEquals(5, $decodedResponse['meta']['pagination']['count']);
+
+        foreach ($groupTypes as $groupType) {
+            $this->assertDatabaseHas('group_types', [
+                'name' => $groupType->name,
+            ]);
+        }
```
DFurnes, line 30: We could probably remove this assertion, since it's effectively just testing that Laravel's model factories work. I think we can generally trust that the framework (and more broadly, anything in the "arrange" portion of the test) works as expected, and limit assertions to whatever application logic is running (the response).

tests/Http/GroupTypeTest.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public function testUniqueGroupTypeNameIndex()
+    {
+        $this->expectException(QueryException::class);
+
+        $groupType = factory(GroupType::class)->create();
+
+        factory(GroupType::class)->create([
+            'name' => $groupType->name,
+        ]);
```
DFurnes, line 46: Similarly, I think we could safely omit this test & instead test that the endpoint or command we'd be using to create these records returns a good error (e.g. if there was a "create" endpoint, we might check `$response->assertStatus(422)`).

Speaking of, how are we planning on creating new group types when this is in prod?

DFurnes:
> Add new record to `group_types`, e.g.
> 
> ```
> INSERT INTO `group_types` (`name`, `created_at`, `updated_at`)
> VALUES
> 	('Test Group Type', '2020-03-24 18:33:56', '2020-03-24 18:33:56');
> ```

Another handy tool for this is Laravel's "tinker"  REPL, which lets you execute arbitrary code for debugging or testing on your local machine. To do this with Tinker on your Homestead:

```
vagrant@homestead:~/Sites/rogue$ php artisan tinker
Psy Shell v0.9.12 (PHP 7.3.15-3+ubuntu18.04.1+deb.sury.org+1 — cli) by Justin Hileman
>>> GroupType::create(['name' => 'Test Group Type']);
[!] Aliasing 'GroupType' to 'Rogue\Models\GroupType' for this Tinker session.
=> Rogue\Models\GroupType {#3200
     name: "Test Group Type",
     updated_at: "2020-05-28 21:59:20",
     created_at: "2020-05-28 21:59:20",
     id: 1,
   }
```

tests/Http/GroupTypeTest.php
aaronschachter: Yeah, I figured a `POST / request` of some sort would be a better test. I'm thinking only allowing admins, and also was thinking about modeling how actions and campaigns are currently created via server-side blade templates, vs mutations. mainly for sake of convenience. we'll have the same issue when its time to create `groups` once that model is in place. Do you have a strong preference for all GraphQL? The server-side approach I'm thinking would be even more TODO's, but feels like it'd go faster... maybe I'm wrong though.

DFurnes: My preference would definitely be to use GraphQL mutations, since it better sets us up for extracting this admin interface in the future (and we'll probably need GraphQL stuff anyways for the user-facing views).

That said, I can totally understand that it's faster to throw together some of these CRUD-y UIs in this one single repository rather than jumping around, so that's a fair trade-off if you want to make it! Just will add a lil' bit to that epic down the line if we can ever get to it. 🙏

One alternate idea to keep things quick & simple – would an Artisan command work for adding these (like `php artisan rogue:group-type:add` or something?)

aaronschachter: The artisan command is a great idea --thanks! I liked the idea of adding the web controller to have an example for when we create/edit groups next sprint, but I think it'll be easy enough to reference the actions or campaigns controllers.

PR #1029, opened 2020-06-01
Check if group_type exists before creating

aaronschachter:
### What's this PR do?

This pull request is a bit of a band-aid for builds failing on QA right now. The initial problem began when trying to build after merging #1028 today, Monday. Our weekend refresh overwrote the Rogue production QA `migrations` table  with production values, so when the build ran today and executed the `migrate` Artisan command, it failed with:

```
Base table or view already exists: 1050 Table 'group_types  
  ' already exists  
```
Because this table never made it to prod last summer Friday.

This addition to the migration should prevent it from failing, if we start from the `prod` migrations. I tried rolling back a migration to fix, and ended up borking the `referrer_user_id` on signups and posts -- causing the `migrate` fail to instead fail with ` Duplicate column name 'referre  
  r_user_id' `

SO -- order of operations would be:
* Refresh QA db from prod
* Merge this PR
* It should work

The other alternative would be manually dropping the table in the QA db... 

👀 

🍌 

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/database/migrations/2020_05_27_000000_create_group_types_table.php
+++ b/database/migrations/2020_05_27_000000_create_group_types_table.php
@@ -13,6 +13,12 @@ class CreateGroupTypesTable extends Migration
      */
     public function up()
     {
+        if (Schema::hasTable('group_types')) {
+            info('group_types table already exists.');
+
+            return;
+        }
+
         Schema::create('group_types', function (Blueprint $table) {
             $table->increments('id');
             $table->string('name')->unique();


aaronschachter:
This should be fine for local development! But I'll check in with folks about refreshing Rogue DB before I merge this.

DFurnes:
```
 Base table or view already exists: 1050 Table 'group_types' already exists
```

To make sure I'm understanding correctly, our weekend refresh pulled down the existing `group_types` schema but not the correct record in the `migrations` table, and that's why this is trying to create the table from scratch again and failing?

DFurnes:
If so, it sounds like we should update our QA Refresh job to sync `migrations` as well. We shouldn't ever have a case where the database schema and migrations table are out of sync.

aaronschachter:
I think the issue is that we don't wipe out the QA db tables completely, https://github.com/DoSomething/rogue/pull/1026 never made it to prod last week.

Refresh happened on Sunday, copying over production `migrations` data. Today's build had new migrations to run, but couldn't create the `group_types` table because it already exists.

DFurnes:
Oooh wait, I see – so we'd created the `group_types` table here on QA, but then reset `migrations` to prod (where we hadn't yet created that table). Tricky! I think we want to make sure the QA refresh drops existing tables before importing the dump in that case...

DFurnes:
Ah yeah sorry – commented at the same time! I agree! 👍 Let's edit that job before we re-run?

aaronschachter:
We tried to modify the [refresh job to drop tables](https://dosomething.slack.com/archives/CUQMTP0RM/p1591033559002600) before importing.. ended up just manually dropping the table on QA for now to avoid merging this in (it'd make our migrations for creating tables look inconsistent, and ideally we shouldn't need to check if tables exist in each `up` method)

PR #1030, opened 2020-06-02
Create + edit Group Types via web

aaronschachter:
### What's this PR do?

This pull request allows admins to create or edit a Group Type (added in #1026) via Rogue admin UI.

👀 

We're using a web controller in Laravel to quickly provide support for add/edit operations, vs creating external API endpoints + GraphQL mutations + React components per discussion in https://github.com/DoSomething/rogue/pull/1026#discussion_r432152226

### Relevant tickets

### Checklist


app/Http/Controllers/Web/GroupTypesController.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Update the specified resource in storage.
+     *
+     * @param  \Rogue\Models\GroupType  $groupType
+     * @param  \Illuminate\Http\Request  $request
+     */
+    public function update(GroupType $groupType, Request $request)
+    {
+        $this->validate($request, $this->rules);
```
DFurnes, line 69: Let's add an override to the [`unique`](https://laravel.com/docs/5.5/validation#rule-unique) rule here to exclude this resource from the uniqueness constraint (since otherwise we'd have to change the name every time we hit this form since it'd conflict with itself).

(I know this doesn't really break things now since this is the only field, but this could make for a confusing surprise if we were to ever add any other fields to this form!)

aaronschachter: Ah - is what's going on over in the `CampaignsController` what you're referring to?

https://github.com/DoSomething/rogue/blob/d85d8429316aabb3834a12e6a504e81c3d252bbd/app/Http/Controllers/Web/CampaignsController.php#L80-L82

I'm still wrapping my head around validation in Laravel... ideally once we start making Groups editable (added in #1031) - we want a unique index on `group_type_id` and `name`, and been wondering about how to handle that in the form (vs just relying on the scary server error error broken screen)

aaronschachter: I'm going to merge this for now, since approved and can fix this up in another PR -- there's some more on deck.

PR #1032, opened 2020-06-02
Adding group_type_id to the campaigns table

lindsayrmaher:
### What's this PR do?

This pull request creates a migration to add the `group_type_id` as a foreign key on the campaigns table! It also adds the property as fillable so we are able to edit it later on.

👀 

We are going to allow admins the ability to add a group type id on a campaign so that it can be associated with a particular group. (right now this is specific to our OVRD campaigns for Voter Registration)

### Relevant tickets

References [Pivotal # 173101101](https://www.pivotaltracker.com/story/show/173101101).

### Checklist

### DIFF
--- a/database/migrations/2020_06_02_204137_add_group_type_id_to_campaigns_table.php
+++ b/database/migrations/2020_06_02_204137_add_group_type_id_to_campaigns_table.php
@@ -1,0 +1,33 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+class AddGroupTypeIdToCampaignsTable extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('campaigns', function (Blueprint $table) {
+            $table->unsignedInteger('group_type_id')->nullable()->index()->comment('The group type id the group is for.');
+            $table->foreign('group_type_id')->references('id')->on('group_types');
+        });
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('campaigns', function (Blueprint $table) {
+            $table->dropColumn('group_type_id');
+        });
+    }
+}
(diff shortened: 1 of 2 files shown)


database/migrations/2020_06_02_204137_add_group_type_id_to_campaigns_table.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('campaigns', function (Blueprint $table) {
+            $table->unsignedInteger('group_type_id')->nullable()->index()->comment('The group type id the group is for.');
+            $table->foreign('group_type_id')->references('id')->on('group_types');
```
aaronschachter, line 18: It wouldn't hurt to an index on the `group_type_id` field we're adding here, as we'll eventually want to filter a `GET /campaigns` index query by `group_type_id`.

We should just be able to additionally add:
```
$table->index('group_type_id');
```

aaronschachter: This shouldn't block merging though, could make sense once we add to the `indexes` array in the `Campaign` model.

lindsayrmaher: 👍  I might add it in on this PR just to keep track of it

katiecrane: Doesn't this line already make it an index with `->index()`?

lindsayrmaher: That's what i thought as well @katiecrane

lindsayrmaher: I was wondering if i then also need to add it to the index array on the model

katiecrane: Ooooo that array looks to be out of date as we have more that just `id` indexed on this table

aaronschachter: Argh I totally missed the `index` call in thsi line! 🤦‍♂️ I think it'd be fine to hold off on modifying the campaign `indexes` in a separate PR. sorry for the false alarm!

PR #1031, opened 2020-06-02
Adds Group model and GET /api/v3/groups endpoint

aaronschachter:
### What's this PR do?

This pull request adds a `Group` model, and related `GET /api/v3/groups` and `GET /api/v3/groups/:id` endpoints.

### How should this be reviewed?

👀 

### Any background context you want to provide?

The [technical spec](https://docs.google.com/document/u/1/d/1Uv08So_A-SOxty5XfYPcq3SxrFV5B41FsM63D4ECPKA/edit) goes into detail on how this resource will be used.

### Relevant tickets

References [Pivotal #172541936](https://www.pivotaltracker.com/story/show/172541936)

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.

PR #1033, opened 2020-06-03
Create + edit Groups via web

aaronschachter:
### What's this PR do?

This pull request adds support for staff and admins to create, view, and edit a Group (added in #1031) via the admin UI, by first navigating to the Group Type page to create/edit a Group for.

Screen Shot 2020-06-03 at 11 07 23 AM

👀 

* This is missing tests for now, to help keep this 👹 of a PR this reviewable.

* I mentioned this in https://github.com/DoSomething/rogue/pull/1030#discussion_r434707291, but couldn't figure out how to add form validation to prevent the server error if you try to create/edit a group with a duplicate name within the group type. Because this is an internal tool, I think we'll be fine to fix this up in a future PR, again to cut down on complexity / review time.

### Relevant tickets

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/GroupsController.php
+++ b/app/Http/Controllers/Web/GroupsController.php
@@ -16,8 +16,10 @@ class GroupsController extends Controller
         $this->middleware('auth');
         $this->middleware('role:admin,staff');
 
+        // @TODO: We should validate that name field is unique within the group type.
         $this->rules = [
-            'name' => 'required|unique:groups,group_type_id',
+            'name' => 'required',
+            'goal' => 'nullable|integer'
         ];
     }
 
--- a/app/Http/Controllers/Web/GroupsController.php
+++ b/app/Http/Controllers/Web/GroupsController.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace Rogue\Http\Controllers\Web;
+
+use Rogue\Models\Group;
+use Illuminate\Http\Request;
+use Rogue\Http\Controllers\Controller;
+
+class GroupsController extends Controller
+{
+    /**
+     * Create a controller instance.
+     */
+    public function __construct()
+    {
+        $this->middleware('auth');
+        $this->middleware('role:admin,staff');
+
+        $this->rules = [
+            'name' => 'required|unique:groups,group_type_id',
+        ];
+    }
+
+    /**
+     * Create a new group.
+     */
+    public function create($groupTypeId)
+    {
+        return view('groups.create')->with(['groupTypeId' => (int) $groupTypeId]);
+    }
+
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     */
+    public function store(Request $request)
+    {
+        $request->validate($this->rules);
+
+        $group = Group::create($request->all());
+
+        // Log that a group was created.
+        info('group', ['id' => $group->id]);
+
+        return redirect('groups/' . $group->id);
+    }
+
+    /**
+     * Edit an existing group.
+     *
+     * @param  \Rogue\Models\Group  $group
+     */
+    public function edit(Group $group)
+    {
+        return view('groups.edit')->with([
+            'group' => $group,
+        ]);
+    }
+
+    /**
+     * Update the specified resource in storage.
+     *
+     * @param  \Rogue\Models\Group  $group
+     * @param  \Illuminate\Http\Request  $request
+     */
+    public function update(Group $group, Request $request)
+    {
+        $this->validate($request, $this->rules);
+
+        $group->update($request->all());
+
+        // Log that a group was updated.
+        info('group_updated', ['id' => $group->id]);
+
+        return redirect('groups/' . $group->id);
+    }
+}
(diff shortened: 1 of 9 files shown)


resources/views/groups/create.blade.php
```diff
@@ -0,0 +1,34 @@
...
+                    {{ csrf_field()}}
+
+                    <div class="form-item">
+                        <label class="field-label">Name</label>
+                        @include('forms.text', ['name' => 'name', 'placeholder' => 'Group name, e.g. NYC Chapter'])
+                    </div>
+
+                    <div class="form-item">
+                        <label class="field-label">Goal</label>
+                        @include('forms.text', ['name' => 'goal', 'placeholder' => 'Optional group goal, e.g. 200'])
```
mendelB, line 23: Do we need this to be a `number` field / do we need this validated as a number on the controller?

routes/web.php
```diff
@@ -22,6 +22,8 @@
 Route::get('campaigns/{id}/actions/create', 'ActionsController@create');
 Route::resource('campaigns', 'CampaignsController', ['except' => ['index', 'show']]);
 Route::resource('group-types', 'GroupTypesController', ['except' => ['index', 'show']]);
+Route::get('group-types/{id}/groups/create', 'GroupsController@create');
+Route::resource('groups', 'GroupsController', ['except' => 'show']);
```
mendelB, line 26: Do we want to `except` the standalone `create` route as well, or are we cool with creating Groups without a group type ID? (It looks like the create action expects a `$groupTypeId`)

app/Http/Controllers/Web/GroupsController.php
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     */
+    public function store(Request $request)
+    {
+        $request->validate($this->rules);
+
+        $group = Group::create($request->all());
```
mendelB, line 41: I'm shaky on this, but is it safe to pass the request data wholesale to the create/update calls? (Are we fine since we've defined a `Group::fillable` list of whitelisted values?)

app/Http/Controllers/Web/GroupsController.php
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Create a controller instance.
+     */
+    public function __construct()
+    {
+        $this->middleware('auth');
+        $this->middleware('role:admin,staff');
+
+        $this->rules = [
+            'name' => 'required|unique:groups,group_type_id',
```
mendelB, line 20: I poked around the docs, but I'm still kinda shaky on how this unique rule works. Is the `group_type_id` saying the `name` is only unique within children of the same group type?

aaronschachter: I'm just going to remove this `unique` because I couldn't get it working, and add a TODO to revisit. We want a unique index on `group_type_id` and `name` for groups, this attempts to validate before we try to save the group -- and it didn't do the trick.

app/Http/Controllers/Web/GroupsController.php
aaronschachter: Yup, I think the `fillable` handles this concern. This controller is mostly copying by example from other controllers in the repo (although, that shouldn't exactly be the best excuse for assuming it's safe!):

https://github.com/DoSomething/rogue/blob/f660e391e988b84aad379b11066c3d3c9138a366/app/Http/Controllers/Web/ActionsController.php#L71

routes/web.php
aaronschachter: Navigating to `/groups/create` will fail the same way navigating to `/actions/create` will fail:
```
Type error: Too few arguments to function Rogue\Http\Controllers\Web\GroupsController::create(), 0 passed and exactly 1 expected
```
We are not cool with creating Groups without a Group Type ID -- but this does prevent it from happening. This is also following by example from the ActionsController, where we don't want to create an Action without a Campaign.

resources/views/groups/create.blade.php
aaronschachter: Added this into the controller in 9a26433c45365a49c9b0d0d30bfa7055574d8c19

routes/web.php
mendelB: Ah, gotcha. Wouldn't it make sense than to explicitly exclude it from the resource since it's not a route we intend to support, instead of just letting it break? Either way, certainly not a blocker here, and I understand that this is a follow-by-example.

app/Http/Controllers/Web/GroupsController.php
mendelB: lol but yes, the `fillable` seems like a good safety measure! (Though perhaps worth adding a `group_type_id` validation rule as well? 🤔)

routes/web.php
aaronschachter: Speaking of `gotcha` -- if we do exclude `create`, we'll get the React rendered route `/groups/{id}`, with an error of 
```
Error: GraphQL error: Variable "$id" of non-null type "Int!" must not be null.
```
All this nitty gritty feels like good examples of why we have the outstanding TODO to move these Laravel web controllers into React components GraphQL migrations:
https://github.com/DoSomething/rogue/blob/f660e391e988b84aad379b11066c3d3c9138a366/routes/web.php#L19-L22

But we decided to go this route for sake of keeping dev in a single repo (and hoping to get most groups backend work done this sprint). refs https://github.com/DoSomething/rogue/pull/1026#discussion_r432152226

mendelB: Oy, _gotcha_ -- looking forward to that migration!

PR #1036, opened 2020-06-04
Add Group Type Dropdown to Create/Edit Campaign Pages

lindsayrmaher:
### What's this PR do?

This pull request adds a dropdown field to our create and edit pages in Rogue for a campaign.

Should we add any text underneath the field to clarify what it's for? (similar to the contentful campaign id field)

👀 

Screen Shot 2020-06-04 at 5 02 46 PM

Screen Shot 2020-06-04 at 4 55 34 PM

Screen Shot 2020-06-04 at 5 02 56 PM

We need to add a way for admins to add a group type to a campaign in order to eventually display it on the campaign in phoenix.

### Relevant tickets

References [Pivotal # 173101101](https://www.pivotaltracker.com/n/projects/2417735/stories/173101101).

### Checklist


resources/views/campaigns/grouptype-select.blade.php
```diff
@@ -0,0 +1,12 @@
...
+        <option value="" {{ ! old($name) && empty($value) ? 'selected' : '' }} disabled>{{ $placeholder or '–' }}</option>
+        @foreach($options as $option)
+            @if (old($name))
+                <option value="{{ $option->id}}" {{ old($value) === $option->id? 'selected': '' }}>{{ $option->name }}</option>
+            @else
+                <option value="{{ $option->id }}" {{ isset($value) && $value === $option->id ? 'selected': '' }}>{{ $option->name  }}</option>
+            @endif
+        @endforeach
+    </select>
+</div>
```
mendelB, line 12: This is arguably _more_ work, so just throwing the thought out there: I wonder if we could map over the `GroupType`s in the controller to obtain a similar key-value pairing that we have on other selections like this (e.g. the [`PostType::labels`](https://github.com/DoSomething/rogue/blob/01124649b964cd157f116ae0d2419934b27b5ca7/app/Types/PostType.php#L19-L31)) so that we could just use the existing [forms.select.blade.php](https://github.com/DoSomething/rogue/blob/01124649b964cd157f116ae0d2419934b27b5ca7/resources/views/forms/select.blade.php#L1) form? 

e.g. (rough idea offhand) 
```php
GroupTypes::all()->map(function($groupType) {
  return [$groupType->id => $groupType->name];
)->flatten();
```

And this could even be a new `GroupType::labels()` convenience method on the model 🤔

aaronschachter: Using the existing `forms.select.blade` would be ideal - great suggestion @mendelB !

aaronschachter: This file should be removed per latest changes.

app/Http/Controllers/Web/CampaignsController.php
```diff
@@ -34,7 +36,11 @@ public function __construct()
      */
     public function create()
     {
-        return view('campaigns.create')->with('causes', Cause::labels());
+        $group_type_labels = GroupType::all()->map(function ($groupType) {
+            return [$groupType->id => $groupType->name];
+        })->flatten();
```
aaronschachter, line 41: Let's add this as helper to the `GroupType` model per @mendelB's suggestion, to DRY and potentially to re-use in other forms (e.g. if we ever wanted to allow editing a Group's `group_type_id`).

```
    /**
     * Returns labeled list of values.
     *
     * @return array
     */
    public static function labels()
    {
        return $this::all()->map(function ($groupType) {
            return [$groupType->id => $groupType->name];
        })->flatten();
    }
```

PR #1037, opened 2020-06-05
Adds tests for Group web controller + fixes validation TODOs

aaronschachter:
### What's this PR do?

This pull request fixes up validation TODOs from #1033 and #1030, and adds some basic test coverage that was missing in #1033 

👀 

🏺 

### Relevant tickets

Last PR for [Pivotal #172541916](https://www.pivotaltracker.com/story/show/172541916).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/GroupTypesController.php
+++ b/app/Http/Controllers/Web/GroupTypesController.php
@@ -4,6 +4,7 @@ namespace Rogue\Http\Controllers\Web;
 
 use Rogue\Models\GroupType;
 use Illuminate\Http\Request;
+use Illuminate\Validation\Rule;
 use Rogue\Http\Controllers\Controller;
 
 class GroupTypesController extends Controller
@@ -17,7 +18,7 @@ class GroupTypesController extends Controller
         $this->middleware('role:admin');
 
         $this->rules = [
-            'name' => 'required|unique:group_types',
+            'name' => 'required',
         ];
     }
 
@@ -36,9 +37,11 @@ class GroupTypesController extends Controller
      */
     public function store(Request $request)
     {
-        $request->validate($this->rules);
+        $values = $this->validate($request, array_merge_recursive($this->rules, [
+            'name' => [Rule::unique('group_types')],
+        ]));
 
-        $groupType = GroupType::create($request->all());
+        $groupType = GroupType::create($values);
 
         // Log that a group type was created.
         info('group_type', ['id' => $groupType->id]);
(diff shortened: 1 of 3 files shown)


app/Http/Controllers/Web/GroupTypesController.php
```diff
@@ -66,11 +69,11 @@ public function edit(GroupType $groupType)
      */
     public function update(GroupType $groupType, Request $request)
     {
-        // @TODO: Override unique rule.
-        // @see https://github.com/DoSomething/rogue/pull/1030#discussion_r434628475
-        $this->validate($request, $this->rules);
+        $values = $this->validate($request, array_merge_recursive($this->rules, [
+            'name' => [Rule::unique('group_types')->ignore($groupType->id)],
+        ]));
```
katiecrane, line 74: I'm confused about why we have to do this here since we're not explicitly pulling in the `'name' => [Rule::unique('group_types')]` here 🤔 (also confused about this in the campaign example linked from the linked comment!)

tests/Http/Web/GroupTest.php
```diff
@@ -0,0 +1,43 @@
...
+    /** @test */
+    public function testStaffCanCreateGroup()
+    {
+        $groupType = factory(GroupType::class)->create();
+        $name = $this->faker->sentence;
+
+        // Verify redirected to new resource.
+        $this->actingAsStaff()
+            ->postJson('groups', ['group_type_id' => $groupType->id, 'name' => $name])
+            ->assertStatus(302);
```
katiecrane, line 41: Does this one need the database assertion and non-duplication verification like the admin test?

app/Http/Controllers/Web/GroupTypesController.php
aaronschachter: There's a "Forcing A Unique Rule To Ignore A Given ID" section in the docs on the Laravel [unique](https://laravel.com/docs/5.5/validation#rule-unique) rule that I found helpful per the changes requested in https://github.com/DoSomething/rogue/pull/1030#discussion_r434628475:

From the docs example of updating an existing user's profile:

> Of course, you will want to verify that the e-mail address is unique. However, if the user only changes the name field and not the e-mail field, you do not want a validation error to be thrown because the user is already the owner of the e-mail address.

tests/Http/Web/GroupTest.php
aaronschachter: Could add it in, but it seemed repetitive to add those tests as this is really just more to confirm staff has access to create/edit.

app/Http/Controllers/Web/GroupTypesController.php
katiecrane: Hmm okay that makes sense but I'm still confused because it doesn't seem like that unique rule that we define in the `store` method would get applied here anyway, so I'm confused why we have to ignore the rule. Does that make sense?

tests/Http/Web/GroupTest.php
katiecrane: sounds good!

app/Http/Controllers/Web/GroupTypesController.php
aaronschachter: The `name` column in `group_types` tables has a unique index, does that help clarify? https://github.com/DoSomething/rogue/blob/50b4d5faee4cc393976f29446a45ff9653561304/database/migrations/2020_05_27_000000_create_group_types_table.php#L16-L20

Once we add another editable field to this form, e.g. "Should collect School ID", then we'd run into trouble trying to update a Group Type when not updating the name, but the additional fields.

katiecrane: AH yes that does! thank you!

PR #1040, opened 2020-06-09
Adds index to posts table on group_id, action_id, status

aaronschachter:
### What's this PR do?

This pull request adds a compound index referred to in #1038, and allows filtering a `GET /posts` index query by `group_id`. This should improve performance when we eventually regularly query the database to get a count of the total completed voter registration posts for a specific `group_id`.

It also modifies the visible scope of the `Post` model to return all completed voter-reg posts, to simplify the logic needed for the calculation (it'd be complicated and more costly to try to join on `signups` to determine whether the authenticated user has signed up with the `group_id` we're running the calculation for)

👀 

A reason to show any voter-reg referral regardless of status is that we eventually want to show a referrer ALL referrals broken down by different status (including non-completed status) in https://www.pivotaltracker.com/story/show/173230283

### Relevant tickets

References [Pivotal #173019860](https://www.pivotaltracker.com/story/show/173019860).

### Checklist

### DIFF
--- a/tests/Http/PostTest.php
+++ b/tests/Http/PostTest.php
@@ -1165,18 +1165,16 @@ class PostTest extends TestCase
     {
         $referrerUserId = $this->faker->unique()->northstar_id;
 
-        // Add two completed voter reg referrals for our referrer, which should be visible to them.
-        $firstCompletedVoterRegReferral = factory(Post::class)->states('voter-reg', 'register-form')->create(['referrer_user_id' => $referrerUserId]);
-        $secondCompletedVoterRegReferral = factory(Post::class)->states('voter-reg', 'register-OVR')->create(['referrer_user_id' => $referrerUserId]);
+        // A referrer can see all of their voter registration referrals.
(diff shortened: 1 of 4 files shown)


tests/Http/PostTest.php
```diff
@@ -1169,21 +1169,21 @@ public function testPostsIndexAsVoterRegistrationReferrer()
...
+        // Add pending referrals for a different referrer, which shouldn't be visible.
+        factory(Post::class)->states('photo', 'pending')->create(['referrer_user_id' => $this->faker->unique()->northstar_id]);
 
         // Add a pending photo post, which shouldn't be visible.
         factory(Post::class)->states('photo', 'pending')->create();
 
         $response = $this->withAccessToken($referrerUserId)->getJson('api/v3/posts');
         $response->assertStatus(200);
-        $response->assertJsonCount(2, 'data');
+        $response->assertJsonCount(4, 'data');
```
katiecrane, line 1186: There are checks below for `register-form` and `register_OVR` showing up, should we add one for `step-1`?

aaronschachter: Updated, good call!

PR #1041, opened 2020-06-09
Set the default SCOUT_DRIVER to 'null'.

DFurnes:
### What's this PR do?

This pull request sets the default `SCOUT_DRIVER` to `'null'`. This fixes an issue where we'd pass the _actual_ `null` value (from the empty value in `.env`) when trying to create a driver and it'd explode whenever the application tried to boot:

```
Type error: Too few arguments to function Illuminate\Support\Manager::createDriver(), 0 passed
```

When running this application with a fresh `.env`, things should no longer error out.

This was a bug introduced in #1022.

### Relevant tickets

References [Pivotal #173252821](https://www.pivotaltracker.com/story/show/173252821).

### Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -29,10 +29,10 @@ NORTHSTAR_CLIENT_SECRET=
 
 GRAPHQL_URL=https://graphql-dev.dosomething.org
 
-# Search Engine
+# Algolia (only use if testing indexing):
+SCOUT_DRIVER=null
 ALGOLIA_APP_ID=
 ALGOLIA_SECRET=
-SCOUT_DRIVER=
 
 # Slack '#notify-badass-members' integration. If testing, create
 # a new Incoming Webhook that posts messages to your DMs here:


DFurnes:
> EDIT: Ohhh, or is it that it literally uses the empty `null` value from the unset .env variable and _not_ the default string?

Yeah, which surprised me! Since we had the variable defined in `.env` but empty, it wouldn't use the "fallback" default value from the `conifg/scout.php`, and instead just passes along that (literal) `null` value to the Manager class.

mendelB:
> > EDIT: Ohhh, or is it that it literally uses the empty `null` value from the unset .env variable and _not_ the default string?
> 
> Yeah, which surprised me! Since we had the variable defined in `.env` but empty, it wouldn't use the "fallback" default value from the `conifg/scout.php`, and instead just passes along that (literal) `null` value to the Manager class.

Ah, jeepers. Good to know for the future!

PR #1042, opened 2020-06-09
Group Type Id Tests

lindsayrmaher:
### What's this PR do?

This pull request adds the `group_type_id` to the CampaignsTransformer and adds some tests to test for updating a campaigns `group_type_id`

👀 

We want to be able to double check that the group_type_id updates correctly. 

### Relevant tickets

References [Pivotal # 173101101](https://www.pivotaltracker.com/story/show/173101101).

### Checklist


tests/Http/CampaignTest.php
```diff
@@ -29,6 +29,7 @@ public function testCreatingACampaign()
             'impact_doc' => 'https://www.google.com',
             'start_date' => $firstCampaignStartDate,
             'end_date' => $firstCampaignEndDate,
+            'group_type_id' => 2,
```
aaronschachter, line 32: To figure out why this test was breaking, I looked into the `storage/logs` to try to find any errors and see that a foreign key constraint fails. We need to create a `GroupType` factory in order to get a valid `group_type_id` value into the database that we can use for this field.

```
[2020-06-09 21:31:18] testing.ERROR: SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`rogue_test`.`campaigns`, CONSTRAINT `campaigns_group_type_id_foreign` FOREIGN KEY (`group_type_id`) REFERENCES `group_types` (`id`)
```

tests/Http/CampaignTest.php
```diff
@@ -288,6 +289,39 @@ public function testUpdatingACampaignWithContentfulId()
...
+     * @return void
+     */
+    public function testUpdatingACampaignWithGroupTypeId()
+    {
+        // Create a campaign to update.
+        $campaign = factory(Campaign::class)->create();
+
+        // Update the group type id.
+        $response = $this->withAdminAccessToken()->patchJson('api/v3/campaigns/' . $campaign->id, [
+            'group_type_id' => 1,
```
aaronschachter, line 305: This also fails because there isn't a `group_types` row with an id of 1.

tests/Http/CampaignTest.php
```diff
@@ -308,6 +342,26 @@ public function testUpdatingACampaignWithInvalidStatus()
...
+    public function testUpdatingACampaignWithInvalidStatusWithGroupTypeId()
+    {
+        // Create a campaign to update.
+        $campaign = factory(Campaign::class)->create();
+
+        $response = $this->withAdminAccessToken()->patchJson('api/v3/campaigns/' . $campaign->id, [
+            'group_type_id' => "4", // This should be an integer
+        ]);
+
+        $response->assertStatus(422);
```
aaronschachter, line 360: It's very odd this one's failing with a 200 status - we should be receiving a 422 if the group_type_id doesn't exist. I wonder if changing the `group_type_id` to something alphanumeric like `cars` would cause the 422 ...

PR #1044, opened 2020-06-11
Algolia requires UNIX timestamps to filter by dates.

DFurnes:
### What's this PR do?

This pull request transforms any dates in the Campaign's `toSearchableArray` payload into UNIX timestamps, which allows us to [filter on those fields using Algolia](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-date/).

👀 

I considered creating `start_date_timestamp` and `end_date_timestamp` fields, like shown in Algolia's docs, but that seems like something that'd just unnecessarily increase our index size.

Can you think of any reasons we might want or need the ISO date in Algolia?

### Relevant tickets

References [Pivotal #172922602](https://www.pivotaltracker.com/story/show/172922602).

### Checklist

### DIFF
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -298,6 +298,11 @@ class Campaign extends Model
             ]);
         });
 
+        // Transform dates into UNIX timestamps:
+        foreach ($this->dates as $date) {
+            $array[$date] = optional($this->{$date})->timestamp;
+        }
+
         // Only send data we want to search against.
         return Arr::only($array, [
             'accepted_count',


weerd:
I think [Scout Extended package](https://github.com/algolia/scout-extended/blob/master/src/Transformers/ConvertDatesToTimestamps.php) would help with this since it has a transformation layer, but not something we can use until we upgrade Rogue's Laravel version 😢

app/Models/Campaign.php
```diff
@@ -298,6 +298,11 @@ public function toSearchableArray()
             ]);
         });
 
+        // Transform dates into UNIX timestamps:
+        foreach ($this->dates as $date) {
```
mendelB, line 302: Is `$this->dates` a `Searchable` method that automatically includes all date fields?

mendelB:
> Can you think of any reasons we might want or need the ISO date in Algolia?

Am I correct that we're not actually using the Algolia data other then searching by it? (Not that we couldn't just format the unix timestamps as they're used in any case...)

DFurnes:
> Am I correct that we're not actually using the Algolia data other then searching by it? (Not that we couldn't just format the unix timestamps as they're used in any case...)

Yep, that's the plan at the moment!

app/Models/Campaign.php
DFurnes: That's a [property on the `Campaign` model](https://github.com/DoSomething/rogue/blob/306e88221fa34510133786f1f4b3dc314dedaa10/app/Models/Campaign.php#L18-L23) (and any Eloquent model)! It's used by Laravel to know which fields to parse from the database into `Carbon` instances.

weerd:
> I considered creating start_date_timestamp and end_date_timestamp fields, like shown in Algolia's docs, but that seems like something that'd just unnecessarily increase our index size.

Oh just thought of something... did you mean increasing the index size because we'd have `start_date` and `end_date` along with `start_date_timestamp` and `end_date_timestamp`?

If that is what you meant (and I'm not misunderstanding) we could maybe just rename `start_date` and `end_date` to be suffixed with `_timestamp` before we send the array to Algolia 🤷‍♂️  It would make it more apparent that it's specifically a timestamp.

DFurnes:
Yeah yeah – if we can't figure out a reason we'd use the objecty-ISO version of these fields it seems like it'd just be taking up space in our index and not really providing any value.

And yeah, that could definitely work. I wasn't sure whether renaming the field would make it clearer or not tbh. 🤔 I could see a good argument either way – same name is probably more discoverable, but the `_timestamp` name definitely makes the type less of a surprise.

PR #1048, opened 2020-06-18
Use regexp for Groups name filter instead of equals

aaronschachter:
### What's this PR do?

This pull request modifies the `filters[name]` in a `GET /groups` index request to return all groups that have names that contain the filter value, instead of looking for an exact match.

Example response:
```
// http://rogue.test/api/v3/groups?filter[name]=new

{
  "data": [
    {
      "id": 1,
      "group_type_id": 1,
      "name": "New York City",
      "goal": 200,
      "created_at": "2020-03-24T18:33:56+00:00",
      "updated_at": "2020-06-03T17:40:52+00:00"
    },
    {
      "id": 114,
      "group_type_id": 1,
      "name": "Newark, OH",
      "goal": null,
      "created_at": "2020-06-18T18:10:41+00:00",
      "updated_at": "2020-06-18T18:10:41+00:00"
    },
    {
      "id": 222,
      "group_type_id": 1,
      "name": "The New School, NY",
      "goal": null,
      "created_at": "2020-06-18T18:10:41+00:00",
      "updated_at": "2020-06-18T18:10:41+00:00"
    },
   ...
```

👀 

We'll be asking users to begin typing their group name on the Group Finder, and want to return matches as they begin typing.

### Relevant tickets

References [Pivotal #173232032](https://www.pivotaltracker.com/story/show/173232032).

### Checklist

### DIFF
--- a/app/Http/Controllers/GroupsController.php
+++ b/app/Http/Controllers/GroupsController.php
@@ -33,6 +33,10 @@ class GroupsController extends ApiController
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Group::$indexes);
 
+        if (isset($filters['name'])) {
+            $query->where('name', 'REGEXP', $filters['name']);
+        }
+
         return $this->paginatedCollection($query, $request);
     }
 
--- a/app/Models/Group.php
+++ b/app/Models/Group.php
@@ -22,7 +22,7 @@ class Group extends Model
      *
      * @var array
      */
-    public static $indexes = ['id', 'group_type_id', 'name'];
+    public static $indexes = ['id', 'group_type_id'];
 
     protected static function boot()
     {
--- a/docs/endpoints/groups.md
+++ b/docs/endpoints/groups.md
@@ -6,6 +6,20 @@
 GET /api/v3/groups
 ```
 
+### Optional Query Parameters
+
+- **filter[id]** _(integer)_
+
+  - Filter results by Group ID or ID's.
+
+- **filter[group_type_id]** _(integer)_
+
+  - Filter results by Group Type ID or ID's.
+
+- **filter[name]** _(string)_
+
+  - Filter results by names that include the given filter, e.g. `filter[name]=New`.
+
 Example Response:
 
 ```
--- a/tests/Http/GroupTest.php
+++ b/tests/Http/GroupTest.php
@@ -16,40 +16,23 @@ class GroupTest extends TestCase
     public function testGroupTypeIndex()
     {
         $groupType = factory(GroupType::class)->create();
-        $groupTypeId = $groupType->id;
+        $groupNames = [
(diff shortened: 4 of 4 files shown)


app/Http/Controllers/GroupsController.php
```diff
@@ -33,6 +33,10 @@ public function index(Request $request)
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Group::$indexes);
 
+        if (isset($filters['name'])) {
+            $query->where('name', 'REGEXP', $filters['name']);
```
mendelB, line 37: This reminded me of Dave's question when I was adding a `REGEXP` where filter in https://github.com/DoSomething/rogue/pull/985#discussion_r373617249 

(We ended up filtering against the built-in list of causes which wouldn't work here tho...)

Could we use SQL `LIKE` though to achieve the same filter?

aaronschachter: The reason I went with `REGEXP` instead of `LIKE` was to avoid any gotchas with case sensitivity (we don't want our search to be case sensitive). Maybe using the `Str::slug` like in https://github.com/DoSomething/rogue/commit/4411570e342c27a42ca63509f5fa3a260b622e59 could help avoid adding a regex to search by?

This PR sure got me thinking about the Algolia work the platforms team is taking on, as a later iteration of the Groups resource will have an additional `state` field that we'll want to filter by for group types that have thousands of groups...

mendelB: Ah ok, hmm tho I suppose if you use `Str::slug` you'll need to like split/join on the hyphens? Maybe worth just using `preg_replace` which I think was another method I tried in that PR. ([This](https://stackoverflow.com/a/659030) seems to be a highly recommended spell).

And totes, Algolia sounds like a neat way to filter down the line!

aaronschachter: I'll add this into #1049.

mendelB: ~I suppose if you use Str::slug you'll need to like split/join on the hyphens?~ Nope, you can use a space as a second argument, Mendel, get a grip.

https://laravel.com/docs/5.5/helpers#method-str-slug

aaronschachter: I'm not so sure this will get us what we want. Some of these MFOL groups in #1047 have commas within the name. So if i search for "County, NJ" but run this filter through `Str::slug` -- I get a search string of `country nj` which won't match any of the rows that contain the substring `County, NJ`. It's for sure questionable whether a user would add commas into the search string at all, but we'd need to strip our results of the commas as well -- which is feeling a bit like over-engineering. I might give`preg_replace` a try instead per 75de1925df325650338bcccec570d42a3d66ea51, to allow commas within the search string.

PR #1050, opened 2020-06-22
Adds proof-of-concept Algolia search interface.

DFurnes:
### What's this PR do?

This pull request adds a proof-of-concept campaign search interface, powered by Algolia & GraphQL.

This is a "frontend" proof-of-concept, to be reviewed alongside DoSomething/graphql#247.

We're hoping to use Algolia as a managed search API (rather than manage our own Solr or ElasticSearch infrastructure) but we want to continue to query data dependencies consistently with GraphQL (especially when they eventually cross system boundaries, like loading information from Rogue & Contentful).

The nice thing about using GraphQL to make these queries, rather than something like Algolia's [InstantSearch](https://www.algolia.com/products/instantsearch/), is that we can keep the same exact query interface no matter how we're displaying data (and Algolia only needs to know about things that impact filtering, rather than adding one more place that data needs to be synced & duplicated).

### Relevant tickets

References [Pivotal #172922602](https://www.pivotaltracker.com/story/show/172922602).

### Checklist

### DIFF
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -303,6 +303,12 @@ class Campaign extends Model
             $array[$date] = optional($this->{$date})->timestamp;
         }
 
+        // If a campaign doesn't have an end date, we'll set a far-future one in
+        // Algolia to allow filtering. More context: <...>
+        if (empty($array['end_date'])) {
+            $array['end_date'] = 2147483647; // January 19th 2038
+        }
+
         // Only send data we want to search against.
         return Arr::only($array, [
             'accepted_count',
--- a/resources/assets/components/CampaignsTable.js
+++ b/resources/assets/components/CampaignsTable.js
@@ -10,15 +10,14 @@ import SortableHeading from './utilities/SortableHeading';
 
 const CAMPAIGNS_QUERY = gql`
   query CampaignsIndexQuery(
-    $isOpen: Boolean!
-    $orderBy: String!
+    $isOpen: Boolean
+    $filter: String
     $cursor: String
   ) {
-    campaigns: paginatedCampaigns(
+    campaigns: searchCampaigns(
       isOpen: $isOpen
-      orderBy: $orderBy
-      after: $cursor
-      first: 80
+      term: $filter
+      cursor: $cursor
     ) {
       edges {
         cursor
@@ -27,9 +26,6 @@ const CAMPAIGNS_QUERY = gql`
           internalTitle
           pendingCount
           acceptedCount
-          causes {
-            name
-          }
         }
       }
       pageInfo {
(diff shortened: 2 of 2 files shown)


app/Models/Campaign.php
```diff
@@ -303,6 +303,12 @@ public function toSearchableArray()
             $array[$date] = optional($this->{$date})->timestamp;
         }
 
+        // If a campaign doesn't have an end date, we'll set a far-future one in
+        // Algolia to allow filtering. More context: <...>
+        if (empty($array['end_date'])) {
+            $array['end_date'] = 2147483647; // January 19th 2038
+        }
```
DFurnes, line 310: This was a surprise to me: because of [restrictions on combining ANDs and ORs](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/in-depth/combining-boolean-operators/#combining-ands-and-ors), we can't simply say "find me campaigns where the start date is in the past, and the end_date exists & is in the future".

To allow us to query for "open" campaigns, I add a far-future "end date" here whenever the `end_date` field is null, so that we can [make a simple date comparison query for Algolia](https://github.com/DoSomething/graphql/pull/247/files#diff-a8806c93ceaf69be83f446966bbf9974R903-R904). 

This "January 19th 2038" date is just the date version of the maximum 32-bit integer (which, as far as I understand, is also the maximum date that we can safely encode into a UNIX timestamp). And, as much as it pains me to say, I'm guessing that we'll probably re-write this application _sometime_ befro 2038 rolls around. 😉

resources/assets/components/CampaignsTable.js
```diff
@@ -27,9 +26,6 @@ const CAMPAIGNS_QUERY = gql`
           internalTitle
           pendingCount
           acceptedCount
-          causes {
-            name
-          }
```
DFurnes, line 32: I removed this from the query since these values were only used for client-side filtering.

resources/assets/components/CampaignsTable.js
```diff
@@ -81,11 +47,11 @@ const CampaignsTable = ({ isOpen, filter }) => {
   const [orderBy, setOrderBy] = useState('pending_count,desc');
 
   const { error, loading, data, fetchMore } = useQuery(CAMPAIGNS_QUERY, {
-    variables: { isOpen, orderBy },
+    variables: { filter, isOpen /* orderBy */ },
```
DFurnes, line 50: One problem I ran into is that Algolia (somewhat surprisingly) [does not support sorting](https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/#backend-implementation-replicas) results out of the box. The suggested workaround seems to be to create separate replica indexes for each possible sort method (e.g. one where items are sorted by ID, one where they're sorted by title, etc.)

This is probably not a deal-breaker, but is worth considering when estimating our usage & definitely worth at least figuring out a proof-of-concept of here so we know how feasible this is for future product builds.

resources/assets/components/CampaignsTable.js
```diff
@@ -97,13 +63,6 @@ const CampaignsTable = ({ isOpen, filter }) => {
     });
   };
 
-  // If we're filtering results & can load more, do so automatically:
-  useEffect(() => {
-    if (filter !== '' && !loading && hasNextPage) {
-      handleViewMore();
-    }
-  }, [filter, endCursor]);
-
```
DFurnes, line 106: Since this was really a "hack" to deal with the fact that we were loading _all_ campaigns and searching client-side, I removed this (since it feels relatively unlikely for folks to search & not find what they're looking for on the first page).

resources/assets/components/CampaignsTable.js
DFurnes: Another option – if we don't plan to use sorting elsewhere – would be to separate _paging_ through campaigns (via the existing `paginatedCampaigns` query) from _searching_ for campaigns (which could maybe display in a separate "search auto-fill" interface when using the search field, and not be sortable).

app/Models/Campaign.php
aaronschachter: The terminators will take care of it!

PR #1055, opened 2020-06-24
Add city, state, and external_id fields to Groups

aaronschachter:
### What's this PR do?

This pull request three new text attributes to the `Group` model:`city`, `state`, and `external_id`. These will be populated when we create groups for our partner NASSP (import by [CSV](https://dosomething.slack.com/archives/CTVPG6L4R/p1592424800487000) will be added in a future PR) 

It also adds support to filter a `GET /posts` index request by the `state` and `external_id` fields.

Do the field names make sense? I figured I'd keep it simple with `city` and `state` vs `addr_city` and `addr_state` (to keep consistent with the Northstar fields). Our [schools DB / Schools GraphQL type also names these fields `city` and `state`](https://github.com/DoSomething/graphql/blob/b9fc0bd1478c7ffe270d5b4c664146e3bb24ef86/src/repositories/schools.js#L55), so it felt reasonable to keep consistent with those (vs users) :

Because the NASSP has significantly many more groups than our MFOL group type, the NASSP campaign group finders will ask users to first select their state, which will filter groups by name when searching for name after selecting the state (similar to how our [School Finder](https://dosomething.gitbook.io/phoenix-documentation/development/content-types/current-school-block) works).

We'll likely want to make the `external_id` read only in the edit form, but keeping it editable for now until we import real data.

### Relevant tickets

### Checklist


database/migrations/2020_06_24_000100_add_city_state_external_id_to_groups_table.php
```diff
@@ -0,0 +1,40 @@
...
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('groups', function (Blueprint $table) {
+            $table->dropUnique(['group_type_id', 'name']);
+            $table->string('external_id')->nullable()->index()->after('name');
+            $table->string('city')->nullable()->after('external_id');
+            $table->string('state', 10)->nullable()->after('city');
```
katiecrane, line 20: Does this limit to 10 characters? Asking because `Rhode Island`.

katiecrane: Ah I see in the docs down below that this will be stored like `RI`!

PR #1056, opened 2020-06-25
Adds command to import NASSP groups

aaronschachter:
### What's this PR do?

This pull request adds a console command to import a CSV from the National Association of Secondary School Principals (NASSP) as three different Group Types and their respective Groups. It also displays the `city` and `state` fields added in #1055 in our GroupsTable component, to help verify imported NASSP data.

Screen Shot 2020-06-25 at 10 49 33 AM
 

👀 

The NCES code is used as an ID for NASSP, so we save it to our `external_id` field. The file does not provide a "universal ID" -- if it did, we could that field to link these groups to their respective School in GraphQL. 

### Relevant tickets

### Checklist

### DIFF
--- a/app/Console/Commands/ImportNasspGroupsCommand.php
+++ b/app/Console/Commands/ImportNasspGroupsCommand.php
@@ -1,0 +1,128 @@
+<?php
+
+namespace Rogue\Console\Commands;
+
+use League\Csv\Reader;
+use Rogue\Models\Group;
+use Rogue\Models\GroupType;
+use Illuminate\Console\Command;
+
+class ImportNasspGroupsCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'rogue:nassp-groups-import {path}';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Create National Association of Secondary School Principals group types and groups from a CSV.';
+
+    /**
+     * Create a new command instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    /**
+     * @param string
+     * @return string
+     */
+    public function sanitize($value)
+    {
+        $result = trim($value);
+
+        return $result == 'NULL' ? null : $value;
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return mixed
+     */
+    public function handle()
+    {
+        $path = $this->argument('path');
+
+        info('rogue:nassp-groups-import: Loading csv from ' . $path);
+
+        // Make a local copy of the CSV.
+        $temp = tempnam(sys_get_temp_dir(), 'command_csv');
+
+        file_put_contents($temp, fopen($this->argument('path'), 'r'));
+
+        // Load CSV contents.
+        $csv = Reader::createFromPath($temp, 'r');
+
+        $csv->setHeaderOffset(0);
+
+        $numImported = 0;
+        $numFailed = 0;
+        $numSkipped = 0;
+
+        $nhsGroupType = GroupType::firstOrCreate([
+            'name' => 'National Honor Society',
+        ]);
+        $njhsGroupType = GroupType::firstOrCreate([
+            'name' => 'National Junior Honor Society',
+        ]);
+        $nascGroupType = GroupType::firstOrCreate([
+            'name' => 'National Student Council',
+        ]);
+
+        info('rogue:nassp-groups-import: Beginning import...');
+
+        /**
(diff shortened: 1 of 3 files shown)


app/Console/Commands/ImportNasspGroupsCommand.php
```diff
@@ -0,0 +1,133 @@
...
+        foreach ($records as $record) {
+            $name = trim($record['LABEL NAME']);
+
+            if (trim($record['NHS']) == 'NHS') {
+                $groupTypeId = $nhsGroupType->id;
+            } elseif (trim($record['NJHS']) == 'NJHS') {
+                $groupTypeId = $njhsGroupType->id;
+            } elseif (trim($record['NASC']) == 'NASC') {
+                $groupTypeId = $nascGroupType->id;
+            } else {
```
aaronschachter, line 104: The bottom 2 rows in this screenshot are an example of a school that appears twice in the CSV -- the first row will create it as a group within the NASC group type (National Student Council), and the second row will create it as a group within the NHS group type (National Honors Society). Both of these groups will share the same `external_id`, 480735000001 (these columns not visible in screenshot)

Screen Shot 2020-06-25 at 11 02 05 AM

app/Console/Commands/ImportNasspGroupsCommand.php
```diff
@@ -0,0 +1,133 @@
...
+     * Returns a trimmed CSV value, or null if equal to the string "NULL".
+     *
+     * @param string
+     * @return string
+     */
+    public function sanitize($value)
+    {
+        $result = trim($value);
+
+        return $result == 'NULL' ? null : $value;
```
mendelB, line 47: Should we be returning the `$result` instead of the `$value` here?

app/Console/Commands/ImportNasspGroupsCommand.php
```diff
@@ -0,0 +1,133 @@
...
+            'name' => 'National Student Council',
+        ]);
+
+        info('rogue:nassp-groups-import: Beginning import...');
+
+        /**
+         * Without specifying these column names, this command fails with error:
+         * "The header record must be an empty or a flat array with unique string value."
+         */
+        $records = $csv->getRecords(['LABEL NAME', 'NHS', 'NJHS', 'NASC', 'NASSP', 'ADDRESS1', 'ADDRESS2', 'CITY', 'STATE', 'POSTAL', 'COUNTRY', 'NCES']);
```
mendelB, line 93: Do we also need to specify the records that we're not using in this script?

app/Console/Commands/ImportNasspGroupsCommand.php
```diff
@@ -0,0 +1,133 @@
...
+        $csv = Reader::createFromPath($temp, 'r');
+
+        $csv->setHeaderOffset(0);
+
+        $numImported = 0;
+        $numFailed = 0;
+        $numSkipped = 0;
+
+        $nhsGroupType = GroupType::firstOrCreate([
+            'name' => 'National Honor Society',
```
aaronschachter, line 78: I think we'll need a boolean attribute like `filter_by_state` on the `GroupType` model that we'd set to `true` here, to indicate that a Group Finder should first require user to select a `state` field to filter groups by. We won't always want to filter by state, for example -- March For Our Lives only has 230 groups, and their source dataset didn't contain `city` and `state` fields. Because it's only the 230 groups, it's preferable to have the finder filter just by the name field only.

I held off on adding this field to the `GroupType` model to keep this PR reviewable. But been pondering the field name, `filter_by_state`. It's more to indicate "hey we have valid state fields that would work better for users finding their group" but I don't like the names `has_state` or `has_location` much either.

app/Console/Commands/ImportNasspGroupsCommand.php
aaronschachter: Argh yes -- 🦅 👀 thank you

app/Console/Commands/ImportNasspGroupsCommand.php
mendelB: Makes sense to me! I like the explicit `filter_by_state` option better as well.

app/Console/Commands/ImportNasspGroupsCommand.php
aaronschachter: Yeah, without adding them -- the wrong values get used. I think this may be an issue with the source CSV column headers having null values (there's like 3 blank columns but sometimes weird stuff appears for some rows here ). The source file isn't ready for a production import yet, so I may have some time to clean this up - esp because it will need to be revisited in [per comment below](https://github.com/DoSomething/rogue/pull/1056#discussion_r445749156).

PR #1057, opened 2020-06-25
Adds filter_by_state field to Group Types

aaronschachter:
### What's this PR do?

This pull request adds a boolean `filter_by_state` attribute to the `GroupType` model. If this is set, a group finder for the group type should display a "Select State" dropdown first, in order to filter the groups search by state as well as name.

We'll want to set this field to `true` in the NASSP import per https://github.com/DoSomething/rogue/pull/1056#discussion_r445749156.

Screen Shot 2020-06-25 at 12 30 11 PM

👀 

☑️ 

### Relevant tickets

References [Pivotal #173231765](https://www.pivotaltracker.com/story/show/173231765).

### Checklist

### DIFF
--- a/app/Http/Controllers/Web/GroupTypesController.php
+++ b/app/Http/Controllers/Web/GroupTypesController.php
@@ -40,6 +40,8 @@ class GroupTypesController extends Controller
         $values = $this->validate($request, array_merge_recursive($this->rules, [
             'name' => [Rule::unique('group_types')],
         ]));
+        // @see ActionsController->fillInOmittedCheckboxes
+        $values['filter_by_state'] = $request->has('filter_by_state');
 
         $groupType = GroupType::create($values);
 
@@ -72,6 +74,8 @@ class GroupTypesController extends Controller
         $values = $this->validate($request, array_merge_recursive($this->rules, [
             'name' => [Rule::unique('group_types')->ignore($groupType->id)],
         ]));
+        // @see ActionsController->fillInOmittedCheckboxes
+        $values['filter_by_state'] = $request->has('filter_by_state');
 
         $groupType->update($values);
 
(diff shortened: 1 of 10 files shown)


app/Http/Controllers/Web/GroupTypesController.php
```diff
@@ -40,6 +40,8 @@ public function store(Request $request)
         $values = $this->validate($request, array_merge_recursive($this->rules, [
             'name' => [Rule::unique('group_types')],
         ]));
+        // @see ActionsController->fillInOmittedCheckboxes
+        $values['filter_by_state'] = $request->has('filter_by_state');
```
DFurnes, line 44: Now that we're using this in two places, I wonder if it'd be helpful to generalize as a helper?

```php
function withOmittedBooleans(Request $request, $model): Request
{
    $booleans = array_keys(((new $class)->casts), 'boolean');

    foreach ($booleans as $field) {
        $request[$field] = $request->has($field);
    }

    return $request;
}
```

Then this could just read:

```php
$request = withOmittedBooleans($request, GroupType::class);

$this->validate($request, // ...
```

aaronschachter: That's an awesome idea! Just to clarify, you're thinking this goes in the [app/helpers.php](https://github.com/DoSomething/rogue/blob/master/app/helpers.php) ? Those functions have a mix of camelCase and snake_case

aaronschachter: I'll add this in a followup PR.

DFurnes: Yes! And oh good catch – let's do `with_omitted_booleans`.

PR #1059, opened 2020-06-29
Cleanup NASSP import

aaronschachter:
### What's this PR do?

Now that we have a [final source file to import](https://dosomething.slack.com/archives/CTVPG6L4R/p1593459055107200?thread_ts=1593452429.105500&cid=CTVPG6L4R) for the NASSP groups, this pull request makes a few tweaks to the `rogue:nassp-groups-import` console command added in #1056:

* Create an additional new group type, `National Association of Secondary School Principals`, which will contain each unique school/name/city row within the file. For this reason, we no longer need to skip a row if it doesn't have any values set in the `NHS`, `NJHS`, or `NASC` columns.

* Modifies column list per our final import file. I've uploaded the file to a [gist](https://gist.githubusercontent.com/aaronschachter/9cd40a703a43e22077e78e12e011a7de/raw/1d689400d64f461d51dce0e99d2b16334c36b10e/nassp.csv) and have been testing the import using that URL.

* Rename the `external_id` field to `school_id`, and save the `universalid` column that's been added on our end to the import file. Refs [Slack thread](https://dosomething.slack.com/archives/CTVPG6L4R/p1593462442109800?thread_ts=1593452429.105500&cid=CTVPG6L4R)

    * Change the Groups transformer and create/edit forms accordingly

 
### How should this be reviewed?

👀 

### Any background context you want to provide?

I plan to add documentation to start detailing some of these import commands that we may still use (and perhaps it's time to delete ones we know we'll never again use), but holding off on adding to this PR to help keep it reviewable.

### Relevant tickets

References [Pivotal #173408737](https://www.pivotaltracker.com/story/show/173408737).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [x] Added appropriate feature/unit tests.

PR #1060, opened 2020-06-30
Groups admin cleanup

aaronschachter:
### What's this PR do?

This pull request makes a few tweaks to the admin UI to make the spot checking the NASSP import (#1059) easier:

* Displays Group `state`, `city` attributes if present, links to the School page if `schoolId` is present
* Displays Groups associated with a School on a School page
* Adds a `/groups` route on web that renders the group-types list (available at `/group-types`, which is longer to type!)

Group page:

Group page

School page:

School page
 

👀 

🦉 

### Relevant tickets

### Checklist

### DIFF
--- a/resources/assets/Application.js
+++ b/resources/assets/Application.js
@@ -39,6 +39,9 @@ const Application = () => {
           <Route path="/campaigns/:id/:status">
             <ShowCampaign />
           </Route>
+          <Route path="/groups" exact>
+            <GroupTypeIndex />
+          </Route>
           <Route path="/group-types" exact>
             <GroupTypeIndex />
           </Route>
--- a/resources/assets/pages/ShowGroupType.js
+++ b/resources/assets/pages/ShowGroupType.js
@@ -63,9 +63,7 @@ const ShowGroupType = () => {
                 className="text-field"
                 onChange={event => setGroupState(event.target.value)}
               >
-                <option key="null" value={null}>
-                  -- Select State --
-                </option>
+                <option value={''}>-- Select State --</option>
 
                 {usaStateOptions.map(state => (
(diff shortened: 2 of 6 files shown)


resources/assets/Application.js
```diff
@@ -39,6 +39,9 @@ const Application = () => {
           <Route path="/campaigns/:id/:status">
             <ShowCampaign />
           </Route>
+          <Route path="/groups" exact>
+            <GroupTypeIndex />
+          </Route>
           <Route path="/group-types" exact>
             <GroupTypeIndex />
           </Route>
```
mendelB, line 47: Should we just remove this route 🤷

resources/assets/pages/ShowGroupType.js
```diff
@@ -63,10 +63,14 @@ const ShowGroupType = () => {
                 className="text-field"
                 onChange={event => setGroupState(event.target.value)}
               >
-                <option value={null}>-- Select State --</option>
+                <option key="null" value={null}>
```
weerd, line 66: Why do you have to set a key and value here? Just curious; don't remember well enough if this is needed and why 🤷‍♂️

resources/assets/Application.js
aaronschachter: I think it's fine to keep since we have a `/group-types/:id`

resources/assets/pages/ShowGroupType.js
aaronschachter: We don't need the key -- thought we'd need it to avoid the React unique key warning. But we do want to set a value here to clear the `state` filter in the `groups` GraphQL query -- setting it to an empty string instead of null gets this working correctly.

PR #1065, opened 2020-07-09
Removing Disabled Attribute on Form Dropdowns to allow for deselecting

lindsayrmaher:
### What's this PR do?

This pull request adds an `isOptional` flag to the form select blade template to decide whether the disabled attribute is added to the default option.

👀 

For most of the dropdowns on the site, we want to be able to require a user to select an option and use the default as strictly a placeholder. In the case of the groupType dropdown, we are adding the ability for users to deselect a group type if needed.

### Relevant tickets

References [Pivotal # 173454237](https://www.pivotaltracker.com/story/show/173454237).

### Checklist

### DIFF
--- a/resources/views/forms/select.blade.php
+++ b/resources/views/forms/select.blade.php
@@ -1,6 +1,6 @@
 <div class="select">
     <select name="{{ $name }}">
-        <option value="" {{ ! old($name) && ! isset($value) ? 'selected' : '' }}>{{ $placeholder or '–' }}</option>
+        <option value="" {{ ! old($name) && ! isset($value) ? 'selected' : '' }} {{ ! isset($isOptional) ? 'disabled' : ''}}>{{ $placeholder or '–' }}</option>
         @foreach($options as $option => $label)
             @if (old($name))
                 <option value="{{ $option }}" {{ old($name) === $option ? 'selected': '' }}>{{ $label }}</option>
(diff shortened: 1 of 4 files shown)


resources/views/forms/select.blade.php
```diff
@@ -1,6 +1,6 @@
 <div class="select">
     <select name="{{ $name }}">
-        <option value="" {{ ! old($name) && ! isset($value) ? 'selected' : '' }} disabled>{{ $placeholder or '–' }}</option>
+        <option value="" {{ ! old($name) && ! isset($value) ? 'selected' : '' }} {{ ! isset($isOptional) ? 'disabled' : ''}}>{{ $placeholder or '–' }}</option>
```
aaronschachter, line 3: Would this work too? I think it could help readability a bit. I know the current implementation matches the check for `value` before it - but it took me a minute to figure out what was going on.

```
$isOptional ? '' : 'disabled'
```

aaronschachter: Or Dave's suggestion in [Slack](https://dosomething.slack.com/archives/CUQMU4Q6B/p1594325875005100?thread_ts=1594325742.004500&cid=CUQMU4Q6B) I think also helps with less operators by using a `required` parameter vs`isOptional`:
```
isset($required) ? 'disabled' : ''
```

lindsayrmaher: Mostly I was trying not to have to update the `actions` blade templates for this PR since that's the only other place we use the `form.select`. If we do the first change we'd have to add `isOptional` as a variable on both of those templates. 

And as far as the second option, that would make more sense readability wise, but I again would need to set the `required` variable on the `actions` templates. 

Do you think it's worth it for the sake of future work and having it be readable? Or maybe we could even just change the variable name to just `optional` so that it makes more sense inside the `isset`

PR #1066, opened 2020-07-09
Adds group attributes to Customer.io events

aaronschachter:
### What's this PR do?

This pull request adds `group_id`, `group_name`, `group_type_id`, and `group_type_name` attributes to Customer.io events for signups and posts. This enables support for displaying a group's info within Customer.io message templates.

Example message template: 
`Thanks for signing up with your {{event.group_name}} chapter of {{event.group_type_name}}!`

Example message text:
`Thanks for signing up with your Monmouth County, NJ chapter of March For Our Lives!`

👀 

This PR also modifies `PostObserver` from #1064 to use the new `$post->group` attribute added in this PR, to DRY.

### Relevant tickets

References [Pivotal #173481072](https://www.pivotaltracker.com/story/show/173481072).

### Checklist

### DIFF
--- a/app/Models/Group.php
+++ b/app/Models/Group.php
@@ -57,6 +57,14 @@ class Group extends Model
     }
 
     /**
+     * Get the posts associated with this group.
+     */
+    public function posts()
+    {
+        return $this->hasMany(Post::class);
+    }
+
+    /**
      * Get the signups associated with this group.
      */
     public function signups()
--- a/app/Models/Group.php
+++ b/app/Models/Group.php
@@ -49,12 +49,10 @@ class Group extends Model
     public static $sortable = ['name'];
 
     /**
-     * Get the Group Type model associated with the group_type_id attribute.
-     *
-     * @return Rogue\Models\GroupType
+     * Get the group type associated with this group.
      */
-    public function getGroupTypeAttribute()
+    public function group_type()
     {
-        return GroupType::find($this->group_type_id);
+        return $this->hasOne(GroupType::class, 'id');
     }
 }
--- a/app/Models/Group.php
+++ b/app/Models/Group.php
@@ -53,6 +53,6 @@ class Group extends Model
      */
     public function group_type()
     {
-        return $this->hasOne(GroupType::class, 'id');
+        return $this->belongsTo(GroupType::class);
     }
 }
--- a/app/Models/Group.php
+++ b/app/Models/Group.php
@@ -47,4 +47,14 @@ class Group extends Model
      * @var array
      */
     public static $sortable = ['name'];
+
+    /**
+     * Get the Group Type model associated with the group_type_id attribute.
+     *
+     * @return Rogue\Models\GroupType
+     */
+    public function getGroupTypeAttribute()
+    {
+        return GroupType::find($this->group_type_id);
+    }
 }
--- a/app/Models/Group.php
+++ b/app/Models/Group.php
@@ -55,4 +55,12 @@ class Group extends Model
     {
         return $this->belongsTo(GroupType::class);
     }
+
+    /**
+     * Get the signups associated with this group.
+     */
+    public function signups()
+    {
+        return $this->hasMany(Signup::class);
+    }
 }
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -253,6 +253,16 @@ class Post extends Model
     }
 
     /**
+     * Get the Group model associated with the group_id attribute.
+     *
+     * @return Rogue\Models\Group
+     */
+    public function getGroupAttribute()
+    {
+        return $this->group_id ? Group::find($this->group_id) : null;
+    }
+
+    /**
      * Create a hard-to-guess "hash" ID.
      *
      * @return string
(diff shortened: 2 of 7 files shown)


app/Models/Group.php
```diff
@@ -47,4 +47,14 @@ class Group extends Model
...
     public static $sortable = ['name'];
+
+    /**
+     * Get the Group Type model associated with the group_type_id attribute.
+     *
+     * @return Rogue\Models\GroupType
+     */
+    public function getGroupTypeAttribute()
+    {
+        return GroupType::find($this->group_type_id);
```
mendelB, line 58: I don't think this makes all that much difference, but this made me think of Rails Active Record relationships where there are convenience methods for relationships, [it looks](https://laravel.com/docs/5.5/eloquent-relationships#one-to-one) like we could do something similar here with a `groupType` method and a `hasOne(GroupType)`.

EDIT: Looks like we use this on [Posts#review](https://github.com/DoSomething/rogue/blob/7b16ef1157d1442b944dcd7a020dc5b3bdb36b14/app/Models/Post.php#L155-L161)!

app/Models/Post.php
```diff
@@ -371,6 +384,10 @@ public function toBlinkPayload()
             'source_details' => $this->source_details,
             'details' => $this->details,
             'referrer_user_id' => $this->referrer_user_id,
+            'group_id' => $this->group_id,
+            'group_name' => isset($group) ? $group->name : null,
+            'group_type_id' => isset($group) ? $group->group_type_id : null,
+            'group_type_name' => isset($group) ? $group->group_type->name : null,
```
mendelB, line 390: Would 'grouping' (ha) this and the below list of attributes in `getReferralPostEventPayload` into like a `getGroupPayload` -or something- helper be useful or overkill? (Maybe useful for the Signup as well?)

app/Models/Group.php
aaronschachter: I did try to use [`hasOne`](https://laravel.com/docs/5.5/eloquent-relationships#one-to-one) in a `group()` method per the Laravel docs, but it'd result in:

```
ErrorException: Undefined property: Illuminate\Database\Eloquent\Relations\HasOne::$name
```
when trying to access the name property of the return.

Looking more closely at the docs, I think this doesn't work because even though we do define a foreign key on Signups and Posts for the `group_id` fields -- there aren't any fields on the `Group` model that reference the `Signup` or `Post` models.

app/Models/Post.php
aaronschachter: I like this idea.. maybe this could be a static method on the Group model, since we won't always have a group on a signup or a post -- but would still want to return null values for all of these array keys.

app/Models/Group.php
mendelB: Ahhhh right -- looks like in this case it's a `belongsTo` relationship! I wonder if that method would work!

app/Models/Post.php
mendelB: Ooh, that makes good sense to me!

app/Models/Group.php
aaronschachter: That still results in a similar error:
```
ErrorException: Undefined property: Illuminate\Database\Eloquent\Relations\BelongsTo::$name
```
also tried to specify the foreign/locak keys within the method and still get the error.

It sure feels like I'm doing something wrong though, because I'd expect this to work. @DFurnes -- anything you can 👀 as far as any 🐛 in getting these Eloquent relationships working here? I thought maybe it was because they'd only be used for query building, but the docs sure make it seem like this should work.

aaronschachter: > It sure feels like I'm doing something wrong though

I was! I didn't realize the relationship methods had to be added on each side of the relationship 🤦 . Fixed this up in https://github.com/DoSomething/rogue/pull/1066/commits/8aa8d792cd47aefbdf86c2853a354b185e4285e8, and going to make similar changes to the `Signup` and `Post` models (that's what I thought you were referring to at first when I first read this comment.. which was way too early to be reading 🌅 ☕ )

mendelB: 🙏  #childNumberOneAndChildNumberTwo

app/Models/Group.php
```diff
@@ -47,4 +47,12 @@ class Group extends Model
      * @var array
      */
     public static $sortable = ['name'];
+
+    /**
+     * Get the group type associated with this group.
+     */
+    public function group_type()
+    {
+        return $this->hasOne(GroupType::class, 'id');
```
mendelB, line 56: I am surprised tho that a one-way `belongsTo` didn't do the trick 🤔

app/Models/Post.php
aaronschachter: I'm going to hold off on this for now.. the one thing I wouldn't be thrilled about is the `group_` fields appearing after the `created_at` and `updated_at` fields when viewing the events with Customer.io. Minor, but can always revisit.

aaronschachter: Ah, testing this and Customer.io orders the payload properties alphabetically anyway. I'll add this in a cleanup PR.

PR #1071, opened 2020-07-20
Upgrading Laravel Framework

weerd:
### What's this PR do?

This pull request upgrades Rogue from Laravel 5.5 to Laravel 5.8!

👀 

I didn't separate these commit by commit but I did try to leave useful notes as comments for some interesting tidbits or changes I got stuck on or where a bit more involved!

🌵 

### Relevant tickets

References [Pivotal #172383643](https://www.pivotaltracker.com/story/show/172383643).

### Checklist


app/Http/Middleware/TrustProxies.php
```diff
@@ -8,23 +8,16 @@
...
+     * @var int
      */
-    protected $headers = [
-        Request::HEADER_FORWARDED => null, // Not set on AWS or Heroku.
-        Request::HEADER_X_FORWARDED_FOR => 'X_FORWARDED_FOR',
-        Request::HEADER_X_FORWARDED_HOST => null, // Not set on AWS or Heroku.
-        Request::HEADER_X_FORWARDED_PORT => 'X_FORWARDED_PORT',
-        Request::HEADER_X_FORWARDED_PROTO => 'X_FORWARDED_PROTO',
-    ];
+    protected $headers = Request::HEADER_X_FORWARDED_ALL;
```
weerd, line 22: I'm less familiar with Trusted Proxy and want to make sure the new bit property is fine that should encompass the prior series of headers.

config/trustedproxy.php
```diff
@@ -1,27 +1,50 @@
...
+     * within TrustedProxy to trust any proxy
+     * that connects directly to your server,
+     * a requirement when you cannot know the address
+     * of your proxy (e.g. if using ELB or similar).
      *
-     * The "**" character is syntactic sugar within TrustedProxy to trust not just any
-     * proxy that connects directly to your server, but also proxies that connect to
-     * those proxies, and all the way back until you reach the original source IP.
+     */
+    'proxies' => '*', // [<ip addresses>,], '*', '<ip addresses>,'
```
weerd, line 18: According to the [`TrustedProxies` Class](https://github.com/fideloper/TrustedProxy/blob/9beebf48a1c344ed67c1d36bb1b8709db7c3c1a8/src/TrustProxies.php#L71) in the package, the `**` value is deprecated in favor of `*`. I checked and we use an env variable on each Rogue env on Heroku and specify it as `**`. By defaulting the config to `*` it should cover the same things and we can remove the env variables from each env on Heroku.

weerd:
Noting here that after the upgrades and running tests I was running into an odd error with tests that use the `withAdminAccessToken()` method.

Instead of the expected series of method calls, after the 5.6 updates, there is a resulting call to the `__isset()` magic method in the `HasAttributes` trait utilized by the `RemoteUser` class in our `Gateway` package.

The only thing I could find related in Laravel 5.6 changes was this merged [pull request](https://github.com/laravel/framework/pull/24042) that adds an `__isset()` to the `Optional` class (we do use this class within Rogue and within the method calls related to `withAdminAccessToken()`).

---

*Update*: This issue was resolved by @DFurnes via an update to [Gateway](https://github.com/DoSomething/gateway/pull/129).

app/Http/Controllers/PostsController.php
```diff
@@ -175,12 +175,14 @@ public function update(PostRequest $request, Post $post)
...
+        $validatedRequest = $request->validated();
+
         // But don't allow user's to review their own posts.
         if (! Gate::allows('review', $post)) {
-            unset($request['status']);
+            unset($validatedRequest['status']);
         }
 
-        $this->posts->update($post, $request->validated());
+        $this->posts->update($post, $validatedRequest);
```
weerd, line 185: This was a funky bug. Some of the logic and return values for validation in 5.7 have changed a bit and `$request->validated()` was causing the previously unset `status` array item in the `Gate` conditional above to be popped back into the array, causing the test to fail for post owner's not being able to review their own post.

So I revised a logic order a bit and call `$request->validated()` before the `Gate` check, so the final `$validatedRequest` array passed to the `update()` method only contains the correct, allowed items.

Although I think this could be simplified to just pass through the request array as:

```php
$this->posts->update($post, $request->toArray());
```

I didn't change it to the simpler approach above in case I was missing something with _why_ we need to use the `validated()` method, but given that this is data passed through the `PostRequest` it should already be valid, and not sure what `validated()` is adding (especially with the [changes in 5.7](https://github.com/laravel/framework/blob/2555bf6ef6e6739e5f49f4a5d40f6264c57abd56/src/Illuminate/Validation/Validator.php#L312-L319) where when `validate()` is called, the data is sent through the `validated()` method at the end of the process as well. 🤷‍♂️

app/Policies/PostPolicy.php
```diff
@@ -28,7 +29,7 @@ public function viewAll($user, Post $post)
      * @param  Rogue\Models\Post $post
      * @return bool
      */
-    public function show($user, Post $post)
+    public function show($user = null, Post $post)
```
weerd, line 32: @placeholder... fill in details regarding this bug fix.

routes/api.php
```diff
@@ -10,60 +10,60 @@
...
 
     // group types
-    $this->get('group-types', 'GroupTypesController@index');
-    $this->get('group-types/{groupType}', 'GroupTypesController@show');
+    Route::get('group-types', 'GroupTypesController@index');
+    Route::get('group-types/{groupType}', 'GroupTypesController@show');
 
     // users
-    $this->delete('users/{id}', 'UsersController@destroy');
+    Route::delete('users/{id}', 'UsersController@destroy');
```
weerd, line 68: After update dependencies for upgrade from 5.7 to 5.8, after running composer update was getting the following error in the `routes/api.php` file:

```
Call to undefined method Illuminate\Routing\RouteFileRegistrar::post()
```

Simple fix is to just use the `Route` alias.

config/database.php
```diff
@@ -1,30 +1,7 @@
...
 
-// Convert Heroku's `REDIS_URL` to standard `REDIS_HOST`, `REDIS_PORT`,
-// and `REDIS_PASSWORD` environment variables:
-if (env('REDIS_URL')) {
-    $url = parse_url(env('REDIS_URL'));
-
-    putenv('REDIS_HOST='.$url['host']);
-    putenv('REDIS_PORT='.$url['port']);
-    putenv('REDIS_PASSWORD='.$url['pass']);
-}
```
weerd, line 11: [Laravel 5.8 now supports a `REDIS_URL` variable](https://github.com/laravel/laravel/pull/5037/) and will parse it correctly, so no need to do this manually anymore.

.env.example
```diff
@@ -13,20 +14,27 @@ DB_USERNAME=homestead
...
+FILESYSTEM_DRIVER=public
+QUEUE_CONNECTION=sync
 SESSION_DRIVER=redis
-STORAGE_DRIVER=public
-QUEUE_DRIVER=sync
 
-REDIS_HOST=127.0.0.1
-REDIS_PASSWORD=null
-REDIS_PORT=6379
+REDIS_URL=redis://h:null@127.0.0.1:6379
```
weerd, line 21: See [this comment](https://github.com/DoSomething/rogue/pull/1071/files#r461733624).

app/Listeners/SendTaggedNotification.php
```diff
@@ -2,8 +2,8 @@
 
 namespace Rogue\Listeners;
 
-use Notification;
 use Rogue\Events\PostTagged;
+use Illuminate\Support\Facades\Notification;
```
weerd, line 6: This seemed weird to me. Not sure why the code in the event listener worked find even though the `use` statement was borked, referring to a non-existent, global `Notification` class 🤷‍♂️

app/Models/Campaign.php
```diff
@@ -20,7 +20,7 @@ class Campaign extends Model
      *
      * @var array
      */
-    protected $dates = ['deleted_at', 'start_date', 'end_date'];
+    protected $dates = ['start_date', 'end_date'];
```
weerd, line 23: Since this is the first instance in the alphabetical list of file reviews, commenting here; the framework not automagically ✨ converts `deleted_at` to Carbon date instances, so no need to specify it within the `$dates` array property :)

app/Notifications/SlackTagNotification.php
```diff
@@ -5,21 +5,12 @@
...
-    query SlackNotificationQuery($userId: String!, $adminId: String!) {
-        user(id: $userId) {
-            displayName
-        }
-        admin: user(id: $adminId) {
-            displayName
-        }
-    }
-';
-
```
weerd, line 22: Removed this query from here and added a new method to the `GraphQL` class, so that we can easily [mock the request](https://github.com/DoSomething/rogue/pull/1071/files#diff-506a2b9afeee7eb30e7ce345deedc128R69-R76).

app/Services/GraphQL.php
```diff
@@ -98,4 +98,31 @@ public function getUserById($userId)
...
+            }
+        }';
+
+        $variables = [
+            'userId' => $userId,
+            'adminId' => $adminId,
+        ];
+
+        return $this->query($query, $variables);
+    }
```
weerd, line 127: Query pulled out from `SlackTagNotification.php` class, so that it can be mocked when testing.

config/database.php
```diff
@@ -116,22 +105,40 @@
...
 
+        'client' => env('REDIS_CLIENT', 'predis'),
+
         'cluster' => false,
 
+        // Not sure if fine to replace cluster option above with this new options item added prior to 5.8.
+        // 'options' => [
+        //     'cluster' => env('REDIS_CLUSTER', 'predis'),
+        //     'prefix' => env('REDIS_PREFIX', Str::slug(env('APP_NAME', 'laravel'), '_').'_database_'),
+        // ],
```
weerd, line 123: I think updating this to the new `options` should be fine, but wanted to flag it for thoughts... 🤔

tests/Http/TagsTest.php
```diff
@@ -56,10 +58,11 @@ public function testTaggingAPostWithoutActivityScope()
 
         // Apply the tag to the post
         $response = $this->postJson('api/v3/posts/' . $post->id . '/tags', [
-            'tag_name' => 'Good Submission',
+            'tag_name' => 'Group Photo',
```
weerd, line 61: Changed a number of these tag names to be `Group Photo` since any tag with `Good *` will trigger a Slack notification. Most of our devs don't see this happen locally since they don't have the `SLACK_WEBHOOK_INTEGRATION_URL` env variable set, but I did and was seeing notifications go through on the test channel when running `phpunit`.

tests/Http/TagsTest.php
```diff
@@ -17,6 +16,9 @@ class TagsTest extends TestCase
      */
     public function testTaggingAndUntaggingAPost()
     {
+        // Supress sending Slack notifications for "good submission" tagging.
+        $this->withoutNotifications();
```
weerd, line 20: I left this test tagging post with `Good Submission` and using the `withoutNotifications()` method as an example so we know it's an option to suppress notifications when testing and since this test mentioned creating a new event anyhoo.

.env.example
```diff
@@ -13,20 +14,27 @@ DB_USERNAME=homestead
 DB_PASSWORD=secret
 
 CACHE_DRIVER=redis
+FILESYSTEM_DRIVER=public
+QUEUE_CONNECTION=sync
```
weerd, line 18: I went ahead and changed these based on what the framework now uses since I liked that they follow more closely to the filename and the array setting they are used in. I can make sure to update these on Heroku prior to deploy (add them before deploy, and remove the old named ones after successful deploy).

app/Providers/AppServiceProvider.php
```diff
@@ -17,19 +16,23 @@
...
      *
      * @return void
      */
     public function boot()
     {
-        $this->app->singleton(GateContract::class, function ($app) {
-            return new CustomGate($app, function () use ($app) {
-                return call_user_func($app['auth']->userResolver());
-            });
-        });
```
weerd, line 31: Related to [this comment](https://github.com/DoSomething/rogue/pull/1071/files#r459600449), since Laravel now allows guest users we no longer need the `CustomGate` class.

weerd:
Just FYI, this test is super finicky from recent experience and planning on rewriting it once I'm done with upgrades.

image

.env.example
DFurnes: We'll need to update these in our Terraform config [here](https://github.com/DoSomething/infrastructure/blob/246a5c8333226d50419a9affc50617084448431f/components/s3_bucket/main.tf#L203) and [here](https://github.com/DoSomething/infrastructure/blob/246a5c8333226d50419a9affc50617084448431f/components/sqs_queue/main.tf#L53). (We can start by adding the "new" ones alongside the old ones, and then remove the old ones after we've updated all of our apps to use the new environment variable names).

DFurnes: I'm a little surprised by the `QUEUE_DRIVER` to `QUEUE_CONNECTION` change... it seems a little inconsistent with the "driver" terminology used elsewhere. I'm curious what the rationale (for Taylor & co.) was behind that!

.env.example
```diff
@@ -13,20 +14,27 @@ DB_USERNAME=homestead
...
-REDIS_PASSWORD=null
-REDIS_PORT=6379
+REDIS_URL=redis://h:null@127.0.0.1:6379
+
+# AWS
+AWS_ACCESS_KEY=
+AWS_REGION=us-east-1
+AWS_SECRET_KEY=
+AWS_S3_BUCKET=
+AWS_S3_URL=
```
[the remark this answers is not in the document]
DFurnes: This is another one we'll need to update in our Terraform config ([here](https://github.com/DoSomething/infrastructure/blob/246a5c8333226d50419a9affc50617084448431f/components/s3_bucket/main.tf#L204-L205) and [here](https://github.com/DoSomething/infrastructure/blob/246a5c8333226d50419a9affc50617084448431f/components/iam_app_user/main.tf#L31-L32)). Like the others, we'll want to keep that old environment variables until we've confirmed everything is updated to use these new names.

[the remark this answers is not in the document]
DFurnes: Since we're changing these now, I'd like to use `AWS_ACCESS_KEY_ID` in place of `AWS_ACCESS_KEY` & `AWS_SECRET_ACCESS_KEY` in place of `AWS_SECRET_KEY` (for consistency with [Laravel's defaults](https://github.com/laravel/laravel/blob/a7b70fcc310fee9d78448e078fef9003150ee8aa/.env.example) and [Amazon's CLI environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html).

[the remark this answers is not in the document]
DFurnes: I think we'll also want to maintain separate `AWS_REGION` variables for different services (e.g. `AWS_S3_REGION` and `AWS_SQS_REGION`) since those might be in different regions. (I mean, 100% unlikely, but it's possible!)

app.json
```diff
@@ -77,10 +77,13 @@
     "ASHES_USERNAME": {
       "required": true
     },
-    "ROGUE_API_KEY": {
+    "AWS_S3_BUCKET": {
+      "required": true
+    },
+    "AWS_S3_URL": {
```
DFurnes, line 83: Do review apps use S3? I'm curious if we should actually have these use the local filesystem `FILESYSTEM_DRIVER`, since we can't (automatically) provision a new S3 bucket for each review app & so they might clobber each other...

app/Http/Controllers/PostsController.php
DFurnes: Oh wow, glad we had a test to cover this! :relieved: 

The reason we use the returned `$validatedRequest` is because it trims out any additional properties not covered by the [validator](https://github.com/DoSomething/rogue/blob/8bcfe954aa20233afe0377cedd45d5a03bb218de/app/Http/Requests/PostRequest.php) (sort-of like how Mass Assignment protection works, but we actually use it)! So for example, if someone included `diego => 'hacker'` in their post payload, it'd be excluded from `$validatedRequest`.

app/Http/Middleware/TrustProxies.php
DFurnes: I think we want to use the `Request::HEADER_X_FORWARDED_AWS_ELB` constant here (which, I think, like our previous manual settings here) would ignore the `Forwarded` and `X-Forwarded-Host` values that aren't used on Heroku or AWS...)

We might be fine with `_ALL`... but I'd just want to double-check that those extra headers can't be misused.

DFurnes: Ah yep, here's Heroku's [documentation on how to configure this](https://devcenter.heroku.com/articles/getting-started-with-laravel#trusting-the-load-balancer) which suggests `Request:: HEADER_X_FORWARDED_AWS_ELB`:

> It is very important to also prevent Laravel from trusting the `Forwarded` and `X-Forwarded-Host` headers, because Heroku’s router does not set those, but Symfony’s Request component trusts them out of the box once a trusted proxy is set.

app/Listeners/SendTaggedNotification.php
DFurnes: This is one of those very spooky things about facades! Laravel actually intercepts the "class not found" event emitted by PHP and subs in the facade, so I think(?!) excluding the import altogether is the intended usage. (But personally, I find that super confusing so glad we added this import in!)

app/Notifications/SlackTagNotification.php
DFurnes: Ah, got it.

I prefer keeping queries colocated with their usage if we can (since it's easier to reason about what data is being fetched, and how changing a query would affect anything that depends on it). I wonder if there's a way we could keep this query here and instead mock the request "further down" based on the `SlackNotificationQuery` name (like we do in Cypress tests!)

DFurnes: Not something we need to sort out now, but I'd love to consider thinking about this...

app/Http/Middleware/TrustProxies.php
weerd: Ah ok! Thanks for the info on this @DFurnes 🙇

app/Policies/PostPolicy.php
DFurnes: I think you can typehint this to `Illuminate\Contracts\Auth\Authenticatable` (as `?Authenticatable`) which both our local users (by [extending](https://github.com/DoSomething/rogue/blob/8bcfe954aa20233afe0377cedd45d5a03bb218de/app/Models/User.php#L7) [this class](https://github.com/laravel/framework/blob/78eb4dabcc03e189620c16f436358d41d31ae11f/src/Illuminate/Foundation/Auth/User.php#L5)) and remote users ([here](https://github.com/DoSomething/gateway/blob/0ae0f829e96e0e674c4906cc3444ec57eacf5b11/src/Server/RemoteUser.php#L6)) share.

app/Providers/AppServiceProvider.php
```diff
@@ -56,7 +58,7 @@ public function boot()
         }, 'The :attribute must be a valid ISO-3166-2 region code.');
 
         // Attach the user & request ID to context for all log messages.
-        Log::getMonolog()->pushProcessor(function ($record) {
+        Log::getLogger()->pushProcessor(function ($record) {
```
DFurnes, line 61: Could you double-check that these are still writing as expected? I think I remember running into trouble with this & needing to extract into a [`ContextFormatter`](https://github.com/DoSomething/northstar/blob/master/app/Logging/ContextFormatter.php) class (which is then applied to our log stacks [here](https://github.com/DoSomething/northstar/blob/4442295cce2affd0c99408f9d576e46bb2dd384e/config/logging.php)).

config/database.php
```diff
@@ -58,34 +35,46 @@
...
             'charset' => 'utf8',
             'collation' => 'utf8_unicode_ci',
             'prefix' => '',
+            'prefix_indexes' => true,
             'strict' => false,
             'engine' => null,
+            // Following was added in 5.8, not sure if needed.
+            // 'options' => extension_loaded('pdo_mysql') ? array_filter([
+            //     PDO::MYSQL_ATTR_SSL_CA => env('MYSQL_ATTR_SSL_CA'),
+            // ]) : [],
```
DFurnes, line 62: I'd love to do a bit more digging on this... betting it makes sense to keep, but curious what it's for!

The [commit where it was added](https://git.io/JJzZw) is not very helpful...

config/database.php
DFurnes: Yep, I think this is fine!

tests/Http/TagsTest.php
DFurnes: Let's disable this in tests across the board by setting `SLACK_WEBHOOK_INTEGRATION_URL` to an empty string in our `phpunit.xml` config! That way we don't need to suppress test-by-test.

tests/Http/TagsTest.php
DFurnes: Left a comment [below](#discussion_r462356992)... I think we can keep all these as-is by swapping out the Slack webhook URL when running tests!

app/Policies/PostPolicy.php
weerd: Ah true! Ok, will give this a shot!

.env.example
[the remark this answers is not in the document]
weerd: > Since we're changing these now, I'd like to use AWS_ACCESS_KEY_ID in place of AWS_ACCESS_KEY & AWS_SECRET_ACCESS_KEY in place of AWS_SECRET_KEY (for consistency with Laravel's defaults and Amazon's CLI environment variables.

Sounds good! I was debating changing those or not since I was already changing a bunch and wasn't sure how much of a pain it'd be to change. But you're right, since we're doing the rest we might as well and definitely prefer to be consistent with Laravel's defaults 👍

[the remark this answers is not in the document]
weerd: @DFurnes changing these up and doing your recommendation of `AWS_S3_REGION` and `AWS_SQS_REGION` but wondering what's the call for the [aws config file](https://github.com/DoSomething/rogue/blob/8bcfe954aa20233afe0377cedd45d5a03bb218de/config/aws.php#L20). It just generally specifies `region`, I'm assuming it's the S3 one, but want to confirm 🤔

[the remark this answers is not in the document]
weerd: Also I guess for [this one](https://github.com/DoSomething/rogue/blob/8bcfe954aa20233afe0377cedd45d5a03bb218de/config/services.php#L48), but we don't use SES so... 🤷 

Maybe 3 items? `AWS_REGION`, `AWS_S3_REGION`, & `AWS_SQS_REGION`?

config/database.php
weerd: So @DFurnes doesn't matter that the `cluster` was previously set as a boolean `false`, but now it defaults to a string of `predis`?

tests/Http/TagsTest.php
weerd: Ah nice, ok. 

Only caveat is if we did want to write a specific test for Notifications we wouldn't be able to since it'd be dependent on this env var being present 🤔

config/database.php
DFurnes: We're not using clustering, but I'd assume the default options would be fine regardless... This doesn't seem to match what's shown [in the documentation](https://laravel.com/docs/6.x/redis#configuration), though, so maybe we should follow that instead.

.env.example
weerd: Ya, I wondered that myself... only explanation I could come up with is that with the "connections" config settings, each connection specifies its own `driver` so I suspect it would have been weird to specify a top level queue driver and then within another driver property, so:

```php
config('database.connections.pgsql.driver`);
```
vs

```php
config('database.driver.pgsql.driver`);
```

🤷‍♂️

config/database.php
weerd: Ah, that's the docs for 6.x which I didn't get to yet! Haha, I'll update and then change it in the next PR 😂 Good find.

weerd: Also just noting those docs are missing the Redis URL option 🤔 Might need to submit a PR 😉

.env.example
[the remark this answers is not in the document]
DFurnes: Hmm.. I'm not sure if we actually need that config file or not... might be interesting to try removing it & see if anything breaks. And yeah, I think those look good. Can't hurt to add an `AWS_SES_REGION` I guess and just not set it (maybe it defaults to `us-east-1`)?

app/Notifications/SlackTagNotification.php
weerd: 🤔 that's fair! And your [suggestion](https://github.com/DoSomething/rogue/pull/1071#discussion_r462356992) seems like it would resolve the testing issues I was having.

config/database.php
weerd: Ya, I think I had checked the commit and didn't see much info there, but didn't get a chance to dig further 🤷‍♂️

app.json
weerd: Oh fair. I was only updating the variable name that was already there, but good to question whether it should even be there!

I checked against the [5.8 defaults](https://github.com/laravel/laravel/blob/5.8/config/filesystems.php#L16) and looks like `local` is set as the default but we have it set to `s3`. Probably something we changed a while back but I think defaulting to `local` makes more sense so I can remove the `AWS_*` env vars for review apps and if `FILESYSTEM_DRIVER` defaults to `local` we should be in a better place!

weerd: Updated! I also realized there were a bunch of old `ASHES_*` env variables for the review apps which I [removed as well](https://github.com/DoSomething/rogue/pull/1071/commits/cddadb2cd8695add8ce6d8d7cc16b92074aee544).

app/Providers/AppServiceProvider.php
weerd: Yep, I think you're right and not working as expected. I hadn't really reviewed those items in the logs before, but they're mostly all `null` I think. It didn't throw errors so I had assumed it was working. Looking into fixing it now!

weerd: Ok, without changing anything and testing it out, it looks like the `user_id` is populated fine, but `client_id` and `request_id` end up `null` in most logs. I think `request_id` being `null` is fine, since this is set by Heroku right? But would have expected `client_id` to be set... or is it ok that's it's `null` on my local?

weerd: Since these _seemed_ to be logging as expected, I left [a note](https://github.com/DoSomething/rogue/pull/1071/commits/3e16a73b7a1045ce87333f3c73b424be41114fa5) to potentially update this in the future using same approach as Northstar.

.env.example
DFurnes: Ohhh, I bet that's it! :bulb:

app/Notifications/SlackTagNotification.php
weerd: Restored old approach since the PHPUnit env var setting keeps notifications from triggering and solves the need to mock this.

.env.example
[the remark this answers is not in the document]
weerd: I double-checked and we don't change any of the default `aws` config options that are in the package so I do think you're right and this `config/aws.php` file is redundant. Removing and then adding those other `AWS_*` env vars.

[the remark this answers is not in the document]
weerd: Ok, since there's also a setting for `dynamodb` cache setting I didn't want to add `AWS_DYNAMODB_REGION` as well, so I just added:

```
AWS_REGION=us-east-1
AWS_S3_REGION=us-east-1
AWS_SQS_REGION=us-east-1
```

Specific region env var for S3 and SQS since we use those and then everything else can just refer to the default `AWS_REGION`. If we need something more specific in the future we can always revise.

PR #1074, opened 2020-07-27
Rename ActionStats accepted_quantity to impact 

aaronschachter:
### What's this PR do?

This pull request adds a migration to rename the `accepted_quantity` int column on the `action_stats` table to `impact`. For voter registration, we'd like to store the total number of voter-reg posts with status starting with `register-` here -- so I figured it'd be simplest to rename the field to something more generic that can be re-purposed for various post types.

It also fixes the same newfound StyleCI error we saw pop up in https://github.com/DoSomething/phoenix-next/pull/2298, and then fixes new style errors that popped up after that change.

:eyes:

Was debating on whether to add a separate field instead of renaming `accepted_quantity`, but the TLDR from this [Slack discussion](https://dosomething.slack.com/archives/CP2D7UGAU/p1595881689011800) made me think two separate fields would still require documentation to explain the purpose of a potential secondary field like `completed_count`.I'd think we could get away with docs that explain what the `impact` means for each post type (docs to come in a future PR if we feel good about this name change) 

### Relevant tickets

### Checklist

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -6,7 +6,6 @@ enabled:
 disabled:
   - concat_without_spaces
   - no_useless_return
-  - simplified_null_return
   - phpdoc_no_package
   - phpdoc_summary
   - phpdoc_var_without_name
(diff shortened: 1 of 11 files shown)


aaronschachter:
This will entail a few other changes in our repos, e.g. the [GraphQL `SchoolActionStat` schema](https://github.com/DoSomething/graphql/blob/master/src/schema/rogue.js#L410-L411), but this PR should be save to merge because the TFJ campaign that referenced these ActionStats has been closed.

PR #1075, opened 2020-07-28
Actually Validate Referrer User ID in Signups

mendelB:
### What's this PR do?

This pull request actually validates the `referrer_user_id` field in Signups.

In #1073 I mistakenly added the validation to the legacy `SignupRequest` which isn't used in our active `/v3/signups` route.

To atone for my sins, I added tests this time and removed the legacy `SignupRequest` which doesn't seem to be used anyplace.

Shout if we should _not_ be deleting this legacy `SignupRequest`. 

Are the tests ok? TBH I'm never sure whether testing a plain validation makes sense, but clearly in this case, it would have been helpful! So I'm the fool.

I kept in the `nullable` allowance. Not sure how useful this is for us? I sort just copied that idea from the legacy `SignupRequest`.

I tried to above.

### Relevant tickets

References [Pivotal #173514593](https://www.pivotaltracker.com/story/show/173514593).

### Checklist


app/Http/Controllers/SignupsController.php
```diff
@@ -58,6 +58,7 @@ public function store(Request $request)
             'campaign_id' => 'required_without:action_id|integer',
             'action_id' => 'required_without:campaign_id|integer',
             'why_participated' => 'string',
+            'referrer_user_id' => 'nullable|objectid',
```
katiecrane, line 61: I think keeping `nullable` here is fine!

tests/Http/SignupTest.php
```diff
@@ -84,6 +84,58 @@ public function testCreatingASignupWithGroupId()
...
+        $this->mock(Blink::class)->shouldReceive('userSignup');
+
+        $response = $this->withAccessToken($northstarId)->postJson('api/v3/signups', [
+            'campaign_id' => $campaignId,
+            'referrer_user_id' => $referrerUserId,
+        ]);
+
+        $response->assertStatus(422);
+        $response->assertJsonValidationErrors(['referrer_user_id']);
+    }
```
katiecrane, line 137: I agree these tests are helpful for the reasons outlined in your write up!

PR #1076, opened 2020-07-28
If signup has a group_id, save it on the post as well

katiecrane:
### What's this PR do?

When a post is being created, we'll check to see if its signup has a `group_id` set and if it does, we'll save the same `group_id` to the post as well.

I wrote the test failing test first, added the logic in the `PostObserver`, made sure the test was passing as well as testing manually. Did I miss anything? Does this mess with the `school_id` logic at all?

It's all in the ticket!

### Relevant tickets

References [Pivotal #173750109](https://www.pivotaltracker.com/story/show/173750109).

### Checklist

### DIFF
--- a/app/Observers/PostObserver.php
+++ b/app/Observers/PostObserver.php
@@ -15,6 +15,11 @@ class PostObserver
      */
     public function creating(Post $post)
     {
+        // If the post's signup has a group_id, save it on the post as well
+        if ($post->signup->group_id) {
+            $post->group_id = $post->signup->group_id;
+        }
+
         // If we have a group_id but no school_id, save the group's school_id if exists.
         if ($post->group_id && (! $post->school_id) && $group = $post->group) {
             $post->school_id = $group->school_id;
--- a/app/Observers/PostObserver.php
+++ b/app/Observers/PostObserver.php
@@ -16,7 +16,7 @@ class PostObserver
     public function creating(Post $post)
     {
         // If the post's signup has a group_id, save it on the post as well
-        if (!$post->group_id && $post->signup->group_id) {
+        if (! $post->group_id && $post->signup->group_id) {
             $post->group_id = $post->signup->group_id;
         }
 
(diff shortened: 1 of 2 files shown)


tests/Http/PostTest.php
```diff
@@ -1628,4 +1628,51 @@ public function testCreatingVoterRegistrationPost()
...
+
+        // Mock the Blink API call.
+        $this->mock(Blink::class)->shouldReceive('userSignupPost');
+
+        // Create the post!
+        $response = $this->withAccessToken($signup->northstar_id)->postJson('api/v3/posts', [
+            'northstar_id'     => $signup->northstar_id,
+            'campaign_id'      => $signup->campaign_id,
+            'type'             => $action->post_type,
+            'action'           => $action->name,
```
aaronschachter, line 1656: I don't think we need this property... I believe there's an ancient TODO to deprecate support for this `action` parameter once all Contentful entries have an `action_id` field (and can't find the related cleanup Pivotal story)

app/Observers/PostObserver.php
```diff
@@ -15,6 +15,11 @@ class PostObserver
      */
     public function creating(Post $post)
     {
+        // If the post's signup has a group_id, save it on the post as well
+        if ($post->signup->group_id) {
```
aaronschachter, line 19: I think we might want an additional check in this conditional to only set the `post->group_id` if it hasn't been set yet. The chompy import is an example where we create the post with a `group_id` parameter, and the signup wouldn't have existed before creating the post.

tests/Http/PostTest.php
katiecrane: Oh great eye @aaronschachter ! I missed that there was `action` and `action_id` here

app/Observers/PostObserver.php
katiecrane: I added a test for that scenario that passed without any changes, but I added the check for existing `post->group_id` in anyway just to be safe!

PR #1078, opened 2020-07-29
Adds referral_signup_created Customer.io Event

mendelB:
### What's this PR do?

This pull request follows the excellent #1018 by example to dispatch a `referral_signup_created` event to customer.io when a Signup is created with a `referral_user_id` so that we can send a dedicated email to the referrer.

commit-by-commit. You can refer to #1018 for slightly more context, but hopefully, this is super straightforward as-is!

See the Pivotal Ticket for the list of requested payload attributes.

### Relevant tickets

References [Pivotal #173563077](https://www.pivotaltracker.com/story/show/173563077).

### Checklist

### DIFF
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -248,4 +248,25 @@ class Signup extends Model
             }
         }]);
     }
+
+    /**
+     * Gets event payload for a referral signup, on behalf of the referrer user ID.
+     */
+    public function getReferralSignupEventPayload()
+    {
+        $userId = $this->northstar_id;
+        $user = app(GraphQL::class)->getUserById($userId);
+
+        $campaignWebsite = app(GraphQL::class)->getCampaignWebsiteByCampaignId($this->campaign_id);
+
+        return [
+            'id' => $this->id,
+            'user_id' => $userId,
+            'user_display_name' => isset($user) ? $user['displayName'] : null,
+            'campaign_id' => (string) $this->campaign_id,
+            'campaign_title' => $campaignWebsite['title'],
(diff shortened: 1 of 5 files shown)


app/Models/Signup.php
```diff
@@ -248,4 +248,25 @@ public function scopeWithVisiblePosts($query, $types = null)
...
+        $userId = $this->northstar_id;
+        $user = app(GraphQL::class)->getUserById($userId);
+
+        $campaignWebsite = app(GraphQL::class)->getCampaignWebsiteByCampaignId($this->campaign_id);
+
+        return [
+            'id' => $this->id,
+            'user_id' => $userId,
+            'user_display_name' => isset($user) ? $user['displayName'] : null,
+            'campaign_id' => (string) $this->campaign_id,
```
aaronschachter, line 266: No biggie just curious to why is this being cast as a string?

mendelB: I was copying the format from the main Blink event for signups: https://github.com/DoSomething/rogue/blob/8627ed64be6f88d035c556be34e5d913485aa015/app/Models/Signup.php#L201

Buuuut, I _believe_ that's a Blink concern, and this is going directly to Customer.io so I wonder if the rationale is moot?

mendelB: 🤔 Though if c.io is expecting a string, we might need to remain consistent in any case?

aaronschachter: Yeah, makes sense to leave since that's what we're sending for the `campaign_signup` events.

weerd:
Post-merge 👍

PR #1079, opened 2020-07-30
Create ActionStats for completed voter registrations

aaronschachter:
### What's this PR do?

This pull request creates or updates an `ActionStat` for a school upon creating or updating a completed `voter-reg` post. We'll eventually use this data to display a leaderboard of voter registrations by school (details in #[2417735](https://www.pivotaltracker.com/n/projects/2417735)).

It implements this by refactoring the existing Post `getAcceptedQuantitySum` method as `updateOrCreateActionStats`, and by calling it from newly added `created` and `updated` hooks in the PostObserver.

Also adds a static `getCompletedVoterRegStatuses` method to the Post model to DRY getting list of statuses considered completed.

👀 

* We have existing test coverage for creating ActionStats for approved posts in test for the Reviews endpoint, so didn't add it into our model test for now. Might be worth moving it later to keep the tests in one place.

* A followup PR will add a `state` field to the `action_stats` table, in order to filter action stats by state. Holding on adding that in this PR to hopefully keep this easier to review.

### Relevant tickets

References [Pivotal #2417735](https://www.pivotaltracker.com/n/projects/2417735).

### Checklist

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -520,7 +520,7 @@ class Post extends Model
             return;
         }
 
-        return $query->whereIn('status', ['accepted', 'register-form', 'register-OVR'])
+        return $query->whereIn('status', array_merge(['accepted'], self::getCompletedVoterRegStatuses()))
             ->orWhere('northstar_id', auth()->id())
             ->orWhere(function ($query) {
                 $query->whereNotNull('referrer_user_id')
@@ -655,7 +655,7 @@ class Post extends Model
     }
 
     /**
-     * Returns all status values of completed voter registrations.
+     * Returns all status values for completed voter registrations.
      *
      * @return array
      */
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -5,6 +5,7 @@ namespace Rogue\Models;
 use Hashids\Hashids;
 use Rogue\Services\GraphQL;
 use Rogue\Events\PostTagged;
+use Rogue\Models\ActionStat;
 use Rogue\Models\Traits\HasCursor;
 use Illuminate\Support\Facades\Cache;
 use Illuminate\Database\Eloquent\Model;
@@ -590,6 +591,33 @@ class Post extends Model
     }
 
     /**
+     * Updates or creates action stat if completed voter registration post for a school.
+     *
+     * @return int
+     */
+    public function updateOrCreateActionStats()
+    {
+        // If this is a completed voter registration for a group with a school:
+        if ($this->type === 'voter-reg' && $this->group && $this->group->school_id && in_array($this->status, ['register-form', 'register-OVR'])) {
+            $schoolId = $this->group->school_id;
+
+            // Find total number of completed voter reg posts for the school.
+            $schoolCompletedVoterRegPostCount = (new self)->newModelQuery()
+                ->where('school_id', $schoolId)
+                ->where('type', 'voter-reg')
+                ->where('action_id', $this->action_id)
+                ->where('status', 'in', self::getCompletedVoterRegStatuses())
+                ->count();
+
+            // Save to action stat.
+            ActionStat::updateOrCreate(
+                ['action_id' => $this->action_id, 'school_id' => $schoolId],
+                ['impact' => $schoolCompletedVoterRegPostCount]
+            );
+        }
+    }
+
+    /**
      * Get the sum quantity of accepted posts with given action and school ID.
      *
      * @return int
@@ -622,5 +650,12 @@ class Post extends Model
             'created_at' => $this->created_at->toIso8601String(),
             'updated_at' => $this->updated_at->toIso8601String(),
         ], Group::toBlinkPayload($this->group));
+    }
+
+    /**
+     * @return array
+     */
+    public static function getCompletedVoterRegStatuses() {
+
     }
 }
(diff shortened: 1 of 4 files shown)


app/Repositories/PostRepository.php
```diff
@@ -205,14 +204,6 @@ public function reviews(Post $post, $status, $comment = null, $admin = null)
         // Update the "counter cache" on the Post Campaign:
         $post->campaign->refreshCounts();
 
-        // If the Post has a School ID, upsert an ActionStat.
-        if ($post->school_id && $post->action_id) {
```
mendelB, line 209: Why were we checking for the `action_id` here? I noticed we omit that check above?

mendelB:
Random Q -- did we ever intend to start queuing the `ActionStat` generation, or is their performance trivial?

> A followup PR will add a state field to the action_stats table, in order to filter action stats by state. Holding on adding that in this PR to hopefully keep this easier to review.

Also lol does this mean like 'status' `state` or the post location? (assuming the former, but that did confuse me since we have both on posts 😅 )

aaronschachter:
Re: `state`... I do mean post location. I did think about calling `school_state`... but I have a feeling we eventually may want to refactor the `action_stats` table to support storing action stats for groups -- right now we only store by schooll. I think it'd be a mess to have both `school_state` and `group_state` fields, since they'd both be identical... but agree that `state` isn't exactly crystal clear ...

app/Repositories/PostRepository.php
aaronschachter: I think this was maybe to safeguard for when posts are created with an `action` vs `action_id` -- but I'm not sure this was even necessary.

app/Models/Post.php
```diff
@@ -590,21 +590,51 @@ public static function getPostCount(Campaign $campaign, string $status)
...
+        // We currently only save action stats for posts that belong to schools.
+        if (! $this->school_id) {
+            return;
+        }
+
+        // If completed voter-reg post, update school impact as completed count for this action.
+        if ($this->type === 'voter-reg' && in_array($this->status, self::getCompletedVoterRegStatuses())) {
+            $impact = (new self)->newModelQuery()
+                ->where('school_id', $this->school_id)
+                ->where('type', 'voter-reg')
```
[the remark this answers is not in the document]
aaronschachter: I think in theory yes -- but the action ID might change for voter reg posts per election year.

aaronschachter:
> Random Q -- did we ever intend to start queuing the ActionStat generation, or is their performance trivial?

I think for now, it's not too often we see posts with school ID's, so I haven't been too concerned... cc @DFurnes

mendelB:
> Re: `state`... I do mean post location. I did think about calling `school_state`... but I have a feeling we eventually may want to refactor the `action_stats` table to support storing action stats for groups -- right now we only store by schooll. I think it'd be a mess to have both `school_state` and `group_state` fields, since they'd both be identical... but agree that `state` isn't exactly crystal clear ...

Ahh fair. Hmm I suppose `location_state` isn't exactly elegant either 🤔  We call this simply `location` on Posts, which we do have filters for in Phoenix galleries, do you think staying consistent is a win?

aaronschachter:
I see that the posts `location` field contains values like `US-VT` -- but we store `VT` in a `state` field on both the Group Rogue [resource](https://github.com/DoSomething/rogue/blob/master/docs/endpoints/groups.md) and [schools](https://github.com/DoSomething/graphql/blob/9e17f767e24414f89c4382038a92833cec43d7ae/src/schema/schools.js#L21), we also allow filtering by it too. I kind of lean towards `state` being more consistent vs `location`, because of the data and existing `state` filters?

DFurnes:
For naming, I think I'd recommend using `location` for consistency (and because, like Mendel mentioned, `state` has a lot of different meanings). That said, I totally get that `location` is maybe a little to vague! If I were renaming the `location` field today, I think I'd call it `region` since these are ISO-3166-2 region codes (and not necessarily _just_ US states).

DFurnes:
I don't know if it's too late, but I'd love to store the group/school locations as ISO-3166-2 codes as well for consistency.

DFurnes:
> I think for now, it's not too often we see posts with school ID's, so I haven't been too concerned... cc @DFurnes

Yeah, I think keeping things simple & then optimizing when we see a performance issue seems smart! I think right now the biggest performance outlier in this application (ugh) is re-calculating post counts on the `/api/v3/reviews` endpoint.

aaronschachter:
> I don't know if it's too late, but I'd love to store the group/school locations as ISO-3166-2 codes as well for consistency.

It's too late for Monday's launch, both the Great Schools Mongo DB and our `state` field in the Rogue `groups` table are storing the US codes. It'd require corresponding changes in GraphQL and the `UsaStateSelect` Phoenix component too. A good cleanup task would be to update all data in each field and rename.

mendelB:
So are you thinking keeping this as `state` pending an epic cleanup across the batch of 'em?

aaronschachter:
I'm thinking of going with `location`, and storing it as ISO-3166-2. It's one less to cleanup later. Changing the `state` field name in our `Group` model isn't terribly tricky since we can create a new field called `location` and migrate state data over as   ISO-3166-2, so maybe I'll sneak it into the next PR. Eventually we can switch the GraphQL queries to query by `location` instead of `state` (for GreatSchools, we could get away with formatting the field in GraphQL vs running a db query.. which would keep the DB update process simple)

mendelB:
Ahh wicked neat, that sounds like a smart way to go!!

PR #1081, opened 2020-07-30
Upgrading Laravel framework from 5.8 to 6.0

weerd:
### What's this PR do?

This pull request upgrades Laravel from 5.8 to the latest LTS version, 6.0.

👀 

🌵 

### Relevant tickets

References [Pivotal #172383643](https://www.pivotaltracker.com/story/show/172383643).

### Checklist


tests/Http/CampaignTest.php
```diff
@@ -192,19 +192,17 @@ public function testCampaignCursorWithOrderBy()
...
 
         // Then, we'll use the last post's cursor to fetch the remaining two:
         $lastCursor = $response->json()['data'][2]['cursor'];
         $response = $this->withAdminAccessToken()->getJson($endpoint . '&cursor[after]=' . $lastCursor);
-        $data = $response->json()['data'];
 
-        $this->assertArraySubset(['id' => $two->id, 'pending_count' => 2], $data[0]);
-        $this->assertArraySubset(['id' => $one->id, 'pending_count' => 1], $data[1]);
+        $response->assertJson(['data' =>[0 => ['id' => $two->id, 'pending_count' => 2]]]);
+        $response->assertJson(['data' =>[1 => ['id' => $one->id, 'pending_count' => 1]]]);
```
weerd, line 205: Updating to remove use of `assertArraySubset()` which was [deprecated in PHPUnit 8.0](https://github.com/sebastianbergmann/phpunit/issues/3495) and was throwing warnings. This approach seemed to work, if a little harder to read 🤷‍♂️ 

The Laravel Framework copied and has their own version of `assertArraySubset()`, but apparently it's recommended only for use inside the test cases in the framework and not meant to be extended outside of it. Thus I went for an alternative approach which did end up passing the tests.

weerd:
FYI: I'm gonna pull together a list of ENV variables that have been changed/added so we can update both Heroku and Terraform.

.env.example
```diff
@@ -22,11 +22,13 @@ REDIS_URL=redis://h:null@127.0.0.1:6379
...
 # AWS
 AWS_ACCESS_KEY_ID=
-AWS_REGION=us-east-1
+AWS_DEFAULT_REGION=us-east-1
 AWS_SECRET_ACCESS_KEY=
 AWS_S3_BUCKET=
 AWS_S3_REGION=us-east-1
 AWS_S3_URL=
+AWS_SQS_PREFIX=
+AWS_SQS_QUEUE=
```
weerd, line 31: While Laravel keeps these unprefixed, I think we're better off prefixing them so all the `AWS_` vars group up nicely in the Heroku config settings interface and easier to see/edit.

config/filesystems.php
```diff
@@ -50,6 +50,7 @@
 
         'public' => [
             'driver' => 'local',
+            // @TODO: should we change to storage_path('app/public')?
```
weerd, line 53: This is an interesting, but subtle difference. 

`public_path('storage')` points to `/public/storage/`

`storage_path('app/public')` points to `/storage/app/public/`

config/filesystems.php
```diff
@@ -61,8 +62,8 @@
             'secret' => env('AWS_SECRET_ACCESS_KEY'),
             'region' => env('AWS_S3_REGION', 'us-east-1'),
             'bucket' => env('AWS_S3_BUCKET'),
-            'base_url' => 'https://s3-external-1.amazonaws.com',
-            'public_url' => env('AWS_S3_URL'),
+            'url' => env('AWS_S3_URL'),
+            'endpoint' => env('AWS_ENDPOINT'),
```
weerd, line 66: Should I add `AWS_ENDPOINT` to the `.env.example` and would it be `https://s3-external-1.amazonaws.com`? Not sure if I matched these up right.

weerd:
I've updated Heroku and added any _new_ (revised names too) ENV variables for when this all gets merged into the main branch.

.editorconfig
```diff
@@ -19,9 +19,5 @@ indent_size = 2
 [*.md]
 trim_trailing_whitespace = false
 
-<<<<<<< HEAD
 [*.{yml,yaml}]
-=======
-[*.yml]
->>>>>>> bd96b84e... Upgrading Laravel Framework (#1071)
```
weerd, line 26: Look like I missed this one in prior rebase on `all-laravel-upgrades` branch 😬

.env.example
katiecrane: Are these two keys new?

app/Http/Kernel.php
```diff
@@ -55,13 +55,13 @@ class Kernel extends HttpKernel
...
         'auth.basic' => \Illuminate\Auth\Middleware\AuthenticateWithBasicAuth::class,
-        'guard' => \DoSomething\Gateway\Server\Middleware\SetGuard::class,
-        'scopes' => \DoSomething\Gateway\Server\Middleware\RequireScope::class,
         'bindings' => \Illuminate\Routing\Middleware\SubstituteBindings::class,
         'cache.headers' => \Illuminate\Http\Middleware\SetCacheHeaders::class,
         'can' => \Illuminate\Auth\Middleware\Authorize::class,
         'guest' => \Rogue\Http\Middleware\RedirectIfAuthenticated::class,
         'signed' => \Illuminate\Routing\Middleware\ValidateSignature::class,
+        'guard' => \DoSomething\Gateway\Server\Middleware\SetGuard::class,
+        'scopes' => \DoSomething\Gateway\Server\Middleware\RequireScope::class,
```
katiecrane, line 64: Do we not want this section alphabetical?

.env.example
weerd: They exist as `SQS_PREFIX` and `SQS_QUEUE` in our Rogue config in Heroku. New in that I'm renaming them to add the `AWS_` prefix so all AWS related keys group up together in Heroku.

config/app.php
```diff
@@ -37,7 +39,7 @@
     |
     */
 
-    'debug' => env('APP_DEBUG', false),
+    'debug' => (bool) env('APP_DEBUG', false),
```
katiecrane, line 42: Oooo smart!

app/Http/Kernel.php
weerd: yep yep, will update!

config/app.php
weerd: It's like that in updated Framework, so not my official doing, but I do agree it is smart!

config/auth.php
```diff
@@ -104,9 +100,9 @@
     'passwords' => [
         'users' => [
             'provider' => 'users',
-            'email' => 'auth.emails.password',
```
katiecrane, line 107: Did this part just not apply?

config/database.php
```diff
@@ -117,24 +116,24 @@
...
 
         'default' => [
             'url' => env('REDIS_URL'),
-            'host' => env('REDIS_HOST', 'localhost'),
+            'host' => env('REDIS_HOST', '127.0.0.1'),
             'password' => env('REDIS_PASSWORD', null),
-            'port' => env('REDIS_PORT', 6379),
-            'database' => env('REDIS_DB', 0),
+            'port' => env('REDIS_PORT', '6379'),
+            'database' => env('REDIS_DB', '0'),
```
katiecrane, line 128: Hm curious why these need to be strings?

config/filesystems.php
katiecrane: Oooo interesting - seems safer to keep the same for now?

config/filesystems.php
katiecrane: Can you point me to the docs on this bit? I can see if the matching makes sense to me!

config/logging.php
```diff
@@ -89,6 +89,15 @@
             'driver' => 'errorlog',
             'level' => 'debug',
         ],
+
+        'null' => [
+            'driver' => 'monolog',
+            'handler' => NullHandler::class,
+        ],
```
katiecrane, line 96: omg what is `null` logging?

tests/Http/CampaignTest.php
katiecrane: I like this approach!

.env.example
katiecrane: Ah so just missing from the `.env.example` previously - love the renaming!

katiecrane: Sorry I'm not sure where these are referenced in the code, but I don't see any updates and would have expected to see somewhere `SQS_PREFIX` being changed to `AWS_SQS_PREFIX` or did that already happen?

weerd: They get used in the [`config/queue.php`](https://github.com/DoSomething/rogue/pull/1081/files#diff-baaf2a02f4c1142ceaa77e6baf4442d0R56-R57).

katiecrane: Oh right - turns out I had already checked on that and marked as viewed 🤦‍♀️

config/app.php
katiecrane: It was all you, Diego!

config/filesystems.php
weerd: Ya, agreed. I might just make a ticket on the Platforms board to discuss the merits.

weerd: Ticket created! <https://www.pivotaltracker.com/story/show/174198565>

config/filesystems.php
weerd: Ok, I found these 2 PRs related to the change:

<https://github.com/laravel/laravel/pull/5267>
<https://github.com/laravel/laravel/pull/5276>

Also randomly amusing is how the PRs relate and one ends in `67` versus `76` 😆

weerd: So from what I gather they're technically both the same, just that it should have originally been `endpoint` but for backwards compatibility they're keeping both 🤔 So I think the env should be `AWS_S3_URL` for both?

config/auth.php
weerd: It was an old 5.2 vestige that was never removed: <https://github.com/laravel/laravel/blob/5.2/config/auth.php>

config/logging.php
weerd: They go nowhere! Haha <https://github.com/laravel/laravel/pull/5106>

config/database.php
weerd: I wasn't sure, but looked into it and [this PR explains it](https://github.com/laravel/laravel/pull/5191)... found the reasoning interesting in the quoted example provided in the description!

weerd: tldr; Maintains consistency of value type if an env variable for this exists, it'll be a string.

config/auth.php
katiecrane: a-ha!

config/database.php
katiecrane: oh interesting!

config/filesystems.php
katiecrane: Oh no I think I'm more confused now - are we planning to use `AWS_S3_URL` for both `url` and `endpoint`?

weerd: Ok digging further Laravel never really had a `base_url` property here. It was added in #25 and it doesn't actually look like it's used anywhere. So I think it's safe to keep both `url` and `endpoint` and have them both use `AWS_S3_URL` which is an actual env var we do use. Thoughts?

katiecrane: That sounds good to me!!

tests/Http/CampaignTest.php
DFurnes: Agreed, I think this is great!

PR #1086, opened 2020-08-05
Adds VoterRegistrationSeeder

aaronschachter:
### What's this PR do?

This pull request adds an additional [DatabaseSeeder class](https://laravel.com/docs/7.x/seeding#running-seeders) to populate the `action_stats` data, making it a whole lot easier to test pagination (#1085).

```
php artisan db:seed --class=VoterRegistrationSeeder
```

It also adds a new text file containing 1000 valid school ID's for use when generating a fake school ID.

:eyes:

I've added a few calls to the static methods in `app/Types` to help DRY copy/pasting strings throughout the codebase. I think it'd be nice to start DRY thes as we find them: 

Examples:

https://github.com/DoSomething/rogue/blob/3a68beb14f392ef0bbf11054c2e9985002af571c/app/Models/Post.php#L671)

https://github.com/DoSomething/rogue/blob/3a68beb14f392ef0bbf11054c2e9985002af571c/app/Models/Post.php#L627

https://github.com/DoSomething/rogue/blob/3a68beb14f392ef0bbf11054c2e9985002af571c/app/Http/Requests/PostRequest.php#L80-L92

### Relevant tickets

### Checklist


database/faker/FakerSchoolId.php
```diff
@@ -4,31 +4,15 @@
...
      * Return a random School ID.
      *
      * @return mixed
      */
     public function school_id()
     {
-        return static::randomElement(static::$ids);
+        $ids = file('database/faker/schools.txt');
+
+        return trim($ids[array_rand($ids)]);
```
katiecrane, line 16: Sorry I feel like I've asked a similar question about `school_id`s before, but why do we need a specific list of ids for this rather than choosing a random string?

app/Types/PostType.php
```diff
@@ -4,9 +4,9 @@
 
 class PostType extends Type
 {
-    private const TEXT = 'text';
-    private const PHOTO = 'photo';
-    private const VOTER_REG = 'voter-reg';
+    public const TEXT = 'text';
+    public const PHOTO = 'photo';
+    public const VOTER_REG = 'voter-reg';
```
katiecrane, line 9: I know you were asking this and I'm not sure the answer of public vs. private here, but as it stands this file now has a mismatch of public and private!

database/faker/FakerSchoolId.php
aaronschachter: No need to apologize! It's so we seed with real school ID's, so GraphQL queries like the ones added in #1085 to view school names don't break locally.

app/Types/PostType.php
aaronschachter: Yep! I wanted to make sure this is okay before changing them all in a future PR, to keep them consistent and to keep the changes down on this one.

database/faker/FakerSchoolId.php
katiecrane: Ah that does sound super familiar!

database/seeds/DatabaseSeeder.php
```diff
@@ -30,22 +29,18 @@ public function run()
...
                         'northstar_id' => $signup->northstar_id,
-                        'group_id' => $group->id,
                     ]));
 
                     $signup->posts()->saveMany(factory(Post::class, rand(2, 4))->states('photo', 'pending')->create([
                         'action_id' => $photoAction->id,
                         'signup_id' => $signup->id,
                         'campaign_id' => $signup->campaign_id,
                         'northstar_id' => $signup->northstar_id,
-                        'group_id' => $group->id,
```
katiecrane, line 48: I thought you had just added these - why don't we want `group_id` anymore?

aaronschachter: I did! It'd be ideal to add these for a campaign with a group type. I think it'd be cleaner to add a separate `GroupSeeder` similar to this to create photo posts for schools/groups, which also generates action stats.

katiecrane: Ah got it - that makes sense!

app/Types/PostType.php
katiecrane: Got it! So I don't see any issues with these being `public` but I don't know for sure

aaronschachter: cc @DFurnes - I'd think this change should be okay to make -- if so, i'll open a follow up PR to make all of these types public to avoid confusion.

DFurnes: These are built with [php-enum](https://github.com/myclabs/php-enum)'s `Enum` class, which I think should expose these via static methods (e.g. `PostType::VOTER_REG()` to get this enum from outside this class).

I _think_ the reason they don't suggest defining these as public is because that would allow you to change an enum at runtime by accident, whereas accessing it via the method keeps it read-only. That said, if that doesn't work I don't think it's a big deal to change!

DFurnes: Although hmm, wouldn't `const` do that? If so, yeah I don't see a reason not to make public! 🤔 The only other advantage of using the static method is it returns a type-checkable instance (I think), but we don't use that here.

aaronschachter: The static method worked! While we're here, I'd love to eventually add a `app/Types/PostStatus` to DRY all of the hardcoded `'accepted'`, `'rejected'` status value strings with similar const / static methods.

database/faker/FakerSchoolId.php
aaronschachter: Added a comment to document! 📝

app/Types/PostType.php
DFurnes: Yes please! 🙌

aaronschachter: I thought calling the static method within the factory for voter-reg posts worked, but this [unit test failed:](https://github.com/DoSomething/rogue/blob/67805b04e81c2636f171e46e701a4957e9ad2b83/tests/Unit/Models/PostModelTest.php#L197-L202)
```
Failed asserting that a row in the table [action_stats] matches the attributes {
    "action_id": 21,
    "impact": 7,
    "school_id": "101757",
    "location": "US-FL"
}.

The table is empty..

```
Something's odd with using the static method and type checking, because this function (which gets called in PostObserver) doesn't run to create action stats:

https://github.com/DoSomething/rogue/blob/67805b04e81c2636f171e46e701a4957e9ad2b83/app/Models/Post.php#L604-L617

Changed the factory back to just return `voter-reg` and the unit test passes again.

PR #1090, opened 2020-08-06
Refactoring and simplifying slack notifications.

weerd:
### What's this PR do?

This pull request refactors and simplifies how we implement the Slack notifications in Rogue.

Not actually sure why we originally took the approach of using an Event and Listener for this instead of just the Notification class and `Notifiable` trait on the `Post` model as indicated in the [docs](https://laravel.com/docs/6.x/notifications#slack-notifications). It may have been that we _used_ to have additional events fire or something, but not really sure. Does not seem to be the case anymore and seemed like it could be simplified!

I noticed this while doing the updates to upgrade Rogue to Laravel 6.0 but refrained from making the changes during those upgrades to reduce overhead of changes.

The naming of the notification also seemed odd to me since the file can potentially notify a number of channels, not just slack, so I renamed the class to `PostTagged` which felt more relevant to the action that causes the notification! 😄 

:eyes:

You can see some of the test messages posted while I was refactoring in the `#notify-slack-webhook-tests` channel.

Just some summer cleaning! 🛀🏼    

### Relevant tickets

🌵 

### Checklist


app/Notifications/PostTagged.php
```diff
@@ -71,52 +85,36 @@ public function via($notifiable)
...
-        ]);
+        $data = $this->queryForUserAndReviewer();
 
         $userName = $data['user']['displayName'];
         $adminName = $data['admin']['displayName'];
 
         return (new SlackMessage)
             ->from('Rogue', ':tonguecat:')
-            ->content($adminName . ' just tagged this post as "' . $this->tag->tag_name . '":')
+            ->content($adminName.' just tagged this post as "'.$this->tag->tag_name . '":')
```
aaronschachter, line 102: This looks like a handy addition to https://github.com/DoSomething/communal-docs/tree/master/Contributing#php-code

app/Notifications/PostTagged.php
```diff
@@ -71,52 +85,36 @@ public function via($notifiable)
...
-        if (! config('services.slack.url')) {
-            return;
-        }
-
-        // Get the user & admin reviewer's names for the notification.
-        $data = app(GraphQL::class)->query(SLACK_NOTIFICATION_QUERY, [
-            'userId' => $this->post->northstar_id,
-            'adminId' => $this->adminId,
-        ]);
+        $data = $this->queryForUserAndReviewer();
```
aaronschachter, line 95: No biggie, but why not just make the query here vs create another function?

app/Notifications/PostTagged.php
weerd: Oh ya. It should also be part of style checking.

app/Notifications/PostTagged.php
weerd: A ya, good question!

So for one, I just thought it reads a little better and looks a little clearer/cleaner. When reviewing the `toSlack()` method you don't really need to worry about seeing the call to `GraphQL` class and `query()` with the constant passed to it etc. All you need to grok is that you're getting `$data` and that data is from a "queryForUserAndReviewer". If you care to know what's involved in that you can check the `queryForUserAndReviewer()`.

But that's more an opinionated reason. For a more practical reason, this `Notification` class is not just a Slack notification class (which is why I changed the name), but meant to potentially notify a number of channels after a post is tagged. 

We happen to only use the `toSlack()` method, but if we did decide we also want to maybe email the marketing team the posts tagged as "good", or also log to database, we could add those channels in this class as well. They would each have their own `to*()` methods, and they'd have access to calling `$this->queryForUserAndReviewer()` to get that same data, instead of having to repeat it in each method.

PR #1082, opened 2020-07-30
Adds location field to Group and ActionStat resources

aaronschachter:
### What's this PR do?

This pull request adds a migration to add `location` column to the `action_stats` and `groups` tables, and populates the `location` field with the `state` values as  ISO-3166-2 codes, to keep consistent with the `location` field values found on the `posts` table (refs https://github.com/DoSomething/rogue/pull/1079#issuecomment-666484121).

It also modifies the `updateOrCreateActionStats` method introduced in #1079 to write a post group's location to the related action stat, and adds support to filter to the `action_stats` table by `location`. This will be used to display leaderboards by US state.

```
// http://rogue.test/api/v3/action-stats?filter[location]=US-TX

{
  "data": [
    {
      "id": 72,
      "action_id": 11,
      "school_id": "4809500",
      "location": "US-TX",
      "impact": 7,
      "created_at": "2020-07-30T00:07:46+00:00",
      "updated_at": "2020-07-30T21:10:41+00:00"
    }
  ],
  "meta": {
    "pagination": {
      "total": 1,
  ...
```

### How should this be reviewed?

👀 

### Any background context you want to provide?

Once these fields are added to GraphQL and the relevant components in Phoenix and Rogue query for `location` instead of `state`, we can drop the group `state` field to clean this up a bit.

### Relevant tickets

References [Pivotal #174021616](https://www.pivotaltracker.com/story/show/174021616).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [x] Added appropriate feature/unit tests.

PR #1091, opened 2020-08-06
Filter Action Stats, Groups by location

aaronschachter:
### What's this PR do?

This pull request adds a filter to the `ActionStatsTable` component to allow filtering by `location`, and modifies the `GroupsTable` component to filter by `location` (instead of `state` -- which we'll soon drop as a column completely once we make the same corresponding changes in the [Phoenix Group Finder](https://github.com/DoSomething/phoenix-next/blob/5673c030db0525da0ee0639ad037bc7b5a83955e/resources/assets/components/CampaignSignupForm/GroupFinder/GroupFinder.js#L15)).

It also adds some cleanup around migrating from the `state` to `location` field:

* Modifies the NASSP and RIDE imports to populate `location` and `filter_by_location` instead of `state` and `filter_by_state`, should we ever need to run those imports again (I do often for database seeding / local testing)

* Modifies the `groups` table to have a unique index on group type + name + city + location (missed this in #1082)

* Modifies the `Group` factory to populate with a location, so that running the `VoterRegistrationSeeder` (#1086) will generate action stats with valid `location` values and allow filtering by location.

* Uses `GroupType.filterByLocation` instead of `filterByState` (refs https://github.com/DoSomething/graphql/pull/267)

👀 

From the `/actions/:id` page, for action created in #1086:

🍁 

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/database/factories/ModelFactory.php
+++ b/database/factories/ModelFactory.php
@@ -205,7 +205,8 @@ $factory->state(Campaign::class, 'voter-registration', function (Generator $fake
 $factory->define(GroupType::class, function (Generator $faker) {
     return [
         'name' => 'National ' . Str::title($faker->unique()->jobTitle) . ' Society',
-        'filter_by_state' => false,
+        'filter_by_location' => true,
+        'filter_by_state' => true,
     ];
 });
 
@@ -218,6 +219,8 @@ $factory->define(Group::class, function (Generator $faker) {
             return factory(GroupType::class)->create()->id;
         },
         'name' => Str::title($faker->unique()->company),
+        'city' => $faker->city,
+        'location' => 'US-' . $faker->stateAbbr,
     ];
 });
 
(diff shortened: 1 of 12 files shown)


database/factories/ModelFactory.php
```diff
@@ -205,7 +205,8 @@
 $factory->define(GroupType::class, function (Generator $faker) {
     return [
         'name' => 'National ' . Str::title($faker->unique()->jobTitle) . ' Society',
-        'filter_by_state' => false,
+        'filter_by_location' => true,
+        'filter_by_state' => true,
```
Vmack, line 209: why is the `filter_by_state` needed here?

aaronschachter: This is the change that corresponds to the 3rd misc item in the PR description:
> Modifies the Group factory to populate with a location, so that running the VoterRegistrationSeeder (#1086) will generate action stats with valid location values and allow filtering by location.

PR #1092, opened 2020-08-07
Adding PHP Code Style Fixer tool 🎉

weerd:
### What's this PR do?

This pull request adds the [PHP CS Fixer Tool](https://cs.symfony.com/) to Rogue. This tool fixes the code in the repository based on specified rules. Similar to what StyleCI does, but it's all controlled within a file in the repo and no need to deal with updates via the StyleCI website to adjust settings.

To make things even easier, I've included the use of a package I created for my personal use [weerd/php-style](https://github.com/weerd/php-style) which allows sharing the styles in different packages and supports rules for `default` PHP fixes and for Laravel projects. It can also be extended to add other rules if desired.

This is all controlled from a new `.php_cs.dist` file which tells the fixer tool which files/directories to go ahead and fix.

You can use the fixer tool using composer by running the following command in the project:

```
$ composer format
```

These specific commits, since there's a commit for having run the tool and it fixed a ton of files (partly due to a change in the Laravel rules where it now sorts imports alphabetically instead of by line-length).

- 🛠   [eeca6ef: Adding the tool and setting it up](https://github.com/DoSomething/rogue/commit/eeca6efb6b29df01e13e816b7d051e872b8cd2f7)
- ⭕️   [1022094: Adding command to check style in CircleCI](https://github.com/DoSomething/rogue/commit/102209413f21c3e44e3a8d45a3a7a17fd5732828)
- 🗑   [bb517e4: Removing StyleCI config since no longer needed](https://github.com/DoSomething/rogue/commit/bb517e40cbca4d06a688d8200a1c1f09756f4d6d)

This approach will remove the headache of needing to interpret what StyleCI changes it is recommending when it finds an error and let the tool format the code for us when it's run, similar to how we handle our JS code. Also, by utilizing the package I created, it'll be easier to share and setup the rules across PHP projects.

### Relevant tickets

References [Pivotal #]().

### Checklist


weerd:
StyleCI is going to fail since I removed the config file. Waiting for approval of this PR before I remove Rogue from the StyleCI site as something to process.

weerd:
The Circle CI error is the finicky one from a test that occasionally fails... (due to an integrity constraint violation due to duplicate NS id).

.circleci/config.yml
```diff
@@ -48,6 +48,9 @@ jobs:
             - node_modules
             - ~/.cache/Cypress
       - run: npm run build
+      - run:
+          name: check php code style
+          command: vendor/bin/php-cs-fixer fix --dry-run --no-interaction --verbose
```
weerd, line 53: Not super used to configuring CircleCI so lemme know if this looks ok. Essentially doing a `--dry-run` since if it tried to fix the code, it'd need to also commit the changes, etc. This at least will error out and signal that there's a style error.

weerd:
I'm curious if we can also add a pre-commit hook to run `composer format` 🤔

app/Console/Commands/ImportMfolGroupsCommand.php
```diff
@@ -79,10 +79,10 @@ public function handle()
             } catch (Exception $e) {
                 $numFailed++;
 
-                info('Error importing group with ' .$name . ':' . $e->getMessage());
+                info('Error importing group with '.$name.':'.$e->getMessage());
```
weerd, line 82: @aaronschachter per your [comment](https://github.com/DoSomething/rogue/pull/1090#discussion_r466743200), this PHP Style Fixer tool will fix all of these instances based on the specified rules! 

I'll also add it to the communal docs, cause it's good practice, but wanted to highlight this 😄

.circleci/config.yml
DFurnes: Looks good to me! 👍

DFurnes:
> I'm curious if we can also add a pre-commit hook to run `composer format` 🤔

That'd be the best!! I think the tricky part with that is that we'd have to run the pre-commit hook inside folk's Vagrant boxes (since that's where we know PHP is set up correctly), but I think usually people are committing their changes outside of Vagrant.

DFurnes:
Love it! I've always had a love-hate relationship with StyleCI and it feels like this gets rid of a lot of the pain points!

I'm curious if you considered Prettier's [PHP Plugin](https://github.com/prettier/plugin-php), and how it compares to PHP CS Fixer? It would be nice to standardize our code formatting with a single tool & workflow, but I'm not sure if it handles as many style cases.

weerd:
> Love it! I've always had a love-hate relationship with StyleCI and it feels like this gets rid of a lot of the pain points!

Ya, same same! I found PHP CS Fixer to be pretty great for linting and formatting and the fact that it can be made shareable with minimal configuration sold me. 

> I'm curious if you considered Prettier's PHP Plugin, and how it compares to PHP CS Fixer? It would be nice to standardize our code formatting with a single tool & workflow, but I'm not sure if it handles as many style cases.

I did a little bit, but not extensively. PHP CS Fixer seems to have become pretty popular amongst the PHP community for code formatting. While I could see having a single tool format both being useful, I'm a bit concerned about Prettier's minimal customizability with use on a framework like Laravel (where the rules might change, like they did with the import rules).

However, I could see an argument for eventually using both. While PHP CS Fixer does prettify code to an extent it's probably not as thorough as Prettier would be, the greater benefit of PHP CS Fixer is its linting ability along with formatting. I could see it being more like the combo we use with ESLint + Prettier for our JavaScript codea. If we do want to add Prettier I think we still could and they'd be both at play to lint, format AND prettify our PHP code.

weerd:
> That'd be the best!! I think the tricky part with that is that we'd have to run the pre-commit hook inside folk's Vagrant boxes (since that's where we know PHP is set up correctly), but I think usually people are committing their changes outside of Vagrant.

🤔 Hmmmm, that is a good point. I can look into this a bit later. An alternative could be to setup the CI to fix and then commit the fixes to the PR branch.

weerd:
I did find [this article](https://madewithlove.com/using-prettier-in-php/) which could be helpful and provides some good information on using Prettier in combo with PHP CS Fixer (and order of operations), if it's something we want to add on at a later date.

DFurnes:
> I did a little bit, but not extensively. PHP CS Fixer seems to have become pretty popular amongst the PHP community for code formatting. While I could see having a single tool format both being useful, I'm a bit concerned about Prettier's minimal customizability with use on a framework like Laravel (where the rules might change, like they did with the import rules).
> 
> However, I could see an argument for eventually using both. While PHP CS Fixer does prettify code to an extent it's probably not as thorough as Prettier would be, the greater benefit of PHP CS Fixer is its linting ability along with formatting. I could see it being more like the combo we use with ESLint + Prettier for our JavaScript codea. If we do want to add Prettier I think we still could and they'd be both at play to lint, format AND prettify our PHP code.

I'm personally less worried about matching the framework's style than just avoiding too much bike-shedding on creating our own "house style" rules. Less customizability is almost a feature!

I'm a little curious what rules PHP CS Fixer has that wouldn't be covered by Prettier's formatting, and how likely we are to run into them in practice. (To be clear: this is _not_ an argument against PHP CS Fixer. It's a genuine question!)

DFurnes:
Another thing I remember from trying PHP CS Fixer waaaaay back was that it was quite slow. That's not a deal-breaker, but might make it less appealing to run as a git commit hook. How is performance like nowadays?

weerd:
> I'm personally less worried about matching the framework's style than just avoiding too much bike-shedding on creating our own "house style" rules. Less customizability is almost a feature!
>
> I'm a little curious what rules PHP CS Fixer has that wouldn't be covered by Prettier's formatting, and how likely we are to run into them in practice. (To be clear: this is not an argument against PHP CS Fixer. It's a genuine question!)

Ah gotcha. Defs agree on the bike-shedding and that's accomplished a bit with the Laravel specific ruleset. I do find sticking to the framework's style was helpful during the upgrade process and comparing against the framework changes for more of the subtle changes.

I can look into this and see how they compare/overlap, etc.

weerd:
> Another thing I remember from trying PHP CS Fixer waaaaay back was that it was quite slow. That's not a deal-breaker, but might make it less appealing to run as a git commit hook. How is performance like nowadays?

It's pretty fast now from what I recall. It formatted those ~250 files in a second or two.

weerd:
@DFurnes made a good suggestion over Slack that this would be a solid candidate for an RFC, and I agree! 

Will write one up and link this PR over on it before deciding on whether to merge or not!

weerd:
Closing out. This was meant to be a prototype and aiming to still make these changes (based on modifications and learnings after being applied to Aurora), but will be crazy to deal w/ merge conflicts since this was first introduced 😬

PR #1095, opened 2020-08-11
Adds ImportGroupsCommand 

aaronschachter:
### What's this PR do?

This pull request refactors the last groups import console command added, `ImportRhodeIslandGroupsCommand`, as a general groups import command, `ImportGroupsCommand`.  You can see commit by commit I started by copying the Rhode Island groups command, and realized it's essentially the same script, just with a few different variables.

* 👀 

I'll eventually deprecate the `ImportMfolGroupsCommand` by adding a `mfol` config entry, held off on it for sake of getting this working (and hopefully keeping the PR reviewable).

### Relevant tickets

References [Pivotal #174264552](https://www.pivotaltracker.com/story/show/174264552).

### Checklist


tests/Console/ImportGroupsCommandTest.php
```diff
@@ -5,18 +5,36 @@
...
+            'name' => 'Automated Test Group Type',
+            'path' => 'tests/Console/example-groups.csv',
+        ]]);
+    }
+
+    public function tearDown(): void
+    {
+        parent::tearDown();
+
+        config(['import.group_types.test' => null]);
```
aaronschachter, line 25: What's bizarre is that I do get a `tearDown` working with `Config::set` over in https://github.com/DoSomething/chompy/blob/b6f9003f5e4f04f39a9740a30ab6235c069bd959/tests/Unit/Models/RockTheVoteReportTest.php#L29-L36

But when I edit this line to
```
 \Config::set('import.group_types.test', false);
``` 

It results in this error:
```

1) Tests\Console\ImportGroupsCommandTest::testImportingGroups
Error: Call to undefined method Illuminate\Support\Facades\Config::set()
```

config/import.php
```diff
@@ -0,0 +1,25 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | This file is a custom addition for setting constants used to import data.
+    |
+    */
+    'group_types' => [
+        'college_board' => [
+            'filter_by_location' => true,
+            'name' => 'The College Board',
+            'path' => 'https://gist.githubusercontent.com/aaronschachter/2933b5d29bbee4a866de2de49a1f35f7/raw/452d267228d2478de6e845982291c7510cab08cf/groups.csv',
```
aaronschachter, line 17: One really strange thing happening when I copy/paste the raw CSV into a Gist is that it seems to be [adding extra rows](https://dosomething.slack.com/archives/CTVPG6L4R/p1597244201343900?thread_ts=1597086544.318300&cid=CTVPG6L4R). It might not be the most reliable place to store a file with 15k rows.

Is there an easy process for getting a raw file over to amazon AWS? There are pros and cons of hardcoding these URLs here, ideally the file path shouldn't change -- this have always been one-time imports, but I run them regularly for testing/dev.

DFurnes: Woah, spooky! I've had good luck providing files over standard input (for example in [this command](https://github.com/DoSomething/northstar/blob/master/app/Console/Commands/DeleteUsersCommand.php)). Then you can just "pipe" the file into the command from your local machine & don't have to worry about hosting it somewhere on the internet:

```sh
cat groups.csv | heroku run php artisan rogue:groups-import
```

DFurnes: For larger files, we could also definitely drop it in this app's S3 bucket but that's a bit more of a hassle.

tests/Console/ImportGroupsCommandTest.php
DFurnes: I think the problem you're running into here is that the application has already been "torn down" and so all the services that the application relies on (like `Config`) have been destroyed. You could sidestep this error by moving this line up above the `parent::tearDown()` call (before that method destroys the config service), and it'd no longer error out but we can actually do one better...

Because the config service (and the whole rest of the application's state) is destroyed and re-built in between every single test, you actually don't even need to reset this at all! Each test method will, by default, get a brand new version of the app (based on the normal config, test environment variables, and anything you do in `setUp`).

You should be safe to remove `tearDown` here entirely! 🔥

config/import.php
aaronschachter: Thanks for the help! This did the trick.

PR #1096, opened 2020-08-13
Fixes exception if campaign doens't have a website.

DFurnes:
### What's this PR do?

This pull request fixes an exception thrown if a campaign doesn't have a website configured:

```
production.ERROR: Trying to access array offset on value of type null {"exception":"[object] (ErrorException(code: 0): Trying to access array offset on value of type null at /app/app/Models/Post.php:353)
```

This is happening because the campaign these posts are created for doesn't have a website, and so it's trying to read a property `'title'` and `'slug'` off of a null response from `campaignWebsiteByCampaignId`.

👀

This is due to the [new way PHP 7.4 treats array-access on null](https://www.php.net/manual/en/migration74.incompatible.php#migration74.incompatible.core.non-array-access).

### Relevant tickets

References [Pivotal #173629223](https://www.pivotaltracker.com/story/show/173629223).

### Checklist

### DIFF
--- a/app/Models/Post.php
+++ b/app/Models/Post.php
@@ -3,6 +3,7 @@
 namespace Rogue\Models;
 
 use Hashids\Hashids;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Rogue\Services\GraphQL;
 use Rogue\Models\Traits\HasCursor;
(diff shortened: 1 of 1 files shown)


DFurnes:
Throwback!! It was true then! 🕰️‼️

PR #1097, opened 2020-08-13
Fix two more places where a query may return null.

DFurnes:
### What's this PR do?

This pull request follows up on #1096 with another two spots where a GraphQL query may return null (and thus trying to access one of the properties from the query response would [now explode](https://www.php.net/manual/en/migration74.incompatible.php#migration74.incompatible.core.non-array-access)).

👀

❗

### Relevant tickets

References [Pivotal #173629223](https://www.pivotaltracker.com/story/show/173629223).

### Checklist

### DIFF
--- a/app/Models/Signup.php
+++ b/app/Models/Signup.php
@@ -256,9 +256,9 @@ class Signup extends Model
         return [
             'id' => $this->id,
             'user_id' => $userId,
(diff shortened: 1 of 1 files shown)


DFurnes:
I think (super trickily) the `isset` check doesn't actually care as long as the variable is defined! We could swap to `! empty()` to catch that case, but this feels a wee bit safer IMO.

PR #1101, opened 2020-08-16
Fixing setup command.

weerd:
### What's this PR do?

This pull request updates the `SetupCommand()` to fix an issue I ran into when running `php artisan rogue:setup` after a complete reset of my Homestead.

Console Error

Seems like the signature of the `runCommand()` has changed and now requires passing a class that implements the `OutputInterface` as the third argument.

I figured the simplest way to resolve this would be to use the `call()` command before hand and precede it with calling the `line()` command to write the message to the console.

👀 

It may be that we'll need to update the other projects that have their own Artisan `setup` command, since they likely also call the `runCommand()` as well.

### Relevant tickets

🌵 

### Checklist

### DIFF
--- a/app/Console/Commands/SetupCommand.php
+++ b/app/Console/Commands/SetupCommand.php
@@ -51,10 +51,13 @@ class SetupCommand extends Command
             $this->setEnvironmentVariable('NORTHSTAR_CLIENT_SECRET', 'Enter the OAuth Client Secret for machine requests');
         });
 
-        $this->runCommand('key:generate', 'Creating application key');
+        $this->line('Creating application key');
+        $this->call('key:generate');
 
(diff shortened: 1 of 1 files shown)


lindsayrmaher:
I updated my homestead and reset up Rogue last week and didn't run into this issue. but i also didn't use the `php artisan rogue:setup` command. But this seems like a reasonable update to make!

weerd:
> I updated my homestead and reset up Rogue last week and didn't run into this issue. but i also didn't use the php artisan rogue:setup command. But this seems like a reasonable update to make!

Ah yeah, it defs requires running the setup command. How did you set up Rogue without it? Just doing the steps it does manually?

lindsayrmaher:
> How did you set up Rogue without it? Just doing the steps it does manually?

Yup yupp!

PR #1103, opened 2020-08-17
General cleanup

weerd:
### What's this PR do?

This pull request wraps up a few cleanup items noticed after the upgrades to Laravel 6.x.

👀 

🌵 

### Relevant tickets

References [Pivotal #174279882](https://www.pivotaltracker.com/story/show/174279882).

### Checklist


composer.json
```diff
@@ -39,9 +39,7 @@
     "laravel/tinker": "^2.0",
     "mockery/mockery": "^1.0",
     "nunomaduro/collision": "^3.0",
-    "phpunit/phpunit": "^8.0",
-    "symfony/css-selector": "3.1.*",
-    "symfony/dom-crawler": "3.1.*"
```
weerd, line 44: These were old and were removed back in Laravel 5.4, but I think we forgot to remove!

app/Http/Kernel.php
```diff
@@ -31,6 +31,7 @@ class Kernel extends HttpKernel
             \Rogue\Http\Middleware\EncryptCookies::class,
             \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
             \Illuminate\Session\Middleware\StartSession::class,
+            // \Illuminate\Session\Middleware\AuthenticateSession::class,
```
DFurnes, line 34: Why is this commented out?

weerd: It's how it is in the [framework code](https://github.com/laravel/laravel/blob/master/app/Http/Kernel.php#L36), as an option to comment back in and apply it. It didn't exist before in code so I added it to match the framework, but assumed it should stay commented out since it didn't exist before.

DFurnes: It looks like this middleware is for invalidating sessions if a user's password changes (which we can't do with this middleware because passwords are stored in a separate service). If it's okay, let's remove this commented line so it doesn't look like it was commented out by mistake.

weerd: Ok, sounds good. I'll make the same edit in Aurora then, since this change went through there but it has the same setup as Rogue to use a separate service. Thanks for the 👀  @DFurnes 😄

PR #1104, opened 2020-08-20
Moves faker out of dev dependencies

aaronschachter:
### What's this PR do?

This pull request adds Faker package out of `require-dev` and into the main `require` within our `composer.json`, so that we're able to run the `DatabaseSeeder` on our dev Rogue DB. This makes will make it possible for staff to fully test an `ActionStatsBlock` in Phoenix, because the current production DB doesn't have enough action stats to paginate through. 

👀 

We could revert this PR once the dev DB is seeded with action stats, if we'd prefer to keep production un-Faker-able.

### Relevant tickets

### Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
     "doctrine/dbal": "~2.5.13",
     "dosomething/gateway": "^3.0",
     "fideloper/proxy": "^4.0",
+    "fzaninotto/faker": "^1.9.1",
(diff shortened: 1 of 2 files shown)


aaronschachter:
I've been burned before by the ol' DB swap in an `env`,  it'd be safer to just revert this PR once I run the seeder on dev.

PR #1114, opened 2020-08-25
Add name filter to Clubs index

mendelB:
### What's this PR do?

This pull request allows us to filter Clubs by the `name` field via the `/clubs` index API endpoint. 

👀 

This follows https://github.com/DoSomething/rogue/pull/1048 by example since we're imitating this filter. 

I did deviate though in using the [SQL `'LIKE'` operator](https://www.w3schools.com/sql/sql_like.asp). The reason we didn't use `LIKE` for Groups was because of case-sensitivity (see https://github.com/DoSomething/rogue/pull/1048#discussion_r443850722), but I discovered that [we use](https://github.com/DoSomething/rogue/blob/45dbce64312e55f09b87e5395d34860d1291c687/config/database.php#L54) `utf8_unicode_ci` collation for our SQL database in Rogue which means we're automatically getting case insensitive querying! ([TIL](https://www.webucator.com/how-to/how-check-case-sensitivity-sql-server.cfm) that the `ci` stands for case insensitive) (I tested some queries against Rogue QA to confirm, and it seems to indeed be the case! (....case 😩)).

### Relevant tickets

References [Pivotal #174300560](https://www.pivotaltracker.com/story/show/174300560).

### Checklist

### DIFF
--- a/app/Http/Controllers/ClubsController.php
+++ b/app/Http/Controllers/ClubsController.php
@@ -33,6 +33,10 @@ class ClubsController extends ApiController
         $filters = $request->query('filter');
         $query = $this->filter($query, $filters, Club::$indexes);
 
+        if (isset($filters['name'])) {
+            $query->where('name', 'LIKE', '%' . $filters['name'] . '%');
+        }
+
         return $this->paginatedCollection($query, $request);
     }
 
--- a/docs/endpoints/clubs.md
+++ b/docs/endpoints/clubs.md
@@ -12,6 +12,10 @@ GET /api/v3/clubs
 
   - Filter results by Club ID or ID's.
 
+- **filter[name]** _(string)_
+
+  - Filter results by names that include the given filter, e.g. `filter[name]=Oakland`.
+
 Example Response:
 
 ```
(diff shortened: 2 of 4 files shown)


aaronschachter:
Whoa, thanks for the details! I'm going to make this change in the groups search too.. to hopefully fix up the [GraphQL error](https://www.pivotaltracker.com/story/show/174472802) we see now and then on Groups search (you flagged it might be the case in [Slack](https://dosomething.slack.com/archives/CUQMTP0RM/p1597859661012700?thread_ts=1597859420.012600&cid=CUQMTP0RM) the other day)

mendelB:
> Whoa, thanks for the details! I'm going to make this change in the groups search too.. to hopefully fix up the [GraphQL error](https://www.pivotaltracker.com/story/show/174472802) we see now and then on Groups search (you flagged it might be the case in [Slack](https://dosomething.slack.com/archives/CUQMTP0RM/p1597859661012700?thread_ts=1597859420.012600&cid=CUQMTP0RM) the other day)

Ah yes sorry @aaronschachter I meant to flag this for you explicitly!

PR #1115, opened 2020-08-26
Fixes Groups search by name

aaronschachter:
### What's this PR do?

This pull request fixes the syntax errors that are thrown when a special character is included in the `name` filter of a Groups index query, with a big thanks to @mendelB for the insights in #1114  

👀 

🌵 

### Relevant tickets

References [Pivotal #174472802](https://www.pivotaltracker.com/story/show/174472802).

### Checklist

### DIFF
--- a/tests/Http/GroupTest.php
+++ b/tests/Http/GroupTest.php
@@ -49,6 +49,13 @@ class GroupTest extends TestCase
 
         $response->assertStatus(200);
         $this->assertEquals(3, $decodedResponse['meta']['pagination']['count']);
+
+        // Test special characters:
+        $response = $this->getJson('api/v3/groups?filter[name]=g\\');
+        $decodedResponse = $response->decodeResponseJson();
+
+        $response->assertStatus(200);
+        $this->assertEquals(0, $decodedResponse['meta']['pagination']['count']);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/Http/GroupTest.php
```diff
@@ -49,6 +49,13 @@ public function testGroupTypeIndex()
 
         $response->assertStatus(200);
         $this->assertEquals(3, $decodedResponse['meta']['pagination']['count']);
+
+        // Test special characters:
+        $response = $this->getJson('api/v3/groups?filter[name]=g\\');
```
mendelB, line 54: BTW I was curious about [your note](https://dosomething.slack.com/archives/CUQMTP0RM/p1597865498014300?thread_ts=1597859420.012600&cid=CUQMTP0RM) re not being able to replicate this error and noticed that while the `g\\` doesn't explicitly trigger the error via the API, the URL encoded `g%5C%5C` does! (I assume GraphQL was sending over the user input URL encoded).

e.g https://activity-dev.dosomething.org/api/v3/groups?filter[name]=g%5C vs https://activity-dev.dosomething.org/api/v3/groups?filter[name]=g\\

DFurnes: Ah I was wondering how `%`s in the input query string would work with this. Do they need to be escaped?

mendelB: Do you mean like would encoded characters match their plain values in the DB? like a group called `Group /` match ?`filter[name]=%5C`

aaronschachter: I'm not sure what @DFurnes is asking either... the test will pass with `g%5C%5C` but fails with `g%5C`

PR #1120, opened 2020-09-16
Add Club ID field to Posts & Signups

mendelB:
### What's this PR do?

This pull request adds a new `club_id` field to the `Post` and `Signup` resources.

- https://github.com/DoSomething/rogue/commit/f20e4d3ea7163d21394e2e57aebee7be8ac52258 migrations for these fields with an [`index`](https://laravel.com/docs/6.x/migrations#creating-indexes), and a [foreign key constraint](https://laravel.com/docs/6.x/migrations#foreign-key-constraints) to ensure they're valid Club ID's.
- https://github.com/DoSomething/rogue/commit/56ee771b03f2432fd05d082f7d1849f42bad7987` Allow this field to be [fillable](https://github.com/DoSomething/rogue/commit/56ee771b03f2432fd05d082f7d1849f42bad7987) and [filtere-able](https://github.com/DoSomething/rogue/blob/d0ba62091d319e03670ea62cf5e9f0e3b3552544/app/Models/Signup.php#L44)
- https://github.com/DoSomething/rogue/commit/f8604f7faad81f7a1a8f2171e5555f1756aee25e Return these fields in API responses via the `Transformer`s for authorized users.
- https://github.com/DoSomething/rogue/commit/485038cdb9ada917edec3db85bf7492886bb23e7, https://github.com/DoSomething/rogue/commit/35e6c73f7f6cf4d3351703ab7b42b1915528277f minor documentation and test updates

### How should this be reviewed? 
👀 

### Any background context you want to provide?
This sets us up to start appending the `club_id` to signups and posts if we see that the user is associated with a club via Northstar (see #1119).

Notabley, we are not adding the ability to create/update these fields via the API since that's out of scope for this feature.

### Relevant tickets

References [Pivotal #174338924](https://www.pivotaltracker.com/story/show/174338924).

### Checklist

- [x] This PR has been added to the relevant Pivotal card.
- [x] Documentation added for new features/changed endpoints.
- [x] Added appropriate feature/unit tests.

PR #1121, opened 2020-09-18
Add User's Club Id to Signups and Posts

mendelB:
### What's this PR do?

This pull request appends the user's `club_id` to their signups and posts. (Branching off of #1120).

commit-by-commit.

1. Is this implementation via a new `SignupObserver` sound?
2. Do you agree with the execution of the GraphQL query (colocated in the SignupObserver) and its mocks?
3. Is it safe to assume that we can simply grab the `club_id` from the post's signup instead of making the GraphQL query ourselves (e.g. what if the signup has no club_id erroneously, we then wouldn't attach a user's `club_id` to the signups posts! But I figured that edge case didn't warrant a separate GraphQL query).
4. Any concerns that making a GraphQL query while saving the signup will disrupt our traffic/bandwidth/devopswords?

So in our initial implementation of the GraphQL client, we discussed (https://github.com/DoSomething/rogue/pull/889#discussion_r297773531) colocating the queries in their respective class implementations. I was interested to see if this would work here, particularly because I wasn't positive how to go about resolving both the existing [`GraphQL::getUserById`](https://github.com/DoSomething/rogue/blob/c84cc8a7fde97c42f94044655552d54c80469813/app/Services/GraphQL.php#L95-L115) method and this new user query. 

I saw that we'd already done this in the [`PostTagged` class](https://github.com/DoSomething/rogue/blob/c84cc8a7fde97c42f94044655552d54c80469813/app/Notifications/PostTagged.php#L65-L74) so I followed by example. It seems like a nice bet! The only issue being the test mock which is more convenient for `GraphQL` methods. But I bumped into [Partial Mocking](http://docs.mockery.io/en/latest/reference/partial_mocks.html) with mockery which seemed to do the trick!

But if you disagree with this implementation LMK. One outstanding question is duplicate queries, which I don't think we have a solid answer for, but for now, I thought this seemed pretty neat!

### Relevant tickets

References [Pivotal #174338924](https://www.pivotaltracker.com/story/show/174338924).

### Checklist

### DIFF
--- a/app/Observers/SignupObserver.php
+++ b/app/Observers/SignupObserver.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Rogue\Observers;
+
+use Rogue\Models\Signup;
+
+class SignupObserver
+{
+
+    /**
+     * Handle the Signup "creating" event.
+     *
+     * @param  \Rogue\Models\Signup  $signup
+     * @return void
+     */
+    public function creating(Signup $signup)
+    {
+    }
+}
--- a/app/Observers/SignupObserver.php
+++ b/app/Observers/SignupObserver.php
@@ -36,7 +36,7 @@ class SignupObserver
      */
     public function creating(Signup $signup)
     {
-        if (!$signup->club_id) {
+        if (config('features.track_club_id') && !$signup->club_id) {
             $data = $this->queryForUser($signup->northstar_id);
 
             if ($club_id = data_get($data, 'user.clubId')) {
--- a/app/Observers/SignupObserver.php
+++ b/app/Observers/SignupObserver.php
@@ -36,5 +36,12 @@ class SignupObserver
      */
     public function creating(Signup $signup)
     {
+        if (!$signup->club_id) {
+            $data = $this->queryForUser($signup->northstar_id);
+
+            if ($club_id = data_get($data, 'user.clubId')) {
+                $signup->club_id = $club_id;
+            }
+        }
     }
 }
--- a/app/Observers/SignupObserver.php
+++ b/app/Observers/SignupObserver.php
@@ -3,9 +3,30 @@
 namespace Rogue\Observers;
 
 use Rogue\Models\Signup;
+use Rogue\Services\GraphQL;
+
+const USER_CLUB_ID_QUERY = '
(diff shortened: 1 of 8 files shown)


app/Observers/SignupObserver.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Handle the Signup "creating" event.
+     *
+     * @param  \Rogue\Models\Signup  $signup
+     * @return void
+     */
+    public function creating(Signup $signup)
+    {
+        if (!$signup->club_id) {
+            $data = $this->queryForUser($signup->northstar_id);
```
DFurnes, line 40: I am a _little_ worried about the devopswords here, since it seems like we'd now effectively be hitting Northstar for any signup that's created (so a spike in traffic to Rogue means a spike in traffic to Northstar). This absolutely feels like the cleanest way to implement this though, so I think it's good to keep & see if it causes us trouble later.

I think this is another good example of where the user/activity domains are somewhat awkward to split... it makes me continue to think about the idea of merging these two applications. 🤔

DFurnes (approved):
Looks good to me! 🎸

app/Observers/SignupObserver.php
mendelB: lol ok I'm going to put devopsword on a t-shirt!

Sounds good -- I'll try and keep a close eye on this particularly around the next broadcast. Do you think it's worth wrapping this in a feature flag to be safe so we can flip it off if chaos starts to abound?

> I think this is another good example of where the user/activity domains are somewhat awkward to split... it makes me continue to think about the idea of merging these two applications. 🤔

👀 👀

DFurnes: Ohhh, yeah feature flagging this would be dajirt if it's not too much work!

mendelB: Wrapped it all up in a (poorly named? but eh) feature flag in  https://github.com/DoSomething/rogue/pull/1121/commits/9e706a9ec227ed11d39f6124c0886bb19a49236e.

DFurnes: 🙌

PR #1123, opened 2020-09-30
Adds group_type_id filter to action-stats index, and general exclude_by_field filter

aaronschachter:
### What's this PR do?

This pull request adds support to filter a list of action stats by a given `group_type_id`, by joining on the `groups` table by `school_id`.

It also adds support for an `exclude_by_field` filter to controllers that use our `FiltersRequests` trait.

...

...

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/app/Http/Controllers/Traits/FiltersRequests.php
+++ b/app/Http/Controllers/Traits/FiltersRequests.php
@@ -88,6 +88,10 @@ trait FiltersRequests
      */
     public function orderBy($query, $orderBy, $indexes)
     {
+        // We specify the table name of our 'id' in case the query is joining on tables.
+        $tableName = $query->getModel()->getTable();
+        $idField = $tableName . '.id';
+
         if ($orderBy) {
             [$column, $direction] = explode(',', $orderBy, 2);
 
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/Traits/FiltersRequests.php
```diff
@@ -34,6 +34,18 @@ public function filter($query, $filters, $indexes)
         if (array_key_exists('exclude', $filters)) {
             $excludedValues = $filters['exclude'];
             unset($filters['exclude']);
+
+            /**
+             * If there is an exclude_by_field filter, remove from $filters and save the value,
+             * Else default to exclude by an 'id' field.
+             * Example: filter[exclude]=1200537,3618101&filter[exclude_field]=school_id.
```
DFurnes, line 41: If possible, I'd love to swap this to a top-level `?exclude[school_id]=1200537,3618101` API that mirrors we'd use our filter query parameter, like `?filter[school_id]=1200537,3618101`.

The original intent behind "namespacing" filters under that `?filter[…]` query parameter is to avoid potential conflicts if a field & an API action had the same name (e.g. imagine if we had a field named `exclude` on actions to determine whether they're displayed to users or not, we'd _want_ to say `filter[exclude]=true` or `filter[exclude]=false` to filter on that `exclude` field's value).

Ideally, every time you say `filter[x]=y` it works the same for any `x` and `y`!

DFurnes: And to maintain backwards compatibility, we could maintain `?filter[exclude]=1` as an alias for `?exclude[id]=1` and then just use `?exclude[field]=value` for newer use-cases that require us to specify excludes by fields other than `id`.

aaronschachter: Makes sense... I'm tempted to back out of adding this `exclude_by_field` into this PR, and taking on the work in a separate PR. This was sort of a red herring that it felt small enough to leave in. 

FWIW - Searching our repos and from what I can tell we only used this in [Ashes](https://github.com/DoSomething/phoenix/blob/11cb8b73af0c1edbe7cb697ff82c835dbb594436/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc#L165).. I can't seem to find any instances in anything we currently use?

DFurnes: Oh you're probably 100% right! I think we did add this feature to power that old behavior in the gallery.

That'd be killer if we could remove this altogether & replace with the `exclude[field]=` option. 🔥

aaronschachter: I did remove all of my additions for the filter -- can add the `exclude[field]` option in a cleanup PR (we don't need it for https://www.pivotaltracker.com/story/show/175047229)

DFurnes (approved):
⚗️

PR #1124, opened 2020-10-01
Update CustomerIo client & dependencies.

DFurnes:
### What's this PR do?

This pull request updates Rogue's `CustomerIo` client to match Northstar's. We'll then use this client's `trackEvent` method to replace places that we rely on Blinks `user-signup`, `user-signup-post`, and `user-signup-post-review` queues (which all just route to this Customer.io endpoint behind-the-scenes).

I updated the `CustomerIo` API client to match Northstar's (based on the work done in DoSomething/northstar#1060 and https://github.com/DoSomething/northstar#1063. Two small differences: We pass a `$userId` instead of a user model (since users don't live here), and I've excluded the methods for updating & deleting customer's since we don't need to do that here.

I've also updated this application's PHP & JavaScript dependencies in 78fae46.

This keeps these two applications consistent & makes the next few tasks nice and easy!

### Relevant tickets

References [Pivotal #175077144](https://www.pivotaltracker.com/story/show/175077144).

### Checklist


app/Services/CustomerIo.php
```diff
@@ -24,6 +24,16 @@ public function __construct()
...
     }
 
+    /**
+     * Is Customer.io enabled for this app?
+     *
+     * @return bool
+     */
+    protected function enabled(): bool
+    {
+        return config('features.blink');
```
DFurnes, line 34: I'm planning to switch the name of this config & environment variable in both applications later (but don't want to introduce a change just yet that'd require changing any settings in order to deploy this).

aaronschachter (approved):

PR #1125, opened 2020-10-02
Fixes bug when filtering Action Stats by location

aaronschachter:
### What's this PR do?

This pull request fixes a bug introduced in #1123: when additionally filtering by `location` as well as `group_type_id` (which can be done in the leaderboard by selecting a location), an error is thrown: ` Integrity constraint violation: 1052 Column 'location' in where clause is ambiguous`

#1123 joins the `action_stats` table on the `groups` table -- both tables have a `location` column. This PR manually constructs the filters in a `GET /action-stats` index query to specify the table name of any potentially ambigious columns we may be filtering by.

👀 

What I'm puzzled about is a related separate ambiguous error we're seeing in Slack:
```
production.ERROR: SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous (SQL: select * from `action_stats` inner join `groups` on `action_stats`.`school_id` = `groups`.`school_id` where `action_id` = 954 and `groups`.`group_type_id` = 5 and (`impact` < 6 or (`impact` = 6 and `id` > 11955)) order by `impact` desc, `action_stats`.`id` asc limit 11 offset 0)
```
Off the top of my head - I have no idea which app is making this query. Will dig in a bit, but figured I'd open this PR to fix the `location` integrity constraint violation, and also see if we still see this `id`  integrity constraint violation upon deploy.

### Relevant tickets

References [Pivotal #175096130](https://www.pivotaltracker.com/n/projects/2417735/stories/175096130).

### Checklist

### DIFF
--- a/app/Http/Controllers/ActionStatsController.php
+++ b/app/Http/Controllers/ActionStatsController.php
@@ -30,12 +30,29 @@ class ActionStatsController extends ApiController
     public function index(Request $request)
     {
         $query = $this->newQuery(ActionStat::class);
+        $tableName = $query->getModel()->getTable();
 
         $filters = $request->query('filter');
-        $query = $this->filter($query, $filters, ActionStat::$indexes);
+
+        /**
+         * Because we may be joining on groups, which also have location and school_id columns, we
+         * specify our table name to avoid integrity constraint violations for ambiguous clauses.
+         */
+
+        if (Arr::has($filters, 'action_id')) {
+            $query->where('action_id', $filters['action_id']);
+        }
+
+        if (Arr::has($filters, 'location')) {
+            $query->where($tableName . '.location', $filters['location']);
+        }
 
         if (Arr::has($filters, 'group_type_id')) {
             $query = $query->inGroupTypeId($filters['group_type_id']);
+        }
+
+        if (Arr::has($filters, 'school_id')) {
+            $query->where($tableName . '.school_id', $filters['school_id']);
         }
 
         // Allow ordering results:
--- a/app/Http/Controllers/ActionStatsController.php
+++ b/app/Http/Controllers/ActionStatsController.php
@@ -30,7 +30,6 @@ class ActionStatsController extends ApiController
     public function index(Request $request)
     {
         $query = $this->newQuery(ActionStat::class);
-        $tableName = $query->getModel()->getTable();
 
         $filters = $request->query('filter');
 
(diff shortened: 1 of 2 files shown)


app/Http/Controllers/ActionStatsController.php
```diff
@@ -32,12 +32,29 @@ public function index(Request $request)
...
+         */
+        $tableName = $query->getModel()->getTable();
+
+        if (Arr::has($filters, 'action_id')) {
+            $query->where('action_id', $filters['action_id']);
+        }
+
+        if (Arr::has($filters, 'location')) {
+            $query->where($tableName . '.location', $filters['location']);
+        }
```
DFurnes, line 48: Instead of hard-coding this solution here for each filter, can we add this table name prefix to the `filter` method itself? We'll always want to be filtering on the "primary" model I think (and that'd make sure we don't have to continue to make one-off change as other cases like this occur).

DFurnes (approved):
This works! I'd love to follow-up with a more generalized solution though.

app/Http/Controllers/ActionStatsController.php
DFurnes: Hmm, the `filter` method actually hands off to [this helper](https://github.com/DoSomething/rogue/blob/989450c162efe4cf444c7acd5ccce98a0e091355/app/helpers.php#L60-L64) so I guess the fix would go there (although TBH I think we could just inline this in the two places it's referenced to make things a little easier to follow).

aaronschachter: Ah forgot that [`filter` method is custom.](https://github.com/DoSomething/rogue/blob/989450c162efe4cf444c7acd5ccce98a0e091355/app/Http/Controllers/Traits/FiltersRequests.php#L27). I'll merge and deploy this for now to stop the noise (and also see if this super mysterious  error goes away) and follow up in a new PR.

DFurnes:
> What I'm puzzled about is a related separate ambiguous error we're seeing in Slack:
> 
> ```
> production.ERROR: SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous (SQL: select * from `action_stats` inner join `groups` on `action_stats`.`school_id` = `groups`.`school_id` where `action_id` = 954 and `groups`.`group_type_id` = 5 and (`impact` < 6 or (`impact` = 6 and `id` > 11955)) order by `impact` desc, `action_stats`.`id` asc limit 11 offset 0)
> ```
> 
> Off the top of my head - I have no idea which app is making this query. Will dig in a bit, but figured I'd open this PR to fix the `location` integrity constraint violation, and also see if we still see this `id` integrity constraint violation upon deploy.

This is coming from the [`HasCursor`](https://github.com/DoSomething/rogue/blob/master/app/Models/Traits/HasCursor.php) trait (which is how we handle cursor-based pagination, which is based filtering results using the ID & sort column rather than a numerical offset for better performance & consistency). Like here, we'll want to prefix `id` with the table name from the query.

DFurnes:
It might also be nice to add a simple test that demonstrates this bug so that we don't accidentally regress in the future!

PR #1126, opened 2020-10-02
Send posts directly to Customer.io.

DFurnes:
### What's this PR do?

This pull request updates Rogue to send `campaign_signup_post` events directly to [Customer.io](https://customer.io).

🎯 I updated the `SendPostToCustomerIo` job to talk directly to Customer.io, via the `trackEvent` method in b1147d0.

🚜 I updated the `toCustomerIoPayload` method to include transformations [previously handled by Blink](https://github.com/DoSomething/blink/blob/dfe6d5ef138bd2fb3d66a105d626f4ab6f62d6c5/src/messages/CampaignSignupPostMessage.js#L52-L55) in fb4aa7c.

🚥 I updated tests to mock this method, rather than Blink's `userSignupPost` method in 9d9c491.

This reduces complexity and will allow us to reduce some Heroku costs in the future.

### Relevant tickets

References [Pivotal #174827032](https://www.pivotaltracker.com/story/show/174827032).

### Checklist

### DIFF
--- a/app/Jobs/SendPostToCustomerIo.php
+++ b/app/Jobs/SendPostToCustomerIo.php
@@ -2,17 +2,25 @@
 
 namespace Rogue\Jobs;
 
-use DoSomething\Gateway\Blink;
 use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Bus\Dispatchable;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Facades\Redis;
 use Rogue\Models\Post;
+use Rogue\Services\CustomerIo;
 
 class SendPostToCustomerIo implements ShouldQueue
 {
     use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;
+
+    /**
+     * Delete the job if its models no longer exist.
+     *
+     * @var bool
+     */
+    public $deleteWhenMissingModels = true;
 
     /**
      * The post to send to Customer.io.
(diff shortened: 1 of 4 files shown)


app/Jobs/SendPostToCustomerIo.php
```diff
@@ -2,60 +2,67 @@
...
-    protected $log;
+    protected $post;
 
     /**
      * Create a new job instance.
      *
      * @return void
      */
-    public function __construct(Post $post, $log = true)
+    public function __construct(Post $post)
```
DFurnes, line 37: We'd previously added this `$log` parameter to reduce log volume when running scripts. Now that we're on a shared Papertrail account (rather than per-app add-ons), this is not a significant usage concern.

app/Jobs/SendPostToCustomerIo.php
```diff
@@ -2,60 +2,67 @@
...
     /**
      * Execute the job.
      *
      * @return void
      */
-    public function handle(Blink $blink)
+    public function handle(CustomerIo $customerIo)
     {
-        // Check if the post still exists before sending (might have been deleted immediately if created in Runscope test).
-        if ($this->post && $this->post->signup) {
```
DFurnes, line 50: I was able to replace this conditional with Laravel's new [`$deleteWhenMissingModels`](https://laravel.com/docs/6.x/queues#ignoring-missing-models), above.

app/Managers/PostManager.php
```diff
@@ -61,7 +61,7 @@ public function create($data, $signupId, $authenticatedUserRole = null)
         );
 
         // Save the new post in Customer.io, via Blink.
-        if (config('features.blink') && $shouldSendToCustomerIo) {
+        if ($shouldSendToCustomerIo) {
```
DFurnes, line 64: Since we [check the feature flag within the `CustomerIo` class](https://github.com/DoSomething/rogue/blob/1f7f1f6a9113d078f9304a3e837653bb3eb73795/app/Services/CustomerIo.php#L47-L55), this check is unnecessary. By removing it, we also get these logged on local environments.

lindsayrmaher (approved):
this makes good sense to me!

PR #1128, opened 2020-10-06
Send signups directly to Customer.io & Gambit.

DFurnes:
### What's this PR do?

This pull request updates Rogue to send signup events directly to Customer.io and Gambit, rather than routing this event through Blink. These API calls power email & SMS transactional messages when a user signs up for a campaign.

I've broken this into individual commits for each portion of this task.

This is blocked from merge/deploy until I make some some configuration changes to add Gambit environment variables & feature flag to each environment. I'm preparing that change now and will update this note when it's ready.

This reduces complexity and will help us save on Heroku costs in the future. 💰 

### Relevant tickets

References [Pivotal #174827024](https://www.pivotaltracker.com/story/show/174827024).

### Checklist

### DIFF
--- a/app/Jobs/Middleware/GambitRateLimit.php
+++ b/app/Jobs/Middleware/GambitRateLimit.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Rogue\Jobs\Middleware;
+
+use Illuminate\Support\Facades\Redis;
+
+class GambitRateLimit
+{
+    /**
+     * Process the queued job.
+     *
+     * @param  mixed  $job
+     * @param  callable  $next
+     * @return mixed
+     */
+    public function handle($job, $next)
+    {
+        // Rate limit to 10 requests/second.
+        $throttler = Redis::throttle('gambit')
+            ->allow(10)
+            ->every(1);
+
+        $throttler->then(
+            function () use ($job, $next) {
+                // Lock obtained, run job...
+                $next($job);
+            },
+            function () use ($job) {
+                // If we  can't obtain a lock, release back to queue...
+                $job->release(10);
+            },
+        );
+    }
+}
--- a/app/Managers/SignupManager.php
+++ b/app/Managers/SignupManager.php
@@ -4,6 +4,7 @@ namespace Rogue\Managers;
 
 use Rogue\Jobs\CreateCustomerIoEvent;
 use Rogue\Jobs\SendSignupToCustomerIo;
+use Rogue\Jobs\SendSignupToGambit;
 use Rogue\Repositories\SignupRepository;
 
 class SignupManager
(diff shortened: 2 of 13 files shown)


app/Jobs/Middleware/GambitRateLimit.php
```diff
@@ -0,0 +1,34 @@
...
+     * Process the queued job.
+     *
+     * @param  mixed  $job
+     * @param  callable  $next
+     * @return mixed
+     */
+    public function handle($job, $next)
+    {
+        // Rate limit to 10 requests/second.
+        $throttler = Redis::throttle('gambit')
```
DFurnes, line 19: I've created a second rate limiter (in addition to `CustomerIoRateLimit`) since this is a separate service and therefore shouldn't count towards the same limit as Customer.io. I figured 10 requests/second was a decent max, but happy to adjust to whatever makes the most sense.

cc: @aaronschachter

app/Managers/SignupManager.php
```diff
@@ -44,11 +45,16 @@ public function create($data, $northstarId, $campaignId)
...
 
-        // Save the new signup in Customer.io, via Blink.
-        if (config('features.blink') && $shouldSendToCustomerIo) {
+        // Send the new signup to Customer.io:
+        if ($shouldSendToCustomerIo) {
             SendSignupToCustomerIo::dispatch($signup);
         }
 
+        // If this wasn't triggered via SMS, send to Gambit:
+        if (!preg_match('/(sms|gambit)/', $signup->source)) {
```
DFurnes, line 54: This is to [replicate this logic in Blink](https://github.com/DoSomething/blink/blob/dfe6d5ef138bd2fb3d66a105d626f4ab6f62d6c5/src/workers/GambitCampaignSignupRelayWorker.js#L50-L56). And just in case we ever update this `source` to read the application name as we do for other sources, I figured we may as well include the string `gambit` in this check.

app/Services/Gambit.php
```diff
@@ -0,0 +1,60 @@
...
+                'id' => $signup->id,
+                'payload' => $payload,
+            ]);
+
+            return;
+        }
+
+        $this->client->post('/api/v2/messages?origin=signup', [
+            'json' => $payload,
+        ]);
```
DFurnes, line 56: For convenience, here's the [API documentation](https://git.io/JUAiB) for this endpoint & here's [where we used to handle this in Blink](https://github.com/DoSomething/blink/blob/dfe6d5ef138bd2fb3d66a105d626f4ab6f62d6c5/src/workers/GambitCampaignSignupRelayWorker.js#L28-L47). We rely on Guzzle throwing exceptions for non-successful responses here, which will log an error to Papertrail & put this into the failed jobs table.

aaronschachter (approved):
🔌

PR #1131, opened 2020-10-08
Fix Guzzle import.

DFurnes:
### What's this PR do?

This pull request fixes a typo'd import in #1129.

### How should this be reviewed?

This fixes a [typo I made](https://github.com/DoSomething/rogue/pull/1129#issuecomment-705849122). You can see the actual namespace [here](https://github.com/guzzle/guzzle/blob/c3927bc669178cd051d1bee595387ef80f31acd8/src/Exception/ClientException.php).

### Any background context you want to provide?

I wonder if we can set up a linter rule to check this? I'm surprised it didn't error out!

### Relevant tickets

References [Pivotal #174827024](https://www.pivotaltracker.com/story/show/174827024).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.

PR #1132, opened 2020-10-09
Properly catch the unexpected 422s here.

DFurnes:
### What's this PR do?

This pull request properly catches the 422's being thrown by Gambit here when a user without a mobile number signs up for a campaign. The previous fix in #1131 didn't work because the `$response` variable was not defined yet when Guzzle throws the exception, so instead you have to grab it from the exception itself.

For reference, here's [Guzzle's 'Exceptions' documentation](http://docs.guzzlephp.org/en/stable/quickstart.html#exceptions).

Third times the charm! I've hooked up my local to Gambit Staging & confirmed this fix actually works. 😅

### Relevant tickets

References [Pivotal #174827024](https://www.pivotaltracker.com/story/show/174827024).

### Checklist

### DIFF
--- a/app/Services/Gambit.php
+++ b/app/Services/Gambit.php
@@ -61,8 +61,8 @@ class Gambit
 
             if ($response->getStatusCode() === 422) {
                 // We expect to get 422s from this endpoint for any users who sign up for
-                // a campaign but don't have a mobile on their profile. These should not
-                // be counted as failures!
+                // a campaign but don't have a mobile on their profile or are unsubscribed
+                // from text messages. These should not be counted as failures!
             } else {
                 throw $exception;
             }
--- a/app/Services/Gambit.php
+++ b/app/Services/Gambit.php
@@ -53,13 +53,17 @@ class Gambit
         }
 
         try {
-            $response = $this->client->post('/api/v2/messages?origin=signup', [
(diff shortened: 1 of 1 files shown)


app/Services/Gambit.php
```diff
@@ -53,13 +53,17 @@ public function relaySignup(Signup $signup)
...
             ]);
         } catch (ClientException $exception) {
-            // We expect to get 422s for any users who sign up for a campaign but don't
-            // have a mobile on their profile. These should not count as failures.
-            if ($response->getStatusCode() !== 422) {
+            $response = $exception->getResponse();
+
+            if ($response->getStatusCode() === 422) {
+                // We expect to get 422s from this endpoint for any users who sign up for
+                // a campaign but don't have a mobile on their profile. These should not
```
aaronschachter, line 64: Another scenario is the user does have a mobile - but they are unsubscribed (`sms_status` is equal to `stop` or `undeliverable`)

aaronschachter (approved):
3️⃣  🤞

app/Services/Gambit.php
DFurnes: Oh thanks for noting that! I updated the comment in 145e265.

PR #1133, opened 2020-10-09
Fix issue sending signup-less posts to Customer.io, and other cleanup.

DFurnes:
### What's this PR do?

This pull request fixes an issue sending posts lacking signups (mostly from testing) to Customer.io. I've also included some cleanup along the way, which I'll annotate my thinking for in comments.

It's probably easiest to review this one commit-by-commit!

🍃

### Relevant tickets

References [Pivotal #175209573](https://www.pivotaltracker.com/story/show/175209573).

### Checklist


app/Models/Group.php
```diff
@@ -68,18 +68,19 @@ public function signups()
...
 
     /**
-     * Returns Customer.io payload attributes for given group, if exists.
+     * Transform this group for Customer.io.
      *
-     * @param Group $group
      * @return array
      */
-    public static function toCustomerIoPayload(?self $group)
+    public function toCustomerIoPayload()
```
DFurnes, line 75: This change allows this to be used consistently with other `toCustomerIoPayload` methods, which all exist on the instance. To use this in scenarios where the group may not exist, we call it with [`optional`](https://laravel.com/docs/8.x/helpers#method-optional), like so:

```php
optional($this->group)->toCustomerIoPayload()
```

app/Models/Post.php
```diff
@@ -336,21 +336,12 @@ public function siblings()
...
     /**
-     * Transform the post model for Blink.
+     * Transform this post for Customer.io.
      *
      * @return array
      */
     public function toCustomerIoPayload()
     {
-        // Blink expects quantity to be a number.
-        $quantity = $this->quantity === null ? 0 : $this->quantity;
```
DFurnes, line 346: We no longer need to play by Blink's validation rules, but this is also something we can handle by simply casting the `$this->quantity` to an integer (since `(int) null` becomes `0`)!

app/Models/Post.php
```diff
@@ -336,21 +336,12 @@ public function siblings()
...
-
-        // Bypass Campaign->cause accessor method so the value isn't converted to an array
-        // which Customer.io does not support.
-        $campaign_cause = optional($this->signup->campaign)->getAttributes()[
-            'cause'
-        ];
```
DFurnes, line 352: I decided to replace this with an `implode(',', $campaign->cause)` call below because I think it makes the intention clearer (that we want this as a CSV), rather than just relying on that implicitly being the format stored in the database. It's also, conveniently, a little bit more concise!

app/Models/Post.php
```diff
@@ -361,32 +352,32 @@ public function toCustomerIoPayload()
             $school = app(GraphQL::class)->getSchoolById($this->school_id);
         }
 
-        // The associated Action for this post.
         $action = $this->actionModel;
+        $signup = optional($this->signup);
```
DFurnes, line 356: This fixes the problems we'd run into with this method & posts without signups, since [`optional`](https://laravel.com/docs/8.x/helpers#method-optional) always returns an object that you can work with (and if the signup doesn't exist, properties or method calls will simply return `null`).

tests/Unit/Models/PostModelTest.php
```diff
@@ -153,26 +153,6 @@ public function testBlinkPayload()
...
-
-        $this->assertEquals($result['group_id'], null);
-        $this->assertEquals($result['group_name'], null);
-        $this->assertEquals($result['group_type_id'], null);
-        $this->assertEquals($result['group_type_name'], null);
-        $this->assertEquals($result['school_id'], null);
-        $this->assertEquals($result['school_name'], null);
-        $this->assertEquals($result['referrer_user_id'], null);
-    }
-
```
DFurnes, line 175: Since we no longer explicitly return each field with a null value, we can't `assertEquals` here. I think this test is safe to omit since the behavior & likely failure scenarios (exceptions if we omit a group) would be caught by other tests.

app/Models/Post.php
```diff
@@ -709,10 +701,10 @@ public function getReferralPostEventPayload()
                 'type' => $this->type,
                 'status' => $this->status,
                 'action_id' => $this->action_id,
-                'created_at' => $this->created_at->toIso8601String(),
-                'updated_at' => $this->updated_at->toIso8601String(),
+                'created_at' => $this->created_at->timestamp,
+                'updated_at' => $this->updated_at->timestamp,
```
DFurnes, line 705: This change is made for consistency with other dates we send to Customer.io. Customer.io can only do "date operations" when given a UNIX timestamp, and so these would otherwise just be treated as plain old strings.

app/Models/Group.php
```diff
@@ -68,18 +68,19 @@ public function signups()
...
     {
         return [
-            'group_id' => $group ? $group->id : null,
-            'group_name' => $group ? $group->name : null,
-            'group_type_id' => $group ? $group->group_type->id : null,
-            'group_type_name' => $group ? $group->group_type->name : null,
+            // Note: These fields will be merged in with post & signup events, and so
+            // it's important to prefix them all with 'group_' to avoid conflicts!
```
aaronschachter, line 79: 👍  💬  ⚠️

aaronschachter (approved):
👏 ❤️

PR #1138, opened 2020-10-15
Render Club Leader ID Safely

mendelB:
### What's this PR do?

This pull request fixes the Club index table to render the Club's `leaderId` instead of the nested leader resource, if the resource is unavailable.

👀 

The leaderId is a required field and must be a valid Northstar ID. However 
1. We don't validate this against NS, it really only needs to be a valid Mongodb ID
2. The user can be deleted and we don't remove the leader id from the club

I noticed that the table was breaking on production since presumably we're being met by one of these scenarios.

I think this is edge-casey enough, for now, to simply safely render the leader ID instead. But perhaps we should think about adding clearing the Leader ID from the club if the user is deleted? And/or properly validate the Leader ID against Northstar?

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/resources/assets/components/ClubsTable.js
+++ b/resources/assets/components/ClubsTable.js
@@ -18,6 +18,7 @@ const CLUBS_QUERY = gql`
           name
           city
           location
+          leaderId
           leader {
             id
             firstName
@@ -84,11 +85,15 @@ const ClubsTable = ({ filter }) => {
             </td>
             <td>{node.city ? `${node.city}, ${node.location}` : null}</td>
             <td>
-              <EntityLabel
-                id={node.leader.id}
-                name={node.leader.firstName}
-                path="users"
-              />
+              {node.leader ? (
+                <EntityLabel
+                  id={node.leader.id}
+                  name={node.leader.firstName}
+                  path="users"
+                />
+              ) : (
+                node.leaderId
+              )}
             </td>
             <td>
               {node.schoolId ? (


DFurnes (approved):

DFurnes:
> I think this is edge-casey enough, for now, to simply safely render the leader ID instead. But perhaps we should think about adding clearing the Leader ID from the club if the user is deleted? And/or properly validate the Leader ID against Northstar?

Yeah, this is one of those things that's kinda-awkward with these two apps! I've generally tried to avoid situations where we have to communicate between Identity & Activity so they're not too tightly coupled, but it leaves us with scenarios like this where we sorta just "trust" potentially bad data.

I think it'd be good to keep an eye on & we can weigh solutions based on how much trouble this causes.

mendelB:
> > I think this is edge-casey enough, for now, to simply safely render the leader ID instead. But perhaps we should think about adding clearing the Leader ID from the club if the user is deleted? And/or properly validate the Leader ID against Northstar?
> 
> Yeah, this is one of those things that's kinda-awkward with these two apps! I've generally tried to avoid situations where we have to communicate between Identity & Activity so they're not too tightly coupled, but it leaves us with scenarios like this where we sorta just "trust" potentially bad data.
> 
> I think it'd be good to keep an eye on & we can weigh solutions based on how much trouble this causes.

That makes good sense! (And reminds me of https://github.com/DoSomething/rogue/pull/1121#discussion_r490982952).

I can't offhand think of anyplace where this could bite us if we're safe about rendering nested leaders (not that we even do that anyplace else right now). Will keep an 👁️ out!

PR #1140, opened 2020-10-15
Fix cursor pagination errors for Action Stats index endpoint

aaronschachter:
### What's this PR do?

This pull request fixes the ` Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous ` error that occurs if querying the `GET /action-stats` index endpoint with `filter[group_type_id]` and `cursor[after]` query parameters set, by specifying the table name of our base model in the `HasCursor` trait, per @DFurnes's helpful comment in https://github.com/DoSomething/rogue/pull/1125#issuecomment-702849789.

It also cleans up the `ActionStatsController` per https://github.com/DoSomething/rogue/pull/1125#discussion_r498941457, so any other endpoints that may end up joining on other tables won't suffer from the same ambiguous column errors.

👀 

Question for @DFurnes -- in https://github.com/DoSomething/rogue/pull/1125#issuecomment-702850642 you mention adding a simple test to demonstrate the bug. I can't think of a way to do this, beyond adding a new query parameter that would avoid specifying the table name in `HasCursor` and `multipleValueQuery`... but this seems like it complicates the code even more for the sake of a test. Did you have a different approach in mind?

### Relevant tickets

References [Pivotal #175272211](https://www.pivotaltracker.com/story/show/175272211).

### Checklist

### DIFF
--- a/tests/Http/ActionStatsTest.php
+++ b/tests/Http/ActionStatsTest.php
@@ -21,7 +21,7 @@ class ActionStatsTest extends TestCase
         // Create five action stats.
         factory(ActionStat::class, 5)->create();
 
-        $response = $this->getJson('api/v3/action-stats');
+        $response = $this->getJson($this->url);
         $decodedResponse = $response->decodeResponseJson();
 
         $response->assertStatus(200);
@@ -68,7 +68,7 @@ class ActionStatsTest extends TestCase
         $this->assertEquals(2, $decodedResponse['meta']['pagination']['count']);
 
         $response = $this->getJson(
-            'api/v3/action-stats?filter[group_type_id]=' . $secondGroupTypeId,
+            $this->url . '?filter[group_type_id]=' . $secondGroupTypeId,
         );
         $decodedResponse = $response->decodeResponseJson();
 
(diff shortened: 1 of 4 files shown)


tests/Http/ActionStatsTest.php
```diff
@@ -141,4 +144,37 @@ public function testLocationFilter()
...
+            'school_id' => $actionStats[0]->school_id,
+        ]);
+
+        // Find a cursor to search by.
+        $response = $this->getJson($this->url);
+
+        $decodedResponse = $response->decodeResponseJson();
+        $firstCursor = $decodedResponse['data'][0]['cursor'];
+
+        // Verify no errors are thrown when additionally filtering by group_type_id.
```
DFurnes, line 168: This is exactly what I was thinking of for a test! 💯

DFurnes (approved):
Looks great!

tests/Http/ActionStatsTest.php
DFurnes: Re: your question in the PR, I think if you had swapped around the order of the commits this test would have failed before the changes introduced in 264c776 right? If so, this test sounds like exactly what we need here to make sure we don't "undo" the fix you've applied here in the future!

PR #1142, opened 2020-10-22
Handle undeliverable 'bad requests'.

DFurnes:
### What's this PR do?

This pull request handles `400 Bad Request` errors from Gambit, which we've been seeing pile up in our "failed jobs" table and continually get retried every morning. Since these are all due to "not currently reachable" errors for users who can't receive SMS messages, there's no reason to mark these as failed jobs.

### How should this be reviewed?

👀

### Any background context you want to provide?

See [recent Rogue errors](https://my.papertrailapp.com/groups/10447062/events?focus=1255803625465982976&q=%28sender%3Adosomething-rogue%20%28production.ERROR%20OR%20status%3D5%29%29%20%20OR%20%28sender%3Afastly%20status%3D5%20app%3Ddosomething-rogue%29) and [this Slack thread](https://dosomething.slack.com/archives/CUQMTP0RM/p1602598318001900).

### Relevant tickets

References [Pivotal #175287738](https://www.pivotaltracker.com/story/show/175287738).

### Checklist

- [ ] This PR has been added to the relevant Pivotal card.
- [ ] Documentation added for new features/changed endpoints.
- [ ] Added appropriate feature/unit tests.

PR #1143, opened 2020-10-22
A little more Blink/Customer.io cleanup.

DFurnes:
### What's this PR do?

This pull request commits a bit more cleanup I tackled while in this application for #1142:

1️⃣ Simplifies the Signup's quantity transformer to match [how we handle this in the Post](https://github.com/DoSomething/rogue/blob/049f197592f0f9055d1cc7cacbca63da7765fc3b/app/Models/Post.php#L364).  7a2fc355f418d49caf480389fae7c35655b86878

🌵 Simplifies how the Signup formats causes, again to match [how we handle this in the Post](https://github.com/DoSomething/rogue/blob/049f197592f0f9055d1cc7cacbca63da7765fc3b/app/Models/Post.php#L369). 1b9c7f0

✏️ Replaces outdated references to Blink in comments. a18a33c

🎭 Removes old Blink mocks from our test suite. 84da2bd

🎏 Uses the new `DS_ENABLE_CUSTOMER_IO` flag configured in DoSomething/infrastructure#279. 79f62c4

I'd recommend reviewing these commit-by-commit!

🍃

### Relevant tickets

N/A

### Checklist


app/Managers/PostManager.php
```diff
@@ -54,13 +54,12 @@ public function create($data, $signupId, $authenticatedUserRole = null)
             $authenticatedUserRole,
         );
 
-        // Send to Blink unless 'dont_send_to_blink' is TRUE
+        // Send to Customer.io unless 'dont_send_to_blink' is TRUE
         $shouldSendToCustomerIo = !(
             array_key_exists('dont_send_to_blink', $data) &&
             $data['dont_send_to_blink']
```
aaronschachter, line 60: If I remember correctly... the only instances we'd send along this parameter are from console commands? Is renaming this parameter in scope for the cleanup?

aaronschachter (approved):
Looking good!

app/Managers/PostManager.php
DFurnes: Oh, that could absolutely be true! I would _looooove_ to rename (or even remove!) this parameter. Let me do a little digging to see where we might still be using it and make a follow-up PR to make those changes if possible.

aaronschachter: I'd imagine we'd still want the property/feature for backfilling, when we wouldn't want to send a confirmation for activity created months (years?) ago... or not?

DFurnes: Oh just saw this! I wonder... I think in general we'd be able to handle this by excluding the post or action from messaging in Customer.io. But I also always just jump at the chance to delete code... 😈 😂

DFurnes: Moving this over into #1144 for further discussion. ⏩

PR #1144, opened 2020-10-22
Remove unused 'dont_send_to_blink' parameters.

DFurnes:
### What's this PR do?

This pull request removes the unused `dont_send_to_blink` parameters from signup & post endpoints. These are no longer accurate (it should really be `dont_send_to_customer_io`), and [no longer used anywhere](https://github.com/search?q=org%3ADoSomething+dont_send_to_blink&type=code) (aside from a very old Ashes script).

If we need to exclude a set of posts from messaging in the future, we can do so with Customer.io trigger rules.

👀

Thanks @aaronschachter for flagging this in https://github.com/DoSomething/rogue/pull/1143#discussion_r510333090!

### Relevant tickets

N/A

### Checklist

### DIFF
--- a/app/Managers/PostManager.php
+++ b/app/Managers/PostManager.php
@@ -54,17 +54,10 @@ class PostManager
             $authenticatedUserRole,
         );
 
-        // Send to Customer.io unless 'dont_send_to_blink' is TRUE
-        $shouldSendToCustomerIo = !(
-            array_key_exists('dont_send_to_blink', $data) &&
-            $data['dont_send_to_blink']
-        );
+        // Send post event(s) to Customer.io for messaging:
+        SendPostToCustomerIo::dispatch($post);
 
-        if ($shouldSendToCustomerIo) {
-            SendPostToCustomerIo::dispatch($post);
-        }
-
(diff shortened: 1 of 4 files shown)


DFurnes:
Just referencing a continuing discussion on [the comment thread](https://github.com/DoSomething/rogue/pull/1143#discussion_r510375320) in #1143:

> @aaronschachter: I'd imagine we'd still want the property/feature for backfilling, when we wouldn't want to send a confirmation for activity created months (years?) ago... or not?

> @DFurnes: Oh just saw this! I wonder... I think in general we'd be able to handle this by excluding the post or action from messaging in Customer.io. But I also always just jump at the chance to delete code... 😈 😂

I don't really have super strong opinions on whether this stays or goes (aside from my predisposition to delete things). If we keep, I'm tempted to move this to a query parameter (like `?muted=true` or something) so that it's not smushed in with "real" fields in the request body.

aaronschachter (approved):
I'm up for removing too if we haven't used it for a long time -- this definitely simplifies our code. We could always sneak something into a post or signup details to filter by within Customer.io should we need to.

DFurnes:
Alright, let's give it a try! Worst case, we've always got the commit history. :octocat:

PR #1145, opened 2020-11-09
Strip tags from user-provided content before sending to Customer.io.

DFurnes:
### What's this PR do?

This pull request strips any HTML tags from user-provided fields (like `caption` or `why_participated`) before sending it to Customer.io. This ensures that, if we _were_ to use these in email templates (we don't at the moment AFAIK), that we would not need to worry about a malicious user injecting content into our email templates.

👀

There's some further context on why this change was made in DoSomething/northstar#1066.

### Relevant tickets

References [Pivotal #173717785](https://www.pivotaltracker.com/story/show/173717785).

### Checklist


katiecrane (approved):
looks good! thanks for adding that note about `caption`!

PR #1146, opened 2020-11-16
Campaign 'Searchable' payload - Add has_website computed attribute for filtering

mendelB:
### What's this PR do?

This pull request adds a computed `has_website` attribute to the payload generated by the `toSearchableArray` method on Campaigns (for Algolia / Laravel Scout) so that we can filter for website campaigns via Algolia.
  
👀 

[Algolia specifies](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-null-or-missing-attributes/) two methods to filter by a missing/null attribute (which we're doing here -- a campaign with a null `campaign_contentful_id` is not a website campaign and vice versa).

The simpler of the two approaches seemed to be to simply add a computed attribute (the other method being to use a new `_tags` array for a built-in "faceted search"). And this is consistent with the existing `hasWebsite` filter [in GraphQL](https://github.com/DoSomething/graphql/blob/533018cbb108b279649e800dcd79e6fc1552a3a6/src/schema/rogue.js#L83-L84) / [Rogue Campaigns](https://github.com/DoSomething/rogue/blob/cff701d37ae979b55cb64a089aa19ea6e551d796/app/Models/Campaign.php#L132-L146).

Once this attribute is in Algolia, we can use a simple `has_website = 1` to filter for website campaigns.

This is all to continue equipping our [Algolia campaigns search](https://github.com/DoSomething/graphql/blob/master/src/dataSources/Algolia.js) on GraphQL with the functionality expected by the [Phoenix Paginated Campaigns gallery](https://github.com/DoSomething/phoenix-next/blob/6f77f7492c3ca15302a6a01a096a1b7a2b8ae23a/resources/assets/components/utilities/PaginatedCampaignGallery/PaginatedCampaignGallery.js).

### Relevant tickets

References [Pivotal #173817560](https://www.pivotaltracker.com/story/show/173817560).

### Checklist

### DIFF
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -326,6 +326,9 @@ class Campaign extends Model
             $array[$date] = optional($this->{$date})->timestamp;
         }
 
+        // Add boolean attribute for filtering website campaigns.
+        $array['has_website'] = !is_null($this->contentful_campaign_id);
+
         // Only send data we want to search against.
         return Arr::only($array, [
             'accepted_count',
@@ -333,6 +336,7 @@ class Campaign extends Model
             'contentful_campaign_id',
             'cause',
             'end_date',
+            'has_website',
             'id',
             'internal_title',
             'pending_count',
--- a/app/Models/Campaign.php
+++ b/app/Models/Campaign.php
@@ -327,7 +327,7 @@ class Campaign extends Model
         }
 
         // Add boolean attribute for filtering website campaigns.
-        $array['has_website'] = !is_null($this->contentful_campaign_id);
+        $array['has_website'] = isset($this->contentful_campaign_id);
 
         // Only send data we want to search against.
         return Arr::only($array, [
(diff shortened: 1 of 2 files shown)


app/Models/Campaign.php
```diff
@@ -326,13 +326,17 @@ public function toSearchableArray()
             $array[$date] = optional($this->{$date})->timestamp;
         }
 
+        // Add boolean attribute for filtering website campaigns.
+        $array['has_website'] = !is_null($this->contentful_campaign_id);
```
lindsayrmaher, line 330: so this is basically saying, if the `contentful_campaign_id` is not null set `has_website` to true? just to make sure i'm understanding this. haven't looked through php in a minute 😅

lindsayrmaher (approved):
looking good to me, just one clarifying question to make sure i'm understanding everything!

app/Models/Campaign.php
mendelB: I'm really grateful you asked this because I second-guessed the double negative but left it in there. You _are_ correct as to the logic, but I updated this to use an [`isset`](https://www.w3schools.com/php/func_var_isset.asp) instead to check that the `contentful_id` has an assigned value and is _not_ `null` in https://github.com/DoSomething/rogue/pull/1146/commits/a7a6501fde8fcae2cdb1aa904bbac0c4630c6b32 for readability!

PR #1150, opened 2020-12-04
Algolia Campaign is_group_campaign Computed Attribute

mendelB:
### What's this PR do?

This pull request adds a new computed `is_group_campaign` attribute to the Campaign model and sends this attribute along to Algolia.

👀 

This will allow us to filter for group / non-group campaigns via GraphQL / the Algolia API

### Relevant tickets

References [Pivotal #175358477](https://www.pivotaltracker.com/story/show/175358477).

### Checklist

### DIFF
--- a/tests/Unit/Models/CampaignTest.php
+++ b/tests/Unit/Models/CampaignTest.php
@@ -4,6 +4,7 @@ namespace Tests\Unit\Models;
 
 use Rogue\Models\Action;
 use Rogue\Models\Campaign;
+use Rogue\Models\GroupType;
 use Tests\TestCase;
 
 class CampaignTest extends TestCase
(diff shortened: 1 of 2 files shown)


tests/Unit/Models/CampaignTest.php
```diff
@@ -61,6 +66,10 @@ public function testToSearchableArray()
             $evergreenWebsiteCampaign->toSearchableArray()['is_evergreen'],
             true,
         );
+        $this->assertEquals(
+            $evergreenWebsiteCampaign->toSearchableArray()['is_group_campaign'],
+            true,
+        );
```
DFurnes, line 72: This is so nitpicky, but what if we used `assertTrue` and `assertFalse` for these checks? I think it might read a little clearer, especially in scenarios like this one where things get split onto multiple lines.

DFurnes (approved):
👥

tests/Unit/Models/CampaignTest.php
mendelB: Oooooh heck yes! That is one satisfying refactor https://github.com/DoSomething/rogue/pull/1150/commits/8a43d8ea443919a91bd2c33c6492659b4d93aaaa

PR #1165, opened 2021-01-27
Bug hunt

aaronschachter:
### What's this PR do?

This pull request looked promising to fix the bug reported in https://github.com/DoSomething/chompy/pull/198 -- the debugging in Chompy displays the posts `details` in the API response. 

However, when my local Chompy then attempts to execute a `PATCH /posts/:id` request to update the post status, it receives an error with status 403: `This action is unauthorized.`

Per the 2nd commit, I thought that might be the issue where Gambit doesn't see `why_participated,` but I can see it breaks a ton of our tests (the 1st commit for `PostPolicy` does not). 

...

...

### Relevant tickets

References [Pivotal #]().

### Checklist

### DIFF
--- a/app/Policies/PostPolicy.php
+++ b/app/Policies/PostPolicy.php
@@ -17,7 +17,7 @@ class PostPolicy
      * @param  App\Models\Post $post
      * @return bool
      */
-    public function viewAll($user, Post $post)
+    public function viewAll(?Authenticatable $user, Post $post)
     {
         return is_staff_user() || is_owner($post);
     }
--- a/app/Policies/SignupPolicy.php
+++ b/app/Policies/SignupPolicy.php
@@ -16,7 +16,7 @@ class SignupPolicy
      * @param  App\Models\Signup $signup
      * @return bool
      */
-    public function viewAll($user, Signup $signup)
+    public function viewAll(?Authenticatable $user, Signup $signup)
     {
         return is_staff_user() || is_owner($signup);
     }


aaronschachter:
@DFurnes Per out chat today, I tried both using the full path to the Gate facade within the PostTransfomer, and also using kebab case `view-all` instead of the camelCase `viewAll` when calling the Gate method and didn't have luck with either.
 
 Looking at the[ Northstar policy methods](https://github.com/DoSomething/northstar/blob/0c5c92e2a8b62498d101686e67bab43650e72eaf/app/Policies/UserPolicy.php#L61-L64), I saw that the Rogue policy methods were missing the nullable type and thinking that could be a clue... but wondering if this potentially sheds any new light on the bug?

DFurnes:
Hmm, I think you're onto something! By default, Laravel disallows anonymous users from being authorized to do things unless you [use those nullable types to add "guest user" support](https://laravel.com/docs/8.x/authorization#guest-users). These requests aren't _really_ anonymous.... but they also (from Gambit & Chompy) don't have a _user_ logged in!

If you add those nullable types, does it run the expected method?

aaronschachter:
Yes, just committed adding a logger into the Post Policy's `viewAll` and confirmed it logs ([which wasn't happening in the branch that I initially was working on during our cal](https://github.com/DoSomething/rogue/compare/bug-hunt)l -- I created a new branch once it seemed like this nullable / machine-user issue was the culprit).

aaronschachter:
Running the signup tests and can see this error -- where the test is passing a `DoSomething\Gateway\Server\RemoteUser`, causing a failure...

```
[2021-01-27 22:42:32] testing.ERROR: Argument 1 passed to App\Policies\SignupPolicy::viewAll() must be an instance of App\Policies\Authenticatable or null, instance of DoSomething\Gateway\Server\RemoteUser given, called in /home/vagrant/Code/rogue/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php on line 706 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Argument 1 passed to App\\Policies\\SignupPolicy::viewAll() must be an instance of App\\Policies\\Authenticatable or null, instance of DoSomething\\Gateway\\Server\\RemoteUser given, called in /home/vagrant/Code/rogue/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php on line 706 at /home/vagrant/Code/rogue/app/Policies/SignupPolicy.php:19)
[stacktrace]
```

aaronschachter:
Tried working around defining the class in https://github.com/DoSomething/rogue/pull/1165/commits/3a281de5a163697ce671786c88b631b1df9e51de by setting a `null` default, which gets tests to pass locally -- but our linter won't allow it. Would we always expect the `viewAll` to be called with a nullable `DoSomething\Gateway\Server\RemoteUser` ? (which I'd imagine won't be the case anymore once app is merged into Northstar...)

